PR #16, opened 2025-05-19
Fix documentation structure and fine tune explanations

rgsl888prabhu:
This is an update to doc after refactoring it to add C and Python API. And address missing items and fixing few structures.

### DIFF
--- a/cpp/src/dual_simplex/phase2.cpp
+++ b/cpp/src/dual_simplex/phase2.cpp
@@ -35,6 +35,28 @@ namespace cuopt::linear_programming::dual_simplex {
 namespace phase2 {
 
 template <typename i_t, typename f_t>
+f_t l2_dual_residual(const lp_problem_t<i_t, f_t>& lp, const lp_solution_t<i_t, f_t>& solution)
+{
+  std::vector<f_t> dual_residual = solution.z;
+  const i_t n                    = lp.num_cols;
+  // dual_residual <- z - c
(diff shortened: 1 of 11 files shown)


docs/cuopt/source/cuopt-cli/cli.rst
```diff
@@ -0,0 +1,144 @@
...
+
+Mixed Integer Programming Example
+#################################
+
+Here's an example of solving a Mixed Integer Programming (MIP) problem using the CLI:
+
+.. code-block:: bash
+
+    echo "* Optimal solution -28
+   * Has 5 nodes
```
chris-maes: I think the comment about the number of nodes is out of date. Looks like it only has 4 nodes now. You might want to remove it.

python/libcuopt/libcuopt/_cli_wrapper.py
```diff
@@ -0,0 +1,27 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES.
```
ramakrishnap-nv, line 1: @tmckayus I had to add this since cuopt_cli was getting added under /libcuopt/bin/ and PATH was not able to pick it up, and adding it under PATH was causing it to loose link to dependent libraries. This fixes the issue. This acts like a wrapper around it with same name and resides in the traditional PATH so it is readily available after installation.

PR #17, opened 2025-05-19
Fix LP termination status 9 errors in batch LP solves

chris-maes:
These errors were caused by all LPs in the batch sharing the same concurrent halt variable. This meant that the first LP to solve, forced all other LPs to halt. So it was possible that both PDLP and Dual Simplex solves returned with a ConcurrentLimit termination status. This should never happen normally.

For now, we fix this by disabling the Concurrent method in batch mode. We set the method to PDLP, if Concurrent was chosen, and tell the user to choose PDLP or Dual Simplex to avoid this warning in the future.

### DIFF
--- a/cpp/src/linear_programming/utilities/cython_solve.cu
+++ b/cpp/src/linear_programming/utilities/cython_solve.cu
@@ -274,6 +274,14 @@ std::pair<std::vector<std::unique_ptr<solver_ret_t>>, double> call_batch_solve(
   // Limit parallelism as too much stream overlap gets too slow
   const int max_thread = compute_max_thread(data_models);
 
+  if (solver_settings->get_parameter<int>(CUOPT_METHOD) == CUOPT_METHOD_CONCURRENT) {
+    CUOPT_LOG_INFO("Concurrent mode not supported for batch solve. Using PDLP instead. ");
+    CUOPT_LOG_INFO(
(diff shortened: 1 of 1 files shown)


cpp/src/linear_programming/utilities/cython_solve.cu
```diff
@@ -274,6 +274,14 @@ std::pair<std::vector<std::unique_ptr<solver_ret_t>>, double> call_batch_solve(
   // Limit parallelism as too much stream overlap gets too slow
   const int max_thread = compute_max_thread(data_models);
 
+  if (solver_settings->get_parameter<int>(CUOPT_METHOD) == CUOPT_METHOD_CONCURRENT) {
+    CUOPT_LOG_INFO("Concurrent mode not supported for batch solve. Using PDLP instead. ");
+    CUOPT_LOG_INFO(
+      "Set the CUOPT_METHOD parameter to CUOPT_METHOD_PDLP or CUOPT_METHOD_DUAL_SIMPLEX to avoid "
```
rg20, line 280: This would be printed even if the user does not specify anything and leave it default right?

chris-maes: Yep. I think that is the correct behavior. IF the user leaves Concurrent as the default setting, we want to switch it and ask them to switch it themselves.

PR #21, opened 2025-05-20
PDLP solution file generation and binary fix

Kh4ster:
This PR fixes two issues:

1. Solution file generated by PDLP presented several JSON issues, was still mentioning ms for the solve time and was not handling the potential missing fields from Dual Simplex. To handle the last and so that user can know, a bool was added to know which solver solved the LP problem
2. The default value for binary time limit was an int resulting in a bad any cast in case it was not set

### DIFF
--- a/cpp/include/cuopt/logger.hpp
+++ b/cpp/include/cuopt/logger.hpp
@@ -72,7 +72,11 @@ inline rapids_logger::logger& default_logger()
 {
   static rapids_logger::logger logger_ = [] {
     rapids_logger::logger logger_{"CUOPT", {default_sink()}};
+#if CUOPT_LOG_ACTIVE_LEVEL >= RAPIDS_LOGGER_LOG_LEVEL_INFO
+    logger_.set_pattern("%v");
+#else
     logger_.set_pattern(default_pattern());
+#endif
     logger_.set_level(default_level());
     logger_.flush_on(rapids_logger::level_enum::info);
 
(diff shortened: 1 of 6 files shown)


cpp/include/cuopt/linear_programming/pdlp/solver_solution.hpp
```diff
@@ -97,6 +97,9 @@ class optimization_problem_solution_t : public base_solution_t {
 
     /** Solve time in seconds */
     double solve_time;
+
+    /** Whether the problem was solved by PDLP or Dual Simplex */
+    bool solved_by_pdlp;
```
aliceb-nv: Minor tangential remark but should the members of `additional_termination_information_t` possess default initializers? This is essentially a plain struct and I see that its instanciation sites initialize members manually, this might lead in the future to missed initializations/garbage output in `write_additional_termination_statistics_to_file` (e.g. if in some future codepath we forget to set `solved_by_pdlp`)

Kh4ster: Yes I think I agree. Should we initialized it with a "zero" value or "nan" value so that if those value appear we know we forgot something?

aliceb-nv: My vote would be for NaN :) 0 may be valid in some context / mislead whoever is reading the solution log into thinking this is expected output

python/cuopt/cuopt/linear_programming/solution/solution.py
```diff
@@ -128,6 +128,8 @@ class Solution:
         Time used for pre-solve
     solve_time: Float64
         Solve time in milliseconds
```
Iroy30: does this need to be seconds now?

Kh4ster: yes good catch

PR #22, opened 2025-05-20
Handle out of memory and out of time limit issues gracefully

rg20:
This PR adds two things: 

1. Introduces a time limit for computing related variables. This function can take a lot of time for very large problems (in this case 20+ M variables).  Ideally, we need to speed up this with a better algo and fine tune for larger problems
2. Catch `std::bad_alloc` exceptions in solve function. Currently, we only catch `cuopt::logic_error`

### DIFF
--- a/cpp/src/mip/problem/problem.cu
+++ b/cpp/src/mip/problem/problem.cu
@@ -793,12 +793,6 @@ void problem_t<i_t, f_t>::compute_related_variables(double time_limit)
   i_t related_var_offset = 0;
   auto start_time        = std::chrono::high_resolution_clock::now();
   for (i_t i = 0;; ++i) {
-    auto current_time = std::chrono::high_resolution_clock::now();
-    if (std::chrono::duration_cast<std::chrono::seconds>(current_time - start_time).count() >
-        time_limit) {
-      CUOPT_LOG_DEBUG("Early exit in computing related variables!");
-      break;
-    }
     i_t slice_size = min(max_slice_size, n_variables - i * max_slice_size);
     if (slice_size <= 0) break;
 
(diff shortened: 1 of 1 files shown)


cpp/src/mip/problem/problem.cu
```diff
@@ -788,7 +791,14 @@ void problem_t<i_t, f_t>::compute_related_variables()
 
   i_t output_offset      = 0;
   i_t related_var_offset = 0;
+  auto start_time        = std::chrono::high_resolution_clock::now();
   for (i_t i = 0;; ++i) {
+    auto current_time = std::chrono::high_resolution_clock::now();
+    if (std::chrono::duration_cast<std::chrono::seconds>(current_time - start_time).count() >
```
hlinsen: I think the resize of vectors is required here. You could move the time check in the condition slightly below.

cpp/src/linear_programming/solve.cu
```diff
@@ -553,6 +553,11 @@ optimization_problem_solution_t<i_t, f_t> solve_lp(optimization_problem_t<i_t, f
   } catch (const cuopt::logic_error& e) {
     CUOPT_LOG_ERROR("Error in solve_lp: %s", e.what());
     return optimization_problem_solution_t<i_t, f_t>{e, op_problem.get_handle_ptr()->get_stream()};
+  } catch (const std::bad_alloc& e) {
+    CUOPT_LOG_ERROR("Error in solve_lp: %s", e.what());
+    return optimization_problem_solution_t<i_t, f_t>{
+      cuopt::logic_error("Memory allocation failed", cuopt::error_type_t::RuntimeError),
```
chris-maes, line 559: Should a `bad_alloc` be a `cuopt::logic_error`? We might want to create a different cuopt exception for this.

rg20: bad_alloc is thrown by rmm, unless we catch them somewhere and convert them to logic_error, which is what I am doing here

cpp/src/mip/problem/problem.cu
rg20: Not sure I follow.  

Do you mean resize to 0 if we can't find all?

hlinsen: > Do you mean resize to 0 if we can't find all?

Yes

rg20: done

PR #23, opened 2025-05-20
Fix bug in crossover where numerical was returned instead of time limit

chris-maes:
Also adds the --relaxation flag to cuopt_cli to solve the LP relaxation of a MIP

### DIFF
--- a/cpp/cuopt_cli.cpp
+++ b/cpp/cuopt_cli.cpp
@@ -81,6 +81,7 @@ inline auto make_async() { return std::make_shared<rmm::mr::cuda_async_memory_re
  */
 int run_single_file(const std::string& file_path,
                     const std::string& initial_solution_file,
+                    bool solve_relaxation,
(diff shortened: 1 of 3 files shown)


cpp/cuopt_cli.cpp
```diff
@@ -257,8 +263,9 @@ int main(int argc, char* argv[])
   std::string file_name = program.get<std::string>("filename");
 
   const auto initial_solution_file = program.get<std::string>("--initial-solution");
+  const auto solve_relaxation      = program.get<bool>("--relaxation");
```
rg20, line 266: does this handle true false words as well? or just 0 and 1 ?

chris-maes: It works like a flag. If you add `--relaxation` the value is true, if you don't add that option the value is false. I thought it should be different than the other parameter settings (since this is not a parameter).

PR #25, opened 2025-05-21
Fix bug in C API for bool parameters. Fix bug on irish-electricty

chris-maes:
Since C does not have a boolean type, we need to allow `cuOptGetIntParameter` and `cuOptSetIntParameter` to work for boolean parameters as well.

On irish-electricity we had a bug where we were stopping PDLP early, due to a numerical error inside dual simplex. In concurrent mode we should only stop the other solver when we have reached a definitive conclusion on the problem.

Fix a bug in the Python docstrings found by Flora.

Also log whenever we change settings. And set the log pattern to `%v` outside the solve as well.

### DIFF
--- a/cpp/include/cuopt/logger.hpp
+++ b/cpp/include/cuopt/logger.hpp
@@ -94,7 +94,11 @@ inline void reset_default_logger()
 {
   default_logger().sinks().clear();
   default_logger().sinks().push_back(default_sink());
+#if CUOPT_LOG_ACTIVE_LEVEL >= RAPIDS_LOGGER_LOG_LEVEL_INFO
+  default_logger().set_pattern("%v");
+#else
   default_logger().set_pattern(default_pattern());
+#endif
   default_logger().set_level(default_level());
   default_logger().flush_on(rapids_logger::level_enum::info);
 }


cpp/cuopt_cli.cpp
```diff
@@ -86,6 +86,7 @@ int run_single_file(const std::string& file_path,
 {
   const raft::handle_t handle_{};
   cuopt::linear_programming::solver_settings_t<int, double> settings;
+  cuopt::default_logger().set_pattern("%v");
```
rg20: We are better off doing this in the default_logger() construction.   That way we don't have to set this everywhere

replace default_pattern() with "%v"

cpp/src/linear_programming/cuopt_c.cpp
```diff
@@ -492,6 +496,14 @@ cuopt_int_t cuOptSetIntegerParameter(cuOptSolverSettings settings,
     static_cast<solver_settings_t<cuopt_int_t, cuopt_float_t>*>(settings);
   try {
     solver_settings->set_parameter<cuopt_int_t>(parameter_name, parameter_value);
+  } catch (const std::invalid_argument& e) {
```
rg20, line 495: There is a bool in C. You just need to include `stdbool.h` right?

chris-maes: stdbool.h just creates a typedef that defines a `bool` to be an `_Bool` which I beleive is an `int` and the  `true` and `false` constants. If you look at C code `bool`s are not really standard. I thought about bringing in `stdbool.h` and creating `cuOptGet/SetBoolParameter` functions, but these would just be syntactic sugar. I think using `int`s for `bool`s is more aligned with how C is done in practice.

cpp/cuopt_cli.cpp
chris-maes: I wasn't sure where to do this. Is this the right place in `cuopt/include/logger.hpp`?

```
inline rapids_logger::logger& default_logger()
{
  static rapids_logger::logger logger_ = [] {
    rapids_logger::logger logger_{"CUOPT", {default_sink()}};
    logger_.set_pattern(default_pattern());  /// <- replace with %v?
    logger_.set_level(default_level());
    logger_.flush_on(rapids_logger::level_enum::info);

    return logger_;
  }();
  return logger_;
}
```

chris-maes: Looks like this work. Done. Thank you.

cpp/include/cuopt/logger.hpp
```diff
@@ -72,7 +72,11 @@ inline rapids_logger::logger& default_logger()
 {
   static rapids_logger::logger logger_ = [] {
     rapids_logger::logger logger_{"CUOPT", {default_sink()}};
+#if CUOPT_LOG_ACTIVE_LEVEL >= RAPIDS_LOGGER_LOG_LEVEL_INFO
```
rg20, line 75: we need this in reset function below as well, I think

chris-maes: Ah. Good catch, thanks. Sorry I'm not familiar with how the logging works.

chris-maes: Done.

PR #26, opened 2025-05-22
Fix non-trivial presolve not reporting Optimality

aliceb-nv:
The non-trivial presolve did not report optimality if the problem was fully reduced. Fixed by this PR

Closes https://github.com/rapidsai/cuopt/issues/2507

### DIFF
--- a/cpp/src/dual_simplex/presolve.cpp
+++ b/cpp/src/dual_simplex/presolve.cpp
@@ -24,7 +24,9 @@
 namespace cuopt::linear_programming::dual_simplex {
 
 template <typename i_t, typename f_t>
-i_t remove_empty_cols(lp_problem_t<i_t, f_t>& problem, i_t& num_empty_cols)
+i_t remove_empty_cols(lp_problem_t<i_t, f_t>& problem,
+                      i_t& num_empty_cols,
+                      presolve_info_t<i_t, f_t>& presolve_info)
 {
   constexpr bool verbose = false;
   if (verbose) { printf("Removing %d empty columns\n", num_empty_cols); }
@@ -35,15 +37,22 @@ i_t remove_empty_cols(lp_problem_t<i_t, f_t>& problem, i_t& num_empty_cols)
   // sum_{k != j} c_k * x_k + c_j * l_j if c_j > 0
   // or
   // sum_{k != j} c_k * x_k + c_j * u_j if c_j < 0
+  presolve_info.removed_variables.reserve(num_empty_cols);
+  presolve_info.removed_values.reserve(num_empty_cols);
+  presolve_info.removed_reduced_costs.reserve(num_empty_cols);
   std::vector<i_t> col_marker(problem.num_cols);
   i_t new_cols = 0;
   for (i_t j = 0; j < problem.num_cols; ++j) {
     if ((problem.A.col_start[j + 1] - problem.A.col_start[j]) == 0) {
       col_marker[j] = 1;
+      presolve_info.removed_variables.push_back(j);
+      presolve_info.removed_reduced_costs.push_back(problem.objective[j]);
       if (problem.objective[j] >= 0) {
+        presolve_info.removed_values.push_back(problem.lower[j]);
         problem.obj_constant += problem.objective[j] * problem.lower[j];
         assert(problem.lower[j] > -inf);
       } else {
+        presolve_info.removed_values.push_back(problem.upper[j]);
         problem.obj_constant += problem.objective[j] * problem.upper[j];
         assert(problem.upper[j] < inf);
       }
@@ -52,6 +61,7 @@ i_t remove_empty_cols(lp_problem_t<i_t, f_t>& problem, i_t& num_empty_cols)
       new_cols++;
     }
   }
+  presolve_info.remaining_variables.reserve(new_cols);
 
   problem.A.remove_columns(col_marker);
   // Clean up objective, lower, upper, and col_names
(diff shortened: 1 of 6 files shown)


cpp/src/mip/solver.cu
```diff
@@ -108,13 +109,14 @@ solution_t<i_t, f_t> mip_solver_t<i_t, f_t>::run_solver()
   if (!presolve_success) {
     CUOPT_LOG_INFO("Presolve is infeasible, returning infeasible solution!");
```
chris-maes: Very minor nitpick. This might read better as
CUOPT_LOG_INFO("Problem proven infeasible in presolve")

cpp/src/mip/solver.cu
```diff
@@ -98,6 +98,7 @@ solution_t<i_t, f_t> mip_solver_t<i_t, f_t>::run_solver()
   if (context.problem_ptr->empty) {
     CUOPT_LOG_INFO("Problem fully reduced at trivial presolve");
```
chris-maes: Very minor nitpick. This might read better as
CUOPT_LOG_INFO("Problem fully reduced in presolve")

cpp/src/mip/solver.cu
```diff
@@ -108,13 +109,14 @@ solution_t<i_t, f_t> mip_solver_t<i_t, f_t>::run_solver()
   if (!presolve_success) {
     CUOPT_LOG_INFO("Presolve is infeasible, returning infeasible solution!");
     solution_t<i_t, f_t> sol(*context.problem_ptr);
-    sol.set_problem_infeasible();
+    sol.set_problem_fully_reduced();
     context.problem_ptr->post_process_solution(sol);
     return sol;
   }
   if (context.problem_ptr->empty) {
     CUOPT_LOG_INFO("Problem fully reduced at presolve");
```
chris-maes: Very minor nitpick. This might read better as
CUOPT_LOG_INFO("Problem full reduced in presolve");

cpp/src/mip/solution/solution.cuh
```diff
@@ -148,7 +148,7 @@ class solution_t {
   f_t h_user_obj           = 0.;
   f_t h_infeasibility_cost = 0.;
   bool is_feasible         = false;
-  bool is_problem_infeasible{false};
+  bool is_problem_fully_reduced{false};
```
hlinsen, line 151: I think we should always update copy and assignment constructors with these vars.

cpp/src/mip/solver.cu
```diff
@@ -98,6 +98,7 @@ solution_t<i_t, f_t> mip_solver_t<i_t, f_t>::run_solver()
   if (context.problem_ptr->empty) {
     CUOPT_LOG_INFO("Problem fully reduced at trivial presolve");
     solution_t<i_t, f_t> sol(*context.problem_ptr);
+    sol.set_problem_fully_reduced();
     context.problem_ptr->post_process_solution(sol);
```
chris-maes, line 102: Dumb question: is this where the solution is populated? Where does presolve store the information about the solution when the problem is fully reduced in presolve?

cpp/src/mip/solution/solution.cuh
aliceb-nv: Oh good catch, thank you Hugo!

cpp/src/mip/solver.cu
aliceb-nv: In the case of fully presolved problems the solution is populated by post_process_solution() (which tracks the transformations made by presolve) and additional solving stats information is added when `get_solution()` is called to return the final `mip_solution_t` : https://github.com/NVIDIA/cuopt/blob/branch-25.05/cpp/src/mip/solution/solution.cu#L560-L606

cpp/src/mip/solution/solution.cuh
```diff
@@ -72,10 +72,10 @@ class solution_t {
   void set_infeasible();
   // gets the is_feasible
   bool get_feasible();
-  // gets the is_problem_infeasible
-  bool get_problem_infeasible();
```
akifcorduk, line 76: Why do we remove the infeasible options? The problem can be infeasible right?

cpp/src/mip/solver.cu
```diff
@@ -106,15 +107,16 @@ solution_t<i_t, f_t> mip_solver_t<i_t, f_t>::run_solver()
   dm.timer              = timer_;
   bool presolve_success = dm.run_presolve(timer_.remaining_time());
   if (!presolve_success) {
-    CUOPT_LOG_INFO("Presolve is infeasible, returning infeasible solution!");
+    CUOPT_LOG_INFO("Problem proven infeasible in presolve");
     solution_t<i_t, f_t> sol(*context.problem_ptr);
-    sol.set_problem_infeasible();
+    sol.set_problem_fully_reduced();
```
akifcorduk, line 112: If we decide to keep infeasible, here it makes more sense to keep set_problem_infeasible

cpp/src/mip/solution/solution.cuh
aliceb-nv: I decided to cover both the case of infeasibility and optimality after presolve with `set_problem_fully_reduced`. `get_solution()` then determines, based on the status of the solution after post_process, which status to return (Optimal or Infeasible)
This ended up making the code more concise so that's why I went with this :)

https://github.com/NVIDIA/cuopt/pull/26/files#diff-115aa63d798001b25f7e001080d8ff9136923b2509328d8ccd079131e4dbc2b8L589-R613

PR #27, opened 2025-05-22
Bug fixes on preprocess, line segment returning worse solution, temporary fix add B&B solution directly

akifcorduk:
This PR fixes bugs in

- Preprocess in set callback function, the checks were done on the original assignment before preprocess, thus rejecting almost all set_solution calls.
- Line segment returning worse solutions because of bug.
- A side effect of LS returning worse solutions is that B&B added solutions are not added to the population as LS makes them worse.


aliceb-nv (approved):
LGTM, thanks!

cpp/src/mip/problem/problem.cu
```diff
@@ -630,6 +611,25 @@ bool problem_t<i_t, f_t>::pre_process_assignment(rmm::device_uvector<f_t>& assig
                  temp_assignment.begin(),
                  assignment.begin());
   handle_ptr->sync_stream();
+  auto has_nans = cuopt::linear_programming::detail::has_nans(handle_ptr, assignment);
```
hlinsen: This will cause errors if not caught before

cpp/src/mip/problem/problem.cu
```diff
@@ -630,6 +611,25 @@ bool problem_t<i_t, f_t>::pre_process_assignment(rmm::device_uvector<f_t>& assig
...
+
+  auto has_integrality_discrepancy = cuopt::linear_programming::detail::has_integrality_discrepancy(
+    handle_ptr, integer_indices, assignment, tolerances.integrality_tolerance);
+  if (has_integrality_discrepancy) {
+    CUOPT_LOG_DEBUG("Solution discarded due to integrality discrepancy");
+    return false;
+  }
+
+  auto has_variable_bounds_violation =
+    cuopt::linear_programming::detail::has_variable_bounds_violation(handle_ptr, assignment, this);
```
hlinsen, line 628: Does it solve the issue if you pass original problem bounds here?

akifcorduk: Probably, i didn't try it. But it was not accepting the solutions we outputted with get_solution.

hlinsen: Because bounds are scaled so I think this is the issue

hlinsen (approved):
Thanks Akif!

spoorendonk:
Port complete, validated, pushed to spoorendonk/cuopt delta-api-main.

PR #29, opened 2025-05-22
Bug fix: solution incorrect when simplex presolve removes empty columns

### LINKED ISSUE #13: [BUG] standgub.json from netlib/lp fails if zero-only final column is left in the constraint matrix
The standgub dataset from the netlib/lp suite has a final column in the constraint matrix which is all zeros.  If that column is removed from the dataset, cuopt will find an optimal solution. If that column is left in the dataset, cuopt will return dual infeasible.  Is this behavior correct?

The attached zip file contains a Python runner, a JSON file with the standgub dataset from netlib/lp( converted from MPS using the cuOpt MPS parser to generate a dictionary), and a log of outputs using the runner in both modes.

[standgub.zip](https://github.com/user-attachments/files/20195794/standgub.zip)

Run the problem with the zero column included
```bash
python cuopt_json.py standgub.json
```
Run the problem with the zero column removed
```bash
python cuopt_json.py standgub.json -z
````

### PR DESCRIPTION
**Pull request by chris-maes**
Fixes an issue reported on the following MPS file, where `_dummy` variable does not appear in the constraints. We remove empty columns (and thus `_dummy`) in presolve for dual simplex. But do not take this into account when constructing the solution. 

Further work, is needed to develop an actual LP presolve. 

```
*SENSE:Minimize
NAME          test_export_dict_LP_no_obj
ROWS
 N  OBJ
 G  c1
 E  c2
 L  c3
 G  c4
COLUMNS
    __dummy   OBJ        1.000000000000e+00
    w         c4         1.000000000000e+00
    x         c1         1.000000000000e+00
    x         c2         1.000000000000e+00
    y         c1         1.000000000000e+00
    y         c3        -1.000000000000e+00
    z         c2         1.000000000000e+00
    z         c3         1.000000000000e+00
RHS
    RHS       c1         5.000000000000e+00
    RHS       c2         1.000000000000e+01
    RHS       c3         7.000000000000e+00
    RHS       c4         0.000000000000e+00
BOUNDS
 FX BND       __dummy    0.000000000000e+00
 FX BND       w          0.000000000000e+00
 UP BND       x          4.000000000000e+00
 LO BND       y         -1.000000000000e+00
 UP BND       y          1.000000000000e+00
```

Fixes #13  which was caused by an issue in column scaling when the norm of a column is zero.

### DIFF
--- a/cpp/src/dual_simplex/scaling.cpp
+++ b/cpp/src/dual_simplex/scaling.cpp
@@ -48,7 +48,7 @@ i_t column_scaling(const lp_problem_t<i_t, f_t>& unscaled,
       const f_t x = scaled.A.x[p];
       sum += x * x;
     }
-    f_t col_norm_j = column_scaling[j] = std::sqrt(sum);
+    f_t col_norm_j = column_scaling[j] = sum > 0 ? std::sqrt(sum) : 1.0;
     max                                = std::max(col_norm_j, max);
   }
   settings.log.printf("Scaling matrix. Maximum column norm %e\n", max);

### CODE REVIEW

File: cpp/src/dual_simplex/presolve.cpp [hunk 1]
```diff
@@ -826,6 +842,36 @@ void uncrush_primal_solution(const user_problem_t<i_t, f_t>& user_problem,
...
+
+  i_t k = 0;
+  for (const i_t j : presolve_info.remaining_variables) {
+    uncrushed_x[j] = crushed_x[k];
+    uncrushed_z[j] = crushed_z[k];
+    k++;
+  }
+
+  for (const i_t j : presolve_info.removed_variables) {
+    uncrushed_x[j] = presolve_info.removed_values[j];
```
rg20: Is the indexing correct here? 

I think it should be 

```
k = 0; 
for (const i_t j : presolve_info.removed_variables) {
    uncrushed_x[j] = presolve_info.removed_values[k]; 
    k++;
```

chris-maes: Yes, thank you for catching this! 

I got lucky with the instance that Ishka sent where there was only one removed variable and it occurred at position 0. I verified on my unit test that the this code gives out of memory accesses. Your suggestion is correct.

PR #33, opened 2025-05-23
Fix bug: dual simplex does not fill in additional info

chris-maes:
Adds additional termination info for dual simplex. This allows users to get the same info from either solver when using Concurrent mode.


cpp/src/linear_programming/solve.cu
```diff
@@ -253,10 +255,17 @@ optimization_problem_solution_t<i_t, f_t> convert_dual_simplex_sol(
...
+  info.primal_objective            = solution.user_objective;
+  info.dual_objective              = solution.user_objective;
+  info.gap                         = 0.0;
+  info.relative_gap                = 0.0;
+  info.solve_time                  = duration;
+  info.number_of_steps_taken       = solution.iterations;
+  info.l2_primal_residual          = solution.l2_primal_residual;
+  info.l2_dual_residual            = solution.l2_dual_residual;
+  info.l2_relative_primal_residual = solution.l2_primal_residual / (1.0 + norm_user_objective);
+  info.l2_relative_dual_residual   = solution.l2_dual_residual / (1.0 + norm_rhs);
```
Kh4ster, line 274: Shouldn't we also set:
total_number_of_attempted_steps (same as the other step for Dual Simplex)
max_primal_ray_infeasibility (can be 0 for Dual Simplex)
primal_ray_linear_objective (same)
max_dual_ray_infeasibility (same)
dual_ray_linear_objective (same)
?

chris-maes: Done. Thanks for catching this!

PR #38, opened 2025-05-27
Fix branch and nightly build for libcuopt wheel to build on only python3.12

rgsl888prabhu:
Libcuopt wheel is a wrapper arounf C++ layer and there is nothing specific with respect to python version, so we just need to build for python3 version for different cuda and archs.

This also adds updates to README to add build status


README.md
```diff
@@ -53,4 +55,4 @@ conda install -c rapidsai -c conda-forge -c nvidia \
 We also provide [nightly Conda packages](https://anaconda.org/rapidsai-nightly) built from the HEAD
 of our latest development branch.
 
-Note: cuOpt is supported only on Linux, and with Python versions 3.10 and later.
+Note: cuOpt is supported only on Linux, and with Python versions 3.10 and later.
```
tmckayus, line 58: if we are limiting tests to 3.12 and excluding 3.13, should we say 3.10-3.12 here?

ramakrishnap-nv: Yes, I have updated in another PR where I am addressing VDR and VPR reviews

PR #42, opened 2025-05-28
Add missing nvtx ranges and cuda checks

hlinsen:
This PR adds:
* More comprehensive profiling by populating more methods with nvtx ranges.The bottlenecks shift depending on instance properties and are hard to track.
* Missing checks after kernel calls.


rgsl888prabhu:
@rg20 @Kh4ster  May I get your review on this PR

PR #51, opened 2025-05-30
include Python constraints for pure-Python conda packages

jameslamb:
The `cuopt-sh-client` and `cuopt-server` packages are pure Python.

However, their conda packages are not technically built as `noarch: python` because of some historical limitations in RAPIDS build workflows (to be addresses as part of https://github.com/rapidsai/build-planning/issues/43).

As a result, 1 of those packages is published for each combination of `(Python minor version, CPU architecture)` supported by RAPIDS. See, for example, https://anaconda.org/nvidia/cuopt-server/files?version=25.05.00

image

For as long as we're doing that, those packages should have a constraint on `python`, so that (for example) only 1 package matches for a given `(Python minor version, CPU architecture)`.

This does that.

### DIFF
--- a/conda/recipes/cuopt-server/recipe.yaml
+++ b/conda/recipes/cuopt-server/recipe.yaml
@@ -40,9 +40,6 @@ requirements:
     - pandas>=2
     - python
     - uvicorn ${{ uvicorn_version }}
-  ignore_run_exports:
-    by_name:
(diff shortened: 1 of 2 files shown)


gforsyth (approved):
:tada:

rgsl888prabhu (approved):

PR #62, opened 2025-06-02
Update Source build instructions

rgsl888prabhu:
This simplifies and update source build instructions.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -98,11 +98,13 @@ cd $CUOPT_HOME
 
 #### Building with a conda environment
 
-**Note:** Using a conda environment is the easiest way to satisfy the library's dependencies.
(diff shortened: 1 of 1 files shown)


chris-maes:
We say the following

```

**Note:** Using a conda environment is the easiest way to satisfy the library's dependencies.

- Create the conda development environment:
Please install conda if you don't have it already. You can install it from [https://docs.conda.io/en/latest/miniconda.html](https://docs.conda.io/en/latest/miniconda.html)
```

I think it might be good to 

1) Make a bit stronger statement that "is the easiest way to satisfy the library's dependencies". It sound like conda is really the only way to satisfy these dependencies. So maybe we should say

```
Building from source without conda is very difficult. We highly recommend that users build cuOpt inside a conda environment
```

2) Include instructions for installing miniconda. I worry a bit about pointing users to miniconda installation instructions. Maybe we could add one or two lines that show how to install miniconda?

rg20 (approved):

PR #68, opened 2025-06-03
Pdlp fix batch cuda graph

Kh4ster:
This PR aims at fixing the invalid operation while there is a graph capture we sometimes see when using batch solve.

Solution is to use a regular instead of a non-blocking stream to make sure that if any operation (like a cudaFree from Thrust) is being launched on the default stream, it will wait for all other operations on other stream to finish first, preventing any cudaMalloc/Free while another stream might be doing a CUDA Graph capture.


aliceb-nv (approved):

rgsl888prabhu:
@Kh4ster Just wanted to confirm if this needs to go to 25.05

Kh4ster:
@rgsl888prabhu if we can have it in 25.05 yes that's better, that avoid having a hard to read / understand and impossible to fix error when running in batch mode

tmckayus:
> I think there are few other bug fixes that needs to go in, may be we can add this as well.

I agree

PR #74, opened 2025-06-05
Fix data initialization in create depot node for max travel time feature

hlinsen:
This PR fixes:
- Fixes data init in create depot node when wait time is modeled. This is used for move eval in 2-opt and OX.
- Fixes an assertion when computed arrival time is checked and wait time is modeled
- Removes deprecated tests


jameslamb (approved):
Giving you a `cuopt-ci-codeowners` approval. The CMake changes look ok to me.

PR #75, opened 2025-06-05
Fix inversion of constraint bounds in conditional bounds presolve

kaatish:
Updating the constraint bounds based on minimal and maximal activity can sometimes cause the update constraint lower bounds to be greater than constraint upper bounds even if the constraint is feasible within tolerance.

This PR fixes the constraint bounds to the same value if this happens.

### DIFF
--- a/cpp/src/mip/presolve/bounds_presolve.cu
+++ b/cpp/src/mip/presolve/bounds_presolve.cu
@@ -350,82 +350,36 @@ void bound_presolve_t<i_t, f_t>::set_updated_bounds(const raft::handle_t* handle
 template <typename i_t, typename f_t>
 void bound_presolve_t<i_t, f_t>::calc_and_set_updated_constraint_bounds(problem_t<i_t, f_t>& pb)
 {
-  i_t cross_bound_count_0 = thrust::transform_reduce(
-    pb.handle_ptr->get_thrust_policy(),
-    thrust::make_counting_iterator(0),
-    thrust::make_counting_iterator(pb.n_constraints),
-    [cnst_lb = make_span(pb.constraint_lower_bounds),
-     cnst_ub = make_span(pb.constraint_upper_bounds)] __device__(i_t idx) -> i_t {
-      return cnst_lb[idx] > cnst_ub[idx];
-    },
-    0,
-    thrust::plus<i_t>{});
-
   calculate_activity_on_problem_bounds(pb);
 
-  i_t infeas_cnst_count = thrust::transform_reduce(
-    pb.handle_ptr->get_thrust_policy(),
-    thrust::make_counting_iterator(0),
-    thrust::make_counting_iterator(pb.n_constraints),
-    [eps     = 1e-12,
-     min_act = make_span(upd.min_activity),
-     max_act = make_span(upd.max_activity),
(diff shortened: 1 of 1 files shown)


cpp/src/mip/presolve/bounds_presolve.cu
```diff
@@ -350,18 +352,29 @@ void bound_presolve_t<i_t, f_t>::calc_and_set_updated_constraint_bounds(problem_
...
+                    cnst_lb = make_span(pb.constraint_lower_bounds),
+                    cnst_ub = make_span(pb.constraint_upper_bounds)] __device__(i_t idx) -> i_t {
+                     auto min_a    = min_act[idx];
+                     auto max_a    = max_act[idx];
+                     auto c_lb     = cnst_lb[idx];
+                     auto c_ub     = cnst_ub[idx];
+                     auto new_c_lb = max(c_lb, min_a);
+                     auto new_c_ub = min(c_ub, max_a);
+                     i_t infeas    = check_infeasibility<i_t, f_t>(
+                       min_a, max_a, new_c_lb, new_c_ub, pb.tolerances.absolute_tolerance, 0);
```
chris-maes: pb.tolerances.absolute_tolerance is pretty large: 1e-4 if I remember correctly. I think ideally this would be around 1e-6. Maybe we should introduce a new tolerance here that we can remove later once we bring the absolute tolerance down to 1e-6.

chris-maes:
Thanks for the fix @kaatish . Would you also be able to add a tolerance to `check_var_bounds_sanity` and `check_constraint_bounds_sanity`? We  should not throw an error unless the lower bound is greater than the upper bound plus a tolerance.

We also should consider if we should declare the problem infeasible (rather than throwing an exception) if `check_var_bounds_sanity` or `check_constraint_bounds_sanity` are triggered on the original problem from the user.

cpp/src/mip/presolve/bounds_presolve.cu
kaatish: relative_tolerance default value is set to 1e-6. Do you want to just use that? If not, what should I name the new tolerance as?

chris-maes: Maybe we should add a tolerance called `presolve_absolute_tolerance = 1e-6`. We can remove it later when we set the `absolute_tolerance = 1e-6`.

chris-maes:
> > Thanks for the fix @kaatish . Would you also be able to add a tolerance to `check_var_bounds_sanity` and `check_constraint_bounds_sanity`? We should not throw an error unless the lower bound is greater than the upper bound plus a tolerance.
> 
> Same as above? Tolerance of 1e-6?

Yes, I would use 1e-6. You can refer to it as `presolve_absolute_tolerance`. 

> > We also should consider if we should declare the problem infeasible (rather than throwing an exception) if `check_var_bounds_sanity` or `check_constraint_bounds_sanity` are triggered on the original problem from the user.
> 
> I can return a bool from check_bounds_sanity and return false from run_presolve if the check fails. What you are proposing is to add those checks (apart from the checks we already have) before we run trivial_presolve?

No need to add any additional checks. I'm just proposing we mark the problem as infeasible (rather than throwing an exception) if the check fails before presolve. During presolve if the bounds cross, we should already return Infeasible before calling this check.

PR #87, opened 2025-06-10
Create a bash script to run mps files in parallel

rg20:
This PR adds a bash script to run all mps files in a given directory in parallel. This should replace run_mip and run_pdlp scripts.  

The script uses  file locking to enable all free gpus to grab the next available mps file so that load balancing is ensured.

### DIFF
--- a/benchmarks/linear_programming/run_mps_files.sh
+++ b/benchmarks/linear_programming/run_mps_files.sh
@@ -1,0 +1,203 @@
+#!/bin/bash
+
+# SPDX-FileCopyrightText: Copyright (c) 2021-2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+# This script runs MPS (Mathematical Programming System) files in parallel across multiple GPUs
+# It supports running linear programming (LP) and mixed-integer programming (MIP) problems
+#
+# Usage:
+#   ./run_mps_files.sh --path /path/to/mps/files --ngpus 4 --time-limit 3600 --output-dir results
+#
+# Arguments:
+#   --path        : Directory containing .mps files (required)
+#   --ngpus       : Number of GPUs to use (default: 1)
+#   --time-limit  : Time limit in seconds for each problem (default: 360)
+#   --output-dir  : Directory to store output log files (default: current directory)
+#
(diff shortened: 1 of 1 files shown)


benchmarks/linear_programming/run_mps_files.sh
```diff
@@ -0,0 +1,276 @@
+#!/bin/bash
+
+# SPDX-FileCopyrightText: Copyright (c) 2021-2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
ramakrishnap-nv:
```suggestion
# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```

benchmarks/linear_programming/run_mps_files.sh
```diff
@@ -0,0 +1,276 @@
...
+
+Optional Arguments:
+    --ngpus N           Number of GPUs to use (default: 1)
+    --time-limit N      Time limit in seconds for each problem (default: 360)
+    --output-dir PATH   Directory to store output log files (default: current directory)
+    --relaxation       Run relaxation instead of solving the MIP
+    --mip-heuristics-only  Run mip heuristics only
+    --write-log-file   Write log file
+    --num-cpu-threads  Number of CPU threads to use
+    --batch-num        Batch number
```
ramakrishnap-nv, line 77: May be more details on batch number and n-batches and does it support LP or MIP or both.

benchmarks/linear_programming/run_mps_files.sh
```diff
@@ -0,0 +1,276 @@
...
+    --n-batches        Number of batches
+    --log-to-console   Log to console
+    -h, --help         Show this help message and exit
+
+Examples:
+    # Run all MPS files in /data/lp using 2 GPUs with 1 hour time limit
+    $(basename "$0") --path /data/lp --ngpus 2 --time-limit 3600
+
+    # Run with specific GPU devices using CUDA_VISIBLE_DEVICES
+    CUDA_VISIBLE_DEVICES=0,2,3 $(basename "$0") --path /data/mip
```
ramakrishnap-nv, line 87: Since mps files don't have option to set solver settings, do you want to make this into config + dataset ? Does it makes sense to have a config to set solver settings for each  mps ?

ramakrishnap-nv (commented):
@rg20 This is an awesome addition, just had few comments.

benchmarks/linear_programming/run_mps_files.sh
rg20: It definitely does not make sense to have different setting for each mps, because this script is purely for benchmarking purpose.  

With regards to settings, I am thinking this would be benchmarking with all the default settings.    That's why I did not add any parameter to it.

ramakrishnap-nv: Makes sense

benchmarks/linear_programming/run_mps_files.sh
ramakrishnap-nv: Do we need to expand more on these options ?

rg20: added notes and example on the batch num 

We have run_mip.cpp script that we have been using, I moved all the options from there here. If need more, we can add. But for now this is good, I think.

ramakrishnap-nv (approved):

PR #92, opened 2025-06-10
Fix to decouple thin client requirements from cuopt main module

rgsl888prabhu:
Termination status is being pulled from cuOpt which makes cuopt main library a dependency for cuopt thin client. Just for temporary, a copy of these termination status is being added to solution to detach thin client.

We will circle back and fix this in 25.08. I will create an issue to follow-up for this.

### DIFF
--- a/python/cuopt/cuopt/linear_programming/solution/solution.py
+++ b/python/cuopt/cuopt/linear_programming/solution/solution.py
@@ -13,11 +13,51 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from cuopt.linear_programming.solver.solver_wrapper import (
-    LPTerminationStatus,
-    MILPTerminationStatus,
-    ProblemCategory,
-)
+from enum import IntEnum
+
+
+# TODO: Remove this once we have a way to not include main cuOpt
+# libs in in thin client
+class ProblemCategory(IntEnum):
+    """
+    Problem category enum
+    """
+
+    LP = 0
+    MIP = 1
+    IP = 2
+
+
+# TODO: Remove this once we have a way to not include main cuOpt
+# libs in in thin client
+class LPTerminationStatus(IntEnum):
+    """
+    LP termination status enum
+    """
+
+    NoTermination = 0
+    NumericalError = 6
+    Optimal = 1
+    PrimalInfeasible = 2
+    DualInfeasible = 3
+    IterationLimit = 4
+    TimeLimit = 5
+    PrimalFeasible = 7
+
+
+# TODO: Remove this once we have a way to not include main cuOpt
+# libs in in thin client
+class MILPTerminationStatus(IntEnum):
+    """
+    MILP termination status enum
+    """
+
+    NoTermination = 0
+    Optimal = 1
+    FeasibleFound = 8
+    Infeasible = 2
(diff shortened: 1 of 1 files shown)


python/cuopt/cuopt/linear_programming/solution/solution.py
```diff
@@ -13,11 +13,51 @@
...
+
+
+# TODO: Remove this once we have a way to not include main cuOpt
+# libs in in thin client
+class LPTerminationStatus(IntEnum):
+    """
+    LP termination status enum
+    """
+
+    NoTermination = 0
```
rg20: these are hard-coded right? 

I think you can use CUOPT_NO_TERMINATION 

Most of these constants already defined in solver.solver_parameters.pyx. Instead we can just have a constants.pyx where we define all constants. Wdyt?

rg20: I am doing similar things on JuMP interface as well

ramakrishnap-nv: @rg20 issue is we are trying to connect things back to cuOpt, which we should avoid since it will make cuOpt a necessary item on thin client.

ramakrishnap-nv: solver.solver_parameters.pyx belongs to main cuOpt library.

chris-maes: I don't think it is viable to hard code these status codes. I think it very likely that these status codes will change in the future. 

Where are these status codes being used in the thin client? Could we have the thin client use a string instead?

ramakrishnap-nv: Yes, I agree, that’s why I want to fix this in 25.08, and I had thought about completely detaching this and making it string based. But this might require many changes and I wanted to make changes with least amount of items being moved since we are doing hot fix and didn’t want to add too many changes last minute.

ramakrishnap-nv: I have made PR changes to accommodate strings

rg20 (approved):

rgsl888prabhu:
@jameslamb May I get review on this PR

python/cuopt/cuopt/linear_programming/solution/solution.py
```diff
@@ -13,12 +13,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from cuopt.linear_programming.solver.solver_wrapper import (
```
chris-maes: I don't think we want to remove these status codes for the entire Python API. Just for the thin client right?

ramakrishnap-nv: Solution.py is shared, so we can't have it on both, else we need to create another Solution folder for thin client.

rgsl888prabhu:
> This changes the entire Python API to use strings instead of status codes for termination status. I don't think that is what we want. We should just change the thin client to use strings. No changes need to be made in solution.py, solver_wrapper.pyx, or the tests of the Python API.

solution.py is packaged with thin client as well, so it would affect thin client if we keep it as it is.

PR #96, opened 2025-06-11
Replace accidental O(N^2) BFRT with O(N log N). 13% better on NETLIB

chris-maes:
The current bound-flipping ratio test (BFRT) makes an O(N) pass through the remaining breakpoints for each breakpoint resulting in O(N^2) run time, where N is the number of breakpoints.

This PR places the breakpoints in a heap in O(N) time. And then removes the minimum breakpoint O(1) and reorders the heap O(log N) for each of the N breakpoints. Thus, the run time should be O(N log N).

On the NETLIB LP test set this results in a 67% reduction of time spent in the BFRT and a 13% reduction in overall solve time.

The entering variable chosen by the new code should be the same as the previous code (except when ties are broken differently when constructing the heap).


chris-maes:
This is still a work in progress. As I have not yet removed the old code, since I am still using it to run benchmarks on the MIPLIB test set.

aliceb-nv:
Should this draft PR be deleted since https://github.com/NVIDIA/cuopt/pull/192 exists?

chris-maes:
> Should this draft PR be deleted since #192 exists?

Yes. I will close. Thank you.

PR #107, opened 2025-06-17
Fix issue when problem has an empty problem in pdlp

### LINKED ISSUE #103: [BUG] LP solve crashes on a tiny problem
**Describe the bug**
This model has no constraints other than the variable bounds. Dual simplex solves the problem but PDLP throws an exception.  So the default concurrent mode crashes.  

```
echo "NAME
> ROWS
>  N  OBJ
> COLUMNS
>     x1        OBJ       -3
> RHS
> RANGES
> BOUNDS
>  MI bounds    x1
>  UP bounds    x1        2
> ENDATA" >trivial.mps
```

`cuopt_cli trivial.mps 
`
**Expected behavior**
It crashes for method 0 and 1

Kh4ster:
Fixes #103 

- Matrix with an empty problem no longer triggers an assert.
- If in PDLP mode, a LOG_INFO tells the user he should use DualSimplex instead and returns a NumericalError
- If in Concurrent mode, PDLP will stop but Dual Simplex will solve the problem

### DIFF
--- a/cpp/src/linear_programming/solve.cu
+++ b/cpp/src/linear_programming/solve.cu
@@ -347,13 +347,27 @@ optimization_problem_solution_t<i_t, f_t> run_dual_simplex(
 }
 
 template <typename i_t, typename f_t>
+static optimization_problem_solution_t<i_t, f_t> run_pdlp_solver(
+  detail::problem_t<i_t, f_t>& problem,
+  pdlp_solver_settings_t<i_t, f_t> const& settings,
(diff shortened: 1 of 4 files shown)


chris-maes:
@Kh4ster This was already merged. But what about instead of just saying we can't handle a problem with no constraints in PDLP, why don't we just add an empty constraint (e.g 0 * x == 0,  or -inf <= 0 * x <= inf )

PR #109, opened 2025-06-17
Fix bug in optimization_problem_solution_t::copy_from

### LINKED ISSUE #108: [BUG] Output primal solution is wrong when pdlp fails but dual simplex works
We are copying from dual simplex solution to pdlp solution in this case, however the sizes are not correct on the pdlp side. This is a bug in `optimization_problem_solution_t::copy_from`

rg20:
This PR closes #108 

When PDLP fails, the vectors are not resized appropriately.  This is causing wrong values sent to julia interface.  This could result in crashes for large cases.


Kh4ster (approved):
All good, nice catch!

PR #110, opened 2025-06-18
Fix batch graph capture issue caused by pinned memory allocator

Kh4ster:
This PR aims at fixing the intermittent error we are seeing in CI on the Python Batch PDLP test.

Previous PR were trying to achieve the same while not solving the root cause: work being pushed to the default stream.

As described here: https://github.com/NVIDIA/cccl/issues/5027 creating a thrust::host pinned memory allocator is incorrectly launching work on the GPU on the default stream which might happen during a graph capture, creating an error.

To fix it, we now use a std::unique_ptr with a custom allocator to allocate host pinned memory.

This allows us to go back to a non blocking stream when using batch mode.


aliceb-nv (approved):
LGTM, awesome :)

PR #119, opened 2025-06-23
Fix compile error when building with clang.

legrosbuffle:
`move_score_t` is not trivially default-constructible, clang is unhappy with a shared array of `move_score_t`:

`error: initialization is not supported for __shared__ variables`

Use an aligned char array instead.


rgsl888prabhu:
Thank you @legrosbuffle for the PR

aliceb-nv (approved):
Thank you for the PR :)

PR #120, opened 2025-06-24
Use cusparsespmv_preprocess() now that Raft implements it.

vitor1001:
## Description
This removes a temporary hack.

## Checklist

- [X] I am familiar with the [Contributing Guidelines](https://github.com/NVIDIA/cuopt/blob/HEAD/CONTRIBUTING.md).
- Testing 
   - [X] New or existing tests cover these changes
   - [ ] Added tests
   - [ ] NA
- Documentation
   - [ ] Added new documentation 
   - [X] NA


cpp/src/linear_programming/cusparse_view.cu
```diff
@@ -31,96 +31,6 @@
...
-    dynamic_load_runtime::function<cusparseSpMV_preprocess_sig>("cusparseSpMV_preprocess");
-  if (func.has_value()) {
-    RAFT_CUSPARSE_TRY(cusparseSetStream(handle, stream));
-    RAFT_CUSPARSE_TRY(
-      (*func)(handle, opA, alpha, matA, vecX, beta, vecY, float_type, alg, externalBuffer));
-  }
-}
-#endif
-
 // This cstr is used in pdhg
```
Kh4ster: If we use RAFT cusparsespmv_preprocess will the above logic be maintained?

cpp/src/linear_programming/cusparse_view.hpp
```diff
@@ -77,17 +77,17 @@ class cusparse_view_t {
...
-  // cusparse view of auxillirary space needed for some spmv computations
+  // cusparse view of auxiliary space needed for some spmv computations
   cusparseDnVecDescr_t tmp_primal;
   cusparseDnVecDescr_t tmp_dual;
 
   // reuse buffers for cusparse spmv
-  rmm::device_uvector<uint8_t> buffer_non_transpose;
-  rmm::device_uvector<uint8_t> buffer_transpose;
+  rmm::device_uvector<f_t> buffer_non_transpose;
+  rmm::device_uvector<f_t> buffer_transpose;
```
Kh4ster: Since we use resize on those, I think it will overallocate if we use f_t (double in most cases) instead of uint8_t or am I mistaking?

vitor1001:
> Please run : https://github.com/NVIDIA/cuopt/blob/branch-25.08/ci/check_style.sh (or just the pre-commit part) to pass the check-style

Thanks for the tip, done!

vitor1001:
> @vitor1001 Sorry maybe it's on my side then. I have shared a screenshot with the review comments I left, don't you see those in you go up the thread or in the file section?

@Kh4ster I confirm those are not visible. Maybe there is a "finish review" or "publish comments" button?

About the comment: Raft doesn't do dlopen() explicitly ([link](https://github.com/rapidsai/raft/blob/0baf228551cbccd82d22dde43065581c31605882/cpp/include/raft/sparse/detail/cusparse_wrappers.h#L432-L454)). That said, I guess if one have a similar runtime/header mismatch Raft will be broken and maybe a better fix is having Raft CMake checking for the mismatch and asking the user to fix the environment? Such work-arounds sound dangerous, is it still necessary?

Kh4ster:
@rgsl888prabhu regarding the comments I put on the PR as review, do you know why external contributors would not see those?

Kh4ster:
@vitor1001 sorry for the delay in the response I was on PTO.

I don't know why you can't see my review comment but this is an issue, we will try to understand why.

Regarding dl_open:
I agree that the result is "dangerous" but this is the only way we had to make it work at the time.
@rgsl888prabhu do you confirm because we rely on RAPIDS CI, we may have a mismatch the compiled CUDA version and the runtime CUDA version? Or can you confirm that we do not support anything < 12.4 anymore?

Regarding my last comment @vitor1001 let me put it here as screenshot:
image

rgsl888prabhu:
> > @vitor1001 Sorry maybe it's on my side then. I have shared a screenshot with the review comments I left, don't you see those in you go up the thread or in the file section?
> 
> @Kh4ster I confirm those are not visible. Maybe there is a "finish review" or "publish comments" button?
> 
> About the comment: Raft doesn't do dlopen() explicitly ([link](https://github.com/rapidsai/raft/blob/0baf228551cbccd82d22dde43065581c31605882/cpp/include/raft/sparse/detail/cusparse_wrappers.h#L432-L454)). That said, I guess if one have a similar runtime/header mismatch Raft will be broken and maybe a better fix is having Raft CMake checking for the mismatch and asking the user to fix the environment? Such work-arounds sound dangerous, is it still necessary?

I agree, those comments are pending and need to be submitted, that's why you can see them but not us.

PR #121, opened 2025-06-24
Several fixes needed to compile cuOpt with LLVM

vitor1001:
- add std:: prefix to min() and max();
- add missing headers;
- avoid calling std::optional::value() on device code since it can throw exceptions;
- add "template" keyword to help parsing;
- use a constexpr static method to define init_data(), to avoid using a not-yet-finished classes.

## Checklist

- [X] I am familiar with the [Contributing Guidelines](https://github.com/NVIDIA/cuopt/blob/HEAD/CONTRIBUTING.md).
- Testing 
   - [X] New or existing tests cover these changes
   - [ ] Added tests
   - [ ] NA
- Documentation
   - [ ] Added new documentation 
   - [X] NA

### DIFF
--- a/cpp/src/linear_programming/utils.cuh
+++ b/cpp/src/linear_programming/utils.cuh
@@ -398,7 +398,7 @@ struct relative_residual_t {
 
     // Used for best primal so far, count how many constraints are violated
     if (abs_.has_value() && nb_violated_constraints_.has_value()) {
-      if (residual >= abs_.value() + rel_ * rhs) atomicAdd(nb_violated_constraints_.value(), 1);
+      if (residual >= *abs_ + rel_ * rhs) atomicAdd(*nb_violated_constraints_, 1);
     }
     return residual - rel_ * rhs;
   }
--- a/cpp/src/mip/presolve/load_balanced_bounds_presolve.cu
+++ b/cpp/src/mip/presolve/load_balanced_bounds_presolve.cu
@@ -30,7 +30,7 @@
 #include <utilities/device_utils.cuh>
 
 #include <cub/cub.cuh>
-#include <nvtx3/nvtx3.hpp>
+//#include <nvtx3/nvtx3.hpp>
 #include <raft/common/nvtx.hpp>
 #include "load_balanced_bounds_presolve.cuh"
 #include "load_balanced_bounds_presolve_helpers.cuh"
--- a/cpp/src/routing/ges/eject_until_feasible.cu
+++ b/cpp/src/routing/ges/eject_until_feasible.cu
@@ -31,7 +31,7 @@ template <typename i_t = int, typename f_t = float>
 __device__ inline i_t get_lane_id()
 {
   i_t id;
-  asm("mov.s32 %0, %laneid;" : "=r"(id));
+  asm("mov.s32 %0, %%laneid;" : "=r"(id));
   return id;
 }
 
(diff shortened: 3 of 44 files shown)


rgsl888prabhu:
Thank you @vitor1001 for the PR, one of the reviewers should be reviewing it shortly.

cpp/src/routing/ges/eject_until_feasible.cu
```diff
@@ -31,7 +31,7 @@ template <typename i_t = int, typename f_t = float>
 __device__ inline i_t get_lane_id()
```
hlinsen: It is a duplicate. Do you mind removing this occurance?

cpp/src/linear_programming/utils.cuh
```diff
@@ -398,7 +398,7 @@ struct relative_residual_t {
 
     // Used for best primal so far, count how many constraints are violated
     if (abs_.has_value() && nb_violated_constraints_.has_value()) {
-      if (residual >= abs_.value() + rel_ * rhs) atomicAdd(nb_violated_constraints_.value(), 1);
+      if (residual >= *abs_ + rel_ * rhs) atomicAdd(*nb_violated_constraints_, 1);
```
hlinsen, line 401: Here there is no harm in keeping the wrapper. We should actually use cuda::std::optional but it is only available starting 12.2 I think.

vitor1001:
note: updated the PR with check-style fixes

cpp/src/mip/presolve/load_balanced_bounds_presolve.cu
```diff
@@ -30,7 +30,7 @@
 #include <utilities/device_utils.cuh>
 
 #include <cub/cub.cuh>
-#include <nvtx3/nvtx3.hpp>
+// #include <nvtx3/nvtx3.hpp>
```
rg20: Delete commented line

vitor1001: Good catch, removed.

cpp/src/linear_programming/utils.cuh
vitor1001: It is just a wrapper. Calling std::optional::value() on a not valued object raises an exception, while calling std::optional::operator*() is UB (never throws). Since this code runs on the device and the exception handling is host-only, this is required to compile on Clang.

hlinsen: I know but I don't think it throws on device unless you use cuda::std::optional. However if it stops compilation in clang we can remove it.

cpp/src/routing/ges/eject_until_feasible.cu
vitor1001: Good point, I just saw there was the same thing on Raft. Done

cpp/src/linear_programming/utils.cuh
vitor1001: I can confirm it is indeed necessary on Clang.

cpp/src/routing/ges/eject_until_feasible.cu
hlinsen: Thanks!

hlinsen (approved):

hlinsen:
> @hlinsen it looks like the single Python test breakage is a flaky test. Doesn't look related to any of the changes.

I confirm it is not related. I am re running the tests.

PR #125, opened 2025-06-24
Relax pinnings on several dependencies, remove nvidia channel

bdice:
## Description
This relaxes some pinnings on numba-cuda, numba, and pandas to ensure that the latest nightly of cudf can be used. I also removed the `nvidia` channel as it is no longer needed.

This also fixes cuopt_cli test which was failing due to missing `pyenv`


bdice:
Good call @gforsyth, normally I catch myself on that but not this time!

rgsl888prabhu:
Thank you @bdice for the PR.

bdice:
Sorry. Then my advice about removing `nvidia` only applies to RAPIDS and not cuOpt... I'll revert those changes.

rgsl888prabhu:
@bdice I am working on fixing the error in tests, will push it to this PR and also revert changes to README

PR #132, opened 2025-06-25
Disable debug symbol in cuopt_cli

### LINKED ISSUE #94: [BUG] In cuopt_cli is broken in 25.08 due to audit wheel modifying some symbols
**Describe the bug**

NOTE: This happens only in case of wheel package.
cuopt_cli before audit wheel is not tripped and works fine, but after audit wheel, it segfaults as soon as it is run.

This requires additional work, so creating this issue as follow-up.

rgsl888prabhu:
## Description
cuopt_cli was failing due to unknown reason when it as being stripped wheel repair, so debug symbols was enabled  so it will not be stripped. But it seems there was an update to the patch which fixed this. 

## Issue
closes #94 

## Checklist

- Testing 
   - [ ] Added tests
   - [ ] NA
- Documentation
   - [ ] Added new documentation 
   - [x] NA


rgsl888prabhu:
> This PR and the linked issue don't describe the issue in much detail so it's hard for me to say much... but this is a good change anyway. Changing the CMake build type to `Debug` (https://cmake.org/cmake/help/v4.1/variable/CMAKE_BUILD_TYPE.html) if you needed debugging symbols is preferable to manually adding `-g` and changing the optimization level like this... that'll be portable to more compilers and influence other decisions CMake makes.
> 
> And I don't think we should be shipping packages with debug symbols to users anyway.
> 
> It may also not have been `auditwheel repair` that was stripping symbols. `scikit-build-core` does it by default: https://github.com/scikit-build/scikit-build-core/blob/a7d5bf423013100892321f3289123fa9afba1f15/README.md#L250-L251
> 
> You should also check if cutting this would allow for a stricter limit on binary size. For example, it looks like the `cuopt` wheels are around 40MiB compressed:
> 
> ```
> ----- package inspection summary -----
> file size
>   * compressed size: 39.743M
>   * uncompressed size: 97.652M
>   * compression space saving: 59.3%
> ```
> 
> ([build link](https://github.com/NVIDIA/cuopt/actions/runs/15880182269/job/44779538957?pr=132#step:12:451))
> 
> So you could probably reduce the upper limit on size from 100 MiB to something like 50, to help detect unexpected package size growth sooner:
> 
> https://github.com/NVIDIA/cuopt/blob/63fbb6b22c8949798fffe8cb34ace85ad203f2bb/python/cuopt/pyproject.toml#L122-L128

@jameslamb at the time of testing, it was segfaulting only after repair, I tested before that step and it was working. So for time being, enable debug symbol for cli to unblock. And when I tried to repro and fix as follow-up, it seemed like it got fixed. So added changes to remove it in this PR.

rgsl888prabhu:
Seems like issue still persists.

cpp/CMakeLists.txt
```diff
@@ -75,7 +75,7 @@ set(CUOPT_CXX_FLAGS "")
 set(CUOPT_CUDA_FLAGS "")
 
 if(CMAKE_COMPILER_IS_GNUCXX)
-  list(APPEND CUOPT_CXX_FLAGS -Werror -Wno-error=deprecated-declarations)
```
ramakrishnap-nv: Test comment

rk888prabhu: Test comment

cpp/CMakeLists.txt
```diff
@@ -307,7 +307,6 @@ endif()
 
 
 
-list(APPEND CUOPT_CXX_FLAGS -g -O0)
```
rk888prabhu, line 307: test comment

rg20, line 310: list(APPEND CUOPT_CXX_FLAGS -O0)

rgsl888prabhu:
Moving this PR to 25.10

rgsl888prabhu:
closing this for now, will reopen for 25.10

PR #145, opened 2025-06-26
Fix compile error when using `clang` with C++20.

legrosbuffle:
The extents type of the left hand side of those assignment are `extents<i_t, FJ_MOVE_SIZE, raft::dynamic_extent>`. But the extents type created on the right hand side by `raft::make_extents<i_t>((i_t)FJ_MOVE_SIZE, (i_t)fj.pb_ptr->n_variables))` are `extents<int, dynamic_extents, dynamic_extents>`.

The former is constructible from the latter, but the documentation specific that this conversion is *explicit*:

"This constructor is explicit if `((Extents != std::dynamic_extent && OtherExtents == std::dynamic_extent) || ...)` is true"

The above condition holds in our case: (`FJ_MOVE_SIZE != dynamic_extent && dynamic_extents == dynamic_extent)` for the first extent.

This ctor being explicit translates to the ctor of the mdspan being explicit, so in C++20, when the compiler enforces explicitness, we need to give it the exact extents type.


hlinsen (approved):
Thanks! lgtm

PR #146, opened 2025-06-26
Fix out-of-bound access in `compute_max_variable_violation`.

legrosbuffle:
When [there are free variables](https://github.com/NVIDIA/cuopt/blob/63fbb6b22c8949798fffe8cb34ace85ad203f2bb/cpp/src/mip/problem/problem.cu#L1234), the lower/upper bounds and assignment arrays are not necessarily the same size as that the original problem.

In that case, when computing max violation we're accessing out-of-bounds data.

This is already covered by `bound_standardization_test`.

I'm not quite sure about whether `assert(problem_ptr->variable_lower_bounds.size() >= num_variables);` should actually be an equality (i.e., how/if the mapping from variables to bounds is handled). But at least this is strictly better than out-of-bound accesses.


legrosbuffle:
> I'm not quite sure about whether assert(problem_ptr->variable_lower_bounds.size() >= num_variables); should actually be an equality (i.e., how/if the mapping from variables to bounds is handled). But at least this is strictly better than out-of-bound accesses.


Actually I'm seeing that same pattern of OOB accesses in other places elsewhere, so I'm suspecting that there is a larger pattern of broken invariants with the model. I'm opening a bug to discuss that.

cpp/src/mip/solution/solution.cu
```diff
@@ -541,10 +541,13 @@ f_t solution_t<i_t, f_t>::compute_max_int_violation()
 template <typename i_t, typename f_t>
 f_t solution_t<i_t, f_t>::compute_max_variable_violation()
 {
+  const auto num_variables = view().assignment.size();
+  assert(problem_ptr->variable_lower_bounds.size() >= num_variables);
```
akifcorduk: The standard is to use `cuopt_assert` to convey the error message. There, we also have the control to enable and disable asserts easier.

akifcorduk: I also think it is not possible to infer any asserts here. The size might be greater or lower. The reason is that, we might be eliminating some vars in presolve, or we might be adding some vars because of free vars.

rgsl888prabhu:
@akifcorduk what's the next course of action here ?

akifcorduk:
I couldn't reproduce this issue. It might have been fixed by one of the PRs or this is specific to the custom environment that the OP is using.

legrosbuffle:
> @legrosbuffle could you give me the instructions to reproduce this issue? Data, machine, compiler, settings etc.

I thought you had  a repro here: https://github.com/NVIDIA/cuopt/actions/runs/16776602702/job/47507156031?pr=258#step:10:2580 ?

For the data, this simply happens in several of the existing unit tests (see the bug for details: https://github.com/NVIDIA/cuopt/issues/150. 

Compiler & settings: Unfortunately we're using a custom toolchain (based on clang clang + libc++) and the only machines with GPUs I have access to require using that toolchain, so I can't give you a usable command-line for repro. But I can reproduce in two different ways: with asserts on (which triggers bound cheking errors in the `span`), or when running with address sanitizer on . Note that this is not the first time that I'm triggering asserts that you don't seem to be able to reproduce (https://github.com/rapidsai/raft/pull/2732#issuecomment-3035188498).

PR #149, opened 2025-06-26
Fix several bugs appeared in unit testing of JuMP interface

### LINKED ISSUE #147: [BUG] Wrong objective on trivial LP problem without constraints
**Describe the bug**
The objective value is being reported wrong

```
echo "NAME
OBJSENSE
 MAX
ROWS
 N  OBJ
COLUMNS
    x1        OBJ       1
RHS
RANGES
BOUNDS
 MI bounds    x1
 UP bounds    x1        1
ENDATA" >trivial.mps 

cuopt_cli trivial.mps
```

**Expected behavior**
The objective should be 1. But dual simplex returns -1.

### LINKED ISSUE #148: [BUG] Wrong handling of integer variables with non-integer bounds
**Describe the bug**
When fractional value is specified as a bound for integer variable on trivial problem, the resulting solution is not integer

```
echo "NAME
ROWS
 N  OBJ
COLUMNS
    MARKER    'MARKER'                 'INTORG'
    x1        OBJ       2
    MARKER    'MARKER'                 'INTEND'
RHS
RANGES
BOUNDS
 LO bounds    x1        1.5
 PL bounds    x1
ENDATA" > integer.mps

cuopt_cli integer.mps
```

**Expected behavior**
The objective value should be 4. But cuopt returns 3

### LINKED ISSUE #166: [BUG] Building cuOpt on pre-Volta infra produces unrelated error message(s)
When cuOpt is built from scratch on a pre-Volta GPU infrastructure, the build process produces an error where compile fails first, and this is not quite relevant to the root cause of the error (ie. pre-Volta systems not being supported in the first place).

It would be better if the build process checks for the underlying GPU infra and bail out right away with a suitable error message that says pre-Volta GPUs are not supported.

See below for an example error when cuOpt was built on a Pascal-based system.

`/home/mjs/git/cuopt/cpp/src/mip/feasibility_jump/load_balancing.cuh(
   453): error: identifier "__match_any_sync" is undefined
         uint32_t mask = __match_any_sync(__activemask(), delta);
                         ^
             detected during instantiation of "void
   cuopt::linear_programming::detail::load_balancing_mtm_compute_candid
   ates<i_t,f_t>(cuopt::linear_programming::detail::fj_t<i_t,
   f_t>::climber_data_t::view_t) [with i_t=int, f_t=double]" at line
   1540 of
   /home/mjs/git/cuopt/cpp/src/mip/feasibility_jump/feasibility_jump_ke
   rnels.cu

   1 error detected in the compilation of
   "/home/mjs/git/cuopt/cpp/src/mip/feasibility_jump/feasibility_jump_k
   ernels.cu".
   [59/209] Building CUDA object
   CMakeFiles/cuopt.dir/src/routing/local_search/compute_insertions.cu.
   o
   ninja: build stopped: subcommand failed.`

rg20:
## Description

While testing the JuMP interface, there were several corner cases showed up. Especially for tiny problems and problems without constraints. This PR has several such fixes.  

In addition, following things are added: 

- Implement `cuOptGetVersion()` in C-API
- Change the max allowed absolute gap tolerance to INFINITY 

## Issue

closes #148 
closes #147 
closes #166 

## Checklist

- Testing 
   - [x] Added tests
   - [ ] NA
- Documentation
   - [ ] Added new documentation 
   - [x] NA


cpp/include/cuopt/linear_programming/cuopt_c.h
```diff
@@ -93,6 +93,13 @@ int8_t cuOptGetFloatSize();
  */
 int8_t cuOptGetIntSize();
 
+/**
+ * @brief Get the version of the library.
+ *
+ * @return The version of the library.
+ */
+const char* cuOptGetVersion();
```
chris-maes: Instead of returning a char* (and having to deal with the user freeing memory) I would change the signature of the function to:
```
void cuOptGetVersion(cuopt_int_t *major_ptr, cuopt_int_t *minor_ptr, cuopt_int_t *bugfix_ptr);

cuopt_int_t major, minor, bugfix;
cuOptGetVersion(&major, &minor, &bugfix);

File: cpp/src/linear_programming/cuopt_c.cpp [hunk 2]
```diff
@@ -59,6 +61,18 @@ int8_t cuOptGetFloatSize() { return sizeof(cuopt_float_t); }
 
 int8_t cuOptGetIntSize() { return sizeof(cuopt_int_t); }
 
+const char* cuOptGetVersion()
+{
+  static char version[64];
+  std::snprintf(version,
```
chris-maes: See comment above. Won't `version` be put on the stack and go out of scope after the call?

rg20: This is declared static.  Won't go out of the scope until the program finishes.

cpp/src/linear_programming/cuopt_c.cpp
```diff
@@ -797,7 +811,10 @@ cuopt_int_t cuOptGetSolutionBound(cuOptSolution solution, cuopt_float_t* solutio
         solution_and_stream_view->mip_solution_ptr);
     *solution_bound_ptr = mip_solution->get_solution_bound();
   } else {
-    return CUOPT_INVALID_ARGUMENT;
+    optimization_problem_solution_t<cuopt_int_t, cuopt_float_t>* optimization_problem_solution =
```
chris-maes: `cuOptGetSolutionBound` is only for MIP. If you want to access the dual objective we should create a `cuOptGetDualObjective` for LP.

cpp/src/mip/presolve/trivial_presolve.cuh
```diff
@@ -234,7 +234,6 @@ void update_from_csr(problem_t<i_t, f_t>& pb)
 
   //  update objective_offset
   pb.presolve_data.objective_offset +=
-    pb.presolve_data.objective_scaling_factor *
```
chris-maes, line 237: Didn't we need to add this in to fix another bug? I believe @aliceb-nv suggested it.

chris-maes:
How are you handling problems with zero constraints.

Kh4ster (approved):
LGTM!

chris-maes (changes requested):
Please create a new function to get the dual objective in the C API before merging. I also think we should verify the changes around the objective are correct. As this undos past bug fixes.

cpp/src/linear_programming/cuopt_c.cpp
rg20: @chris-maes would you say the solution bound does not make sense at all for LP (at least in cuOpt context)? 

or can we use the dual objective value as the solution bound for reporting for LP?  

Also does dual objective make sense for MIP? Can we use the solution bound as the dual objective?

cpp/src/mip/presolve/trivial_presolve.cuh
rg20: As long as the conversion from maximize to minimize and vice versa are done consistently, its okay. I did both in this PR. This was needed to handle the dual simplex calculations correctly. 

In PDLP and MIP heuristics, we are not touching this constant. But in dual simplex presolve the constant is being changed (I guess when variables are fixed).  For that to be correct, I had to switch the convention.

ramakrishnap-nv (approved):

cpp/src/dual_simplex/solve.cpp
```diff
@@ -96,14 +96,14 @@ template <typename i_t, typename f_t>
...
-  const f_t user_obj = obj * lp.obj_scale + lp.obj_constant;
+  const f_t user_obj = compute_user_objective(lp, obj);
   return user_obj;
 }
 
 template <typename i_t, typename f_t>
 f_t compute_user_objective(const lp_problem_t<i_t, f_t>& lp, f_t obj)
 {
-  const f_t user_obj = obj * lp.obj_scale + lp.obj_constant;
+  const f_t user_obj = lp.obj_scale * (obj + lp.obj_constant);
```
chris-maes, line 106: I think you are changing the convention for how the scaling is done here. I'm not sure why the current convention was chosen. Can you check with @aliceb-nv  or @akifcorduk before changing this convention?

cpp/src/linear_programming/cuopt_c.cpp
```diff
@@ -59,6 +61,18 @@ int8_t cuOptGetFloatSize() { return sizeof(cuopt_float_t); }
 
 int8_t cuOptGetIntSize() { return sizeof(cuopt_int_t); }
 
+const char* cuOptGetVersion()
```
chris-maes: I'd prefer it if the signature of this function was
`void cuOptGetVersion(cuopt_int_t *major_ptr, cuopt_int_t* minor_ptr, cuopt_int_t *patch_ptr)`

chris-maes (approved):
Removing request changes so you can merge while I am out of office. But I left some comments that would be great if you could address. 

Also, it wasn't clear to me if the code handled the case where no constraints are present. If so, could you point me to where this is handled?

cpp/src/dual_simplex/solve.cpp
akifcorduk: The reason we chose the current convention was to handle `obj_scale` values other than `1` or `-1`. The new way is equivalent only if `obj_scale` is 1 or -1.

akifcorduk (approved):
Looks good but one note: the objective offset needs to be correctly handled in case the scaling factor is neither 1 nor -1.

rg20:
> Removing request changes so you can merge while I am out of office. But I left some comments that would be great if you could address.
> 
> Also, it wasn't clear to me if the code handled the case where no constraints are present. If so, could you point me to where this is handled?

in MIP empty problem is handled here:[ ](https://github.com/NVIDIA/cuopt/blob/d204b1d28d481107636d9f63a9fd75b5a0567c90/cpp/src/mip/solve.cu#L84) 

in LP, dual simplex is solving the problem. I guess initial assignment solves the problem?

PR #157, opened 2025-06-27
experiment to return service time breakdown in results

tmckayus:
Adding experimental code to break down service time.

### DIFF
--- a/python/cuopt_self_hosted/cuopt_sh_client/cuopt_self_host_client.py
+++ b/python/cuopt_self_hosted/cuopt_sh_client/cuopt_self_host_client.py
@@ -435,7 +435,8 @@ class CuOptServiceSelfHostClient:
         complete = False
         reqId = None
         done = threading.Event()
-
+        first_time = True
+        
         def poll_for_logs(reqId, logging_callback, done):
             while not done.is_set():
                 self._get_logs(reqId, logging_callback)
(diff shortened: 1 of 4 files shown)


python/cuopt_server/cuopt_server/utils/job_queue.py
```diff
@@ -269,8 +269,12 @@ def get_solution_for_id(id, delete=True):
                 s = saved_results[id]
                 if delete:
                     del saved_results[id]
-                return s.wait() + (mime_type,)
+                if timestamps:
+                    return s.wait(timeout=10) + (mime_type, s.get_timestamps())
```
tmckayus, line 273: This doesn't seem to be having any effect and I'm not sure why. The goal was to just hardcode in a wait time so that we don't have to poll at such a low interval.  Will investigate more ....

python/cuopt_self_hosted/cuopt_sh_client/cuopt_self_host_client.py
```diff
@@ -503,8 +504,10 @@ def stop_threads(log_t, inc_t, done):
                         raise TimeoutError(json.dumps(response))
                     else:
                         break
-
-                time.sleep(self.polling_interval)
+                if first_time:
```
tmckayus, line 507: skip the sleep on the first poll ....

PR #173, opened 2025-07-01
Concurrent LP in get_relaxed_lp (used in FP) and lay foundations for future PDLP warm start

Kh4ster:
This PR aims at multiple improvements in MIP:

- Allow for concurrent solve (PDLP and Dual Simplex) in get_relaxed_lp which should improve FP
	- This is done by adding a solve LP interface that takes as input the MIP problem representation (instead of the LP one)
	- This interface is now used everywhere. If we need to solve an LP with the LP problem representation, the problem will be converted (which was already done eventually in any case)
	- Added handling of the "inside_mip" flag for both PDLP and Dual Simplex
- Proper handling of PDLP with warm start data through an additional field. This is in case previous problem was solved by Dual Simplex because of a concurent solve
	- Test added and also comments
- Added a few missing NVTX ranges

Eventually we want to add plug warm start in FP for both PDLP and Dual Simplex.

### DIFF
--- a/cpp/include/cuopt/linear_programming/solve.hpp
+++ b/cpp/include/cuopt/linear_programming/solve.hpp
@@ -29,6 +29,11 @@
 
 namespace cuopt::linear_programming {
 
+namespace detail {
+template <typename i_t, typename f_t>
+class problem_t;
+}
+
 /**
  * @brief Linear programming solve function.
  * @note Both primal and dual solutions are zero-initialized. For custom initialization, see
@@ -52,6 +57,31 @@ optimization_problem_solution_t<i_t, f_t> solve_lp(
   pdlp_solver_settings_t<i_t, f_t> const& settings = pdlp_solver_settings_t<i_t, f_t>{},
   bool problem_checking                            = true,
   bool use_pdlp_solver_mode                        = true);
+
+/**
+ * @brief Linear programming solve function.
+ * @note Both primal and dual solutions are zero-initialized. For custom initialization, see
+ * op_problem.initial_primal/dual_solution
+ *
+ * @tparam i_t Data type of indexes
+ * @tparam f_t Data type of the variables and their weights in the equations
+ *
+ * @param[in] op_problem  An optimization_problem_t<i_t, f_t> object with a
+ * representation of a linear program
+ * @param[in] settings  A pdlp_solver_settings_t<i_t, f_t> object with the settings for the PDLP
+ * solver.
+ * @param[in] problem_checking  If true, the problem is checked for consistency.
+ * @param[in] use_pdlp_solver_modes  If true, the PDLP hyperparameters coming from the
+ * pdlp_solver_mode are used (instead of the ones comming from a potential hyper-params file).
+ * @return optimization_problem_solution_t<i_t, f_t> owning container for the solver solution
+ */
+template <typename i_t, typename f_t>
+optimization_problem_solution_t<i_t, f_t> solve_lp(
+  detail::problem_t<i_t, f_t>& problem,
+  pdlp_solver_settings_t<i_t, f_t> const& settings = pdlp_solver_settings_t<i_t, f_t>{},
+  bool problem_checking                            = true,
+  bool use_pdlp_solver_mode                        = true,
+  bool inside_mip                                = false);
 
 /**
  * @brief Linear programming solve function.
--- a/cpp/include/cuopt/linear_programming/solve.hpp
+++ b/cpp/include/cuopt/linear_programming/solve.hpp
@@ -81,7 +81,7 @@ optimization_problem_solution_t<i_t, f_t> solve_lp(
   pdlp_solver_settings_t<i_t, f_t> const& settings = pdlp_solver_settings_t<i_t, f_t>{},
   bool problem_checking                            = true,
   bool use_pdlp_solver_mode                        = true,
-  bool inside_mip                                = false);
+  bool inside_mip                                  = false);
 
 /**
  * @brief Linear programming solve function.
--- a/cpp/include/cuopt/linear_programming/solve.hpp
+++ b/cpp/include/cuopt/linear_programming/solve.hpp
@@ -59,7 +59,8 @@ optimization_problem_solution_t<i_t, f_t> solve_lp(
   bool use_pdlp_solver_mode                        = true);
 
 /**
- * @brief Linear programming solve function.
+ * @brief Linear programming solve function. Used in the context of a MIP when the input is a
+ * detail::problem_t.
  * @note Both primal and dual solutions are zero-initialized. For custom initialization, see
  * op_problem.initial_primal/dual_solution
  *
@@ -73,6 +74,7 @@ optimization_problem_solution_t<i_t, f_t> solve_lp(
  * @param[in] problem_checking  If true, the problem is checked for consistency.
  * @param[in] use_pdlp_solver_modes  If true, the PDLP hyperparameters coming from the
  * pdlp_solver_mode are used (instead of the ones comming from a potential hyper-params file).
+ * @param[in] inside_mip  If true, the problem is being solved in the context of a MIP.
  * @return optimization_problem_solution_t<i_t, f_t> owning container for the solver solution
  */
 template <typename i_t, typename f_t>
@@ -84,7 +86,8 @@ optimization_problem_solution_t<i_t, f_t> solve_lp(
   bool inside_mip                                  = false);
 
 /**
- * @brief Linear programming solve function.
+ * @brief Linear programming solve function. This is a wrapper around the solve_lp function taking a
+ * detail::problem_t as input.
  * @note Both primal and dual solutions are zero-initialized. For custom initialization, see
  * op_problem.initial_primal/dual_solution
  *
--- a/cpp/src/linear_programming/pdlp.cu
+++ b/cpp/src/linear_programming/pdlp.cu
@@ -128,56 +128,64 @@ pdlp_solver_t<i_t, f_t>::pdlp_solver_t(problem_t<i_t, f_t>& op_problem,
     set_initial_dual_solution(dual_sol);
   }
 
-  if (settings.get_pdlp_warm_start_data().last_restart_duality_gap_dual_solution_.size() != 0) {
-    set_initial_primal_solution(settings.get_pdlp_warm_start_data().current_primal_solution_);
-    set_initial_dual_solution(settings.get_pdlp_warm_start_data().current_dual_solution_);
-    initial_step_size_     = settings.get_pdlp_warm_start_data().initial_step_size_;
-    initial_primal_weight_ = settings.get_pdlp_warm_start_data().initial_primal_weight_;
-    total_pdlp_iterations_ = settings.get_pdlp_warm_start_data().total_pdlp_iterations_;
-    pdhg_solver_.total_pdhg_iterations_ =
-      settings.get_pdlp_warm_start_data().total_pdhg_iterations_;
-    pdhg_solver_.get_d_total_pdhg_iterations().set_value_async(
-      settings.get_pdlp_warm_start_data().total_pdhg_iterations_, stream_view_);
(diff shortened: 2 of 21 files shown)


cpp/include/cuopt/linear_programming/solve.hpp
```diff
@@ -54,7 +59,35 @@ optimization_problem_solution_t<i_t, f_t> solve_lp(
...
+ * @param[in] settings  A pdlp_solver_settings_t<i_t, f_t> object with the settings for the PDLP
+ * solver.
+ * @param[in] problem_checking  If true, the problem is checked for consistency.
+ * @param[in] use_pdlp_solver_modes  If true, the PDLP hyperparameters coming from the
+ * pdlp_solver_mode are used (instead of the ones comming from a potential hyper-params file).
+ * @param[in] inside_mip  If true, the problem is being solved in the context of a MIP.
+ * @return optimization_problem_solution_t<i_t, f_t> owning container for the solver solution
+ */
+template <typename i_t, typename f_t>
+optimization_problem_solution_t<i_t, f_t> solve_lp(
```
rg20, line 80: If this is used purely in MIP context.  It is cleaner to rename this to `solve_relaxed_lp` or `solve_lp_relaxation`

cpp/src/linear_programming/solve.cu
```diff
@@ -295,7 +295,8 @@ optimization_problem_solution_t<i_t, f_t> convert_dual_simplex_sol(
 template <typename i_t, typename f_t>
 std::tuple<dual_simplex::lp_solution_t<i_t, f_t>, dual_simplex::lp_status_t, f_t, f_t, f_t>
 run_dual_simplex(dual_simplex::user_problem_t<i_t, f_t>& user_problem,
-                 pdlp_solver_settings_t<i_t, f_t> const& settings)
+                 pdlp_solver_settings_t<i_t, f_t> const& settings,
```
rg20, line 298: We should consider changing `pdlp_solver_settings` to `lp_settings`

cpp/src/linear_programming/solve.cu
```diff
@@ -591,27 +601,37 @@ optimization_problem_solution_t<i_t, f_t> solve_lp(optimization_problem_t<i_t, f
...
     CUOPT_LOG_ERROR("Error in solve_lp: %s", e.what());
     return optimization_problem_solution_t<i_t, f_t>{
       cuopt::logic_error("Memory allocation failed", cuopt::error_type_t::RuntimeError),
-      op_problem.get_handle_ptr()->get_stream()};
+      problem.handle_ptr->get_stream()};
   }
 }
 
+template <typename i_t, typename f_t>
+optimization_problem_solution_t<i_t, f_t> solve_lp(optimization_problem_t<i_t, f_t>& op_problem,
```
rg20, line 621: Is this for pure LP?

cpp/src/linear_programming/solve.cu
```diff
@@ -455,26 +462,26 @@ void run_dual_simplex_thread(
...
+    run_dual_simplex(problem, settings, inside_mip));
 }
 
 template <typename i_t, typename f_t>
 optimization_problem_solution_t<i_t, f_t> run_concurrent(
-  optimization_problem_t<i_t, f_t>& op_problem,
   detail::problem_t<i_t, f_t>& problem,
-  pdlp_solver_settings_t<i_t, f_t> const& settings)
+  pdlp_solver_settings_t<i_t, f_t> const& settings,
+  bool inside_mip)
```
rg20, line 478: Is inside_mip currently used anywhere? or is it for future purposes?

cpp/include/cuopt/linear_programming/solve.hpp
Kh4ster: I agree that for now this is just used in the context of MIP but some people might want to solve an LP using the detail::problem representation instead of the optimization_problem one. What do you think?

cpp/src/linear_programming/solve.cu
Kh4ster: Same comment as above, yes currently this used in the context LP as its using the optimization_problem_t representation.

cpp/src/linear_programming/solve.cu
Kh4ster: Yes I'm passing it to both PDLP and Dual Simplex

cpp/include/cuopt/linear_programming/solve.hpp
Kh4ster: @akifcorduk do you have an opinion there?

akifcorduk: I think renaming it kind of pins it to MIP use-cases, I agree with Nicolas that it is just another LP interface that accepts `problem_t`. In the MIP context, the naming doesn't matter too much because we have other functions like `run_lp_with_vars_fixed`

cpp/src/linear_programming/pdlp.cu
```diff
@@ -128,56 +128,65 @@ pdlp_solver_t<i_t, f_t>::pdlp_solver_t(problem_t<i_t, f_t>& op_problem,
...
-               settings.get_pdlp_warm_start_data().last_restart_duality_gap_primal_solution_.size(),
-               stream_view_);
-    raft::copy(restart_strategy_.last_restart_duality_gap_.dual_solution_.data(),
-               settings.get_pdlp_warm_start_data().last_restart_duality_gap_dual_solution_.data(),
-               settings.get_pdlp_warm_start_data().last_restart_duality_gap_dual_solution_.size(),
-               stream_view_);
+  if (settings.has_pdlp_warm_start_data()) {
+    const auto& warm_start_data = settings.get_pdlp_warm_start_data();
+    if (!warm_start_data.solved_by_pdlp_) {
+      CUOPT_LOG_INFO(
```
akifcorduk: I wouldn't put INFO logs here, it would bloat the regular solution logs. DEBUG would be more appropiate.

cpp/src/linear_programming/pdlp.cu
```diff
@@ -128,56 +128,65 @@ pdlp_solver_t<i_t, f_t>::pdlp_solver_t(problem_t<i_t, f_t>& op_problem,
...
+        "Warm start data coming from a solution which was not solved by PDLP, skipping warm start");
+    } else if (pdlp_hyper_params::restart_strategy ==
+               static_cast<int>(
+                 pdlp_restart_strategy_t<i_t, f_t>::restart_strategy_t::TRUST_REGION_RESTART)) {
+      CUOPT_LOG_INFO(
```
akifcorduk: Same: INFO is only for solution progress info to be conveyed to the user.

cpp/src/linear_programming/solve.cu
```diff
@@ -515,7 +523,8 @@ optimization_problem_solution_t<i_t, f_t> run_concurrent(
       sol_dual_simplex.get_termination_status() == pdlp_termination_status_t::PrimalInfeasible ||
       sol_dual_simplex.get_termination_status() == pdlp_termination_status_t::DualInfeasible) {
     CUOPT_LOG_INFO("Solved with dual simplex");
-    sol_pdlp.copy_from(op_problem.get_handle_ptr(), sol_dual_simplex);
+    // TODO confirm with Akif that using the problem handle ptr is not a problem
+    sol_pdlp.copy_from(problem.handle_ptr, sol_dual_simplex);
```
akifcorduk, line 526: I think using problem handle_ptr should be okay given that there are no races across host threads. Even a different handle is used, the thread is `default_per_thread` so it should be fine.

cpp/src/mip/relaxed_lp/relaxed_lp.cu
```diff
@@ -95,17 +95,18 @@ optimization_problem_solution_t<i_t, f_t> get_relaxed_lp_solution(
...
   CUOPT_LOG_DEBUG(
     "running LP with n_vars %d n_cstr %d", op_problem.n_variables, op_problem.n_constraints);
   // before LP flush the logs as it takes quite some time
   cuopt::default_logger().flush();
-  // temporarily add timer
-  auto start_time = std::chrono::high_resolution_clock::now();
-  lp_solver.set_inside_mip(true);
-  auto solver_response = lp_solver.run_solver(start_time);
+  // TODO check that we do want to do problem checking here
+  auto solver_response = solve_lp(op_problem, settings, true, true, true);
```
akifcorduk: I would name the boolean argument, so that it is easier to read.

cpp/src/mip/relaxed_lp/relaxed_lp.cu
```diff
@@ -95,17 +95,18 @@ optimization_problem_solution_t<i_t, f_t> get_relaxed_lp_solution(
                        if (!isfinite(x) || i >= prev_size) { return 0.0; }
                        return x;
                      });
-    lp_solver.set_initial_primal_solution(lp_state.prev_primal);
-    lp_solver.set_initial_dual_solution(lp_state.prev_dual);
+    settings.set_initial_primal_solution(lp_state.prev_primal.data(),
```
akifcorduk, line 98: So currently, we are not using the full warm start data yet and this code is a base/skeleton for adding the future warm start data?

Kh4ster: Excatly, because we could need to add the warm start data to the lp_state which is also tied to the mip problem

Iroy30 (commented):
Needs changes in the server warmstart population
https://github.com/NVIDIA/cuopt/blob/e0dc6037b47401f0f5ff9a14878d54a760b21c1a/python/cuopt_server/cuopt_server/utils/linear_programming/solver.py#L315

cpp/src/linear_programming/pdlp.cu
Kh4ster: Agreed, changed

cpp/src/linear_programming/solve.cu
Kh4ster: Thanks, removed the comment

cpp/src/mip/relaxed_lp/relaxed_lp.cu
Kh4ster: I agree. But regarding the comment, should we do problem checking here? Maybe just when we are in assert/debug mode

akifcorduk: At this point we don't need to do problem checks because we are not modifiying the problem. It is done in places where we modify the problem.

Kh4ster: It did help me in the debug process, will add it in only assert/debug mode

cpp/src/linear_programming/solve.cu
```diff
@@ -537,42 +545,48 @@ optimization_problem_solution_t<i_t, f_t> run_concurrent(
...
       raft::common::nvtx::range fun_scope("Check problem representation");
       // This is required as user might forget to set some fields
-      problem_checking_t<i_t, f_t>::check_problem_representation(op_problem);
-      problem_checking_t<i_t, f_t>::check_initial_solution_representation(op_problem, settings);
+      if (!inside_mip) {
+        problem_checking_t<i_t, f_t>::check_problem_representation(*problem.original_problem_ptr);
+        problem_checking_t<i_t, f_t>::check_initial_solution_representation(
+          *problem.original_problem_ptr, settings);
+      } else {
+        problem.check_problem_representation(true, true);
```
akifcorduk: I think this might be unnecessary as we are doing the problem check in the problem cstr. We could maybe convert it to an assert? I think we should do the runtime problem check only at the beginning from the user, later problem checks should be asserts.

akifcorduk (approved):

cpp/src/linear_programming/solve.cu
Kh4ster: I agree we should do it in as assert. But I believe we should keep the checks for two reasons:
1. We also check the initial solutions which is not done in the problem constructor
2. Even if we check in the problem constructor this is more for the case where the problem has been modified in an invalid way, we then need to check the problem again

PR #178, opened 2025-07-02
New heuristic improvements

### LINKED ISSUE #140: [FEA] Improvements in MIP heuristics

akifcorduk:
## Description

This PR improves the heuristic quality and performance by following improvements:
- Multi Armed Bandit algorithm for learning and selecting the more promising crossovers for a given problem.
- Caching the integer-fixed problem and filling that problem structure rather than recreating the problem each time.
- Improved bounds prop rounding by doing double probing on a bulk. Which means we propagate two sets of variable values which are promising. This is possible with almost no additional performance hit because most of the data are already in memory. It is only a few more vectors and few more access per propagation.
- Update multi probe bounds presolve such that it can take 
- Diversity level adjustments by time progress.
- Weight management such that the best solution is unlikely to be infeasible. We already normalize the weights but this is an added measure to prevent infeasible solutions flooding the population.
- Improvement on bounds propagation recombiner, such that we do balanced rounding by trying to represent both parents. This is achieved by passing the probing candidates and counting the used candidates of each parent.
- Improvement in all recombiner such that we use at least some objective variables in the recombiners. Before there was no guarantee that there was an objective component in the recombiners.
- Dynamic selection of number of different variables used in the recombiners depending on if the recombiner brings the offspring back to one of the parents.
- Improvement on line segment recombiner by doing a delta vector only on the subspace of different vars between parents.
- Initialize the distance variables in FP to the distance of previous projection.
- Improve line segment search such that it does a middle-first, binary search rather than a sequential search of the points.
- Improve line segment search solution saving mechanism.
- Fix a bug in get_two_random function. 
- Small improvements in feasibility jump.

Added functionalities:
- Accepting multiple initial solutions instead of one.
- Recombiner and diversity configurations files for better control.
- FJ rounding of fractional solutions by adding weights on non-integrality. If within the time limit, all are not rounded, second stage only does rounding moves.

The improvements resulted in an average gap to optimality after 10mins on H100:
Regular run (B&B and heuristics):  from 30% to ~23%
Heuristics only: from 33% to ~25%
On presolved instances: from 26% to ~19%

## Issue
partially closes #140 

## Checklist

- [ X] I am familiar with the [Contributing Guidelines](https://github.com/NVIDIA/cuopt/blob/HEAD/CONTRIBUTING.md).
- Testing 
   - [ X] New or existing tests cover these changes
   - [ ] Added tests
   - [ ] NA
- Documentation
   - [ X] The documentation is up to date with these changes
   - [ ] Added new documentation 
   - [ ] NA


cpp/src/mip/problem/problem.cu
```diff
@@ -1039,13 +1049,79 @@ void problem_t<i_t, f_t>::fix_given_variables(problem_t<i_t, f_t>& original_prob
                                               const rmm::device_uvector<i_t>& variables_to_fix,
                                               const raft::handle_t* handle_ptr)
 {
-  i_t TPB = 64;
-  fix_given_variables_kernel<i_t, f_t><<<n_constraints, TPB, 0, handle_ptr->get_stream()>>>(
```
hlinsen, line 1053: We can completely remove the kernel

cpp/src/mip/problem/problem.cu
```diff
@@ -1039,13 +1049,79 @@ void problem_t<i_t, f_t>::fix_given_variables(problem_t<i_t, f_t>& original_prob
...
-    original_problem.view(),
-    view(),
-    raft::device_span<f_t>{assignment.data(), assignment.size()},
-    raft::device_span<i_t>{const_cast<i_t*>(variables_to_fix.data()), variables_to_fix.size()});
+  rmm::device_uvector<f_t> reduction_in_rhs(n_constraints, handle_ptr->get_stream());
```
hlinsen: Should we move the allocations in a struct? If it is called multiple times with large sizes the overhead could be non-negligible.

cpp/src/mip/diversity/diversity_config.hpp
```diff
@@ -0,0 +1,44 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2024-2025 NVIDIA CORPORATION & AFFILIATES. All rights
```
hlinsen:
```suggestion
 * SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights
```

cpp/src/mip/diversity/diversity_manager.cu
```diff
@@ -490,15 +565,26 @@ void diversity_manager_t<i_t, f_t>::main_loop()
...
-    auto new_solutions      = generate_more_solutions();
-    auto current_population = population.population_to_vector();
-    population.clear();
-    current_population.insert(current_population.end(),
-                              std::make_move_iterator(new_solutions.begin()),
-                              std::make_move_iterator(new_solutions.end()));
-    population.find_diversity(current_population, use_avg_diversity);
-    population.add_solutions_from_vec(std::move(current_population));
+    bool halve_population = false;
+    if (halve_population) {
```
hlinsen: Should `halve_population` go into settings?

cpp/src/mip/local_search/local_search_config.hpp
```diff
@@ -0,0 +1,29 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2024-2025 NVIDIA CORPORATION & AFFILIATES. All rights
```
hlinsen:
```suggestion
 * SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights
```

cpp/src/mip/local_search/local_search.cu
```diff
@@ -85,36 +85,48 @@ void local_search_t<i_t, f_t>::generate_fast_solution(solution_t<i_t, f_t>& solu
...
+    is_feas = run_fj_annealing(solution, timer, baseline_objective);
     // we should remove this logic once, FJ is fixed
-    if (is_feasible_before_annealing && !solution.get_feasible()) {
-      solution.copy_from(solution_before_annealing);
-      solution.handle_ptr->sync_stream();
-    }
-    cuopt_func_call(bool is_feasible_after_annealing = solution.get_feasible());
-    cuopt_assert(!is_feasible_before_search || is_feasible_after_annealing,
-                 "Annealing should not change feasibility");
+    // if (is_feasible_before_annealing && !solution.get_feasible()) {
```
hlinsen: dead code in this section

cpp/src/mip/diversity/population.cu
```diff
@@ -157,14 +171,23 @@ std::vector<solution_t<i_t, f_t>> population_t<i_t, f_t>::get_external_solutions
   return return_vector;
 }
 
+template <typename i_t, typename f_t>
+bool population_t<i_t, f_t>::is_better_than_best_feasible(solution_t<i_t, f_t>& sol)
+{
+  bool obj_better = problem_ptr->maximize ? sol.get_user_objective() > best_feasible_objective
```
hlinsen: It seems to be the only place where we have a special check for maximization. Aren't we always converting to minimization? Should it be used here https://github.com/akifcorduk/cuopt/blob/new_heuristic_improvements/cpp/src/mip/diversity/diversity_manager.cu#L852-L853?

cpp/src/mip/diversity/diversity_manager.cu
```diff
@@ -544,48 +663,234 @@ diversity_manager_t<i_t, f_t>::recombine_and_local_search(solution_t<i_t, f_t>&
...
+                         true /* use integer fixed problem */);
   cuopt_assert(population.test_invariant(), "");
   cuopt_assert(lp_offspring.test_number_all_integer(), "All must be integers after LP");
   f_t lp_qual = lp_offspring.get_quality(population.weights);
   CUOPT_LOG_DEBUG("After LP offspring sol cost:feas %f : %d", lp_qual, lp_offspring.get_feasible());
   f_t offspring_qual = std::min(offspring.get_quality(population.weights), lp_qual);
   recombine_stats.update_improve_stats(
     offspring_qual, sol1.get_quality(population.weights), sol2.get_quality(population.weights));
+  add_mab_reward(recombine_stats.get_last_attempt(),
+                 min(sol1.get_quality(population.weights), sol2.get_quality(population.weights)),
```
hlinsen:
```suggestion
                 std::min(sol1.get_quality(population.weights), sol2.get_quality(population.weights)),
```

hlinsen:
```suggestion
                 std::min(sol1.get_quality(population.weights), sol2.get_quality(population.weights)),
```

cpp/src/mip/diversity/recombiners/recombiner_configs.hpp
```diff
@@ -0,0 +1,104 @@
...
+  static constexpr size_t initial_n_of_vars_from_other = 200;
+  static constexpr size_t max_different_var_limit      = 10000;
+  static constexpr size_t min_different_var_limit      = 20;
+  static size_t max_n_of_vars_from_other;
+  static constexpr double n_var_ratio_increase_factor = 1.1;
+  static constexpr double n_var_ratio_decrease_factor = 0.99;
+  static void increase_max_n_of_vars_from_other()
+  {
+    max_n_of_vars_from_other =
+      min(size_t(max_n_of_vars_from_other * n_var_ratio_increase_factor), max_different_var_limit);
```
hlinsen:
```suggestion
      std::min(size_t(max_n_of_vars_from_other * n_var_ratio_increase_factor), max_different_var_limit);
```

cpp/src/mip/diversity/recombiners/recombiner_configs.hpp
```diff
@@ -0,0 +1,104 @@
...
```
hlinsen:
```suggestion
      std::min(size_t(max_n_of_vars_from_other * n_var_ratio_increase_factor), max_different_var_limit);
```

cpp/src/mip/diversity/recombiners/recombiner_configs.hpp
```diff
@@ -0,0 +1,104 @@
...
+  static constexpr size_t initial_n_of_vars_from_other       = 200;
+  static constexpr size_t max_different_var_limit            = 10000;
+  static constexpr size_t min_different_var_limit            = 20;
+  static size_t max_n_of_vars_from_other;
+  static constexpr double n_var_ratio_increase_factor = 1.1;
+  static constexpr double n_var_ratio_decrease_factor = 0.99;
+  static void increase_max_n_of_vars_from_other()
+  {
+    max_n_of_vars_from_other =
+      min(size_t(max_n_of_vars_from_other * n_var_ratio_increase_factor), max_different_var_limit);
```
hlinsen:
```suggestion
      std::min(size_t(max_n_of_vars_from_other * n_var_ratio_increase_factor), max_different_var_limit);
```

cpp/src/mip/diversity/recombiners/recombiner_configs.hpp
```diff
@@ -0,0 +1,104 @@
...
+    CUOPT_LOG_DEBUG("Increased max_n_of_vars_from_other in FP recombiner to %lu",
+                    max_n_of_vars_from_other);
+  }
+  static void decrease_max_n_of_vars_from_other()
+  {
+    max_n_of_vars_from_other =
+      max(size_t(max_n_of_vars_from_other * n_var_ratio_decrease_factor), min_different_var_limit);
```
hlinsen:
```suggestion
      std::max(size_t(max_n_of_vars_from_other * n_var_ratio_decrease_factor), min_different_var_limit);
```

cpp/src/mip/local_search/feasibility_pump/feasibility_pump.cu
```diff
@@ -248,140 +254,15 @@ bool feasibility_pump_t<i_t, f_t>::linear_project_onto_polytope(solution_t<i_t,
...
-}
-
 // round will use inevitable infeasibility while propagating the bounds
 template <typename i_t, typename f_t>
 bool feasibility_pump_t<i_t, f_t>::round(solution_t<i_t, f_t>& solution)
 {
   bool result;
   CUOPT_LOG_DEBUG("Rounding the point");
   timer_t bounds_prop_timer(std::min(2., timer.remaining_time()));
-  const f_t lp_run_time_after_feasible = std::min(3., timer.remaining_time() / 20.);
```
hlinsen, line 383: It should still be std::min

cpp/src/mip/problem/problem.cu
akifcorduk: I was trying to avoid bloating the problem definition, but let me create a separate struct for those problem fixing helpers.

cpp/src/mip/diversity/population.cu
akifcorduk: Here we are checking for the user objective, but you are right that we can check the internal objective. We still need the additional variable of `best_feasible_objective` and we can't use `best_feasible.get_objective()` because the best feasible might be changing depending on the quality (think small infeasibility).

The code block you posted checks the quality instead of objective, which is different.

cpp/src/mip/problem/problem.cu
akifcorduk: Done

cpp/src/mip/diversity/population.cu
akifcorduk: Done

akifcorduk:
/ok to test [b61f302](https://github.com/NVIDIA/cuopt/pull/178/commits/b61f30200be0fa8137cccb3a2766f9ec1ac73b7f)

akifcorduk:
/ok to test [4a3a0b5](https://github.com/NVIDIA/cuopt/pull/178/commits/4a3a0b590daeed4ed2223936981e3893545da6c9)

akifcorduk:
/ok to test [f057b04](https://github.com/NVIDIA/cuopt/pull/178/commits/f057b0478767effcbf55d3def5e5d8c69a637d13)

benchmarks/linear_programming/cuopt/initial_solution_reader.hpp
```diff
@@ -33,13 +34,14 @@ class solution_reader_t {
...
     while (std::getline(file, line)) {
       std::stringstream ss(line);
       std::string var_name;
       std::string value_str;
-
       // Read var_name
-      if (!std::getline(ss, var_name, ',')) continue;
+      if (!std::getline(ss, var_name, '\t')) continue;
       // Read value
       if (!std::getline(ss, value_str)) continue;
```
aliceb-nv: I remember running into occasional issues when reading some MIPLIB initial solution files with this parsing code since the objective value isn't always present or in the first line (or be tab-separated). 
I think it may be better to explicitly check if var_name == `=obj=` (the miplib script does this), and do `ss >> var_name >> value_str` to properly parse whitespaces

cpp/src/linear_programming/utilities/problem_checking.cu
```diff
@@ -113,8 +113,10 @@ void problem_checking_t<i_t, f_t>::check_initial_solution_representation(
   const optimization_problem_t<i_t, f_t>& op_problem,
   const mip_solver_settings_t<i_t, f_t>& settings)
 {
-  if (settings.initial_solution_.get() != nullptr) {
-    check_initial_primal_representation(op_problem, settings.get_initial_solution());
+  if (settings.initial_solutions.size() > 0) {
+    for (const auto& initial_solution : settings.initial_solutions) {
+      check_initial_primal_representation(op_problem, *initial_solution);
+    }
```
aliceb-nv: Very minor but the if on `.size() > 0` is redundant since it is followed by a for loop

cpp/src/mip/diversity/diversity_manager.cu
```diff
@@ -143,33 +182,34 @@ void diversity_manager_t<i_t, f_t>::average_fj_weights(i_t i)
...
-      population.run_solution_callbacks(sol);
-      initial_sol_vector.emplace_back(std::move(sol));
-    } else {
-      CUOPT_LOG_ERROR(
-        "Error cannot add the provided initial solution! \
-    Assignment size %lu \
-    initial solution size %lu",
-        sol.assignment.size(),
-        init_sol.size());
+  if (context.settings.initial_solutions.size() > 0) {
```
aliceb-nv: Same; 'if' unnecessary

cpp/src/mip/diversity/diversity_manager.cu
```diff
@@ -179,16 +219,19 @@ void diversity_manager_t<i_t, f_t>::add_user_given_solution(
 template <typename i_t, typename f_t>
 void diversity_manager_t<i_t, f_t>::generate_initial_solutions()
 {
-  add_user_given_solution(initial_sol_vector);
+  add_user_given_solutions(initial_sol_vector);
+  bool skip_initial_island_generation =
+    initial_sol_vector.size() > diversity_config_t::n_sol_for_skip_init_gen || from_dir;
```
aliceb-nv, line 222: Is `from_dir` still useful?

cpp/src/mip/diversity/diversity_manager.cu
```diff
@@ -544,48 +664,236 @@ diversity_manager_t<i_t, f_t>::recombine_and_local_search(solution_t<i_t, f_t>&
...
     return std::make_pair(sol, success);
   }
 }
 
+template <typename i_t, typename f_t>
+recombiner_enum_t diversity_manager_t<i_t, f_t>::select_mab_recombiner()
+{
+  if (recombiner_enum_t::SIZE == 0) {
+    CUOPT_LOG_ERROR("select_mab_recombiner called with no recombiners defined.");
+    cuopt_expects(false, error_type_t::RuntimeError, "No recombiners available to select in MAB.");
```
aliceb-nv, line 740: Minor nit: this could be a `static_assert`

cpp/src/mip/diversity/diversity_manager.cu
```diff
@@ -544,48 +664,236 @@ diversity_manager_t<i_t, f_t>::recombine_and_local_search(solution_t<i_t, f_t>&
...
+    // Calculate reward based on your existing logic
+    double reward = 0.0;
+    if (is_better_than_best_feasible) {
+      reward = 8.0;
+    } else if (is_better_than_best_of_parents) {
+      double factor;
+      if (std::abs(offspring_quality - best_feasible_quality) /
+            (std::abs(best_feasible_quality) + 1.0) >
+          1.0) {
+        factor = 1 / 3;
```
aliceb-nv: Should be `1. / 3` or similar to avoid it being treated as an integer division and truncated down to 0

cpp/src/mip/diversity/diversity_manager.cu
```diff
@@ -544,48 +664,236 @@ diversity_manager_t<i_t, f_t>::recombine_and_local_search(solution_t<i_t, f_t>&
...
+      } else if (std::abs(offspring_quality - best_feasible_quality) /
+                   (std::abs(best_feasible_quality) + 1.0) >
+                 0.2) {
+        factor = 1 / 2;
```
aliceb-nv: Ditto

cpp/src/mip/diversity/population.cu
```diff
@@ -245,6 +270,37 @@ void population_t<i_t, f_t>::run_solution_callbacks(solution_t<i_t, f_t>& sol)
...
+    // fixme
+    weighted_violation_of_best = max(weighted_violation_of_best, 1e-10);
+    f_t quality_difference     = best_feasible().get_quality(weights) - best().get_quality(weights);
+    CUOPT_LOG_DEBUG("quality_difference %f best_feasible_quality %f best_quality %f",
+                    quality_difference,
+                    best_feasible().get_quality(weights),
+                    best().get_quality(weights));
+    if (quality_difference < 1e-10) { return; }
+    // make the current best infeasible 10% worse than feasible
+    f_t increase_ratio = (quality_difference * 1.1) / weighted_violation_of_best;
```
aliceb-nv: We might want to turn that 1.1 into a constexpr like `max_infeasibility_weight` for tuning

cpp/src/mip/diversity/population.cu
```diff
@@ -550,7 +592,8 @@ template <typename i_t, typename f_t>
 void population_t<i_t, f_t>::halve_the_population()
 {
   raft::common::nvtx::range fun_scope("halve_the_population");
-  if (current_size() <= max_solutions / 2) { return; }
+  // try 3/4 here
+  if (current_size() <= (max_solutions * 3) / 4) { return; }
```
aliceb-nv: Minor nit: could be a constant as well

cpp/src/mip/diversity/population.cu
```diff
@@ -591,6 +634,24 @@ size_t population_t<i_t, f_t>::find_free_solution_index()
...
+  population_start_time = timer.elapsed_time();
+  initial_threshold     = var_threshold;
+}
+
+template <typename i_t, typename f_t>
+void population_t<i_t, f_t>::adjust_threshold(cuopt::timer_t timer)
+{
+  double time_ratio = (timer.elapsed_time() - population_start_time) /
+                      (timer.get_time_limit() - population_start_time);
+  // time_ratio *= time_ratio;
```
aliceb-nv: nit: dead code

cpp/src/mip/diversity/recombiners/bound_prop_recombiner.cuh
```diff
@@ -35,96 +35,219 @@ class bound_prop_recombiner_t : public recombiner_t<i_t, f_t> {
...
+       int_tol,
+       probing_values = probing_values.data(),
+       seed           = cuopt::seed_generator::get_seed()] __device__(i_t idx) {
+        f_t guiding_val = guiding_view.assignment[idx];
+        f_t other_val   = other_view.assignment[idx];
+        cuopt_assert(guiding_view.problem.check_variable_within_bounds(idx, guiding_val), "");
+        cuopt_assert(other_view.problem.check_variable_within_bounds(idx, other_val), "");
+        f_t avg_val = (other_val + guiding_val) / 2;
+        if (guiding_view.problem.is_integer_var(idx)) {
+          raft::random::PCGenerator rng(seed, 0, 0);
```
aliceb-nv: The second argument of `PCGenerator` should probably be set to `idx`, otherwise the random decision will be the same for all integers in this call

cpp/src/mip/diversity/recombiners/bound_prop_recombiner.cuh
```diff
@@ -35,96 +35,219 @@ class bound_prop_recombiner_t : public recombiner_t<i_t, f_t> {
...
+          // assign some floating value, so that they can be rounded by bounds prop
+          f_t lb = guiding_view.problem.variable_lower_bounds[idx];
+          f_t ub = guiding_view.problem.variable_upper_bounds[idx];
+          if (integer_equal<f_t>(lb, ub, int_tol)) {
+            cuopt_assert(false, "The var values must be different in A and B!");
+          } else if (isfinite(lb)) {
+            offspring_view.assignment[idx] = lb + 0.1;
+          } else {
+            offspring_view.assignment[idx] = ub - 0.1;
+          }
```
aliceb-nv, line 97: What if `ub - lb < 0.1` ?

cpp/src/mip/diversity/recombiners/line_segment_recombiner.cuh
```diff
@@ -36,28 +36,99 @@ class line_segment_recombiner_t : public recombiner_t<i_t, f_t> {
...
+      if (same_as_parents) {
+        ls_recombiner_config_t::increase_max_n_of_vars_from_other();
+      } else {
+        ls_recombiner_config_t::decrease_max_n_of_vars_from_other();
+      }
+    }
+    if (better_cost_than_parents || better_feasibility_than_parents) {
+      CUOPT_LOG_DEBUG("Offspring is feasible or better than both parents");
+      return std::make_pair(offspring, true);
+    }
```
aliceb-nv, line 131: Nit: there's a fair amount of commonality in the code between the three recombiners - maybe `recombine()` could be turned into a member of the parent `recombiner_t` class? To write this common prologue and epilogue once (maybe using CRTP to call `ls/fp/bp_recombiner_config_t` members from the parent)

cpp/src/mip/diversity/recombiners/recombiner.cuh
```diff
@@ -71,33 +71,79 @@ template <typename i_t, typename f_t>
...
+      std::vector<i_t> objective_indices_not_in_subproblem;
+      std::set_difference(objective_indices.begin(),
+                          objective_indices.end(),
+                          objective_indices_in_subproblem.begin(),
+                          objective_indices_in_subproblem.end(),
+                          std::back_inserter(objective_indices_not_in_subproblem));
+      std::shuffle(objective_indices_not_in_subproblem.begin(),
+                   objective_indices_not_in_subproblem.end(),
+                   rng_host);
+      for (auto var : objective_indices_not_in_subproblem) {
```
aliceb-nv, line 136: Was there an issue with thrust's `set_difference`? Or is it just easier to perform things host-side

cpp/src/mip/local_search/line_segment_search/line_segment_search.cu
```diff
@@ -122,23 +228,82 @@ bool line_segment_search_t<i_t, f_t>::search_line_segment(solution_t<i_t, f_t>&
...
+  cuopt_func_call(test_point_is_within_bounds(solution, point_2));
+  rmm::device_uvector<f_t> delta_vector(solution.problem_ptr->n_variables,
+                                        solution.handle_ptr->get_stream());
+
+  thrust::transform(solution.handle_ptr->get_thrust_policy(),
+                    point_2.data(),
+                    point_2.data() + solution.problem_ptr->n_variables,
+                    point_1.data(),
+                    delta_vector.begin(),
+                    [] __device__(const f_t& a, const f_t& b) { return a - b; });
```
aliceb-nv: Minor nit: I really doubt this has any noticeable effect in practice, but basic scalar types (like int/float/double/char/enums/etc) should be passed by value since passing by const ref/pointer just adds overhead

cpp/src/mip/problem/problem.cu
```diff
@@ -47,6 +47,8 @@
 
 #include <unordered_set>
 
+#include <cuda_profiler_api.h>
+
```
aliceb-nv: debugging leftover?

cpp/src/mip/problem/problem_helpers.cuh
```diff
@@ -179,6 +180,8 @@ __global__ void kernel_check_transpose_validity(raft::device_span<const f_t> coe
     __syncthreads();
     // Would want to assert there but no easy way to gtest it, so moved it to the host
     if (!shared_found) {
+      DEVICE_LOG_DEBUG(
+        "For cstr %d, var %d, value %f was not found in the transpose", constraint_id, col, value);
```
aliceb-nv, line 184: Did you run into a cusparse transpose issue?

aliceb-nv (approved):
Thanks for the great work Akif!! Incredible improvement numbers

cpp/src/mip/diversity/diversity_manager.cu
akifcorduk: Yes, we are doing test for improvement on a given population. It will be relevant as long as we are testing the diversity improvements. Once we have more stable/deterministic version, we can even add a CI test for that.

cpp/src/mip/diversity/diversity_manager.cu
akifcorduk: Yes you are right :) This function is changed in my next PR so it is more generic. Not to create merge conflicts, i leave it as it currently.

cpp/src/mip/diversity/diversity_manager.cu
akifcorduk: Yes :) I realized this later, but i think factor being zero for these cases is better as benchmarks show. For now, i will set factor to 0 for those and I hope to find better configs for the next PR.

cpp/src/mip/diversity/recombiners/bound_prop_recombiner.cuh
akifcorduk: This lambda is only executed for integers, so the diff should be at least 1.

cpp/src/mip/diversity/recombiners/recombiner.cuh
akifcorduk: It was easier to implement and i think probably more performant. We would've done 3 different set_difference i think.

cpp/src/mip/local_search/line_segment_search/line_segment_search.cu
akifcorduk: This must be either Cursor or some copy paste error. Fixed :)

cpp/src/mip/problem/problem.cu
akifcorduk: yes, fixed.

cpp/src/mip/problem/problem_helpers.cuh
akifcorduk: Not related to cusparse, but in integer fixed problem. We are computing the integer fixed problem from the original problem and dynamically fill it, instead of recreating the problem every time. This was to make sure everything matches.

cpp/src/mip/diversity/recombiners/line_segment_recombiner.cuh
akifcorduk: This is some larger change. Let's do it in another PR. I created an issue on that:
https://github.com/NVIDIA/cuopt/issues/208

Issue #177, opened 2025-07-02
[BUG] cudaErrorIllegalAddress


Gorniker:
Hello, I'm encountering an almost persistent (9/10 runs) cudaErrorIllegalAddress error running vehicle routing optimization. According to logs, it happens at "compute_insertions.cu" line 875. 
The error occurs both with initial_solution and without one. It also remains with CUDA_LAUNCH_BLOCKING=1. 
I got the same error on 2 different servers. 
Environment:
1)NVIDIA GeForce RTX 4090 (24GB), driver 570.133.20
2)L4, driver 575
CUDA: 12.8
cuOpt: 25.05
Python: 3.12.3
Memory Usage: ~1.4GB/24GB when error occurs

Data files are attached.  

[problem_without_initial.json](https://github.com/user-attachments/files/21018172/problem_without_initial.json)

[problem_with_initial.json](https://github.com/user-attachments/files/21018176/problem_with_initial.json)

[debug_logs.txt](https://github.com/user-attachments/files/21018177/debug_logs.txt)

PR #179, opened 2025-07-02
Disable IX for PDP

hlinsen:
We need a special path that does not reverse nodes that are in pair.
Closes: https://github.com/NVIDIA/cuopt/issues/177


cpp/src/routing/diversity/diverse_solver.hpp
```diff
@@ -1126,16 +1162,21 @@ struct solve {
         recombine_options.insert(recombiner_t::AEAX);
       }
       // True for VRP
-      if (consider_expensive_recombiners) {
+      if (solution::request_type == request_t::VRP &&
+          consider_expensive_recombiners) {
         recombine_options.insert(recombiner_t::EAX);
```
rg20: EAX is completely disabled for PDP?

hlinsen: It is enabled for single route solutions

hlinsen:
> Are we using a different clang style file now? It is hard to find the file where the actual code change happens. If there is no mistake on the styling, i think we should reformat whole repo with a PR.
> 
> Also could you please open an issue to make sure we implement the IX for PDP? Do we know the quality impact of disabling IX? If it is a significant decrease in quality, we should implement the fix instead of disabling it.

I think it was never meant to be enabled. I can check on homberger.

cpp/src/routing/util_kernels/runtime_checks.cu
```diff
@@ -292,7 +292,7 @@ void solution_t<i_t, f_t, REQUEST>::global_runtime_checks(
   [[maybe_unused]] bool check_feasible,
   [[maybe_unused]] const std::string_view where)
 {
-  // #define DEBUG_RUNTIME_CHECKS 1
+#define DEBUG_RUNTIME_CHECKS 1
```
akifcorduk: We should disable runtime tests before merging.

PR #180, opened 2025-07-02
Add nightly container support 

### LINKED ISSUE #123: [FEA] Add nightly container builder to repo
Add container builder and also add a workflow to create nightly container for development.

rgsl888prabhu:
## Description
Adds workflow for container build, test and push for nightly and release.

Nightly container would have the tag for example 25.8.0a-cuda12.8-py3.12

And these will be published to https://hub.docker.com/r/nvidia/cuopt/tags and ngc internal registry.

This PR also removes several unused workflows and also update license header.

## Issue
closes #123

## Checklist

- Testing 
   - [x] Added tests
   - [ ] NA
- Documentation
   - [x] Added new documentation 
   - [ ] NA


ci/docker/Dockerfile
```diff
@@ -0,0 +1,125 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+ARG CUDA_VER=12.8.0
+ARG CUOPT_VER=25.8.0
+ARG PYTHON_VER=3.12.11
+ARG PYTHON_SHORT_VER=3.12
```
jameslamb:
```suggestion
ARG CUDA_VER=unset
ARG CUOPT_VER=unset
ARG PYTHON_VER=unset
ARG PYTHON_SHORT_VER=unset
```

If these are all intended to be overridden by explicitly-passed values in CI, it'd be better to not use realistic values as defaults.

Using realistic defaults means that bugs of the form "forgot to pass in the Python version" may not be caught, as wrong values will be silently used. For more discussion on this idea:

* https://github.com/rapidsai/docker/pull/689#discussion_r1684921202
* https://github.com/rapidsai/ci-imgs/pull/102

ci/docker/Dockerfile
```diff
@@ -0,0 +1,125 @@
...
+
+FROM nvidia/cuda:${CUDA_VER}-runtime-ubuntu22.04 AS cuda-libs
+
+# Install cuOpt
+FROM nvidia/cuda:${CUDA_VER}-base-ubuntu22.04 AS python-env
```
jameslamb:
```suggestion
FROM nvidia/cuda:${CUDA_VER}-runtime-${LINUX_VER} AS cuda-libs

# Install cuOpt
FROM nvidia/cuda:${CUDA_VER}-base-${LINUX_VER} AS python-env
```

I think you should add another build argument `LINUX_VER`, to pass in this value. That way:

* this `Dockerfile` could be reused to build for multiple Ubuntu versions
* all configuration could happen outside of the `Dockerfile` (fewer places to update when you change versions)

Would you consider that?

ci/docker/Dockerfile
```diff
@@ -0,0 +1,125 @@
...
+
+RUN wget https://www.python.org/ftp/python/${PYTHON_VER}/Python-${PYTHON_VER}.tgz \
+    && tar xzf Python-${PYTHON_VER}.tgz \
+    && cd Python-${PYTHON_VER} \
+    && ./configure --enable-optimizations \
+    && make -j$(nproc) \
+    && make install \
+    && cd .. \
+    && rm -rf Python-${PYTHON_VER} Python-${PYTHON_VER}.tgz \
+    && python3 -m pip install --upgrade pip
```
jameslamb: Why are we building Python from source here? If there isn't a particular reason, please just add `python3-dev python3-pip` to the install list.

That would also (I hope) allow you to remove all the installs and then removals of libraries that I'm guessing were only being installed to support building Python from scratch, simplifying the `Dockerfile` and making builds faster.

ci/docker/Dockerfile
```diff
@@ -0,0 +1,125 @@
...
+    libreadline-dev \
+    libffi-dev \
+    libsqlite3-dev \
+    libbz2-dev \
+    && apt-get autoremove -y \
+    && apt-get clean
+
+RUN ln -sf /usr/local/bin/python3 /usr/local/bin/python
+RUN groupadd -r cuopt && useradd -r -g cuopt cuopt
+RUN chown -R cuopt:cuopt /usr/local/lib/python3.12/site-packages
```
jameslamb:
```suggestion
RUN <<EOF
ln -sf /usr/local/bin/python3 /usr/local/bin/python
groupadd -r cuopt
useradd -r -g cuopt cuopt
chown -R cuopt:cuopt /usr/local/lib/python3.12/site-packages
EOF
```

Each additional `RUN` statement adds a new layer to the image. More layers = bigger images and slower pulls, because there is some fixed overhead per image.

It can sometimes be helpful to separate things into multiple steps if some things are cache-friendly and others aren't. But in this case, these are all very-cacheable steps and should be combined. Please do that.

ci/docker/Dockerfile
```diff
@@ -0,0 +1,125 @@
...
+
+ENV PIP_EXTRA_INDEX_URL="https://pypi.nvidia.com https://pypi.anaconda.org/rapidsai-wheels-nightly/simple"
+
+SHELL ["/bin/bash", "-c"]
+RUN if [[ "${CUDA_VER}" =~ ^12\. ]]; then \
+        cuda_suffix="cu12"; \
+        python -m pip install nvidia-cuda-runtime-cu12==$(echo ${CUDA_VER} | cut -d'.' -f1-2).*; \
+    fi && \
+    python -m pip install --user --no-cache-dir cuopt-server-${cuda_suffix}==${CUOPT_VER} cuopt-sh-client==${CUOPT_VER} && \
+    python -m pip list
```
jameslamb:
```suggestion
ARG CUDA_VER
ARG CUOPT_VERSION

SHELL ["/bin/bash", "-c"]

RUN <<EOF
cuda_suffix=cu$(echo ${CUDA_VER} | cut -d'.' -f1)
cuda_major_minor=$(echo ${CUDA_VER} | cut -d'.' -f1-2)

python -m pip install \
  --extra-index-url https://pypi.nvidia.com \
  --extra-index-url https://pypi.anaconda.org/rapidsai-wheels-nightly/simple \
  --user \
  --no-cache-dir \
  "cuopt-server-${cuda_suffix}==${CUOPT_VERSION}" \
  "cuopt-sh-client==${CUOPT_VERSION}" \
  "nvidia-cuda-runtime-${cuda_suffix}==${cuda_major_minor}.*" \
 
python -m pip list
EOF
```

I have several suggestions here:

1. fix the handling of the CUDA suffix here. You have multiple competing definitions of the same thing here... an `ARG` called `cuda-suffix` which is never used and a hard-coded shell variable `cuda_suffix`
2. do you really want the nightly index to be used by default in all future `pip {cmd}` calls in uses of this image? If not, those values should be inline instead of persisted in the environment variable `PIP_EXTRA_INDEX_URL`
3. this repo no longer builds against CUDA 11 on `branch-25.08` (https://github.com/NVIDIA/cuopt/pull/32), so any "if CUDA 12" stuff should be removed
4. `pip install` calls should be combined into a single call, to minimize the risk of breaking the environment (https://github.com/rapidsai/cuopt/pull/2318#discussion_r1984159089)
5. to re-use an `ARG` after a new `FROM`, you have to repeat it... `CUDA_VER` is referenced but missing, which means it's being interpolated to an empty string

ci/docker/Dockerfile
```diff
@@ -0,0 +1,125 @@
...
+    libncurses5-dev \
+    libgdbm-dev \
+    libnss3-dev \
+    libssl-dev \
+    libreadline-dev \
+    libffi-dev \
+    libsqlite3-dev \
+    libbz2-dev \
+    ca-certificates \
+    && rm -rf /var/lib/apt/lists/*
```
jameslamb:
```suggestion
RUN apt-get update \
 && apt-get install -y \
        build-essential \
        ca-certificates \
        gcc \
        git \
        libffi-dev \
        libgdbm-dev \
        libncurses5-dev \
        libnss3-dev \
        libreadline-dev \
        libsqlite3-dev \
        libssl-dev \
        libbz2-dev \
        wget \
        zlib1g-dev \
 && rm -rf /var/lib/apt/lists/*

```

Some recommendations here:

1. pass `--no-install-recommends` to every `apt-get install` call, to minimize how many packages are downloaded (makes builds faster and images smaller, and reduces the risk of conflicts)
2. alphabetize this list so it is easier for humans to read
3. (minor, style) left-align the `&&` so the commands are a little easier to read

ci/docker/Dockerfile
```diff
@@ -0,0 +1,125 @@
...
+        python -m pip install nvidia-cuda-runtime-cu12==$(echo ${CUDA_VER} | cut -d'.' -f1-2).*; \
+    fi && \
+    python -m pip install --user --no-cache-dir cuopt-server-${cuda_suffix}==${CUOPT_VER} cuopt-sh-client==${CUOPT_VER} && \
+    python -m pip list
+
+USER root
+
+RUN apt-get purge -y gcc git \
+    && apt-get update \
+    && apt-get install -y unzip \
```
jameslamb:
```suggestion
    && apt-get install -y --no-install-recommends unzip \
```

If we're going to install this, use `--no-install-recommends` for the reasons I mentioned above. But also:

* why are we installing `unzip`, if it's not used elsewhere in the `Dockerfile`? could it be installed further up?
* why are we removing `gcc` and `git` here? If it's because we don't need compilers any more in the image, should `build-essential` be included here too?

Adding a comment over this `RUN` block would be useful, to help explain why it's here. I'm not really sure what the purpose is.

ci/docker/Dockerfile
```diff
@@ -0,0 +1,125 @@
...
+
+# Create a .bash_profile that sources .bashrc if it exists
+RUN echo 'if [ -f ~/.bashrc ]; then . ~/.bashrc; fi' > /home/cuopt/.bash_profile
+
+COPY --from=cuda-libs /usr/local/cuda/lib64/libnvrtc* /usr/local/cuda/lib64/
+COPY --from=cuda-libs /usr/local/cuda/lib64/libnvJitLink* /usr/local/cuda/lib64/
+
+ENV CUOPT_SERVER_PORT=8000
+
+ENTRYPOINT ["python", "-m", "cuopt_server.cuopt_service"]
```
jameslamb:
```suggestion
ENTRYPOINT ["python", "-m", "cuopt_server.cuopt_service"]

```

This file should end with a newline. Please consider adding the `pre-commit` hook that we use across RAPIDS to enforce and auto-fix this: https://github.com/rapidsai/dependency-file-generator/blob/8de48ce0d1e704ccedddba3e0a847db52bfffd1f/.pre-commit-config.yaml#L5-L8

*(could be a separate PR)*

ci/docker/Dockerfile
```diff
@@ -0,0 +1,125 @@
...
+COPY ./VERSION /home/cuopt/VERSION
+
+FROM install-env AS cuopt-final 
+
+ARG PYTHON_SHORT_VER
+
+# Set environment variables in .bashrc for all future shells
+RUN echo 'export PATH="/usr/local/cuda/bin:/usr/bin:/usr/local/bin:/usr/local/nvidia/bin/:/home/cuopt/.local/lib/python3.12/site-packages/libcuopt/bin/:/home/cuopt/.local/bin:$PATH"' >> /home/cuopt/.bashrc && \
+    echo 'export LD_LIBRARY_PATH="/usr/lib/x86_64-linux-gnu:/usr/lib/aarch64-linux-gnu:/usr/local/cuda/lib64:/usr/local/nvidia/lib:/usr/local/nvidia/lib64:/usr/lib/wsl/lib:/usr/lib/wsl/lib/libnvidia-container:/usr/lib/nvidia:/usr/lib/nvidia-current:/home/cuopt/.local/lib/python3.12/site-packages/libcuopt/lib/:/home/cuopt/.local/lib/python3.12/site-packages/rapids_logger/lib64:${LD_LIBRARY_PATH}"' >> /home/cuopt/.bashrc && \
+    echo 'export PYTHONPATH="/home/cuopt/.local/lib/python${PYTHON_SHORT_VER}/site-packages:/usr/local/lib/python${PYTHON_SHORT_VER}/site-packages:${PYTHONPATH:-}"' >> /home/cuopt/.bashrc
```
jameslamb: Why was it necessary to patch `PATH`, `LD_LIBRARY_PATH`, and `PYTHONPATH` in this way?

I wouldn't have expected that to be necessary, and it adds some non-trivial complexity here. For example:

* hard-coded `x86_64` will need to be changed if you try to build `arm64` images
* hard-coded `python3.12/` will need to be changed if you change the Python version

ci/docker/test_image.sh
```diff
@@ -0,0 +1,52 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -euo pipefail
+
+chsh -s /bin/bash cuopt
+
+# Install dependencies
+apt-get install -y file bzip2 gcc
```
jameslamb:
```suggestion
apt-get install -y --no-install-recommends \
    bzip2 \
    file \
    gcc
```

I'm going to stop leaving individual comments for these. Can you please go through and ensure that every `apt-get install` uses `--no-install-recommends`?

jameslamb (dismissed):
I took a first pass at reviewing this, but this is a pretty big changeset so stopping now to put up a first round of comments.

Could you please move the unrelated changes (like license comments) to a different PR? You could `@` me for a review there and I'll approve quickly.

I've left a first round of suggested changes, I'll go through again once some of those have been addressed.

ci/docker/Dockerfile
ramakrishnap-nv: Yep, that would be a good option to add. I will update the matrix.

ci/docker/Dockerfile
ramakrishnap-nv: Yeah, agree. I will update these

ci/docker/Dockerfile
ramakrishnap-nv: I think, when I was working on this dockerfile, the python version was not yet available so I had to resort to this. But I agree, this would reduce lot of items from installing and removing.

ci/docker/Dockerfile
ramakrishnap-nv: Yeah, my intention was to keep them separate so build would start from last change (in case of local/dev). But since there are already well defined steps, I can move those.

ci/docker/Dockerfile
ramakrishnap-nv:
```
this repo no longer builds against CUDA 11 on branch-25.08 (https://github.com/NVIDIA/cuopt/pull/32), so any "if CUDA 12" stuff should be removed
```
I had left this to reuse in case of CUDA 13 support in future.

ci/docker/Dockerfile
ramakrishnap-nv: gcc and git were required for python installation, since that might go away, these would not be required.

ci/docker/Dockerfile
jameslamb: If you do find you need support for multiple Python version, then still I think building with an installer like `pyenv` might be easier than compiling CPython from source.

Here's an example of how to do that: https://github.com/rapidsai/ci-imgs/blob/df5652c6b90e5f97964e9748106e73a910901af6/ci-wheel.Dockerfile#L205-L240

ci/docker/Dockerfile
ramakrishnap-nv: unzip and few other dependency were required for downloading dataset and using few notebooks.

Since the default user is cuOpt and not root (this was required since this is also run as a service). So users don't have to modify and add few basic tools.

ci/docker/Dockerfile
jameslamb: For now, I'd remove that and only add in any new `if` logic once you see a new CUDA version that requires it.

ci/docker/Dockerfile
ramakrishnap-nv: Yeah, python3.12 was a mistake, should have used PYTHON_SHORT_VER.

I had to add so many PATH and LD_LIBRARY_PATH since there was a use case where nvidia-smi was not being found or some library was not being which gets loaded from nvidia runtime. So I had to add both x86 and arm PATH just to be on safer side.

ci/docker/test_image.sh
ramakrishnap-nv: Thank you @jameslamb I will address these comments and get back to you.

ci/docker/test_image.sh
```diff
@@ -0,0 +1,54 @@
...
+
+set -euo pipefail
+
+chsh -s /bin/bash cuopt
+
+# Install dependencies
+apt-get update
+DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommends file bzip2 gcc
+apt-get clean
+rm -rf /var/lib/apt/lists/*
```
jameslamb:
```suggestion
```

For test code, I don't think you need to spend time deleting things like this. I'd remove this, for simplicity.

ci/docker/test_image.sh
```diff
@@ -0,0 +1,54 @@
...
+# Download test data
+bash datasets/linear_programming/download_pdlp_test_dataset.sh
+bash datasets/mip/download_miplib_test_dataset.sh
+cd datasets && ./get_test_data.sh --solomon && ./get_test_data.sh --tsp && cd -
```
jameslamb:
```suggestion
pushd ./datasets
./get_test_data.sh --solomon
./get_test_data.sh --tsp
popd
```

Chaining these commands with `&&` will make the logs a bit more difficult to understand if anything breaks, and makes the logic of the test file a bit harder to follow. I recommend just splitting it onto separate lines.

ci/docker/context/README.md
```diff
@@ -0,0 +1 @@
+Add all the files and data for the buildx context to this folder like entrypoint script, and others.
```
jameslamb: I think this file should be moved up a level (`ci/docker/README.md`) and that this information about `context/` should be just one part of it.

I'd like to see it also include how `test_image.sh` is intended to be used (ideally with code snippets someone could copy and paste).

ci/docker/test_image.sh
```diff
@@ -0,0 +1,54 @@
...
+  echo '----------------- CLI TEST START ---------------' && \
+  bash python/libcuopt/libcuopt/tests/test_cli.sh && \
+  echo '----------------- CLI TEST END ---------------' && \
+  echo '----------------- CUOPT TEST START ---------------' && \
+  python -m pytest python/cuopt/cuopt/tests/linear_programming && \
+  python -m pytest python/cuopt/cuopt/tests/routing && \
+  echo '----------------- CUOPT TEST END ---------------' && \
+  echo '----------------- CUOPT SERVER TEST START ---------------' && \
+  python -m pytest python/cuopt_server/cuopt_server/tests/ && \
+  echo '----------------- CUOPT SERVER TEST END ---------------'"
```
jameslamb:
```suggestion
cat > ./test.sh <<EOF
cd ~/cuopt
pip install --user pytest pexpect
export PATH=\$PATH:/home/cuopt/.local/bin
export RAPIDS_DATASET_ROOT_DIR=\$(realpath datasets)
echo '----------------- CLI TEST START ---------------'
bash python/libcuopt/libcuopt/tests/test_cli.sh
echo '----------------- CLI TEST END ---------------'
echo '----------------- CUOPT TEST START ---------------'
python -m pytest python/cuopt/cuopt/tests/linear_programming
python -m pytest python/cuopt/cuopt/tests/routing
echo '----------------- CUOPT TEST END ---------------'
echo '----------------- CUOPT SERVER TEST START ---------------'
python -m pytest python/cuopt_server/cuopt_server/tests/
echo '----------------- CUOPT SERVER TEST END ---------------'
EOF

su - cuopt -c "bash ./test.sh"
```

`&&-ing` this together in a large string makes it more difficult to read and modify. It will also make it more difficult to debug when it fails.

I recommend just writing the logic you want out to a script, then executing that script. That way, you'll get a specific line number when something breaks, and you don't need to worry about missing an `&& \`.

.github/workflows/build_images.yaml
```diff
@@ -0,0 +1,102 @@
...
+        uses: docker/build-push-action@v6
+        with:
+          context: ./ci/docker/context
+          file: ./ci/docker/Dockerfile
+          push: true
+          pull: true
+          build-args: |
+            CUDA_VER=${{ inputs.CUDA_VER }}
+            PYTHON_SHORT_VER=${{ steps.trim.outputs.PYTHON_SHORT }}
+            CUOPT_VER=25.8.0a49
```
gforsyth:
```suggestion
            CUOPT_VER=${{ inputs.CUOPT_VER }}
```
this seems like it shouldn't be hardcoded

.github/workflows/build_images.yaml
```diff
@@ -0,0 +1,102 @@
...
+            LINUX_VER=${{ inputs.LINUX_VER }}
+          tags: nvidia/cuopt:${{ inputs.IMAGE_TAG_PREFIX }}-cuda${{ steps.trim.outputs.CUDA_SHORT }}-py${{ steps.trim.outputs.PYTHON_SHORT }}-${{ matrix.ARCH }}
```
gforsyth, line 97: Should this have `CUOPT_VER` somewhere in the tag?

.github/workflows/build_test_publish_images.yaml
```diff
@@ -0,0 +1,170 @@
...
+
+      - name: Compute matrix
+        id: compute-matrix
+        run: |
+          MATRIX=$(jq -c '.' <<EOF
+          {
+            "arch": ["amd64", "arm64"],
+            "cuda_ver": ["12.8.0"],
+            "python_ver": ["3.12.11"],
+            "linux_ver": ["22.04"]
```
gforsyth: These defaults need to be encoded somewhere, but it would be more flexible if they were workflow inputs with default values.

That way you can both change them for one-off runs, and they'll live at the top of the yaml file, which makes it easier to find them.

.github/workflows/build_test_publish_images.yaml
```diff
@@ -0,0 +1,170 @@
...
+          echo "/tmp/gha-tools" >> "${GITHUB_PATH}"
+
+      - name: Compute cuopt version
+        id: compute-cuopt-ver
+        run: |
+          ver=$(rapids-generate-version)
+          CUOPT_VER=$(echo "$ver" | sed -E 's/^v//' \
+            | sed -E 's/\.0+([0-9])/\.\1/g' \
+            | sed -E 's/\.0+([a-zA-Z].*)/.0\1/g' \
+            | sed -E 's/\b0*([0-9]+)\b/\1/g')
```
gforsyth: Are all of these required?  This seems excessive for the output of a script that we control.  I don't think the first `sed` expression is required at all, as `rapids-generate-versino` strips out leading the leading `v`.  And I'm not clear on what the final `sed` pass is aimed at cleaning up.

If they are all required, I think it would be good to add a few comments explaining what each successive pass of `sed` is doing.

.github/workflows/build_test_publish_images.yaml
```diff
@@ -0,0 +1,170 @@
...
+          echo "CUOPT_VER=$CUOPT_VER" >> $GITHUB_OUTPUT
+          if [[ "$CUOPT_VER" == *"a"* ]]; then
+            IMAGE_TAG_PREFIX=$(echo "$CUOPT_VER" | sed -E 's/([0-9]+\.[0-9]+\.[0-9]+)a.*/\1a/')
+          else
+            IMAGE_TAG_PREFIX="$CUOPT_VER"
+          fi
```
gforsyth: This seems like a roundabout way of calling `rapids-is-release-build` -- can you walk me through what you're aiming to accomplish here so we can figure out a less `sed`-heavy approach?

ci/docker/Dockerfile
```diff
@@ -0,0 +1,99 @@
...
+COPY ./THIRD_PARTY_LICENSES /home/cuopt/THIRD_PARTY_LICENSES
+
+FROM install-env AS cuopt-final 
+
+ARG PYTHON_SHORT_VER
+
+# Set environment variables in .bashrc for all future shells
+RUN echo 'export PATH="/usr/local/cuda/bin:/usr/bin:/usr/local/bin:/usr/local/nvidia/bin/:/home/cuopt/.local/lib/python${PYTHON_SHORT_VER}/dist-packages/libcuopt/bin/:/home/cuopt/.local/bin:$PATH"' >> /home/cuopt/.bashrc && \
+    echo 'export LD_LIBRARY_PATH="/usr/lib/x86_64-linux-gnu:/usr/lib/aarch64-linux-gnu:/usr/local/cuda/lib64:/usr/local/nvidia/lib:/usr/local/nvidia/lib64:/usr/lib/wsl/lib:/usr/lib/wsl/lib/libnvidia-container:/usr/lib/nvidia:/usr/lib/nvidia-current:/home/cuopt/.local/lib/python${PYTHON_SHORT_VER}/dist-packages/libcuopt/lib/:/home/cuopt/.local/lib/python${PYTHON_SHORT_VER}/dist-packages/rapids_logger/lib64:${LD_LIBRARY_PATH}"' >> /home/cuopt/.bashrc && \
+    echo 'export PYTHONPATH="/home/cuopt/.local/lib/python${PYTHON_SHORT_VER}/dist-packages:/usr/local/lib/python${PYTHON_SHORT_VER}/dist-packages:${PYTHONPATH:-}"' >> /home/cuopt/.bashrc
```
gforsyth: I could be missing it, but I don't see anywhere in this dockerfile that things should be installed in `~/.local/lib/python...` -- I think that should be dropped from the `PATH` and `LD_LIBRARY_PATH` adjustments.
And then I think the `PYTHONPATH` modifications can be removed entirely

gforsyth (changes requested):
Hi @rgsl888prabhu -- I've made it through most of this and flagged up a few questions and changes -- I can take another look after we clarify a bit more on some of the question areas.

.github/workflows/build_images.yaml
ramakrishnap-nv: Apologies, yeah, I was testing this as PR and might have left this, will update.

.github/workflows/build_images.yaml
ramakrishnap-nv: The IMAGE_TAG prefix is built from it, and would reduce version till 25.8.0a, since we don't want to create a new image everytime.

.github/workflows/build_test_publish_images.yaml
ramakrishnap-nv: Nice idea, will add it

.github/workflows/build_test_publish_images.yaml
ramakrishnap-nv: Will add comments to this

.github/workflows/build_test_publish_images.yaml
ramakrishnap-nv: The container image name should remain same for nightly, so 25.8.0a187 like this should be 25.8.0a, so this mostly strips elements after a.

ramakrishnap-nv: But I think I can use is release build instead of checking for `a`

rgsl888prabhu:
@gforsyth May I get your review on this

.github/workflows/build_test_publish_images.yaml
```diff
@@ -0,0 +1,185 @@
...
+  pull-requests: read
+  repository-projects: none
+  security-events: none
+  statuses: none
+
+jobs:
+  compute-matrix:
+    runs-on: ubuntu-latest
+    container:
+      image: rapidsai/ci-conda:latest
```
gforsyth:
```suggestion
      image: rapidsai/ci-conda:25.08-latest
```

I wouldn't count on `ci-conda:latest` "just working" in the future (although that's the hope).  Better to point at the specific calver tag

.github/workflows/pr.yaml
```diff
@@ -150,16 +151,18 @@ jobs:
...
+    secrets: inherit
+    uses: rapidsai/shared-workflows/.github/workflows/custom-job.yaml@branch-25.08
+    with:
+      build_type: pull-request
+      node_type: "gpu-l4-latest-1"
+      arch: "amd64"
+      file_to_upload: "docs/cuopt/build/html/"
+      artifact-name: "cuopt_docs"
+      container_image: "rapidsai/ci-conda:25.08-latest"
+      script: "ci/build_docs.sh"
```
gforsyth: This seems like an unrelated change -- can we move it to a separate PR?

.github/workflows/test_images.yaml
```diff
@@ -0,0 +1,62 @@
...
+
+          echo "CUDA_SHORT=$CUDA_SHORT" >> $GITHUB_OUTPUT
+          echo "PYTHON_SHORT=$PYTHON_SHORT" >> $GITHUB_OUTPUT
+
+  test:
+    runs-on: "linux-${{ inputs.ARCH }}-gpu-a100-latest-1"
+    needs: prepare
+    container:
+      image: "nvidia/cuopt:${{ inputs.IMAGE_TAG_PREFIX }}-cuda${{ needs.prepare.outputs.CUDA_SHORT }}-py${{ needs.prepare.outputs.PYTHON_SHORT }}"
+      options: --user root
```
gforsyth, line 54: Is it necessary to run this as `root`?

I see that there are a few `apt install` commands in `test_image.sh` but do they need to be there?  `gcc` should already be installed and `bzip2` can be added to the base `Dockerfile`.

Do the dataset download scripts require elevated permissions for some reason?

ci/docker/test_image.sh
```diff
@@ -0,0 +1,59 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -euo pipefail
+
+chsh -s /bin/bash cuopt
+
+# Install dependencies
+apt-get update
+DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommends file bzip2 gcc
```
gforsyth: Alternatively, what do we need `gcc` for here?

ramakrishnap-nv: Think I was using it for some other issue, which might have been resolved and I didn't think of removing it later.

.github/workflows/pr.yaml
ramakrishnap-nv: Yep, I have merged with main branch, this should be fixed.

.github/workflows/test_images.yaml
ramakrishnap-nv: Need root access since the repo gets checked out in /, so without root access it was failing.

.github/workflows/build_test_publish_images.yaml
ramakrishnap-nv: Updated

.github/workflows/test_images.yaml
gforsyth: The `checkout` action takes a `path:` argument, so we can check it out somewhere else.  I don't think this is a good reason to run this container as `root`.  Also, if we're testing the container and the default user is `cuopt`, I'm not sure how good a test this is if we don't run as that user.

gforsyth: we might not have to check it out to a different path, either -- it might just be a matter of `chown`ing the checked out repo to the UID of the `cuopt` user inside the container

ramakrishnap-nv: gcc is being removed in docker since we wanted to keep it light.

I will explore path towards checking out the code in different path and see if that works.

ramakrishnap-nv: If I need to run as non root, may be I need to install these as well within docker

gforsyth: I think we decided we didn't need `gcc` for the tests anyway, yeah?  So it would just be adding `bzip2` to the docker image, which is a pretty light dependency.

ramakrishnap-nv: Yeah, forgot about gcc

ramakrishnap-nv: Created an issue for follow-up on non root testing https://github.com/NVIDIA/cuopt/issues/238

gforsyth (approved):
Thanks for bearing with me during the various iterations @rgsl888prabhu !

rgsl888prabhu:
> Thanks for bearing with me during the various iterations @rgsl888prabhu !

Thanks to you and @jameslamb it is in much better shape.

Issue #189, opened 2025-07-07
[FEA] Add functionality to create a .deb file from cmake using cpack.


mostroot:
**Is your feature request related to a problem? Please describe.**
It would be convenient if cuOpt could create a deb installation package from cmake. 

**Describe the solution you'd like**
Adding in the cpack instructions into the cmake_lists.txt

**Describe alternatives you've considered**
You could manually create a deb package, but that is much more work.

**Additional context**
I have a working implementation I will point my fork's PR to this issue along with instructions of how to create the deb file.

PR #190, opened 2025-07-07
Feat: rework cmakelists.txt to facilitate deb file creation.

mostroot:
## Description
Adds changes to the top level cmake_lists.txt to enable cpack creation of .deb files. In order to create the runtime and develop .deb files you will need to:
1) Build cuOpt
2) Navigate to cpp/build
3) run  cpack -G DEB

This creates two .deb files for the architecture of the machine that cuOpt is built on. 
For example: `cuOpt_25.08.0_amd64-Development.deb` and `cuOpt_25.08.0_amd64-Runtime.deb`

The runtime deb package contains the cuopt CLI, the mps reader.so and the cuopt.so.
The cuopt_CLI is installed in the `/usr/local/bin` directory
The .so files are installed in the `/usr/local/lib` directory

The development deb package contains the mps reader.so, the cuopt.so and the header files for the project all of which are installed in `/usr/local/lib` in a standard fashion. 

The headers are in `/usr/local/lib/cuopt`

Note: There are items in the deb package creation that I took the liberty to fill, some of them may need changes. 
Specifically these two lines:
`set(CPACK_DEBIAN_DEV_PACKAGE_MAINTAINER "NVIDIA")`
`set(CPACK_DEBIAN_RUNTIME_PACKAGE_MAINTAINER "NVIDIA")`
I assumed that is how you'd like the maintainer referenced.

Additionally the deb packages do not correctly reference their dependencies yet, i'm happy to follow up with that work if this looks likely to be merged.

## Issue
#189 

## Checklist

- Testing 
   - [ ] New or existing tests cover these changes (Not added for the cmake changes)
   - [ ] Added tests (Compiling and deb installation works, unsure how you'd like a test for this)
   - [x] NA
- Documentation
   - [ ] Added new documentation ( Happy to add the documentation just need to pointed to where you'd like the cpack instructions)
   - [ ] NA

### DIFF
--- a/cpp/CMakeLists.txt
+++ b/cpp/CMakeLists.txt
@@ -252,16 +252,49 @@ endif(BUILD_TESTS)
 
 # ##################################################################################################
 # - install targets -------------------------------------------------------------------------------
+
+# allows for CPack component builds and install location
+set(CPACK_DEB_COMPONENT_INSTALL ON)
+set(CPACK_COMPONENTS_ALL runtime dev)
+set(CPACK_PACKAGING_INSTALL_PREFIX "/usr/local")
+
+#If using cpack to create a deb package
+if(CPACK_GENERATOR STREQUAL "DEB")
+  set(_BIN_DEST "bin")
+  set(_LIB_DEST "lib")
+  set(_INCLUDE_DEST "lib/cuopt")
+
+#If building locally use the Default install paths(e.g. for local development or other package types)
+else()
+  set(_BIN_DEST "${CMAKE_INSTALL_BINDIR}")
+  set(_LIB_DEST "${lib_dir}")
+  set(_INCLUDE_DEST  include/cuopt/)
+endif()
+
+# adds the .so files to the runtime deb package
 install(TARGETS cuopt mps_parser
-  DESTINATION ${lib_dir}
-  EXPORT cuopt-exports)
+  DESTINATION ${_LIB_DEST}z
+  COMPONENT runtime
+  EXPORT cuopt-exports
+)
 
+# adds the .so files to the development deb package
+install(TARGETS cuopt mps_parser
+  DESTINATION ${_LIB_DEST}
+  COMPONENT dev
+)
+
+# adds the header files to the development deb package
 install(DIRECTORY include/cuopt/
-  DESTINATION include/cuopt)
+  DESTINATION ${_INCLUDE_DEST}
+  COMPONENT dev
+)
 
+# adds the version header file to the development deb package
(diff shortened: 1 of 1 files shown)


rgsl888prabhu:
Thank you @mostroot for the issue and the PR, I wanted to understand the use case of deb package compared to pip wheel.

And also do we expect underlying dependencies to be available as deb package or expect users to install it by themselves.

As per my knowledge, there few dependencies which are header only, but there are dependencies which are in the form of conda or pip wheel package. So, we might not have all dependencies available as .deb package.

mostroot:
@rgsl888prabhu I looked into the dependency tracking for the deb file, in short its rather complicated. I can specify that it depends on a particular cuda-toolkit version in the file, but because of how the Nvidia cuda-toolkit versions are packaged I can't do something simple like: depends on cuda version 11 or greater. 

In short the cuda-toolkit's are literally named: `cuda-toolkit-12-1` or `cuda-toolkit-12-8` cpack sees those as two completely different package names, not two versions of the same name. So I could set very strict requirements of a particular version like 12-8, but it wouldn't be installable on any system with another version of cuda-toolkit installed. 

Because of this, I think the right route is to have this produce a deb file with no dependencies and assume that the user has cuda and the cuda-toolkit installed correctly. It's not perfect, a user could add specific dependencies for a particular version if they wanted their deployed systems to function correctly, or alternitively you could have a lengthy script that creates a ton of different deb files for each specific version of cuda and let the user pick the right one. I could take a stab at that solution, but I think it would be much more messy.

Happy to find some time to talk through it if you'd like. Let me know where you'd like me to take it.

Also @rgsl888prabhu that's a great one piece avatar you have :)

tmckayus (approved):
looks good to me. Is the debian package always being built?

If the debian building fails, does it derail the whole build?

cpp/CMakeLists.txt
```diff
@@ -252,16 +252,49 @@ endif(BUILD_TESTS)
...
+  set(_BIN_DEST "${CMAKE_INSTALL_BINDIR}")
+  set(_LIB_DEST "${lib_dir}")
+  set(_INCLUDE_DEST  include/cuopt/)
+endif()
+
+# adds the .so files to the runtime deb package
 install(TARGETS cuopt mps_parser
-  DESTINATION ${lib_dir}
-  EXPORT cuopt-exports)
+  DESTINATION ${_LIB_DEST}z
```
ramakrishnap-nv: @mostroot Is there a particular need for use lib64z ?

mostroot: good catch, i'll fix the typo

mostroot: z removed

PR #192, opened 2025-07-08
Performance improvements to dual simplex: BFRT, hypersparsity, basis updates, primal infeasible list

chris-maes:
This PR contains multiple performance improvements to the dual simplex code including:

* O(N^2) -> O(N) bound-flipping ratio test
* Exploiting hypersparsity (sparse factors plus sparse right-hand side) throughout the solver,
* New middle-product form basis update (replaces Forrest-Tomlin since it currently handles hypersparsity better)
* Maintaining and update a list of primal infeasibilites

New or existing test cover these changes. No changes in documentation are necessary.

### DIFF
--- a/cpp/src/mip/presolve/bounds_presolve.cu
+++ b/cpp/src/mip/presolve/bounds_presolve.cu
@@ -202,6 +202,7 @@ termination_criterion_t bound_presolve_t<i_t, f_t>::bound_update_loop(problem_t<
   }
   pb.handle_ptr->sync_stream();
   calculate_infeasible_redundant_constraints(pb);
+  solve_iter = iter;
 
   return criteria;
 }
--- a/cpp/src/mip/presolve/bounds_presolve.cuh
+++ b/cpp/src/mip/presolve/bounds_presolve.cuh
@@ -86,6 +86,7 @@ class bound_presolve_t {
   i_t infeas_constraints_count = 0;
   i_t redund_constraints_count = 0;
   probing_cache_t<i_t, f_t> probing_cache;
+  i_t solve_iter;
 };
 
 }  // namespace cuopt::linear_programming::detail
--- a/cpp/src/mip/presolve/load_balanced_bounds_presolve.cu
+++ b/cpp/src/mip/presolve/load_balanced_bounds_presolve.cu
@@ -245,6 +245,7 @@ void load_balanced_bounds_presolve_t<i_t, f_t>::setup(
                                                            heavy_degree_cutoff,
                                                            problem.cnst_bin_offsets,
                                                            problem.offsets);
+  RAFT_CHECK_CUDA(stream_heavy_cnst);
 
   num_blocks_heavy_vars = create_heavy_item_block_segments(stream_heavy_vars,
                                                            heavy_vars_vertex_ids,
@@ -253,49 +254,33 @@ void load_balanced_bounds_presolve_t<i_t, f_t>::setup(
                                                            heavy_degree_cutoff,
                                                            problem.vars_bin_offsets,
                                                            problem.reverse_offsets);
+  RAFT_CHECK_CUDA(stream_heavy_vars);
 
   tmp_act.resize(2 * num_blocks_heavy_cnst, stream_heavy_cnst);
   tmp_bnd.resize(2 * num_blocks_heavy_vars, stream_heavy_vars);
 
-  std::tie(is_cnst_sub_warp_single_bin, cnst_sub_warp_count) = sub_warp_meta(
-    streams.get_stream(), warp_cnst_offsets, warp_cnst_id_offsets, pb->cnst_bin_offsets, 4);
+  std::tie(is_cnst_sub_warp_single_bin, cnst_sub_warp_count) =
+    sub_warp_meta(stream, warp_cnst_offsets, warp_cnst_id_offsets, pb->cnst_bin_offsets, 4);
 
-  std::tie(is_vars_sub_warp_single_bin, vars_sub_warp_count) = sub_warp_meta(
-    streams.get_stream(), warp_vars_offsets, warp_vars_id_offsets, pb->vars_bin_offsets, 4);
+  std::tie(is_vars_sub_warp_single_bin, vars_sub_warp_count) =
+    sub_warp_meta(stream, warp_vars_offsets, warp_vars_id_offsets, pb->vars_bin_offsets, 4);
 
-  stream.synchronize();
-  streams.sync_all_issued();
+  RAFT_CHECK_CUDA(stream);
+  streams.sync_test_all_issued();
 
   if (!calc_slack_erase_inf_cnst_graph_created) {
-    bool erase_inf_cnst                     = true;
-    calc_slack_erase_inf_cnst_graph_created = build_graph(
-      streams,
-      handle_ptr,
-      calc_slack_erase_inf_cnst_graph,
-      calc_slack_erase_inf_cnst_exec,
-      [erase_inf_cnst, this]() { this->calculate_activity_graph(erase_inf_cnst, true); },
-      [erase_inf_cnst, this]() { this->calculate_activity_graph(erase_inf_cnst); });
+    create_constraint_slack_graph(true);
+    calc_slack_erase_inf_cnst_graph_created = true;
   }
 
   if (!calc_slack_graph_created) {
-    bool erase_inf_cnst      = false;
-    calc_slack_graph_created = build_graph(
-      streams,
-      handle_ptr,
-      calc_slack_graph,
-      calc_slack_exec,
-      [erase_inf_cnst, this]() { this->calculate_activity_graph(erase_inf_cnst, true); },
-      [erase_inf_cnst, this]() { this->calculate_activity_graph(erase_inf_cnst); });
+    create_constraint_slack_graph(false);
+    calc_slack_graph_created = true;
   }
 
   if (!upd_bnd_graph_created) {
-    upd_bnd_graph_created = build_graph(
-      streams,
-      handle_ptr,
-      upd_bnd_graph,
-      upd_bnd_exec,
-      [this]() { this->calculate_bounds_update_graph(true); },
-      [this]() { this->calculate_bounds_update_graph(); });
+    create_bounds_update_graph();
+    upd_bnd_graph_created = true;
   }
 }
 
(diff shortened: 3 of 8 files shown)


KyleFromNVIDIA (approved):

cpp/src/dual_simplex/phase2.cpp
```diff
@@ -1227,35 +2298,68 @@ dual::status_t dual_phase2(i_t phase,
...
+  std::vector<i_t> basic_mark(n);
+  std::vector<i_t> delta_z_mark(n, 0);
+  std::vector<i_t> delta_z_indices;
+  std::vector<f_t> v(m, 0.0);
+  std::vector<f_t> squared_infeasibilities;
+  std::vector<i_t> infeasibility_indices;
+
+  phase2::reset_basis_mark(basic_list, nonbasic_list, basic_mark, nonbasic_mark);
+
+  std::vector<bool> bounded_variables(n, false);
```
aliceb-nv: `std::vector<bool>` tends to be somewhat slower (as it's implemented through a bitmap), I'm not sure if this is anything close to a hotspot but for these use cases a std::vector<int> or std::vector<uint8_t> work better

cpp/src/dual_simplex/phase2.cpp
```diff
@@ -1582,30 +2833,42 @@ dual::status_t dual_phase2(i_t phase,
...
+
+#ifdef CHECK_BASIS_MARK
+    phase2::check_basis_mark(settings, basic_list, nonbasic_list, basic_mark, nonbasic_mark);
 #endif
 
     iter++;
 
-    const f_t obj = compute_objective(lp, x);
-    f_t now       = toc(start_time);
+    // TODO(CMM): Do we also need to clear delta_y?
```
aliceb-nv: Has this TODO been addressed?

cpp/src/dual_simplex/basis_updates.cpp
```diff
@@ -55,6 +63,33 @@ i_t basis_update_t<i_t, f_t>::b_solve(const std::vector<f_t>& rhs,
...
+template <typename i_t, typename f_t>
+i_t basis_update_t<i_t, f_t>::b_solve(const sparse_vector_t<i_t, f_t>& rhs,
+                                      sparse_vector_t<i_t, f_t>& solution,
+                                      sparse_vector_t<i_t, f_t>& Lsol) const
+{
+  const i_t m = L0_.m;
+  assert(row_permutation_.size() == m);
+  assert(rhs.n == m);
+  assert(solution.n == m);
+  assert(Lsol.n == m);
```
aliceb-nv, line 75: Those asserts will only be compiled when building in debug mode which we rarely ever do, we might want to use cuopt_assert instead (which is controlled by the "-a" flag in ./build.sh)

cpp/src/dual_simplex/basis_updates.cpp
```diff
@@ -617,8 +1116,932 @@ i_t basis_update_t<i_t, f_t>::lower_triangular_multiply(const csc_matrix_t<i_t,
...
+  // u = utilde - U0(:, leaving_index)
+  for (i_t p = col_start; p < col_end; ++p) {
+    const i_t i = U0_.i[p];
+    u[i] -= U0_.x[p];
+  }
+
+  i_t u_nz = 0;
+  for (i_t i = 0; i < m; ++i) {
+    if (u[i] != 0.0) { u_nz++; }
+  }
```
aliceb-nv: Just FYI (no need to change as it's minor really): can be replaced by std::count/std::count_if in `<algorithm>`

cpp/src/dual_simplex/basis_updates.cpp
```diff
@@ -617,8 +1116,932 @@ i_t basis_update_t<i_t, f_t>::lower_triangular_multiply(const csc_matrix_t<i_t,
...
+  i_t S_nz_start = S_nz;
+
+  // Scatter u into S
+  for (i_t i = 0; i < m; ++i) {
+    if (u[i] != 0.0) {
+      S_.i[S_nz] = i;
+      S_.x[S_nz] = u[i];
+      S_nz++;
+    }
+  }
```
aliceb-nv: Implementing common util functions for scatter/gather/dot could help streamline the code / make it easier to reason with

cpp/src/dual_simplex/basis_updates.cpp
```diff
@@ -617,8 +1116,932 @@ i_t basis_update_t<i_t, f_t>::lower_triangular_multiply(const csc_matrix_t<i_t,
...
+    std::vector<f_t> Uj(m, 0.0);
+    for (i_t p = U0_.col_start[j]; p < U0_.col_start[j + 1]; ++p) {
+      const i_t i = U0_.i[p];
+      Uj[i]       = U0_.x[p];
+    }
+    l_multiply(Uj);
+    for (i_t i = 0; i < m; ++i) {
+      if (Uj[i] != 0.0) {
+        out.i.push_back(row_permutation_[i]);
+        out.x.push_back(Uj[i]);
```
aliceb-nv, line 2026: We might want to .reserve() `out.i` and `out.x` since those `push_back` are in a tight loop

cpp/src/dual_simplex/basis_updates.hpp
```diff
@@ -85,6 +119,213 @@ class basis_update_t {
...
+
+ private:
+  void clear()
+  {
+    pivot_indices_.clear();
+    pivot_indices_.reserve(L0_.m);
+    for (i_t k = 0; k < L0_.m; ++k) {
+      col_permutation_[k]         = k;
+      inverse_col_permutation_[k] = k;
+    }
```
aliceb-nv: Another minor FYI: std::iota in `<algorithm>` does this :)

cpp/src/dual_simplex/phase2.cpp
```diff
@@ -34,17 +36,465 @@ namespace cuopt::linear_programming::dual_simplex {
...
+  const i_t m           = lp.num_rows;
+  const i_t n           = lp.num_cols;
+  f_t max_abs_obj_coeff = 0.0;
+  for (i_t j = 0; j < n; ++j) {
+    max_abs_obj_coeff = std::max(max_abs_obj_coeff, std::abs(lp.objective[j]));
+  }
+
+  const f_t dual_tol = settings.dual_tol;
+
+  std::srand(static_cast<unsigned int>(std::time(nullptr)));
```
aliceb-nv, line 240: Let's use a deterministic seed for this: `cuopt::seed_generator::get_seed()` provides one

cpp/src/dual_simplex/phase2.cpp
```diff
@@ -34,17 +36,465 @@ namespace cuopt::linear_programming::dual_simplex {
...
+    if (std::abs(delta_y_i) < 1e-12) { continue; }
+    const i_t row_start = A_transpose.col_start[i];
+    const i_t row_end   = A_transpose.col_start[i + 1];
+    for (i_t p = row_start; p < row_end; ++p) {
+      const i_t j = A_transpose.i[p];
+      if (nonbasic_mark[j] >= 0) {
+        delta_z[j] -= delta_y_i * A_transpose.x[p];
+        if (!delta_z_mark[j]) {
+          delta_z_mark[j] = 1;
+          delta_z_indices.push_back(j);
```
aliceb-nv: Unsure how much of a hotspot this is, but we might want to `.reserve()` delta_z_indices as well

cpp/src/dual_simplex/phase2.cpp
```diff
@@ -101,6 +580,223 @@ void compute_dual_solution_from_basis(const lp_problem_t<i_t, f_t>& lp,
...
+                                   const std::vector<i_t>& basic_list,
+                                   const std::vector<f_t>& x,
+                                   i_t entering_index,
+                                   i_t leaving_index,
+                                   std::vector<i_t>& basic_change_list,
+                                   std::vector<f_t>& squared_infeasibilities,
+                                   std::vector<i_t>& infeasibility_indices,
+                                   f_t& primal_inf)
+{
+  const f_t now_feasible = std::numeric_limits<f_t>::denorm_min();
```
aliceb-nv: Are denormals useful here? I think `double` should have enough precision, computations involving denormals usually imply a performance hit
As an aside: we might want to measure the performance/quality impact of flushing denormals to zero at some point for the CPU code, just in case

cpp/src/dual_simplex/phase2.cpp
```diff
@@ -101,6 +580,223 @@ void compute_dual_solution_from_basis(const lp_problem_t<i_t, f_t>& lp,
...
+                                       j,
+                                       primal_inf);
+  }
+}
+
+template <typename i_t, typename f_t>
+void clean_up_infeasibilities(std::vector<f_t>& squared_infeasibilities,
+                              std::vector<i_t>& infeasibility_indices)
+{
+  const f_t now_feasible = std::numeric_limits<f_t>::denorm_min();
```
aliceb-nv: ditto

cpp/src/dual_simplex/phase2.cpp
```diff
@@ -1356,52 +2465,161 @@ dual::status_t dual_phase2(i_t phase,
...
+        constexpr bool use_farkas = false;
+        if constexpr (use_farkas) {
+          std::vector<f_t> farkas_y;
+          std::vector<f_t> farkas_zl;
+          std::vector<f_t> farkas_zu;
+          f_t farkas_constant;
+          std::vector<f_t> my_delta_y;
+          delta_y_sparse.to_dense(my_delta_y);
+
+          // TODO(CMM): Do I use the perturbed or unperturbed objective?
```
aliceb-nv, line 2616: Has this TODO been addressed?

cpp/src/dual_simplex/phase2.cpp
```diff
@@ -1534,10 +2739,65 @@ dual::status_t dual_phase2(i_t phase,
     }
 #endif
 
+    timers.start_timer();
+    // TODO(CMM): Do I also need to update the objective due to the bound flips?
+    // TODO(CMM): I'm using the unperturbed objective here, should this be the perturbed objective?
```
aliceb-nv, line 2767: Has this TODO been addressed?

cpp/src/dual_simplex/sparse_vector.cpp
```diff
@@ -0,0 +1,226 @@
...
+    }
+    i = i_sorted;
+    x = x_sorted;
+  }
+
+  // Check
+#ifdef CHECK_SORT
+  for (i_t k = 0; k < i.size() - 1; ++k) {
+    if (i[k] > i[k + 1]) { printf("Sort error %d %d\n", i[k], i[k + 1]); }
+  }
```
aliceb-nv: Just another tiny FYI: `std::is_sorted` is useful for this (in asserts and similar)

aliceb-nv (commented):
Thanks a lot for the awesome work Chris!
Just a few minor comments :)

rgsl888prabhu (approved):

cpp/src/dual_simplex/phase2.cpp
chris-maes: Ah yes. Good catch. The bitmap might slow down the access. I switched to uint8_t as suggested.

cpp/src/dual_simplex/phase2.cpp
chris-maes: Yes. This can be a hot spot. Nice catch! I added the reserve when I create `delta_z_indices` and left a comment in this loop. Thank you!

cpp/src/dual_simplex/phase2.cpp
chris-maes: I'm not using the denormals for computation here. I was using this small value as a marker in the `squared_infeasibilities` array. If `squared_infeasibilites[j] == now_feasible`, than variable `j` has become feasible, and it can be removed from the `infeasibility_indicies` list.  I think it is fine to just set the value to 0.0.

cpp/src/dual_simplex/phase2.cpp
chris-maes: Done

cpp/src/dual_simplex/basis_updates.cpp
chris-maes: Thanks for the recommendation. I implemented these functions to streamline the code.

cpp/src/dual_simplex/basis_updates.cpp
chris-maes: I moved this to a utility function.

cpp/src/dual_simplex/basis_updates.hpp
chris-maes: I switched to `std::iota`. Thanks

cpp/src/dual_simplex/basis_updates.cpp
chris-maes: I would reserve, but I don't know how many nonzeros are in the product of L and U up front. Note that this is just debug code and not performance critical. So I thought I'd might just let the dynamic array handle it.

cpp/src/dual_simplex/phase2.cpp
chris-maes: Removed denormals. See above.

cpp/src/dual_simplex/phase2.cpp
chris-maes: Yes, it has. Removed the TODO.

cpp/src/dual_simplex/phase2.cpp
chris-maes: Unfortunately, no. I'm still not sure which to use here. Right now the objective value is just for printing in the logs.  So it doesn't affect the solution.

cpp/src/dual_simplex/phase2.cpp
chris-maes: Unfortunately, no. Similar to the above I'm not sure what to do here. Luckily,  it only affects the objective in the logs.

cpp/src/dual_simplex/sparse_vector.cpp
chris-maes: Thanks. I switched to `std::is_sorted`.

aliceb-nv (approved):
Approved! Thanks a lot for the great work Chris!

tmckayus:
This is not critical but is highly, highly desirable from a perception basis

PR #199, opened 2025-07-11
add a simple smoketest for cvxpy integration

tmckayus:
This is a simple smoketest to look for breakage and any abnormal behavior with respect to runtime from cuopt Python API changes. It does not do any exhaustive performance testing or building.

depends on #235

### DIFF
--- a/cpp/src/mip/problem/problem.cu
+++ b/cpp/src/mip/problem/problem.cu
@@ -1208,9 +1208,6 @@ void problem_t<i_t, f_t>::remove_given_variables(problem_t<i_t, f_t>& original_p
 {
   thrust::fill(handle_ptr->get_thrust_policy(), offsets.begin(), offsets.end(), 0);
   cuopt_assert(assignment.size() == n_variables, "Variable size mismatch");
-  std::cout << "variable_map size " << variable_map.size() << std::endl;
-  std::cout << "n_variables " << n_variables << std::endl;
-  printf("variable_map size %zu n_variables %d\n", variable_map.size(), n_variables);
   cuopt_assert(variable_map.size() < n_variables, "Too many variables to fix");
   rmm::device_uvector<f_t> tmp_assignment(assignment, handle_ptr->get_stream());
 


ci/external/cvxpy_smoketest.sh
```diff
@@ -0,0 +1,31 @@
...
+pip install -e .
+python -m pytest --error-for-skips cvxpy/tests/test_conic_solvers.py -k "TestCUOPT"
+EXITCODE="$?"
+if [ "$EXITCODE" -eq 0 ]; then
+    echo PASSED smoketest
+else
+    echo FAILED smoketest
+fi
+popd || exit 1
+exit $EXITCODE
```
jameslamb:
```suggestion
set -e -u -o pipefail

git clone -b cuopt_solver https://github.com/tmckayus/cvxpy
pushd ./cvxpy
pip install pytest-error-for-skips
pip install -e .
python -m pytest --error-for-skips cvxpy/tests/test_conic_solvers.py -k "TestCUOPT"
popd
```

If you use `set -e`, you could simplify this by removing all the exit-code handling. I recommend doing that.

That explicit exit-code handling can be helpful it you want to do more work after the first failure, but I don't think that's the case here.

ci/test_wheel_cuopt.sh
```diff
@@ -59,3 +59,4 @@ timeout 10m bash ./python/libcuopt/libcuopt/tests/test_cli.sh
 # Run Python tests
 RAPIDS_DATASET_ROOT_DIR=./datasets timeout 30m python -m pytest --verbose --capture=no ./python/cuopt/cuopt/tests/
 
+timeout 3m ./ci/external/cvxpy_smoketest.sh
```
jameslamb:
```suggestion
timeout 3m ./ci/external/cvxpy_smoketest.sh
```

3 minutes is a REALLY tight timeout for a script that includes network calls over the internet like cloning a git repo and installing packages from `pypi.org`.

I recommend either increasing this to something like `30m` or, if the goal is just to ensure that test code runs quickly, moving the `timeout` down into `cvxpy_smoketest.sh`.

jameslamb (dismissed):
Left some comments for your consideration.

ci/test_wheel_cuopt.sh
jameslamb: I should also mention, in the description you wrote:

> It does not do any performance testing or building.

Putting such a restrictive timeout on the test run is a form of "performance testing". `cuopt` or `cvxpy` getting much slower would cause that timeout to be exceeded and the job to fail. So I think that statement in the PR description is not quite correct.

ci/external/cvxpy_smoketest.sh
```diff
@@ -0,0 +1,31 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+git clone -b cuopt_solver https://github.com/tmckayus/cvxpy
+pushd cvxpy || exit 1
+pip install pytest-error-for-skips
+pip install -e .
```
jameslamb:
```suggestion
pip install -e '.[CUOPT]'
```

Ah sorry, I should have looked more closely before submitting my review. Looking at your branch (https://github.com/tmckayus/cvxpy/blob/6cbf0aaf1caf49699bf80e0d5a5580fd73760874/pyproject.toml#L72), I see a problem with this approach.

This call to `pip install -e .` will not be able to catch issues like "`cuopt-cu12` and `cvxpy[CUOPT]` have conflicting dependencies". Here, you're allowing `pip` to (for example) upgrade `scipy` in the environment.

To catch issues like that, I recommend the following:

1. switch to `pip install -e '.[CUOPT]'`, to test that the `CUOPT` extra is working
2. on your branch, update to `cuopt`  and `cuopt-sh-client` 25.08 in `cvxpy`'s `pyproject.toml`
3. ensure that the just-downloaded-from-this-CI-run `cuopt-cu12` and `cuopt-sh-client` are used, via something like this: https://github.com/NVIDIA/cuopt/blob/a21dbac5bf8bc319f1a8875ed459bd3a041879c9/ci/test_wheel_cuopt.sh#L29-L33

If you agree with this general suggestion but aren't sure how to implement it, let me know if you'd be ok with me pushing a commit to your branch here. I'd be happy to do that.

ramakrishnap-nv: @jameslamb yeah, please feel free to push updates to this PR. @tmckayus is currently OOO.

ci/test_wheel_cuopt.sh
ramakrishnap-nv: And I will also address other issues you pointed out on 3m run-time. I think @tmckayus added 3m time line similar to our other test to kind trigger a warning in case somethings suspiciously slow, or someone added too many tests. I think 3m is a good amount of time for a smoke test in cuOpt.

ramakrishnap-nv: I will update the PR description.

jameslamb: > I think 3m is a good amount of time for a smoke test in cuOpt

That's fine, but I think it'd be better to apply it to **the tests** instead of `{git clone} + {pip install} + {tests}`. Like this:

```shell
timeout 3m python -m pytest ...
```

(just as in the line right above it)

ramakrishnap-nv: I agree, let me move it just for test, I missed that.

ci/external/cvxpy_smoketest.sh
jameslamb: Ok thanks! I just pushed commits that implement my suggestions.

Specifically:

* ensures that all the just-built-in-CI packages are used (including `cuopt-sh-client`, which is a dependency of `cvxpy[CUOPT]` on that branch)
* made the 3-minute timeout only apply to the tests, not cloning, building, and installing `cvxpy`
* removed the error-trapping (by just consolidating this code into `test_wheel_cuopt.sh`)

ci/test_wheel_cuopt.sh
jameslamb: I did this as part of these changes: https://github.com/NVIDIA/cuopt/pull/199#discussion_r2206124291

ci/external/cvxpy_smoketest.sh
jameslamb: I did this as part of these changes: https://github.com/NVIDIA/cuopt/pull/199#discussion_r2206124291

ramakrishnap-nv: Thank you @jameslamb

ci/test_wheel_cuopt.sh
ramakrishnap-nv: Thank you @jameslamb

ci/test_wheel_cuopt.sh
```diff
@@ -59,4 +59,29 @@ timeout 10m bash ./python/libcuopt/libcuopt/tests/test_cli.sh
 # Run Python tests
 RAPIDS_DATASET_ROOT_DIR=./datasets timeout 30m python -m pytest --verbose --capture=no ./python/cuopt/cuopt/tests/
 
-timeout 3m ./ci/external/cvxpy_smoketest.sh
+# run cvxpy integration tests
```
ramakrishnap-nv, line 76: @jameslamb We would have more integration tests in future, should we move few items to be partly into a bash script and run it from there to keep the test_wheel_cuopt.sh simple ?

jameslamb: Sure we can, and to be fair that's how @tmckayus had it originally before I pushed commits here.

Let me tell you why I did it this way. I was mainly concerned about making sure that the `pip install` that installs `cvxpy` uses the just-built-in-this-CI-run `cuopt` packages.

Doing it in 1 script like this makes that easy to guarantee, because I can just reference variables from the same scope and packages downloaded a few lines higher up. Like this:

```shell
CUOPT_SH_CLIENT_WHEELHOUSE=$(RAPIDS_PY_WHEEL_NAME="cuopt_sh_client" rapids-download-wheels-from-github python)
python -m pip install \
    --extra-index-url=https://pypi.nvidia.com \
    "${CUOPT_MPS_PARSER_WHEELHOUSE}"/cuopt_mps_parser*.whl \
    "${CUOPT_SH_CLIENT_WHEELHOUSE}"/cuopt_sh_client*.whl \
    "${CUOPT_WHEELHOUSE}"/cuopt*.whl \
    "${LIBCUOPT_WHEELHOUSE}"/libcuopt*.whl \
    'pytest-error-for-skips>=2.0.2' \
    "$(echo ./dist/cvxpy*.whl)[CUOPT,testing]"
```

If you want to defer all the `cvxpy` stuff to a separate script, like Trevor had it before, it'd involve some changes to how packages are downloaded.

I can do that here.

ramakrishnap-nv: Since we might be adding few more smoke tests to this, I felt the script might become too huge and it might not be scaled. Yeah, lets move these to a folder "thirdparty-testing" and add a script there to trigger test. Please let me know if your hands are full, I can work on this.

jameslamb: Ok, I will push changes here.

jameslamb: I just pushed changes moving `cvxpy` tests to their own file, as @tmckayus had originally (but using the name `thirdparty-testing/` you asked for).

Hopefully this will set up a pattern that can be reliably replicated when other integration tests like this are added.

I will check back in an hour or so to be sure this worked. I especially want to see evidence in the logs that the just-built-in-CI `cuopt`, `cuopt-mps-parser`, etc. wheels were used, not packages downloaded from the nightly index.

ramakrishnap-nv: Thank you @jameslamb

jameslamb:
I've removed my blocking review. Sorry @rgsl888prabhu , I wasn't able to get this done today, and I'll be unavailable all of next week. Hopefully I've written enough here to explain my concerns with the original approach and show the direction I want to go.

Talk with others in `@rapidsai/ci-codeowners` next week like @msarahan and @gforsyth and someone should be able to help you. Otherwise, if it's not time-sensitive and you want my input, I'll be available to help again in a week.

rgsl888prabhu:
Python tests are failing in cvxpy because of this issue https://github.com/NVIDIA/cuopt/issues/233

rgsl888prabhu:
@jameslamb May I get your review on this PR

ci/thirdparty-testing/run_cvxpy_tests.sh
```diff
@@ -0,0 +1,41 @@
...
+
+# NOTE: installing cvxpy[CUOPT] alongside CI artifacts is helpful to catch dependency conflicts
+echo "installing 'cvxpy' with cuopt"
+python -m pip install \
+    --constraint "${PIP_CONSTRAINT}" \
+    --extra-index-url=https://pypi.nvidia.com \
+    --extra-index-url=https://pypi.anaconda.org/rapidsai-wheels-nightly/simple \
+    'pytest-error-for-skips>=2.0.2' \
+    "$(echo ./dist/cvxpy*.whl)[CUOPT,testing]"
+
```
jameslamb, line 34:
```suggestion
    "$(echo ./dist/cvxpy*.whl)[CUOPT,testing]"

# ensure that environment is still consistent (i.e. cvxpy requirements do not conflict with cuopt's)
pip check

```

Because we're now back to doing a second `pip install` later in this workflow run (after `cuopt` has already been installed), I recommend adding a `pip check` here. This will return a non-0 exit code (failing the workflow run) if any requirements are not met, e.g. if the already-installed `cuopt` depends on `scipy > 2` and `cvxpy` depends on `scipy <2` (just an example).

I should have added that from the beginning, didn't think of it until now. Will you please consider that?

rgsl888prabhu:
> This is looking good! Checked the logs, confirmed that the just-built-in-CI `cuopt` packages are being installed, so this really is testing that the work with `cvxpy`:
> 
> ```
> Processing ./dist/cvxpy-1.8.0-cp310-cp310-linux_x86_64.whl (from cvxpy==1.8.0)
> ...
> Processing /tmp/tmp.YgYbfzhLn4/cuopt_cu12-25.8.0a93-cp310-cp310-manylinux_2_24_x86_64.manylinux_2_28_x86_64.whl (from cvxpy==1.8.0->cvxpy==1.8.0)
> Requirement already satisfied: nvidia-cuda-runtime-cu12<13.0,>=12.8 in /pyenv/versions/3.10.18/lib/python3.10/site-packages (from cvxpy==1.8.0->cvxpy==1.8.0) (12.9.79)
> ...
> Requirement already satisfied: cudf-cu12==25.8.*,>=0.0.0a0 in /pyenv/versions/3.10.18/lib/python3.10/site-packages (from cuopt-cu12>=25.5->cvxpy==1.8.0->cvxpy==1.8.0) (25.8.0a583)
> Processing /tmp/tmp.qP4l2LcWH8/cuopt_mps_parser-25.8.0a93-cp310-cp310-manylinux_2_24_x86_64.manylinux_2_28_x86_64.whl (from cuopt-cu12>=25.5->cvxpy==1.8.0->cvxpy==1.8.0)
> ...
> cuopt-cu12 is already installed with the same version as the provided wheel. Use --force-reinstall to force an installation of the wheel.
> cuopt-mps-parser is already installed with the same version as the provided wheel. Use --force-reinstall to force an installation of the wheel.
> libcuopt-cu12 is already installed with the same version as the provided wheel. Use --force-reinstall to force an installation of the wheel.
> ...
> Installing collected packages: scipy, pycparser, joblib, scs, osqp, hypothesis, cffi, pytest-error-for-skips, clarabel, cvxpy
> 
> Successfully installed cffi-1.17.1 clarabel-0.11.1 cvxpy-1.8.0 hypothesis-6.136.6 joblib-1.5.1 osqp-1.0.4 pycparser-2.22 pytest-error-for-skips-2.0.2 scipy-1.15.3 scs-3.2.7.post2
> ```
> 
> ([build link](https://github.com/NVIDIA/cuopt/actions/runs/16627684149/job/47050317977?pr=199#step:11:2207))
> 
> Notice there that `cuopt` packages don't appear in the "installed" list at the end of that, but instead in earlier log messages saying that the already-installed packages will be used.
> 
> I left one small suggestion, do whatever you want with it and I do not need to re-approve.

Thank you @jameslamb, @tmckayus most of the work was completed by you

Issue #187, opened 2025-07-03
[BUG] Parameter changes are sent to stdout even though CUOPT_LOG_TO_CONSOLE is false


chris-maes:
We call CUOPT_LOG_INFO when changing parameter settings inside solver settings. We should check the value of LOG_TO_CONSOLE before doing this. Or ensure that the logger is correctly setup.

PR #201, opened 2025-07-14
Fix logging levels

akifcorduk:
## Description

This fixes an issue of formatting when writing to a log file, it was always writing with the default format regardless of the log level. Whereas, for builds on debug and higher levels we want timestamps.

### DIFF
--- a/cpp/CMakeLists.txt
+++ b/cpp/CMakeLists.txt
@@ -96,12 +96,9 @@ set(LIBCUOPT_LOGGING_LEVEL
   CACHE STRING "Choose the logging level."
 )
 set_property(
-  CACHE LIBCUOPT_LOGGING_LEVEL PROPERTY STRINGS "TRACE" "DEBUG" "INFO" "WARN" "ERROR" "CRITICAL"
-                                       "OFF"
+  CACHE LIBCUOPT_LOGGING_LEVEL PROPERTY STRINGS "OFF" "CRITICAL" "ERROR"  "WARN" "INFO" "DEBUG" "TRACE"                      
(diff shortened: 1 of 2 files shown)


cpp/CMakeLists.txt
```diff
@@ -96,13 +96,10 @@ set(LIBCUOPT_LOGGING_LEVEL
   CACHE STRING "Choose the logging level."
 )
 set_property(
-  CACHE LIBCUOPT_LOGGING_LEVEL PROPERTY STRINGS "TRACE" "DEBUG" "INFO" "WARN" "ERROR" "CRITICAL"
-                                       "OFF"
+  CACHE LIBCUOPT_LOGGING_LEVEL PROPERTY STRINGS "OFF" "CRITICAL" "ERROR"  "WARN" "INFO" "DEBUG" "TRACE"
```
aliceb-nv: The order of the strings shouldn't matter right? Since this gets translated into a `CUOPT_LOG_ACTIVE_LEVEL=RAPIDS_LOGGER_LOG_LEVEL_${LIBCUOPT_LOGGING_LEVEL}` definition

aliceb-nv (approved):

cpp/CMakeLists.txt
akifcorduk: you are right, it was the pattern setting in the init_logger. i misinterpreted it. reverted, thank!

rg20:
@akifcorduk  there is a logging related issue filed (https://github.com/NVIDIA/cuopt/issues/187). May be worth combining?

PR #203, opened 2025-07-14
Enable doc build workflow and build script for PR and Nightly

### LINKED ISSUE #63: [FEA] Add a job to build doc in CI
Add a new job to build documentation in CI, currently this is disabled.

rgsl888prabhu:
## Description
Enabling doc build for PR and NIghtly workflow, this was disabled earlier since the documentation couldn't be built in github. But with OSS, documentation has changed.

## Issue
closes #63 

## Checklist

- Testing 
   - [ ] Added tests
   - [x] NA
- Documentation
   - [ ] Added new documentation 
   - [x] NA


ci/build_docs.sh
```diff
@@ -19,23 +20,43 @@ set -euo pipefail
...
+./build.sh docs
+
+mkdir -p "${RAPIDS_DOCS_DIR}/cudf/html"
+
+mv docs/cuopt/build/html/* "${RAPIDS_DOCS_DIR}/cuopt/html"
+
+exit ${EXITCODE}
+
+
+
```
jameslamb:
```suggestion
rapids-logger "Build Docs"

./build.sh docs

mkdir -p "${RAPIDS_DOCS_DIR}/cuopt/html"

mv docs/cuopt/build/html/* "${RAPIDS_DOCS_DIR}/cuopt/html"
```

I recommend getting rid of this exit code trapping. It means that, for example, if `build.sh docs` failed, you're still having the script try those following `mkdir` and `mv` commands, which is not what I think you want.

Just relying on the exit statuses + `set -e` is simpler and will make CI failures easier to debug.

Also, I think you meant `cuopt/html`, not `cudf/html`.

dependencies.yaml
```diff
@@ -79,6 +79,11 @@ files:
     includes:
         - cuda
         - cuda_version
+        - depends_on_libcuopt
+        - depends_on_cuopt
+        - depends_on_cuotp_server
```
jameslamb:
```suggestion
        - depends_on_cuopt_server
```

dependencies.yaml
```diff
@@ -464,6 +469,35 @@ dependencies:
...
+        matrices:
+          - matrix:
+              cuda: "12.*"
+              cuda_suffixed: "true"
+            packages:
+              - cuopt-server-cu12==25.8.*,>=0.0.0a0
+          - {matrix: null, packages: [*cuopt_server_unsuffixed]}
+  depends_on_cuopt_sh_client:
+    common:
+      - output_types: conda
```
jameslamb:
```suggestion
      - output_types: [conda, pyproject, requirements]
```

As of now, this `depends_on_cuopt_sh_client` will not return `cuopt-sh-client` for `pyproject` or `requirements`. If that's an intentional choice because it's only used for docs builds  today and docs builds are only done with conda packages, then remove this from it:

```text
      - output_types: requirements
        packages:
          # pip recognizes the index as a global option for the requirements.txt file
          - --extra-index-url=https://pypi.nvidia.com
          - --extra-index-url=https://pypi.anaconda.org/rapidsai-wheels-nightly/simp
```

Otherwise, if you want this to work for `pip install` too, then please accept my inline suggestion here.

Let me know if this is unclear or you have any questions.

.github/workflows/build.yaml
```diff
@@ -152,20 +152,20 @@ jobs:
...
+    secrets: inherit
+    uses: rapidsai/shared-workflows/.github/workflows/custom-job.yaml@branch-25.08
+    with:
+      build_type: ${{ inputs.build_type || 'branch' }}
+      node_type: "gpu-l4-latest-1"
+      sha: ${{ inputs.sha }}
+      branch: ${{ inputs.branch }}
+      date: ${{ inputs.date }}
+      arch: "amd64"
+      container_image: "rapidsai/ci-conda:latest"
```
jameslamb:
```suggestion
      container_image: "rapidsai/ci-conda:25.08-latest"
```

We now have RAPIDS-versioned tags for these images: https://github.com/rapidsai/ci-imgs/pull/287

Since `cuopt` always builds against a single RAPIDS version, I think it'd make sense to start using those tags. That way, when RAPIDS 25.10 development starts, breaking changes in these images won't affect `cuopt` 25.08.

If you agree with this, please:

* look for other `:latest` uses and consider switching them to the RAPIDS-versioned tags
* update https://github.com/NVIDIA/cuopt/blob/branch-25.08/ci/release/update-version-rapids.sh accordingly

.github/workflows/pr.yaml
```diff
@@ -135,16 +136,16 @@ jobs:
...
-  #    run_script: ci/build_docs.sh
+  docs-build:
+    needs: conda-python-build
+    secrets: inherit
+    uses: rapidsai/shared-workflows/.github/workflows/custom-job.yaml@branch-25.08
+    with:
+      build_type: pull-request
+      node_type: "gpu-l4-latest-1"
+      arch: "amd64" 
+      container_image: "rapidsai/ci-conda:latest"
```
```suggestion
      container_image: "rapidsai/ci-conda:25.08-latest"
```

jameslamb (changes requested):
Left some suggestions for your consideration. There are enough that I'm leaving a blocking review.

rgsl888prabhu:
@jameslamb may I get another set of reviews on this ?

.github/workflows/build.yaml
```diff
@@ -152,20 +152,22 @@ jobs:
...
-  #    build_type: ${{ inputs.build_type || 'branch' }}
-  #    node_type: "gpu-l4-latest-1"
-  #    sha: ${{ inputs.sha }}
-  #    branch: ${{ inputs.branch }}
-  #    date: ${{ inputs.date }}
-  #    arch: amd64
-  #    container_image: rapidsai/ci-conda:cuda11.8.0-ubuntu22.04-py3.10
-  #    run_script: ci/build_docs.sh
+  docs-build:
+    if: inputs.build_type == 'nightly' || github.ref_type == 'branch'
```
jameslamb: This will (I think) prevent re-building the docs when a new tag is pushed. Is that what you want?

If you're not sure why this is here and it was just copied from somewhere else, I recommend just removing it.

ci/build_docs.sh
```diff
@@ -16,26 +17,41 @@
...
-pushd docs/cuopt
+rapids-print-env
+
+rapids-logger "Build Docs"
+./build.sh docs
+mkdir -p "${RAPIDS_DOCS_DIR}/cuopt"
+zip -r /tmp/cuopt_docs.zip docs/cuopt/build/html/*
+cp /tmp/cuopt_docs.zip "${RAPIDS_DOCS_DIR}/cuopt/cuopt_docs.zip"
+
+
```
jameslamb:
```suggestion
./build.sh docs
mkdir -p "${RAPIDS_DOCS_DIR}/cuopt"
```

This `file_to_upload` input to `shared-workflows/custom-job.yaml` is passed to the `path:` argument of `actions/upload-artifact`.

https://github.com/rapidsai/shared-workflows/blob/5973ed6605aba3917ed48e7b578a1c523497a3e8/.github/workflows/custom-job.yaml#L167

So the name is a little misleading... it does not actually have to be a **file**... it could be a directory or glob pattern: https://github.com/actions/upload-artifact?tab=readme-ov-file#inputs. Therefore, you don't have to produce a single file by manually `zip`-ing here.

As a result of `zip`-ing this yourself in the CI job, you're producing an artifact here that is double-zipped... a zip archive containing a zip archive of the docs.

I think that should be avoided.

My recommendations:

1. use `file_to_upload: docs/cuopt/build/html/` in the workflow calls for docs builds
2. have `build_docs.sh` write to `docs/cuopt/build/html/`
3. remove everything about `RAPIDS_DOCS_DIR` in `build_docs.sh`
  - *that environment variable only has special meaning for `rapids-upload-docs` ([code link](https://github.com/rapidsai/gha-tools/blob/bc847834b466fbb6da0b10e25ae65f41e61fde4d/tools/rapids-upload-docs)), which uploads to S3 and which you are not using*
4. remove `zip` stuff from `build_docs.sh`

Could you try that please?

ci/build_docs.sh
```diff
@@ -16,26 +17,41 @@
 
 set -euo pipefail
 
+# install zip
+
+apt-get update && apt-get install -y --no-install-recommends zip
+
```
jameslamb:
```suggestion
```

If my other suggestion below is accepted, installing `zip` won't be necessary.

.github/workflows/build.yaml
ramakrishnap-nv: Think it was just a copy paste.

jameslamb:
removed my blocking review, for the reasons mentioned in https://github.com/NVIDIA/cuopt/pull/199#issuecomment-3091042554

Can help when I'm back if no one gets to it sooner, sorry.

gforsyth (changes requested):
Looks good overall, @rgsl888prabhu !

Can you make this addition to the `update-version-rapids.sh` script?  That will handle auto-updating the calver-tagged docker images:

```diff
diff --git a/ci/release/update-version-rapids.sh b/ci/release/update-version-rapids.sh
index 6835e25..72f3e7c 100755
--- a/ci/release/update-version-rapids.sh
+++ b/ci/release/update-version-rapids.sh
@@ -53,6 +53,8 @@ done
 # WORKFLOWS
 for FILE in .github/workflows/*.yaml; do
   sed_runner "/shared-workflows/ s/@.*/@branch-${NEXT_SHORT_TAG}/g" "${FILE}"
+  # CI image tags of the form {rapids_version}-{something}
+  sed_runner "s/:[0-9]*\\.[0-9]*-/:${NEXT_SHORT_TAG}-/g" "${FILE}"
 done

 # CI
```

rgsl888prabhu:
@gforsyth Add the changes

gforsyth (approved):

PR #216, opened 2025-07-17
Heuristic Improvements with solution hash, MAB and simplex root solution

akifcorduk:
## Description
This PR adds improvements to heuristics framework with the following:
- Adds solution hash to check if an integer assignment has been locally searched before. If so, we skip that after few trials.
- Adds dual simplex root relaxation solution to heuristics and halts(or prevents) PDLP root run.
- Implements the multi armed bandit algorithm for the LS, currently disabled because no clear benefit was observed. The experiments will be continued here.
- Generalized the MAB algorithm so that it could be used for any purpose which has discrete options to choose from.

### DIFF
--- a/python/cuopt/cuopt/tests/linear_programming/test_lp_solver.py
+++ b/python/cuopt/cuopt/tests/linear_programming/test_lp_solver.py
@@ -475,6 +475,9 @@ def test_parse_var_names():
         )
 
 
+@pytest.mark.skip(
+    reason="Intermittent failure, new version is being worked on"
+)
 def test_parser_and_batch_solver():
 
     data_model_list = []
@@ -564,6 +567,9 @@ def test_warm_start_other_problem():
         solver.Solve(data_model_obj2, settings)
 
 
+@pytest.mark.skip(
+    reason="Intermittent failure, new version is being worked on"
+)
 def test_batch_solver_warm_start():
 
     data_model_list = []


cpp/src/mip/diversity/assignment_hash_map.cu
```diff
@@ -0,0 +1,136 @@
...
+
+template <typename i_t, typename f_t, int TPB>
+__global__ void reduce_hash_kernel(raft::device_span<size_t> reduction_buffer,
+                                   size_t* global_hash_sum)
+{
+  typedef cub::BlockReduce<size_t, TPB> BlockReduce;
+  __shared__ typename BlockReduce::TempStorage temp_storage;
+  size_t th_hash  = reduction_buffer[threadIdx.x];
+  size_t hash_sum = BlockReduce(temp_storage).Reduce(th_hash, combine_hash(), TPB);
+  if (threadIdx.x == 0) { *global_hash_sum = hash_sum; }
```
aliceb-nv, line 65: If TPB isn't equal to the number of blocks launched for the previous kernel, this is going to cause OoBs (either into uninitialized data, or illegal accesses if n_variables < TPB)

aliceb-nv: More generally, is there a reason why we aren't using CUB::DeviceReduce here?

cpp/src/mip/diversity/assignment_hash_map.cu
```diff
@@ -0,0 +1,136 @@
...
+  if (solution_hash_count[hash] > max_occurance) {
+    CUOPT_LOG_DEBUG("Skipping solution which is encountered %d times", solution_hash_count[hash]);
+    return true;
+  }
+  return false;
+}
+
+#if !MIP_INSTANTIATE_FLOAT && !MIP_INSTANTIATE_DOUBLE
+static_assert(false, "MIP_INSTANTIATE_FLOAT or MIP_INSTANTIATE_DOUBLE must be defined");
+#endif
```
aliceb-nv, line 160: Those atrocious linking errors :) I wish there was a way to throw an error at preprocessor/link time without having to add explicit checks

cpp/src/mip/diversity/assignment_hash_map.cu
```diff
@@ -0,0 +1,136 @@
...
+// we might move this to be a solution member if it is needed
+// currently having an ordered integer array is only needed here
+template <typename i_t, typename f_t>
+void assignment_hash_map_t<i_t, f_t>::fill_integer_assignment(solution_t<i_t, f_t>& solution)
+{
+  thrust::gather(solution.handle_ptr->get_thrust_policy(),
+                 solution.problem_ptr->integer_indices.begin(),
+                 solution.problem_ptr->integer_indices.end(),
+                 solution.assignment.begin(),
+                 integer_assignment.begin());
```
aliceb-nv: I think we ought to be careful with the implicit cast to `size_t` performed here. T[he C++ standard states](https://en.cppreference.com/w/c/language/conversion.html#Real_floating-integer_conversions) that any value that cannot be represented in the target cast type yields undefined behavior: 

> Real floating-integer conversions
A finite value of any real floating type can be implicitly converted to any integer type. Except where covered by boolean conversion above, the rules are:
>
> The fractional part is discarded (truncated towards zero).
If the resulting value can be represented by the target type, that value is used
otherwise, the behavior is undefined.
int n = 3.14; // n == 3
int x = 1e10; // undefined behavior for 32-bit int

I think this mean there might be issues in the case of negative and large (>1e20) assignments. It's probably better to compute the hash in a bitwise manner

cpp/src/mip/diversity/assignment_hash_map.cu
```diff
@@ -0,0 +1,136 @@
...
+
+namespace cuopt {
+namespace linear_programming {
+namespace detail {
+
+struct combine_hash {
+  DI size_t operator()(size_t hash_1, size_t hash_2)
+  {
+    hash_1 ^= hash_2 + 0x9e3779b9 + (hash_1 << 6) + (hash_1 >> 2);
+    return hash_1;
```
aliceb-nv: I think this is a 32-bit hash, with 64-bit values this might not work properly

cpp/src/mip/diversity/diversity_manager.cu
```diff
@@ -407,17 +427,38 @@ solution_t<i_t, f_t> diversity_manager_t<i_t, f_t>::run_solver()
...
+    {
+      std::lock_guard<std::mutex> guard(relaxed_solution_mutex);
+      if (!simplex_solution_exists) {
+        raft::copy(lp_optimal_solution.data(),
+                   lp_optimal_solution_copy.data(),
+                   lp_optimal_solution.size(),
+                   problem_ptr->handle_ptr->get_stream());
+      }
+    }
+    problem_ptr->handle_ptr->sync_stream();
```
aliceb-nv, line 457: Should the sync_stream() be inside the scope of the lock_guard?

aliceb-nv (commented):
Looking good :) Just a few questions regarding the hashing
Thank you Akif!

cpp/src/mip/diversity/assignment_hash_map.cu
akifcorduk: Not really, we could use it. Let me convert it :)

cpp/src/mip/diversity/assignment_hash_map.cu
akifcorduk: Yeah, this is the second time i spend significant amount of time to these issues.

cpp/src/mip/diversity/assignment_hash_map.cu
akifcorduk: Good catch, fixed!

cpp/src/mip/diversity/assignment_hash_map.cu
akifcorduk: Thanks fixed!

cpp/src/mip/diversity/diversity_manager.cu
akifcorduk: Here i don't think so. The copys are ordered so there can't be any race condition on the same stream. We are not signalling anything but acting on a signal here. If the other thread finds a simplex solution and sets the flag and executes a copy, the copy will be completed after this which is safe and what we want.

akifcorduk:
/ok to test [6bcd726](https://github.com/NVIDIA/cuopt/pull/216/commits/6bcd7266316f50818d5de57cef89acb9c5520576)

akifcorduk:
/ok to test [08d8372](https://github.com/NVIDIA/cuopt/pull/216/commits/08d83724b63e6dabe425aa3fd1ed949d956d08e7)

cpp/src/mip/diversity/diversity_manager.cu
aliceb-nv: That makes sense!

cpp/src/mip/diversity/assignment_hash_map.cu
```diff
@@ -81,7 +83,7 @@ void assignment_hash_map_t<i_t, f_t>::fill_integer_assignment(solution_t<i_t, f_
                  solution.problem_ptr->integer_indices.begin(),
                  solution.problem_ptr->integer_indices.end(),
                  solution.assignment.begin(),
-                 integer_assignment.begin());
+                 reinterpret_cast<double*>(integer_assignment.data()));
```
aliceb-nv, line 87: I seriously doubt we will ever use anything but FP64, but in the rare case we instantiate the solver with f_t=FP32, this is going to break/behave badly. We might want to add a static_assert to make sure sizeof(f_t) == sizeof(size_t) in order not to cause issues if users decide to try out FP32

aliceb-nv (approved):
Awesome work Akif!

cpp/src/mip/diversity/assignment_hash_map.cu
akifcorduk: okay let me add the static assert here.

cpp/src/mip/diversity/assignment_hash_map.cu
rg20: Should we just remove FLOAT? do we anticipate using it for MIP?

PR #222, opened 2025-07-22
Remove cuda 11 specific changes from repo

### LINKED ISSUE #220: Dropping CUDA 11?
In RAPIDS 25.08, we are dropping CUDA 11 as noted in [RSN 48]( https://docs.rapids.ai/notices/rsn0048/ ). This is also tracked in issue:

* https://github.com/rapidsai/build-planning/issues/184

This means CI support for CUDA 11 is also being dropped

When does cuOpt plan to drop CUDA 11? Does it make sense to also drop in cuOpt 25.08?

rgsl888prabhu:
## Description
This PR removes CUDA 11 pertaining changes since we have moved to CUDA 12 support only

## Issue
closes #220 

## Checklist

- Testing
   - [ ] Added tests
   - [x] NA
- Documentation
   - [ ] Added new documentation
   - [x] NA


.github/workflows/build.yaml
```diff
@@ -56,7 +55,6 @@ jobs:
     secrets: inherit
     uses: rapidsai/shared-workflows/.github/workflows/conda-python-build.yaml@branch-25.08
     with:
-      matrix_filter: map(select((.CUDA_VER | startswith("12")) and .PY_VER != "3.13"))
```
KyleFromNVIDIA, line 59: Why has `.PY_VER` been dropped here?

build.sh
```diff
@@ -268,13 +268,8 @@ if  [ ${BUILD_ALL_GPU_ARCH} -eq 1 ]; then
     echo "Building for *ALL* supported GPU architectures..."
 else
     if [ ${BUILD_CI_ONLY} -eq 1 ]; then
-        if [[ ${CUDA_VERSION} == 11* ]]; then
-            CUOPT_CMAKE_CUDA_ARCHITECTURES="70-real;80"
-            echo "Building for Volta and Ampere architectures..."
-        else
             CUOPT_CMAKE_CUDA_ARCHITECTURES="RAPIDS"
             echo "Building for Volta, Ampere and Hopper architectures..."
```
KyleFromNVIDIA: Please fix the indentation:

```suggestion
        CUOPT_CMAKE_CUDA_ARCHITECTURES="RAPIDS"
        echo "Building for Volta, Ampere and Hopper architectures..."
```

.github/workflows/pr.yaml
```diff
@@ -183,14 +168,13 @@ jobs:
...
       append-cuda-suffix: false
-      matrix_filter: map(select((.CUDA_VER | startswith("12"))))
   wheel-build-libcuopt:
     needs: wheel-build-cuopt-mps-parser
     secrets: inherit
     uses: rapidsai/shared-workflows/.github/workflows/wheels-build.yaml@branch-25.08
     with:
       # build for every combination of arch and CUDA version, but only for the latest Python
-      matrix_filter: map(select((.CUDA_VER | startswith("12")) and .PY_VER == "3.12"))
+      matrix_filter: map(select(.PY_VER == "3.13"))
```
KyleFromNVIDIA: Why has the `.PY_VER` been changed?

.github/workflows/build.yaml
ramakrishnap-nv: 3.13 version was not supported in 25.05 cuopt, but rapids had it, so had to make sure 3.13 will not get built. Noe we support 3.13  as well, so removed this.

.github/workflows/pr.yaml
ramakrishnap-nv: We just need one version of python, I just changed it to latest python version we support. May be it can be changed to a mapping to choose max version in python. 

Can you please help me with that expression ?

ramakrishnap-nv: Updated this as per CUDF changes 

```
      matrix_filter: group_by([.ARCH, (.CUDA_VER|split(".")|map(tonumber)|.[0])]) | map(max_by(.PY_VER|split(".")|map(tonumber)))

```

KyleFromNVIDIA (approved):

Iroy30 (approved):

.github/workflows/build.yaml
jakirkham: Think we would want to rewrite this to use 3.12

```suggestion
      matrix_filter: map(select((.CUDA_VER | startswith("12")) and .PY_VER != "3.12"))
```

Alternatively we can update this like we did below

```suggestion
      matrix_filter: group_by([.ARCH, (.CUDA_VER|split(".")|map(tonumber)|.[0])]) | map(max_by(.PY_VER|split(".")|map(tonumber)))
```

.github/workflows/build.yaml
```diff
@@ -45,7 +45,6 @@ jobs:
     secrets: inherit
     uses: rapidsai/shared-workflows/.github/workflows/conda-cpp-build.yaml@branch-25.08
     with:
-      matrix_filter: map(select((.CUDA_VER | startswith("12"))))
```
jakirkham, line 48: Recognize we are only testing CUDA 12, but could we keep this?

jakirkham (commented):
Thanks Ram! 🙏

Had a couple comments below

jakirkham:
Could we please also take a look at this code?

https://github.com/NVIDIA/cuopt/blob/9952c265de15d19b397ffb56542b216de6ff3331/cpp/src/linear_programming/utilities/ping_pong_graph.cuh#L48-L63

.github/workflows/build.yaml
ramakrishnap-nv: Hmm, I think this one would build only for python3.13 only, and not other python versions, wouldn't that be a problem ?

.github/workflows/build.yaml
ramakrishnap-nv: Since we might move to cuda 13 in 25.10 or later, I removed it just for future case.

rgsl888prabhu:
> Thanks Ram! 🙏
> 
> Can you please look at this CMake code as well?
> 
> https://github.com/NVIDIA/cuopt/blob/47fb1581e2d6867218a3d8aa27d6f2e97521708a/cpp/CMakeLists.txt#L222-L224

Removed if conditions around it.

.github/workflows/build.yaml
ramakrishnap-nv: > Hmm, I think this one would build only for python3.13 only, and not other python versions, wouldn't that be a problem ?

@jakirkham What is your suggestion on this

rgsl888prabhu:
@jakirkham May I get another set of review and also had few questions on previous reviews

cpp/src/linear_programming/utilities/ping_pong_graph.cuh
```diff
@@ -60,13 +60,13 @@ class ping_pong_graph_t {
       if (total_pdlp_iterations % 2 == 0 && !even_initialized) {
         RAFT_CUDA_TRY(cudaStreamEndCapture(stream_view_.value(), &even_graph));
         // Extra NULL NULL 0 mandatory for cuda 11.8
```
hlinsen: We can remove the cuda 11.8 comment as well.
Could you also update more occurances of `cudagraphInstantiate`:
- https://github.com/NVIDIA/cuopt/blob/branch-25.08/cpp/src/routing/cuda_graph.cuh#L57
- https://github.com/NVIDIA/cuopt/blob/branch-25.08/cpp/src/mip/presolve/load_balanced_bounds_presolve.cu#L206-L209
- https://github.com/NVIDIA/cuopt/blob/branch-25.08/cpp/src/mip/feasibility_jump/feasibility_jump.cu#L788

ramakrishnap-nv: @hlinsen updated the PR

Kh4ster (approved):

akifcorduk (approved):
Looks good on cpp side.

.github/workflows/build.yaml
bdice: This is correct to remove, from what I understand of the above. We need to build a Python package for each supported Python version. I think no further changes are needed here.

build.sh
```diff
@@ -268,13 +268,8 @@ if  [ ${BUILD_ALL_GPU_ARCH} -eq 1 ]; then
...
     if [ ${BUILD_CI_ONLY} -eq 1 ]; then
-        if [[ ${CUDA_VERSION} == 11* ]]; then
-            CUOPT_CMAKE_CUDA_ARCHITECTURES="70-real;80"
-            echo "Building for Volta and Ampere architectures..."
-        else
-            CUOPT_CMAKE_CUDA_ARCHITECTURES="RAPIDS"
-            echo "Building for Volta, Ampere and Hopper architectures..."
-        fi
+        CUOPT_CMAKE_CUDA_ARCHITECTURES="RAPIDS"
+        echo "Building for Volta, Ampere and Hopper architectures..."
```
bdice: This comment is not accurate, because the definition of "RAPIDS" arches changes over time. Currently that includes this set of arches: https://github.com/rapidsai/rapids-cmake/blob/4eb862d7eb00ee3bfe79c001fca79c8685c2dadb/rapids-cmake/cuda/set_architectures.cmake#L59

I would change this:
```suggestion
        CUOPT_CMAKE_CUDA_ARCHITECTURES="RAPIDS"
        echo "Building for RAPIDS supported architectures..."
```

bdice (approved):
One small suggestion to clarify a comment, otherwise LGTM. We should land this soon for 25.08.

jakirkham:
Thanks Ram and thanks to all the reviewers! 🙏

Agree this looks good

PR #223, opened 2025-07-22
Add python API

Iroy30:
## Checklist

- Testing
   - [ ] Added tests
   - [ ] NA
- Documentation
   - [ ] Added new documentation
   - [ ] NA
Working on Adding tests and solution extraction

### DIFF
--- a/python/cuopt/cuopt/linear_programming/problem.py
+++ b/python/cuopt/cuopt/linear_programming/problem.py
@@ -1,0 +1,358 @@
+import cuopt.linear_programming.data_model as data_model
+import cuopt.linear_programming.solver as solver
+
+import enum
+import numpy as np
+
+# The type of a variable is either continuous, binary, or integer
+CONTINUOUS = 'C'
+BINARY = 'B'
+INTEGER = 'I'
+
+# Variable objects hold a reference to the problem they were created from as well as info
+# about there index, lower bound, upper bound, objective coefficient and variable type.
+# You can add and subtract Variables to scalars and other Variables to form LinearExpression objects.
+# You can multiply variables by a scalar to form a LinearExpression object
+class Variable:
+    def __init__(self, problem, index, lb=0.0, ub=float('inf'), obj=0.0, vtype=CONTINUOUS, vname=''):
+        self.problem = problem
+        self.index = index
+        self.LB = lb
+        self.UB = ub
+        self.Obj = obj
+        self.Value = 0.0
+        self.VType = vtype
+        self.VName = vname
+    def getIndex(self):
+        return self.index
+    def getValue(self):
+        return self.Value
+    def getObjectiveCoefficient(self):
+        return self.Obj
+    def setObjectiveCoefficient(self, val):
+        self.Obj = val
+    def setLowerBound(self, val):
+        self.LB = val
+    def getLowerBound(self):
+        return self.LB
+    def setUpperBound(self, val):
+        self.UB = val
+    def getUpperBound(self):
+        return self.UB
+    def setVariableType(self, val):
+        self.VType = val
+    def getVariableType(self):
+        return self.VType
+    def setVariableName(self, val):
+        self.VName = val
+    def getVariableName(self):
+        return self.VName
+    def __add__(self, other):
+        match other:
+            case int() | float():
+                return LinearExpression([self], [1.0], float(other))
+            case Variable():
+                # Change?
+                return LinearExpression([self, other], [1.0, 1.0], 0.0)
+            case LinearExpression():
+                return other + self
+            case _:
+                raise ValueError('Cannot add type %s to variable' % type(other).__name__)
+    def __radd__(self, other):
+        return self + other
+    def __sub__(self, other):
+        match other:
+            case int() | float():
+                return LinearExpression([self], [1.0], -float(other))
+            case Variable():
+                return LinearExpression([self, other], [1.0, -1.0], 0.0)
+            case LinearExpression():
+                # self - other ->   other * -1.0 + self
+                return other * -1.0 + self
+            case _:
+                raise ValueError('Cannot subtract type %s from variable' % type(other).__name__)
+    def __rsub__(self, other):
+        # other - self  -> other + self * -1.0
+        return other + self * -1.0
+    def __mul__(self, other):
+        match other:
+            case int() | float():
+                return LinearExpression([self], [float(other)], 0.0)
+            case _:
+                raise ValueError('Cannot multiply type %s with variable' % type(other).__name__)
+    def __rmul__(self, other):
+        return self * other
+
+
+# LinearExpressions contain a set of variables, the coefficients for the variables, and a constant
+# LinearExpressions can be used to create constraints and the objective in the Problem
+# LinearExpressions can be added and subtracted with other LinearExpressions and Variables
+# LinearExpressions can be multiplied and divided by scalars
+# LinearExpressions can be compared with scalars, Variables, and other LinearExpressions to create Constraints
+class LinearExpression:
+    def __init__(self, vars, coefficients, constant):
+        self.vars = vars
+        self.coefficients = coefficients
+        self.constant = constant
+    def getVars(self):
+        return self.vars
+    def getVar(self, i):
+        return self.vars[i]
+    def getCoefficients(self):
+        return self.coefficients
+    def getCoefficient(self, i):
+        return self.coefficients[i]
+    def getConstant(self):
+        return self.constant
+    def zipVarCoefficients(self):
+        return zip(self.vars, self.coefficients)
+    def getValue(self):
+        value = 0.0
+        for i, var in enumerate(self.vars):
+            value += var.Value * self.coefficients[i]
+        return value
+    def size(self):
+        return len(self.vars)
+    def __len__(self):
+        return len(self.vars)
+    def __add__(self, other):
+        match other:
+            case int() | float():
+                self.constant += float(other)
+                return self
+            case Variable():
+                self.vars.append(other)
+                self.coefficients.append(1.0)
+                return self
+            case LinearExpression():
+                self.vars.extend(other.vars)
+                self.coefficients.extend(other.coefficients)
+                self.constant += other.constant
+                return self
+            case _:
+                raise ValueError("Can't add type %s to Linear Expression" % type(other).__name__)
+    def __radd__(self, other):
+        return self + other
+    def __sub__(self, other):
+        match other:
+            case int() | float():
+                self.constant -= float(other)
+                return self
+            case Variable():
+                self.vars.append(other)
+                self.coefficients.append(-1.0)
+                return self
+            case LinearExpression():
+                self.vars.extend(other.vars)
+                for coeff in other.coefficients:  # Same Time Complexity as extend O(k), k = nelements to append
+                    self.coefficients.append(-coeff)
+                self.constant -= other.constant
+                return self
+            case _:
+                raise ValueError("Can't sub type %s from LinearExpression" % type(other).__name__)
+    def __rsub__(self, other):
+        # other - self  -> other + self * -1.0
+        return other + self * -1.0
+    def __mul__(self, other):
+        match other:
+            case int() | float():
+                self.coefficients = [coeff * float(other) for coeff in self.coefficients]
+                self.constant = self.constant * float(other)
+                return self
+            case _:
+                raise ValueError("Can't multiply type %s by LinearExpresson" % type(other).__name__)
+    def __rmul__(self, other):
+        return self * other
+    def __div__(self, other):
+        match other:
+            case int() | float():
+                self.coefficients = [coeff / float(other) for coeff in self.coefficients]
+                self.constant = self.constant / float(other)
+                return self
+            case _:
+                raise ValueError("Can't divide LinearExpression by type %s" % type(other).__name__)
+    def __le__(self, other):
+        match other:
+            case int() | float():
+                return Constraint(self, CONSTRAINT_LE, float(other))
+            case Variable() | LinearExpression():
+                # expr1 <= expr2   -> expr1 - expr2 <= 0
+                expr = self - other
+                return Constraint(expr, CONSTRAINT_LE, 0.0)
+    def __ge__(self, other):
+        match other:
+            case int() | float():
+                return Constraint(self, CONSTRAINT_GE, float(other))
+            case Variable() | LinearExpression():
+                # expr1 >= expr2   ->  expr1 - expr2 >= 0
+                expr = self - other
+                return Constraint(expr, CONSTRAINT_GE, 0.0)
+    def __eq__(self, other):
+        match other:
+            case int() | float():
+                return Constraint(self, CONSTRAINT_EQ, float(other))
+            case Variable() | LinearExpression():
+                # expr1 == expr2   -> expr1 - expr2 == 0
+                expr = self - other
+                return Constraint(expr, CONSTRAINT_EQ, 0.0)
+
+# The sense of a constraint is either less than or equal, greater than or equal, or equal
+CONSTRAINT_LE = 'L'
+CONSTRAINT_GE = 'G'
+CONSTRAINT_EQ = 'E'
+
+# A constraint contains a linear expression, the sense of the constraint, and the right-hand side of the constraint
+class Constraint:
+    def __init__(self, expr, sense, rhs, name=''):
+        self.vindex_coeff_dict = {}
+        nz = len(expr.getVars())
+        for i in range(nz):
+            v_idx = expr.vars[i].index
+            v_coeff = expr.coefficients[i]
+            self.vindex_coeff_dict[v_idx] = self.vindex_coeff_dict[v_idx] + v_coeff if v_idx in self.vindex_coeff_dict else v_coeff
+        self.Sense = sense
+        self.RHS = rhs - expr.getConstant()
+        self.CName = name
+    def getName(self):
+        return CName
+    def getSense(self):
+        return self.Sense
+    def getRHS(self):
+        return self.RHS
+    def getCoefficient(self, var):
+        v_idx = var.index
+        return vindex_coeff_dict[v_idx]
+
+# The sense of a problem is either minimize or maximize
+MINIMIZE = 0
+MAXIMIZE = 1
+
+# A Problem defines a Linear Program or Mixed Integer Program
+# Variable can be be created by calling addVariable()
+# Constraints can be added by calling addConstraint()
+# The objective can be set by calling setObjective()
+# The problem data is formed when calling optimize()
+class Problem:
+    def __init__(self, model_name=''):
+        self.Name = model_name
+        self.vars = []
+        self.constrs = []
+        self.ObjSense = MINIMIZE
+        self.Obj = None
+        self.ObjCon = 0.0
+        self.Status = -1
+        self.IsMIP = False
+
+        self.rhs = None
+        self.row_sense = None
+        self.row_pointers = None
+        self.column_indicies = None
+        self.values = None
+        self.lower_bound = None
+        self.upper_bound = None
+        self.var_type = None
+
+    def addVar(self, lb=0.0, ub=float('inf'), obj=0.0, vtype=CONTINUOUS, name=''):
+        n = len(self.vars)
+        if vtype == INTEGER or vtype == BINARY:
+            self.IsMIP = True
+        var = Variable(self, n, lb, ub, obj, vtype, name)
+        self.vars.append(var)
+        return var
+
+    def addConstr(self, constr, name=''):
+        n = len(self.constrs)
+        match constr:
+            case Constraint():
+                constr.index = n
+                constr.cname = name
+                self.constrs.append(constr)
+            case _:
+                raise ValueError("addConstraint requires a Constraint object")
+
+    def setObjective(self, expr, sense=MINIMIZE):
+        self.ObjSense = sense
+        match expr:
+            case int() | float():
+                for var in self.vars:
+                    var.setObjectiveCoefficient(0.0)
+                self.ObjCon = float(expr)
+            case Variable():
+                for var in self.vars:
+                    var.setObjectiveCoefficient(0.0)
+                    if var.getIndex() == expr.getIndex():
+                        var.setObjectiveCoefficient(1.0)
+            case LinearExpression():
+                for var, coeff in expr.zipVarCoefficients():
+                    self.vars[var.getIndex()].setObjectiveCoefficient(coeff)
+            case _:
+                raise ValueError('Objective must be a LinearExpression or a constant')
+        self.Obj = expr
+
+    def getObjective(self):
+        return self.Obj
+
+    def getVars(self):
+        return self.vars
+
+    def getConstrs(self):
+        return self.constrs
+
+    @property
+    def NumVars(self):
+        return len(self.vars)
+
+    @property
+    def NumConstrs(self):
+        return len(self.constrs)
+
+    @property
+    def NumNZs(self):
+        nnz = 0
+        for constr in self.constrs:
+            nnz += len(constr.vindex_coeff_dict)
+
+    def optimize(self):
+        # iterate through the constraints and construct the constraint matrix and the rhs
+        m = len(self.constrs)
+        n = len(self.vars)
+        self.row_pointers = [0]
+        self.column_indicies = []
+        self.values = []
+        self.rhs = []
+        self.row_sense = []
+        for constr in self.constrs:
+            self.column_indicies.extend(list(constr.vindex_coeff_dict.keys()))
+            self.values.extend(list(constr.vindex_coeff_dict.values()))
+            self.row_pointers.append(len(self.column_indicies))
+            self.rhs.append(constr.RHS)
+            self.row_sense.append(constr.Sense)
+
+        self.objective = []
+        self.lower_bound, self.upper_bound = [], []
+        self.var_type = []
+
+        for j in range(n):
+            self.objective.append(self.vars[j].getObjectiveCoefficient())
+            self.var_type.append(self.vars[j].getVariableType())
+            self.lower_bound.append(self.vars[j].getLowerBound())
+            self.upper_bound.append(self.vars[j].getUpperBound())
+
+        # Initialize datamodel
+        dm = data_model.DataModel()
+        dm.set_csr_constraint_matrix(np.array(self.values), np.array(self.column_indicies), np.array(self.row_pointers))
+        dm.set_maximize(self.ObjSense)
+        dm.set_constraint_bounds(np.array(self.rhs))
+        dm.set_row_types(np.array(self.row_sense))
+        dm.set_objective_coefficients(np.array(self.objective))
+        dm.set_variable_lower_bounds(np.array(self.lower_bound))
+        dm.set_variable_upper_bounds(np.array(self.upper_bound))
+        dm.set_variable_types(np.array(self.var_type))
+
+        # Initialize  solver_settings
+
+        # Call Solver
+        solution = solver.Solve(dm)
+
+        # Post Solve
+        return solution
--- a/python/cuopt/cuopt/linear_programming/__init__.py
+++ b/python/cuopt/cuopt/linear_programming/__init__.py
@@ -17,6 +17,7 @@ from cuopt.linear_programming import internals
 from cuopt.linear_programming.data_model import DataModel
 from cuopt.linear_programming.solution import Solution
 from cuopt.linear_programming.solver import BatchSolve, Solve
+from cuopt.linear_programming.problem import Problem
 from cuopt.linear_programming.solver_settings import (
     PDLPSolverMode,
     SolverMethod,


python/cuopt/cuopt/linear_programming/problem.py
```diff
@@ -0,0 +1,384 @@
+import cuopt.linear_programming.data_model as data_model
+import cuopt.linear_programming.solver as solver
+import cuopt.linear_programming.solver_settings as solver_settings
+
+import enum
+import numpy as np
+
+# The type of a variable is either continuous, binary, or integer
+CONTINUOUS = 'C'
+BINARY = 'B'
```
chris-maes: When I created the prototype I thought cuOpt distinguished between integer and binary variables. It does not. So I would remove BINARY as a variable type here.

python/cuopt/cuopt/linear_programming/problem.py
```diff
@@ -0,0 +1,384 @@
...
+# The sense of a constraint is either less than or equal, greater than or equal, or equal
+CONSTRAINT_LE = 'L'
+CONSTRAINT_GE = 'G'
+CONSTRAINT_EQ = 'E'
+
+# A constraint contains a linear expression, the sense of the constraint, and the right-hand side of the constraint
+class Constraint:
+    def __init__(self, expr, sense, rhs, name=''):
+        self.vindex_coeff_dict = {}
+        nz = len(expr.getVars())
```
chris-maes: I added a `size` method to `LinearExpr`. It would probably be clearer to write this as
`nz = expr.size()`

python/cuopt/cuopt/linear_programming/problem.py
```diff
@@ -0,0 +1,384 @@
...
+        return vindex_coeff_dict[v_idx]
+    @property
+    def Slack(self):
+        lhs = 0.0
+        for var in self.vars:
+            lhs += var.Value * self.vindex_coeff_dict[var.index]
+        return self.RHS - lhs
+
+# The sense of a problem is either minimize or maximize
+MINIMIZE = 0
```
chris-maes: Maybe for consistency with the other APIs we should have 
`MINIMIZE = 1` and `MAXIMIZE = -1`

python/cuopt/cuopt/linear_programming/problem.py
```diff
@@ -0,0 +1,384 @@
...
+MAXIMIZE = 1
+
+# A Problem defines a Linear Program or Mixed Integer Program
+# Variable can be be created by calling addVariable()
```
chris-maes: The comment refers to `addVariable` but the method is named `addVar`. We should be consistent here.

python/cuopt/cuopt/linear_programming/problem.py
```diff
@@ -0,0 +1,384 @@
...
+# Constraints can be added by calling addConstraint()
```
chris-maes: Same here: documentantion refers to `addConstraint` but the method is called `addConstr`. Here I would suggest renaming the method to `addConstraint`.

python/cuopt/cuopt/linear_programming/problem.py
chris-maes: I would probably avoid the abbreviation and rename the method to `addVariable`.

python/cuopt/cuopt/linear_programming/problem.py
```diff
@@ -0,0 +1,384 @@
...
+
+    @property
+    def NumConstrs(self):
+        return len(self.constrs)
+
+    @property
+    def NumNZs(self):
+        nnz = 0
+        for constr in self.constrs:
+            nnz += len(constr.vindex_coeff_dict)
```
chris-maes: I would add a `size` method or `__len__` to `Constraint` and do this as:
`nnz += len(constr)`

That maintains the abstraction barrier---that is you can change the implementation of constraint to replace the dict without affecting this code.

python/cuopt/cuopt/linear_programming/problem.py
chris-maes: I guess I also added `__len__`. So maybe `nz = len(expr)` is more Pythonic.

python/cuopt/cuopt/linear_programming/problem.py
```diff
@@ -0,0 +1,384 @@
...
+        self.rhs = None
+        self.row_sense = None
+        self.row_pointers = None
+        self.column_indicies = None
+        self.values = None
+        self.lower_bound = None
+        self.upper_bound = None
+        self.var_type = None
+
+    def addVar(self, lb=0.0, ub=float('inf'), obj=0.0, vtype=CONTINUOUS, name=''):
```
chris-maes: rename to `addVariable`

python/cuopt/cuopt/linear_programming/problem.py
```diff
@@ -0,0 +1,384 @@
...
+        n = len(self.vars)
+        if vtype == INTEGER or vtype == BINARY:
+            self.IsMIP = True
+        var = Variable(self, n, lb, ub, obj, vtype, name)
+        self.vars.append(var)
+        return var
+
+    def addConstr(self, constr, name=''):
```
chris-maes: rename to `addConstraint`

python/cuopt/cuopt/linear_programming/problem.py
```diff
@@ -0,0 +1,384 @@
...
+# The objective can be set by calling setObjective()
+# The problem data is formed when calling optimize()
+class Problem:
+    def __init__(self, model_name=''):
+        self.Name = model_name
+        self.vars = []
+        self.constrs = []
+        self.ObjSense = MINIMIZE
+        self.Obj = None
+        self.ObjCon = 0.0
```
chris-maes: Might be better to call this `ObjConstant` to make it more readable.

python/cuopt/cuopt/linear_programming/problem.py
```diff
@@ -0,0 +1,384 @@
...
+        self.vindex_coeff_dict = {}
+        nz = len(expr.getVars())
+        self.vars = expr.vars
+        for i in range(nz):
+            v_idx = expr.vars[i].index
+            v_coeff = expr.coefficients[i]
+            self.vindex_coeff_dict[v_idx] = self.vindex_coeff_dict[v_idx] + v_coeff if v_idx in self.vindex_coeff_dict else v_coeff
+        self.Sense = sense
+        self.RHS = rhs - expr.getConstant()
+        self.CName = name
```
chris-maes: Do you want to add 
`self.index = -1`
here?  So that every constraint has an index?

python/cuopt/cuopt/linear_programming/problem.py
```diff
@@ -0,0 +1,384 @@
...
+        self.SolveTime = solution.get_solve_time()
+
+        if solution.problem_category == 0:
+            self.SolutionStats = solution.get_lp_stats()
+        else:
+            self.SolutionStats = solution.get_milp_stats()
+
+        primal_sol = solution.get_primal_solution()
+        reduced_cost = solution.get_reduced_cost()
+        if len(primal_sol) > 0:
```
chris-maes, line 921: I think you want to add support for getting the Lagrange multipliers associated with the constraints here:

```
dual_variables  = solution.get_dual_solution()
for constr in self.constrs:
       constr.DualValue = dual_variables[constr.index]
```

I would use `DualValue` instead of `Pi` here.

python/cuopt/cuopt/linear_programming/problem.py
```diff
@@ -0,0 +1,384 @@
...
+# You can add and subtract Variables to scalars and other Variables to form LinearExpression objects.
+# You can multiply variables by a scalar to form a LinearExpression object
+class Variable:
+    def __init__(self, problem, index, lb=0.0, ub=float('inf'), obj=0.0, vtype=CONTINUOUS, vname=''):
+        self.problem = problem
+        self.index = index
+        self.LB = lb
+        self.UB = ub
+        self.Obj = obj
+        self.Value = 0.0
```
chris-maes: Before the solve, it might be safer to keep the Value initialized to a sentinel value so that if a user accidentally queries it they discover it.

I would do `self.Value = float('nan')`. And simiarly `self.ReducedCost = float('nan')`

python/cuopt/cuopt/linear_programming/problem.py
```diff
@@ -0,0 +1,384 @@
...
+        else:
+            self.SolutionStats = solution.get_milp_stats()
+
+        primal_sol = solution.get_primal_solution()
+        reduced_cost = solution.get_reduced_cost()
+        if len(primal_sol) > 0:
+            for var in self.vars:
+                var.Value = primal_sol[var.index]
+                if not self.IsMIP:
+                    var.RC =  reduced_cost[var.index]
```
chris-maes: I would use `var.ReducedCost` instead of `var.RC` here.

python/cuopt/cuopt/linear_programming/problem.py
chris-maes: I would rename `CNAME` to `ConstraintName`.

python/cuopt/cuopt/linear_programming/problem.py
```diff
@@ -0,0 +1,384 @@
...
+        n = len(self.vars)
+        self.row_pointers = [0]
+        self.column_indicies = []
+        self.values = []
+        self.rhs = []
+        self.row_sense = []
+        for constr in self.constrs:
+            self.column_indicies.extend(list(constr.vindex_coeff_dict.keys()))
+            self.values.extend(list(constr.vindex_coeff_dict.values()))
+            self.row_pointers.append(len(self.column_indicies))
```
chris-maes, line 963: Is this correct? I think the row_pointers needs to be the cumulative sum of the len of the constraint.

```
nz = 0 
self.row_pointers = [0]
for constr in self.constrs:
     # rest of loop
     nz += len(constr)
     self.row_pointers.append(nz)
```

python/cuopt/cuopt/linear_programming/problem.py
```diff
@@ -0,0 +1,384 @@
...
+    def getSense(self):
+        return self.Sense
+    def getRHS(self):
+        return self.RHS
+    def getCoefficient(self, var):
+        v_idx = var.index
+        return vindex_coeff_dict[v_idx]
+    @property
+    def Slack(self):
+        lhs = 0.0
```
chris-maes, line 636: Instead of making this a function that gets recomputed each time the user access `constr.Slack`, we might want to just compute this once in `post_solve` and store it in `self.Slack`.

python/cuopt/cuopt/linear_programming/problem.py
```diff
@@ -0,0 +1,384 @@
...
+    def __init__(self, expr, sense, rhs, name=''):
+        self.vindex_coeff_dict = {}
+        nz = len(expr.getVars())
+        self.vars = expr.vars
+        for i in range(nz):
+            v_idx = expr.vars[i].index
+            v_coeff = expr.coefficients[i]
+            self.vindex_coeff_dict[v_idx] = self.vindex_coeff_dict[v_idx] + v_coeff if v_idx in self.vindex_coeff_dict else v_coeff
+        self.Sense = sense
+        self.RHS = rhs - expr.getConstant()
```
chris-maes, line 600: Do you want to initialize the values from the solve here?
```
self.Slack = float('nan')
self.DualValue = float('nan')
```

python/cuopt/cuopt/linear_programming/problem.py
```diff
@@ -0,0 +1,384 @@
...
+    def getConstant(self):
+        return self.constant
+    def zipVarCoefficients(self):
+        return zip(self.vars, self.coefficients)
+    def getValue(self):
+        value = 0.0
+        for i, var in enumerate(self.vars):
+            value += var.Value * self.coefficients[i]
+        return value
+    def size(self):
```
chris-maes: Maybe we should remove this and just have `__len__`?

python/cuopt/cuopt/linear_programming/problem.py
```diff
@@ -0,0 +1,384 @@
...
+# A constraint contains a linear expression, the sense of the constraint, and the right-hand side of the constraint
+class Constraint:
+    def __init__(self, expr, sense, rhs, name=''):
+        self.vindex_coeff_dict = {}
+        nz = len(expr.getVars())
+        self.vars = expr.vars
+        for i in range(nz):
+            v_idx = expr.vars[i].index
+            v_coeff = expr.coefficients[i]
+            self.vindex_coeff_dict[v_idx] = self.vindex_coeff_dict[v_idx] + v_coeff if v_idx in self.vindex_coeff_dict else v_coeff
```
chris-maes: My guess is that the speed of this loop can be improved. 

The hash lookup is likely costly.  Suppose you have an array of size n (where n is the number of variables) that is all zero. You could store that array in the problem.

```
workspace = problem.workspace # workspace is all zero and of length variables

# scatter the nonzero coefficients into workspace.
# Any coefficients with the same variable index will get added
self.indices = []
for i in range (nz):
     j = expr.vars[i].index
     x = expr.coefficients[i]
     if workspace[j] == 0.0:
          self.indices.append(j)
     workspace[j] += x

# gather the nonzero coefficients into self.coefficients
final_nz = len(self.indices)
self.coefficients = []
for k in range(final_nz):
    self.coefficients.append(workspace[self.indicies[k]])

# clear workspace
for k in range(final_nz):
    workspace[self.indices[k]] = 0.0
```
This should be fast in C/C++. But might be slow in Python.

python/cuopt/cuopt/linear_programming/problem.py
Iroy30: I was planning to do that, however, how often do users retrieve/inspect it? If not accessed as frequently, I'd rather not increase the problem-to-solution time by bloating post-solve routine

python/cuopt/cuopt/linear_programming/problem.py
```diff
@@ -0,0 +1,384 @@
...
+        self.row_sense = []
+        for constr in self.constrs:
+            self.column_indicies.extend(list(constr.vindex_coeff_dict.keys()))
+            self.values.extend(list(constr.vindex_coeff_dict.values()))
+            self.row_pointers.append(len(self.column_indicies))
+            self.rhs.append(constr.RHS)
+            self.row_sense.append(constr.Sense)
+
+        self.objective = []
+        self.lower_bound, self.upper_bound = [], []
```
chris-maes: It might be faster to create `objective`, `lower_bound` and `upper_bound` as numpy arrays and then directly index into them using var.index to fill them in, rather then growing a Python list using `append` and then converting the list to a numpy array.

python/cuopt/cuopt/linear_programming/problem.py
chris-maes: I'm not sure how often users access the slack. You won't increase the solution time by much if you precompute the constraints slack. But if you are really worried, you could keep your function. I would at least memoize the value you compute though, so if a user does access it multiple times you don't recompute it each time.

python/cuopt/cuopt/linear_programming/problem.py
Iroy30: len(self.column_indicies) for each iteration is equivalent to nz += len(constr) since column_indices keeps adding nz column_index to the list

python/cuopt/cuopt/linear_programming/problem.py
Iroy30: Do we want to keep 'B' for UX and internally add lb=0 and ub=1 vtype="I" or do we want to let users update values externally if they want to use Binary Vars

python/cuopt/cuopt/linear_programming/problem.py
Iroy30: Not sure memorizing Slack makes sense since if a user chooses to add constraints and re-optimize , slack would change. I will evaluate the time-add to compute slack in post-solve and check how significant (or not) it is

python/cuopt/cuopt/linear_programming/problem.py
chris-maes: I would just get rid of BINARY for now. And just use integer.

python/cuopt/cuopt/linear_programming/problem.py
chris-maes: Sure. I think computing slack once is best. You could do something where when you call optimize, you loop through the constraints and set all slacks to NaN to invalidate it.

python/cuopt/cuopt/linear_programming/problem.py
chris-maes: Ah ok. I missed that. Thanks for explaining.

python/cuopt/cuopt/linear_programming/problem.py
ramakrishnap-nv: May it is better to use string enums instead of variable so it is easier to pull in documentation in case users need to use those.

python/cuopt/cuopt/linear_programming/problem.py
```diff
@@ -0,0 +1,384 @@
...
+# The type of a variable is either continuous, binary, or integer
+CONTINUOUS = 'C'
+BINARY = 'B'
+INTEGER = 'I'
+
+# Variable objects hold a reference to the problem they were created from as well as info
+# about there index, lower bound, upper bound, objective coefficient and variable type.
+# You can add and subtract Variables to scalars and other Variables to form LinearExpression objects.
+# You can multiply variables by a scalar to form a LinearExpression object
+class Variable:
```
ramakrishnap-nv, line 76: Lets add documentation underneath __init__ similar to other modules so these are fetched while building doc and also available while working interactively.

python/cuopt/cuopt/linear_programming/problem.py
ramakrishnap-nv: Lets add details of the each variable and type, also a minimal example would be helpful.

python/cuopt/cuopt/linear_programming/problem.py
```diff
@@ -0,0 +1,384 @@
...
+    def __init__(self, problem, index, lb=0.0, ub=float('inf'), obj=0.0, vtype=CONTINUOUS, vname=''):
+        self.problem = problem
+        self.index = index
+        self.LB = lb
+        self.UB = ub
+        self.Obj = obj
+        self.Value = 0.0
+        self.VType = vtype
+        self.VName = vname
+    def getIndex(self):
```
ramakrishnap-nv, line 131: Add documentation to all the functions

python/cuopt/cuopt/linear_programming/problem.py
```diff
@@ -0,0 +1,384 @@
...
+    def getVariableType(self):
+        return self.VType
+    def setVariableName(self, val):
+        self.VName = val
+    def getVariableName(self):
+        return self.VName
+    def __add__(self, other):
+        match other:
+            case int() | float():
+                return LinearExpression([self], [1.0], float(other))
```
ramakrishnap-nv, line 208: lets add details on why [1.0] and [1.0, -1.0] is hard-coded

python/cuopt/cuopt/linear_programming/problem.py
```diff
@@ -0,0 +1,384 @@
...
+        match other:
+            case int() | float():
+                return Constraint(self, CONSTRAINT_EQ, float(other))
+            case Variable() | LinearExpression():
+                # expr1 == expr2   -> expr1 - expr2 == 0
+                expr = self - other
+                return Constraint(expr, CONSTRAINT_EQ, 0.0)
+
+# The sense of a constraint is either less than or equal, greater than or equal, or equal
+CONSTRAINT_LE = 'L'
```
ramakrishnap-nv: Lets use string enum

python/cuopt/cuopt/linear_programming/problem.py
ramakrishnap-nv: Int enum

python/cuopt/cuopt/linear_programming/problem.py
```diff
@@ -0,0 +1,384 @@
...
+    def __init__(self, model_name=''):
+        self.Name = model_name
+        self.vars = []
+        self.constrs = []
+        self.ObjSense = MINIMIZE
+        self.Obj = None
+        self.ObjCon = 0.0
+        self.Status = -1
+        self.IsMIP = False
+        self.Settings = solver_settings.SolverSettings()
```
ramakrishnap-nv: A function to set solver settings

python/cuopt/cuopt/linear_programming/problem.py
chris-maes: I'm not sure that level of documentation is necessary. If you add a variable into an expression it's coefficient is 1.0. If you subtract a variable its coefficient is -1.0.  Adding a comment here might be helpful, but since this is an implementation detail, I don't think needs to be documented.

python/cuopt/cuopt/linear_programming/problem.py
chris-maes: @Iroy30 I would just pass the solver settings into the the call to `solve`.

python/cuopt/cuopt/linear_programming/problem.py
```diff
@@ -0,0 +1,384 @@
...
+        primal_sol = solution.get_primal_solution()
+        reduced_cost = solution.get_reduced_cost()
+        if len(primal_sol) > 0:
+            for var in self.vars:
+                var.Value = primal_sol[var.index]
+                if not self.IsMIP:
+                    var.RC =  reduced_cost[var.index]
+        self.ObjVal = self.Obj.getValue()
+
+    def optimize(self):
```
chris-maes: If you change this to 

`def solve(self, settings=solver_settings.SolverSettings())`

The user can pass in the solver settings before the call to `solve`

chris-maes:
> Awesome work @chris-maes and @Iroy30, I have initial few comments.
> 
> Also lets add pytests to cover all the APIs and also scenarios where problem gets modified in a succession so the addition and deletion can be tested to make sure data is getting updated.

@rgsl888prabhu deletion is not supported with this API. You can only add new variables or add new constraints. 

@Iroy30 for addition of variables and constraints we might want to reconsider the design a bit. Adding a new variable to a problem should probably invalidate the `.Value` and ` .ReducedCost` of existing variables, and the `.Slack` and `.DualValue` of `Constraint`, if the previous problem has been solved. Maybe we should store the solution info in the `Problem` object. `Variable` and `Constraint` can store a pointer to the problem. When you try to access these values you go through the problem. When a user adds a new variable or constraint to the problem we can mark the current solution as dirty.

python/cuopt/cuopt/linear_programming/problem.py
Iroy30: removed binary

python/cuopt/cuopt/linear_programming/problem.py
Iroy30: removed size , updated to len

python/cuopt/cuopt/linear_programming/problem.py
Iroy30: Updated to addVariable

python/cuopt/cuopt/linear_programming/problem.py
Iroy30: Updated to addConstraint

python/cuopt/cuopt/linear_programming/problem.py
Iroy30: removed redundant size method and added len

python/cuopt/cuopt/linear_programming/problem.py

python/cuopt/cuopt/linear_programming/problem.py

python/cuopt/cuopt/linear_programming/problem.py
Iroy30: Added Dual Value

python/cuopt/cuopt/linear_programming/problem.py
Iroy30: Updated naming

python/cuopt/cuopt/linear_programming/problem.py

python/cuopt/cuopt/linear_programming/problem.py
Iroy30: removed size method

python/cuopt/cuopt/linear_programming/problem.py
Iroy30: added

python/cuopt/cuopt/linear_programming/problem.py
Iroy30: renamed

python/cuopt/cuopt/linear_programming/problem.py
ramakrishnap-nv: It would just help next developer understand why we are using these values, just comment would be enough as well.

python/cuopt/cuopt/linear_programming/problem.py
Iroy30: Updated

python/cuopt/cuopt/linear_programming/problem.py
Iroy30: self.Value is set to 0.0 since if the user tries to getValue() on a linear expression, they get 0.0 instead invalid

python/cuopt/cuopt/linear_programming/problem.py

python/cuopt/cuopt/linear_programming/problem.py
Iroy30: added docs

python/cuopt/cuopt/linear_programming/problem.py
Iroy30: added docs

python/cuopt/cuopt/linear_programming/problem.py

python/cuopt/cuopt/linear_programming/problem.py

python/cuopt/cuopt/linear_programming/problem.py

python/cuopt/cuopt/linear_programming/problem.py
Iroy30: We can have this as an evaluation and improvement in the next cycle. It looks promising but not a significant improvement. Needs further perf results.

python/cuopt/cuopt/linear_programming/problem.py
Iroy30: added comments

python/cuopt/cuopt/linear_programming/problem.py
Iroy30: The solver setting is the python SolverSettings class. So any setting/getting will follow the same method 
prob.Settings.set_parameter("param_name", value)

python/cuopt/cuopt/linear_programming/problem.py
Iroy30: The solver setting is the python SolverSettings class. So any setting/getting will follow the same method
prob.Settings.set_parameter("param_name", value)
A user can update it in the following manner:
```
prob = Problem()
prob.Settings.set_parameter("time_limit", 60)
```

python/cuopt/cuopt/linear_programming/problem.py
```diff
@@ -0,0 +1,796 @@
+# SPDX-FileCopyrightText: Copyright (c) 2023-2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.  # noqa
```
ramakrishnap-nv:
```suggestion
# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.  # noqa
```

python/cuopt/cuopt/linear_programming/problem.py
```diff
@@ -0,0 +1,796 @@
...
+import numpy as np
+
+import cuopt.linear_programming.data_model as data_model
+import cuopt.linear_programming.solver as solver
+import cuopt.linear_programming.solver_settings as solver_settings
+
+
+# The type of a variable is either continuous or integer
+class VType(str, Enum):
+    CONTINUOUS = "C"
```
ramakrishnap-nv, line 33: Lets add definition to these similar to https://github.com/NVIDIA/cuopt/blob/455bdb0b4f99b64e978f01b44fe37d7dd558308a/python/cuopt_self_hosted/cuopt_sh_client/thin_client_solver_settings.py#L46

python/cuopt/cuopt/linear_programming/problem.py
```diff
@@ -0,0 +1,796 @@
...
+
+
+# The sense of a model is either MINIMIZE or MAXIMIZE
+class sense(int, Enum):
+    MAXIMIZE = -1,
+    MINIMIZE = 1
+
+
+class Variable:
+    """
```
ramakrishnap-nv, line 77: Lets add more details to the doc, similar to this https://github.com/NVIDIA/cuopt/blob/455bdb0b4f99b64e978f01b44fe37d7dd558308a/python/cuopt_self_hosted/cuopt_sh_client/cuopt_self_host_client.py#L246

python/cuopt/cuopt/tests/linear_programming/test_python_API.py
```diff
@@ -0,0 +1,204 @@
+# SPDX-FileCopyrightText: Copyright (c) 2023-2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.  # noqa
```
ramakrishnap-nv:
```suggestion
# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.  # noqa
```

python/cuopt/cuopt/linear_programming/problem.py
```diff
@@ -0,0 +1,796 @@
...
+
+    def getValue(self):
+        """
+        Returns the value of the expression computed with the
+        current solution.
+        """
+        value = 0.0
+        for i, var in enumerate(self.vars):
+            value += var.Value * self.coefficients[i]
+        return value
```
tmckayus: I might be naive :) but shouldn't this include the addition of the constant as well? We have summed the variables but what about the constant?

tmckayus: or are users supposed to call getValue and getConstant and add them themselves if they want?

tmckayus: if the second, we should change the comment to say that it is the sum of the variables * coeffs

python/cuopt/cuopt/linear_programming/problem.py
Iroy30: Moved settings as a parameter to the problem

python/cuopt/cuopt/linear_programming/problem.py
Iroy30: Moved settings as a parameter to the problem.

python/cuopt/cuopt/linear_programming/problem.py
Iroy30: Thanks for pointing this out. I evaluated other modelers as well and we do need the constant added there

python/cuopt/cuopt/linear_programming/problem.py
Iroy30: Thanks for updating

python/cuopt/cuopt/linear_programming/problem.py
Iroy30: The Variable is not supposed to be initialized directly. The main way to do is to use addVariable() function in the Problem Class, the parameters have been defined there.

tmckayus (approved):
lgtm if tests pass.

I wrote a driver that reads cuopt LP/MIP JSON and runs it through the Python API, similar to the cvxpy test driver.  It solved air05.json from miplib with the same result as cvxpy, and generated the same user_problem_file.

python/cuopt/cuopt/linear_programming/problem.py
chris-maes: @Iroy30 I don't think it is a good idea that a user would silently get a 0.0 value. You want the user to discover that they are calling `.getValue()` before the solve is completed. I think you should set this to NaN.

python/cuopt/cuopt/linear_programming/problem.py
chris-maes: Do you want to add `self.Slack` as well?

python/cuopt/cuopt/linear_programming/problem.py
chris-maes: Do you want to move this computation to the Problem class?

python/cuopt/cuopt/linear_programming/problem.py
```diff
@@ -0,0 +1,886 @@
...
+            for var in self.vars:
+                var.Value = primal_sol[var.index]
+                if not self.IsMIP:
+                    var.ReducedCost = reduced_cost[var.index]
+        if not self.IsMIP:
+            dual_sol = solution.get_dual_solution()
+            if len(dual_sol) > 0:
+                for i, constr in enumerate(self.constrs):
+                    constr.DualValue = dual_sol[i]
+        self.ObjVal = self.Obj.getValue()
```
chris-maes: Add slack computation here?

docs/cuopt/source/cuopt-python/lp-milp/lp-milp-examples.rst
```diff
@@ -0,0 +1,312 @@
...
+
+    # Set objective to maximize
+    problem.setObjective(5 * x + 3 * y, sense=sense.MAXIMIZE)
+
+    # Configure solver settings with callback
+    settings = SolverSettings()
+    # Set the incumbent callback
+    incumbent_callback = IncumbentCallback()
+    settings.set_mip_callback(incumbent_callback)
+    settings.set_parameter("time_limit", 30)  # Allow enough time to find multiple incumbents
```
chris-maes: Let's not use a bare string here. Let's use `CUOPT_TIME_LIMIT` instead.

python/cuopt/cuopt/linear_programming/problem.py
Iroy30: Moved

python/cuopt/cuopt/linear_programming/problem.py
Iroy30: Added slack computation in post solve. Slack, Var.Value, RC and DualValue are all reset when either a new Var or Constr or Obj is added

docs/cuopt/source/cuopt-python/lp-milp/lp-milp-examples.rst
ramakrishnap-nv: @Iroy30 can you please address this review

ramakrishnap-nv: And also other places where we use time_limit? And may be also update test case to do  same thing.

python/cuopt/cuopt/linear_programming/problem.py
Iroy30: set to nan

docs/cuopt/source/cuopt-python/lp-milp/lp-milp-examples.rst
Iroy30: that looks to be the only place in docs

Issue #207, opened 2025-07-15
[BUG] Rare batched PDLP CI bug


Kh4ster:
In CI the Python test "test_parser_and_batch_solver" sometimes fails with
```
a pure virtual method called
terminate called without an active exception
Fatal Python error: Aborted
```

As seen here: https://github.com/NVIDIA/cuopt/actions/runs/16266809031/job/45925573387?pr=201

PR #225, opened 2025-07-23
No cuda graph in batched pdlp

Kh4ster:
Temporarly disable cuda graph in batched PDLP until we find the root cause.

This allows use to re-enable the tests on the Python side

Closes https://github.com/NVIDIA/cuopt/issues/207


akifcorduk (approved):
Thanks Nicolas! Could you enable the python tests again? (and cpp if there are any)

Kh4ster:
@rgsl888prabhu done!

Issue #197, opened 2025-07-09
[BUG] list of parameters in solver_settings.py and solver_parameters.pyx do not match


tmckayus:
The list of parameters imported in solver_settings does not match the full list in solver_parameters.pyx
I don't believe this will prevent values from being set given how the code is written at present (however it might break in the future depending on how this is modified), but it does appear that the toDict() routine will not return the full set of parameters.

Is there a reason these lists are different, or was this just an oversight? The complete list from solver_parameters.pyx is called out in the lp-milp settings doc.

Also, if there is a discrepancy here there may also be one in the client-side solver_settings structure

https://github.com/NVIDIA/cuopt/blob/branch-25.05/python/cuopt/cuopt/linear_programming/solver/solver_parameters.pyx#L74
https://github.com/NVIDIA/cuopt/blob/branch-25.05/python/cuopt/cuopt/linear_programming/solver_settings/solver_settings.py#L18

rgsl888prabhu:
@tmckayus Should we close this ?

Issue #193, opened 2025-07-08
[BUG] Fix naming issue with mip and lp solver settings


rgsl888prabhu:
**Describe the bug**
Thin client uses different naming compared to what is there in SDK, for example

| Thin client | SDK |
|----------------|-------|
| relative_mip_gap | mip_relative_gap |
| absolute_mip_gap | mip_absolute_gap |

**Steps/Code to reproduce bug**

https://github.com/rapidsai/cuopt/blob/6499e52f8d48795b3373c14825d0f9b00cf84313/python/cuopt_self_hosted/cuopt_sh_client/thin_client_solver_settings.py#L207C13-L207C29

**Expected behavior**

Names should be same

tmckayus:
Thanks @rgsl888prabhu, for now I have mapped this in the service so that the service API does not change.

PR #227, opened 2025-07-23
align service api with SDK (compat with deprecations)

tmckayus:
The service API is not aligned completely with the SDK API from 25.05 in relation to solver configs. This change brings them into alignment with deprecation notices and backward compatibility

### DIFF
--- a/python/cuopt_server/cuopt_server/utils/solver.py
+++ b/python/cuopt_server/cuopt_server/utils/solver.py
@@ -88,7 +88,7 @@ def solve_LP_sync(
 
     begin_time = time.time()
 
-    if type(LP_data) is list:
+    if isinstance(LP_data, list):
         for i_data in LP_data:
             validate_LP_data(i_data)
     else:
@@ -100,7 +100,13 @@ def solve_LP_sync(
     if not validation_only:
         # log_file setting is ignored in the service,
         # instead we control it and use it as the basis for callbacks
-        if solver_logging:
+        if isinstance(LP_data, list):
+            # clear log_file setting for all because
+            # we don't support callbacks for batch mode
+            # and otherwise we ignore log_file
+            for i_data in LP_data:
+                i_data.solver_config.log_file = ""
+        elif solver_logging:
             log_dir, _, _ = settings.get_result_dir()
             log_fname = "log_" + reqId
             log_file = os.path.join(log_dir, log_fname)


tmckayus:
closes #193
closes #197

tmckayus:
realized that on the clientside, ThinclientSolverSettings also needs to be fixed up

python/cuopt_server/cuopt_server/utils/linear_programming/solver.py
```diff
@@ -156,7 +165,14 @@ def create_solver(LP_data, warmstart_data):
                 CUOPT_INFEASIBILITY_DETECTION,
                 solver_config.infeasibility_detection,
             )
-        if solver_config.pdlp_solver_mode is not None:
+        if solver_config.solver_mode is not None:
```
ramakrishnap-nv, line 173: May be switch if and elif so we can remove elif once we remove support. I think you have done it for others, this might have missed.

tmckayus: There is a problem with solver_mode and heuristics_only (only these two, the tolerances are all fine). The parameters have defaults that are non-Null. I set the deprecated names to a default of None so that we could know if they are explicitly used. If we test the new parameters first, we'll never know if we are just seeing the default value or not and we will never honor the case where someone sets the legacy parameter to a value.

tmckayus:
> CUOPT_MIP_HEURISTICS_ONLY needs to be updated in docs section.
> 
> Also lets add tests to make sure the naming changes are in good shape.

ok fixed the doc and added a dataset that sets all solver_config values for LP to make sure none are rejected
Testing the precedence of current vs deprecated I think is overkill.

python/cuopt_server/cuopt_server/webserver.py
```diff
@@ -957,7 +957,9 @@ async def postrequest(
     ),
     solver_logs: Optional[bool] = Query(
         default=False,
-        description="If set to True, MIP problems will produce detailed solver logs that can be retrieved from /cuopt/log/{id}",  # noqa
+        description="If set to True, MIP problems will produce detailed solver logs that can be retrieved from /cuopt/log/{id}. "  # noqa
```
ramakrishnap-nv: Lets add this in FAQ in documentation and also in server function doc.

rgsl888prabhu (commented):
Awesome @tmckayus, changes look good.

Have few minor doc related suggestions, I am approving the PR. So you can push those changes and merge once it is ready.

PR #231, opened 2025-07-24
MCPDLP

Kh4ster:
This is still under a lot of work.

This is to allow preliminary reviews.


cpp/src/linear_programming/utilities/batched_dot_product_handler.cuh
```diff
@@ -0,0 +1,75 @@
...
+
+#include <raft/linalg/detail/cublas_wrappers.hpp>
+#include <raft/core/handle.hpp>
+
+#include <vector>
+
+namespace cuopt::linear_programming::detail {
+
+// This class is used to start a batched dot product
+// With large problem size (>10K) and small batch size (<100), this is faster than using Segmented Reduce
```
aliceb-nv: Come to think of it I'm not surprised, iirc SegmentedReduce does a 1 block:1 segment mapping and in your case that's pretty terrible, I'm not surprised parallel device-wide blasdot calls beats it

aliceb-nv: Although I just realized they added a new overload optimized for fixed sizes, I wasn't aware of it, maybe this performs better?
https://github.com/NVIDIA/cccl/pull/3969

Kh4ster: Very good catch!! I will test that right away. It might make my life way simpler

Kh4ster: This is still slower than using multiple dot products :(

aliceb-nv: Dang .-.
Looking at their benchmarks they only test for segment sizes up to 1024 so I guess they don't optimize whatsoever for few-segments scenarios. Would be nice if they said so in their docs!

rgsl888prabhu:
@Kh4ster Shall we move this to 25.10

chris-maes:
@Kh4ster is this fine to move out of the 25.12 release?

PR #234, opened 2025-07-25
Add root node presolve using Papilo

hlinsen:
This PR implements:
- Fetching of Papilo headers through Cmake
- Adapters to presolve/postolve before/after running the solver
- Runtime boolean parameter to enable the feature (true by default)

PR #235, opened 2025-07-25
Fix assert failure on MIP problem reducing to LP after presolve

### LINKED ISSUE #233: [BUG] In assert mode, with a given dataset, MIP fails
**Describe the bug**
This assert is triggered in CI and fails with attached example, this happens in assert mode.

https://github.com/NVIDIA/cuopt/blob/branch-25.08/cpp/src/mip/problem/problem.cu#L1211
```
python: /__w/cuopt/cuopt/cpp/src/mip/problem/problem.cu:1211: void cuopt::linear_programming::detail::problem_t<i_t, f_t>::remove_given_variables(cuopt::linear_programming::detail::problem_t<i_t, f_t>&, rmm::device_uvector<f_t>&, rmm::device_uvector<T>&, const raft::handle_t*) [with i_t = int; f_t = double]: Assertion `variable_map.size() < n_variables&& "Too many variables to fix"' failed.
```

[test.zip](https://github.com/user-attachments/files/21419712/test.zip)

Enable assert mode while building locally
Run attached python example 

**Expected behavior**
Test should pass

aliceb-nv:
The MIP codepath assumes the current problem possesses a non-zero number of integer variables.
This is normally never the case as LP problems are directed to the LP codepath, however some trivial MIP instances lose their integer variables after presolve.

In those cases the problem is solved at the root relaxation, this PR simply fixes assertions and checks to ensure such instances don't trigger failures on the MIP heuristics side.

Also fixes an issue where initial solutions were scaled regardless of the mip scaling setting

Fixes #233


akifcorduk (commented):
I would not run MIP at all if the number of integers is zero but run LP with it and exit. For now you could just run PDLP.

rgsl888prabhu:
@kaatish @hlinsen May I get your review on this PR ?

cpp/src/mip/solver.cu
```diff
@@ -124,6 +127,27 @@ solution_t<i_t, f_t> mip_solver_t<i_t, f_t>::run_solver()
...
   }
 
+  // if the problem was reduced to a LP: run concurrent LP
+  if (context.problem_ptr->n_integer_vars == 0) {
+    CUOPT_LOG_INFO("Problem reduced to a LP, running concurrent LP");
+    pdlp_solver_settings_t<i_t, f_t> settings{};
+    settings.time_limit = timer_.remaining_time();
+    settings.method     = method_t::Concurrent;
+
+    auto opt_sol = solve_lp_with_method<i_t, f_t>(
```
hlinsen, line 137: Calling LP with `solve_lp_with_method` instead of `solve_lp` means we would not get the initial LP problem checks and presolve being added in https://github.com/NVIDIA/cuopt/pull/234.
Should we use it instead? the rest looks good to me!

Kh4ster: I also think we should be calling solve_lp and pass down the corresponding settings to use concurrent unless I missed something

aliceb-nv: Problem checks should be superfluous as they're already performed on the presolved problem. Additionally, the checks in `solve_lp` are only done on the original problem structure (which has already been checked at the time of calling solve_mip)
`solve_lp` requires passing either a `mps_data_model` or the original `optimization_problem_t` instead of the presolved problem, I used `solve_lp_with_method` to avoid making changes to `solve_lp`. Do you think I should add an overload instead?

rgsl888prabhu:
Lets try to merge this PR, since few other PRs are waiting on this.

cpp/src/mip/solver.cu
hlinsen: I think another overload would over complexify things at this point. I prefer sticking with this approach. What we could do is move the try/catch and presolve in solve_lp_with_method in some other PR.

rgsl888prabhu:
Thank you @aliceb-nv @hlinsen

Issue #247, opened 2025-07-29
[BUG] File might not compile with RAFT_CUDA_CHECK() enabled


vitor1001:
There is some code at assignment_hash_map.cu ([link](https://github.com/NVIDIA/cuopt/blob/455bdb0b4f99b64e978f01b44fe37d7dd558308a/cpp/src/mip/diversity/assignment_hash_map.cu#L100)) that has the following line:
```c++
  RAFT_CHECK_CUDA(handle_ptr->get_stream());
```

The issue is that there is no `handle_ptr` variable in the context this was called, it should be `solution.handle_ptr`. Either there is a global variable somewhere called `handle_ptr` (probably a bad idea), or RAFT_CHECK_CUDA is never activated on continuous integration.

rgsl888prabhu:
@kaatish @hlinsen for viz

rgsl888prabhu:
I have assigned it to @akifcorduk since the code changes were added by him, but if any one else has bandwidth lets look at it.

vitor1001:
@rgsl888prabhu do you want me to file two different issues - one for the code that doesn't compile, and another for the CI tests passing nevertheless?

rgsl888prabhu:
I think one should be enough

PR #248, opened 2025-07-29
Unset NDEBUG in cmake in assert mode

hlinsen:
Raft defines `RAFT_CHECK_CUDA` depending on if we run in release/Debug mode https://github.com/rapidsai/raft/blob/branch-25.06/cpp/include/raft/util/cuda_rt_essentials.hpp#L78-L82.
In cuopt we have a special mode "assert" that undefines `NDEBUG` to run with optimizations but still enable runtime checks. https://github.com/rapidsai/raft/blob/branch-25.06/cpp/include/raft/util/cuda_rt_essentials.hpp#L78-L82

The first version of `RAFT_CHECK_CUDA` uses the stream argument therefore in debug mode and assert mode we should catch that error. However unsetting `NDEBUG` in our header results in `NDEBUG` being undefined in cuopt but still defined for our dependencies. So we were calling the wrong implementation of `RAFT_CUDA_CHECK`. To undefine `NDEBUG` in both cuopt and our dependencies we have to do it through cmake.
This catches the compile error.

Closes: https://github.com/NVIDIA/cuopt/issues/247

### DIFF
--- a/cpp/CMakeLists.txt
+++ b/cpp/CMakeLists.txt
@@ -134,6 +134,12 @@ elseif(CMAKE_CUDA_LINEINFO)
   set(CMAKE_CUDA_FLAGS_RELEASE "${CMAKE_CUDA_FLAGS_RELEASE} -lineinfo")
 endif(CMAKE_BUILD_TYPE MATCHES Debug)
 
+# Undefine NDEBUG if assert mode is on
+if(DEFINE_ASSERT)
+  message(STATUS "Undefining NDEBUG with assert mode enabled")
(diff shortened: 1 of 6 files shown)


rgsl888prabhu (approved):

aliceb-nv:
I pushed a fix for the failing cpp tests

hlinsen:
> I pushed a fix for the failing cpp tests

Thanks Alice!

aliceb-nv (approved):
Very good catch with this PR Hugo, thanks! That will help us be much more confident about our robustness :)

PR #256, opened 2025-08-05
Add cudss as a dependency

### LINKED ISSUE #292: [FEA] cuDSS as a cuOpt dependency

rgsl888prabhu:
## Description
Add cudss as a dependency

## Issues
closes #292 

## Checklist

- Testing
   - [ ] Added tests
   - [x] NA
- Documentation
   - [ ] Added new documentation
   - [x] NA

### DIFF
--- a/benchmarks/linear_programming/utils/get_datasets.py
+++ b/benchmarks/linear_programming/utils/get_datasets.py
@@ -16,6 +16,8 @@
 import os
 import argparse
 import urllib.request
+import urllib.parse
+import ssl
 import subprocess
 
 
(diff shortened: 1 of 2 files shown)


ci/build_wheel_libcuopt.sh
```diff
@@ -31,6 +31,15 @@ else
     echo "Building in release mode"
 fi
 
+# Install cudss
+if command -v dnf &> /dev/null; then
+    dnf install libcudss-dev -y
```
tmckayus: are we certain that this runs on a redhat variant?  Is it possible we may run on something with "apt" ?

ramakrishnap-nv: Yes, currently we are running on Rocky linux, but I agre, let me update it such a way that this will switch to apt-get in case we switch to ubuntu.

PR #258, opened 2025-08-06
FP as a local search

akifcorduk:
This PR adds a new local search operator: FP with objective cutoff.

We were using FP as a means of solution generation algorithm. Now we are also using it as a local search by adding a cutting plane at the current fesible point perpendicular to the objective. We rerun FJ at this point to have a different restart point and continue the FP search with the cutting plane added.

This on brings down the average optimality gap by 1-2%.


cpp/src/mip/feasibility_jump/feasibility_jump.cu
```diff
@@ -241,7 +241,7 @@ template <typename i_t, typename f_t>
 void fj_t<i_t, f_t>::copy_weights(const weight_t<i_t, f_t>& weights,
                                   const raft::handle_t* handle_ptr)
 {
-  cuopt_assert(cstr_weights.size() == weights.cstr_weights.size(), "Size mismatch");
+  cstr_weights.resize(weights.cstr_weights.size(), handle_ptr->get_stream());
```
aliceb-nv: The weights corresponding to new constraints will be left uninitialized, they should probably be filled to 1

cpp/src/mip/local_search/local_search.cuh
```diff
@@ -25,7 +25,8 @@
 
 namespace cuopt::linear_programming::detail {
 
-enum ls_method_t { FJ_ANNEALING = 0, FJ_LINE_SEGMENT, RANDOM };
+// make sure RANDOM is always the last
+enum ls_method_t { FJ_ANNEALING = 0, FJ_LINE_SEGMENT, FP_SEARCH, RANDOM };
```
aliceb-nv: I think we should use an enum class instead (since 'RANDOM' is likely to cause clashes in the global namespace), or prefix the enum values
Additionally, we could add a last enum entry `LS_METHODS_SIZE = RANDOM` which would be clearer than using `RANDOM` directly as the enum size

cpp/src/mip/local_search/local_search.cu
```diff
@@ -107,14 +108,19 @@ bool local_search_t<i_t, f_t>::run_local_search(solution_t<i_t, f_t>& solution,
   fj.set_fj_settings(fj_settings);
   fj.copy_weights(weights, solution.handle_ptr);
   bool is_feas;
-  i_t rd = std::uniform_int_distribution(0, 1)(rng);
+  i_t rd = std::uniform_int_distribution(0, 2)(rng);
```
aliceb-nv: We could use the enum values (e.g. ls_method_t::FJ_ANNEALING to ls_method_t::FP_SEARCH) directly as the bounds of the distribution

cpp/src/mip/problem/problem.cu
```diff
@@ -1590,6 +1590,34 @@ f_t problem_t<i_t, f_t>::get_user_obj_from_solver_obj(f_t solver_obj)
...
+  auto h_objective_coefficients = cuopt::host_copy(objective_coefficients);
+  handle_ptr->sync_stream();
+  std::vector<i_t> var_indices;
+  std::vector<f_t> constr_coeffs;
+  for (i_t i = 0; i < n_variables; ++i) {
+    if (h_objective_coefficients[i] != 0) {
+      var_indices.push_back(i);
+      constr_coeffs.push_back(h_objective_coefficients[i]);
+    }
+  }
```
aliceb-nv: We could compute and store the indices of variables with nonzero obj coefficients at problem construction, it might be useful to other heuristics

cpp/src/mip/problem/problem.cu
```diff
@@ -1590,6 +1590,34 @@ f_t problem_t<i_t, f_t>::get_user_obj_from_solver_obj(f_t solver_obj)
...
+  h_constraints.add_constraint(
+    var_indices, constr_coeffs, -std::numeric_limits<f_t>::infinity(), objective);
+  insert_constraints(h_constraints);
+  compute_transpose_of_problem();
```
aliceb-nv, line 1628: Should we err on the side of caution and add a `cuopt_func_call(check_problem_representation(true))` ? It may catch issues in the future as we add more invariants to problem_t

aliceb-nv (commented):
Just a few comments :) Thanks a lot for the great work Akif! Awesome to see further quality improvements
Additionally: Shouldn't this PR be merged to the branch-25.10 branch instead?
I'll approve once this is confirmed

cpp/src/mip/feasibility_jump/feasibility_jump.cu
akifcorduk: Good catch!

cpp/src/mip/local_search/local_search.cuh
akifcorduk: Good catch!

aliceb-nv (approved):
Approving, thanks for the great work Akif!

anandhkb:
GREAT work @akifcorduk 
@afender for viz

PR #259, opened 2025-08-06
Sub-MIP recombiner and B&B global variable changes

akifcorduk:
This PR adds a new recombiner: sub-MIP. We use the B&B solver to solve the subproblem with the time limit of the recombiner. This PR also runs both local searches (line segment and FJ) one after another, instead of running either of them with 50% probability.

This PR also changes all the global variables in B&B to class member variables. This way, multiple B&B instances could run in parallel in the same process, including the sub-MIP recombiner.


benchmarks/linear_programming/cuopt/mip_test_instances.hpp
```diff
@@ -17,66 +17,175 @@
...
-  // "roll3000.mps",
-  "s100.mps",
-  // "supportcase19.mps",
-  // "supportcase22.mps",
-  "swath1.mps"
-  // "swath3.mps",
-  // "timtab1.mps",
-  // "triptim1.mps"
-};
+std::vector<std::string> instances = {"30n20b8.mps",
```
aliceb-nv, line 20: Dumb question: why this subset of MIPLIB2017 in particular?

cpp/src/mip/diversity/diversity_config.hpp
```diff
@@ -19,27 +19,26 @@
...
+static int maximum_island_size                   = 8;
+static bool use_avg_diversity                    = false;
+static double generation_time_limit_ratio        = 0.6;
+static double max_island_gen_time                = 600;
+static size_t n_sol_for_skip_init_gen            = 3;
+static double max_fast_sol_time                  = 10;
+static double lp_run_time_if_feasible            = 15.;
+static double lp_run_time_if_infeasible          = 1;
+static bool halve_population                     = true;
+};  // namespace diversity_config_t
```
aliceb-nv: Was the loss of the const/constexpr qualifier intended?
Mutable global variables may cause issues

cpp/src/mip/diversity/recombiners/recombiner_stats.hpp
```diff
@@ -21,7 +21,7 @@
 
 namespace cuopt::linear_programming::detail {
 
-enum recombiner_enum_t : int { BOUND_PROP = 0, FP, LINE_SEGMENT, SIZE };
+enum recombiner_enum_t : int { BOUND_PROP = 0, FP, LINE_SEGMENT, SUB_MIP, SIZE };
```
aliceb-nv: This should probably be an enum class as well, those are some very common tokens which shouldn't be left in the global namespace in a header (or those enum values should receive a prefix)

cpp/src/mip/diversity/recombiners/sub_mip.cuh
```diff
@@ -0,0 +1,140 @@
...
+    }
+    i_t n_vars_from_guiding = a.problem_ptr->n_integer_vars - n_vars_from_other;
+    if (n_vars_from_other == 0 || n_vars_from_guiding == 0) {
+      CUOPT_LOG_DEBUG("Returning false because all vars are common or different");
+      return std::make_pair(offspring, false);
+    }
+    CUOPT_LOG_DEBUG(
+      "n_vars_from_guiding %d n_vars_from_other %d", n_vars_from_guiding, n_vars_from_other);
+    this->compute_vars_to_fix(offspring, vars_to_fix, n_vars_from_other, n_vars_from_guiding);
+    auto [fixed_problem, fixed_assignment, variable_map] = offspring.fix_variables(vars_to_fix);
```
hlinsen, line 81: Do you know if it would cause size issues if we are to remove fixed variable or apply presolve on the fixed problem?

cpp/src/dual_simplex/branch_and_bound.cpp
```diff
@@ -207,13 +163,13 @@ void graphviz_edge(const simplex_solver_settings_t<i_t, f_t>& settings,
...
-f_t get_upper_bound()
+template <typename i_t, typename f_t>
+f_t branch_and_bound_t<i_t, f_t>::get_upper_bound()
 {
-  global_variables::mutex_upper.lock();
-  const f_t upper_bound = global_variables::upper_bound;
-  global_variables::mutex_upper.unlock();
-  return upper_bound;
+  mutex_upper.lock();
+  const f_t upper_bound_ = upper_bound;
```
chris-maes: Thanks for making these changes @akifcorduk !

One nitpick. We have local variables for lower bound, upper bound, and gap. As well as the member variables that are shared between threads. 

I think you used the convention of adding an underscore suffix to local variables (i.e. `upper_bound_` or `lower_bound_`). I'm used to the exact opposite convention---that member variable have an underscore suffix. So I read `lower_bound_` as a member variable of the class.

To avoid confusion, would you be ok if we used the underscore prefix for the member variables?

cpp/src/dual_simplex/pseudo_costs.hpp
```diff
@@ -49,13 +51,18 @@ class pseudo_costs_t {
...
   void update_pseudo_costs_from_strong_branching(const std::vector<i_t>& fractional,
-                                                 const std::vector<f_t>& root_soln,
-                                                 const std::vector<f_t>& strong_branch_down,
-                                                 const std::vector<f_t>& strong_branch_up);
+                                                 const std::vector<f_t>& root_soln);
   std::vector<f_t> pseudo_cost_sum_up;
   std::vector<f_t> pseudo_cost_sum_down;
   std::vector<i_t> pseudo_cost_num_up;
   std::vector<i_t> pseudo_cost_num_down;
+  // Global variable storing the results of strong branching down
```
chris-maes: Remove `Global` from comment?

cpp/src/dual_simplex/pseudo_costs.hpp
```diff
@@ -49,13 +51,18 @@ class pseudo_costs_t {
...
+  std::vector<f_t> strong_branch_down;
+  // Global variable storing the results of strong branching up
```
chris-maes: Remove `Global` from comment

cpp/src/dual_simplex/branch_and_bound.cpp
aliceb-nv: I agree with Chris :) To me at a first glance, prefix or suffix underscores suggest a member variable

cpp/src/dual_simplex/pseudo_costs.cpp
```diff
@@ -112,7 +96,7 @@ void strong_branch_helper(i_t thread_id,
       }
 
       if (branch == 0) {
-        global_variables::strong_branch_down[k] = obj - root_obj;
+        pc.strong_branch_down[k] = obj - root_obj;
```
chris-maes, line 99: I assume it is ok to have multiple threads access different indices of the same vector. If not maybe we need a lock here. Note I didn't use a lock here initially, so it's probably ok. Just double checking.

cpp/src/mip/diversity/recombiners/sub_mip.cuh
```diff
@@ -0,0 +1,140 @@
...
+      branch_and_bound_solution.resize(branch_and_bound_problem.num_cols);
+      // Fill in the settings for branch and bound
+      branch_and_bound_settings.time_limit = sub_mip_recombiner_config_t::sub_mip_time_limit;
+      branch_and_bound_settings.print_presolve_stats = false;
+      branch_and_bound_settings.absolute_mip_gap_tol = context.settings.tolerances.absolute_mip_gap;
+      branch_and_bound_settings.relative_mip_gap_tol = context.settings.tolerances.relative_mip_gap;
+      branch_and_bound_settings.integer_tol = context.settings.tolerances.integrality_tolerance;
+      // disable B&B logs, so that it is not interfering with the main B&B thread
+      branch_and_bound_settings.log.log = false;
+      dual_simplex::branch_and_bound_t<i_t, f_t> branch_and_bound(branch_and_bound_problem,
```
chris-maes, line 109: I think you should add a callback here. Since any feasible solution in the sub-MIP should be a feasible solution in the original problem. So you want to propagate solutions out of the sub-MIP.

cpp/src/mip/diversity/recombiners/sub_mip.cuh
```diff
@@ -0,0 +1,140 @@
...
+    dual_simplex::mip_solution_t<i_t, f_t> branch_and_bound_solution(1);
+    if (run_sub_mip) {
+      // run sub-mip
+      namespace dual_simplex = cuopt::linear_programming::dual_simplex;
+      dual_simplex::user_problem_t<i_t, f_t> branch_and_bound_problem;
+      dual_simplex::simplex_solver_settings_t<i_t, f_t> branch_and_bound_settings;
+      fixed_problem.get_host_user_problem(branch_and_bound_problem);
+      branch_and_bound_solution.resize(branch_and_bound_problem.num_cols);
+      // Fill in the settings for branch and bound
+      branch_and_bound_settings.time_limit = sub_mip_recombiner_config_t::sub_mip_time_limit;
```
chris-maes, line 98: Probably instead of, or in addition to, a time limit, we should add a node limit to branch and bound

cpp/src/mip/diversity/diversity_manager.cu
```diff
@@ -782,11 +784,19 @@ std::pair<solution_t<i_t, f_t>, bool> diversity_manager_t<i_t, f_t>::recombine(
...
     if (success) { recombine_stats.add_success(); }
     return std::make_pair(sol, success);
+  } else if (recombiner == recombiner_enum_t::SUB_MIP) {
+    auto [sol, success] = sub_mip_recombiner.recombine(a, b, population.weights);
+    recombine_stats.stop_recombiner_time();
+    if (success) { recombine_stats.add_success(); }
+    return std::make_pair(sol, success);
+  } else {
+    CUOPT_LOG_ERROR("Invalid recombiner type: %d", recombiner);
+    return std::make_pair(solution_t<i_t, f_t>(a), false);
```
aliceb-nv: Nitpick of debatable importance: using an switch statement would allow the compiler to emit a compilation error (-Werror) if a recombiner type is missed, instead of a runtime error log

cpp/src/mip/diversity/recombiners/sub_mip.cuh
akifcorduk: What do you mean by "removing fixed variable"?

I don't think it would cause any problems if we applied presolve as long as the conversions are done correctly.

cpp/src/dual_simplex/branch_and_bound.cpp
akifcorduk: Actually, I wanted to get rid of this local variables for reading. I wanted to convert all global/shared variables into atomic. I can do it in this PR, or later in another PR. What's your preference @chris-maes ?

cpp/src/mip/diversity/recombiners/sub_mip.cuh
hlinsen: If we can change the size of the CSR matrix

cpp/src/mip/diversity/recombiners/sub_mip.cuh
akifcorduk: Why is this a case? Does returning early provide any value? Or does the B&B have diminishing returns in submip context when we run it more than certain number of nodes?

benchmarks/linear_programming/cuopt/mip_test_instances.hpp
akifcorduk: These are the instances that are not close to optimal (<5%) and have at least 5 solutions generated. This is for heuristics improvement tests.

cpp/src/mip/diversity/diversity_config.hpp
akifcorduk: Yes this was intended. We are changing those constants depending on a environment config variable. They are wrapped in a namespace, so i doubt they will cause issues.

cpp/src/dual_simplex/pseudo_costs.cpp
akifcorduk: I think it is okay since different indices are accessed.

cpp/src/mip/diversity/recombiners/sub_mip.cuh
akifcorduk: Good catch. I wanted to implement getting the best primal solution, this way we can get all intermediate solutions.

cpp/src/mip/diversity/diversity_config.hpp
aliceb-nv: Okay, makes sense!
Although, ideally to make the semantics clearer I'd see this as a separate struct type, marked as a 'const' member and properly initialized at solver construction
Even if in this context it is unlikely to cause issues, global mutable variables are just inherently alarming especially in a large multithreaded codebase

aliceb-nv: If you keep them as global: please mark them as 'static inline' or 'extern'. As it stands, a static variable declared in a header will be instantiated with internal linkage in each separate translation unit; so if one of them is modified in a given .cu file, other .cu files won't see this change

cpp/src/mip/diversity/recombiners/sub_mip.cuh
akifcorduk: We are changing the size of the CSR matrix already. That's the fixed_problem.

cpp/src/mip/diversity/diversity_config.hpp
akifcorduk: Okay, i like the idea of member, i will do that.

cpp/src/dual_simplex/branch_and_bound.cpp
akifcorduk: Handled it, let's do the atomic changes in another PR.

chris-maes: Sounds great. Thanks

cpp/src/mip/diversity/recombiners/sub_mip.cuh
chris-maes: As we discussed offline, I think using a node limit just makes things more deterministic. We can switch later when the MIP heuristics are deterministic. Fine to use time limit for now.

PR #286, opened 2025-08-11
Merge update version and fix version format bugs

### LINKED ISSUE #262: [BUG] Update version script is not covering all the version format
**Describe the bug**
- [ ] Fix issue with alpha version tag, while releasing it is not removing alpha tags
- [ ] Doc versioning version1.json should get updated only once and shouldn't update again, also fix the issue with broken format.

rgsl888prabhu:
## Description
Merge update version which was earlier split was cuOpt and Rapids into one since we are in sync with versions now. 
And also fix failure in version updates and bugs around it.

## Issue
closes #262 

## Checklist

- Testing
   - [ ] Added tests
   - [x] NA
- Documentation
   - [ ] Added new documentation
   - [x] NA

### DIFF
--- a/.github/workflows/inactivity_reminder.yaml
+++ b/.github/workflows/inactivity_reminder.yaml
@@ -1,0 +1,110 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+name: Inactivity Reminder with Different Times
+
+on:
+  schedule:
+    - cron: '0 9 * * *'  # Runs daily at 09:00 UTC
+
+jobs:
+  remind:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Remind inactive issues and PRs
+        uses: actions/github-script@v6
+        with:
+          script: |
+            const MS_IN_DAY = 24 * 60 * 60 * 1000;
+            const now = new Date();
+
+            // Thresholds
+            const ISSUE_INACTIVITY_DAYS = 21; // 3 weeks
+            const PR_INACTIVITY_DAYS = 7;     // 1 week
+
+            // Always notify this user
+            const defaultNotify = '@anandhkb';
+
+            function isInactive(updatedAt, thresholdDays) {
+              const updatedDate = new Date(updatedAt);
+              return (now - updatedDate) > thresholdDays * MS_IN_DAY;
+            }
+
+            // Fetch all open issues + PRs (paginate if >100)
+            const issuesAndPRs = await github.paginate(
+              github.rest.issues.listForRepo,
+              {
+                owner: context.repo.owner,
+                repo: context.repo.repo,
+                state: 'open',
+                per_page: 100,
+              }
+            );
+
+            for (const item of issuesAndPRs) {
+              // Skip if issue/PR has the skip inactivity reminder label
+              if (item.labels && item.labels.some(label => label.name === 'skip inactivity reminder')) {
+                console.log(`Skipping #${item.number} due to skip inactivity reminder label`);
+                continue;
+              }
+
+              const isPR = !!item.pull_request;
+
+              // Skip if PR is in Draft mode
+              if (isPR && item.draft) {
+                console.log(`Skipping #${item.number} due to Draft PR status`);
+                continue;
+              }
+
+              // Skip if issue (not PR) is an EPIC
+              if (!isPR && item.labels && item.labels.some(label => label.name === 'epic')) {
+                console.log(`Skipping #${item.number} due to EPIC label (issue only)`);
+                continue;
+              }
+
+              const thresholdDays = isPR ? PR_INACTIVITY_DAYS : ISSUE_INACTIVITY_DAYS;
+
+              if (isInactive(item.updated_at, thresholdDays)) {
+                // For issues, only send reminder if they have "awaiting response" label
+                if (!isPR && (!item.labels || !item.labels.some(label => label.name === 'awaiting response'))) {
+                  console.log(`Skipping #${item.number} (issue) - no awaiting response label`);
+                  continue;
+                }
+
+                const assigneeMentions = item.assignees && item.assignees.length > 0
+                  ? item.assignees.map(a => `@${a.login}`).join(' ')
+                  : '';
+                const mentions = assigneeMentions
+                  ? `${defaultNotify} ${assigneeMentions}`.trim()
+                  : defaultNotify;
+                const issueNumber = item.number;
+                const type = isPR ? "pull request" : "issue";
+
+                await github.rest.issues.createComment({
+                  owner: context.repo.owner,
+                  repo: context.repo.repo,
+                  issue_number: issueNumber,
+                  body: [
+                    `🔔 Hi ${mentions}, this ${type} has had no activity for ${thresholdDays} days. Please update or let us know if it can be closed. Thank you!`,
+                    '',
+                    'If this is an "epic" issue, then please add the "epic" label to this issue.',
+                    'If it is a PR and not ready for review, then please convert this to draft.',
+                    'If you just want to switch off this notification, then use the "skip inactivity reminder" label.'
+                  ].join('\n'),
+                });
+
+                console.log(`Posted reminder on #${issueNumber} (${type}) to ${mentions}`);
+              }
+            }
--- a/.github/workflows/issue_automation.yaml
+++ b/.github/workflows/issue_automation.yaml
@@ -1,0 +1,95 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+name: Auto-label and Round-Robin Assign Issues
+
+on:
+  issues:
+    types: [opened]
+
+jobs:
+  auto-label:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Add awaiting response label to new issues
+        uses: actions/github-script@v6
+        with:
+          script: |
+            // Only process issues (not PRs)
+            if (context.payload.issue && !context.payload.issue.pull_request) {
+              const issue = context.payload.issue;
+              const issueNumber = issue.number;
+
+              console.log(`Adding 'awaiting response' label to issue #${issueNumber}`);
+
+              await github.rest.issues.addLabels({
+                owner: context.repo.owner,
+                repo: context.repo.repo,
+                issue_number: issueNumber,
+                labels: ['awaiting response']
+              });
+
+              console.log(`Successfully added 'awaiting response' label to issue #${issueNumber}`);
+            } else {
+              console.log('Skipping - this is a pull request, not an issue');
+            }
+
+  round-robin-assign:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Assign issue round-robin only if unassigned
+        uses: actions/github-script@v6
+        with:
+          script: |
+            // Only process issues (not PRs)
+            if (context.payload.issue && !context.payload.issue.pull_request) {
+              const issue = context.payload.issue;
+              const issueNumber = issue.number;
+
+              // Round-robin assignment logic
+              const assignees = [
+                'kaatish',
+                'rg20',
+                'akifcorduk',
+                'hlinsen',
+                'Kh4ster',
+                'aliceb-nv',
+                'chris-maes',
+                'rgsl888prabhu',
+                'Iroy30',
+                'tmckayus'
+              ];
+
+              // Only assign if no one is assigned yet
+              if (!issue.assignees || issue.assignees.length === 0) {
+                const index = (issueNumber - 1) % assignees.length;
+                const assignee = assignees[index];
+
+                console.log(`Assigning issue #${issueNumber} to @${assignee}`);
+
+                await github.rest.issues.addAssignees({
+                  owner: context.repo.owner,
+                  repo: context.repo.repo,
+                  issue_number: issueNumber,
+                  assignees: [assignee],
+                });
+
+                console.log(`Successfully assigned issue #${issueNumber} to @${assignee}`);
+              } else {
+                console.log(`Issue #${issueNumber} already has assignees, skipping assignment.`);
+              }
+            } else {
+              console.log('Skipping - this is a pull request, not an issue');
+            }


tmckayus (approved):

rgsl888prabhu:
@jameslamb may I get your review on this

ci/release/update-version.sh
```diff
@@ -123,6 +123,49 @@ DEPENDENCIES=(
...
+done
+
+# RAPIDS-specific updates
+# RTD update for RAPIDS dependencies
+rapids_dependencies='cudf cudf-cu12 cuvs cuvs-cu12 libcudf librmm librmm-cu12 rmm rmm-cu12 librmm libraft-headers pylibraft pylibraft-cu12 raft-dask raft-dask-cu12 rapids-dask-dependency'
+for FILE in conda/environments/*.yaml dependencies.yaml; do
+    for dependency in ${rapids_dependencies}; do
+        sed_runner "s/\(${dependency}==\)[0-9]\+\.[0-9]\+/\1${NEXT_SHORT_TAG_PEP440}/" "${FILE}"
+    done
+done
```
jameslamb: This seems to be somewhat duplicating work that's done a few lines further down, like this:

```text
# Fixing RAPIDS dependencies and pyproject.toml
RAPIDS_DEPENDENCIES=(
  cudf
  cuvs
  librmm
  rmm
  pylibraft
  raft-dask
  rapids-dask-dependency
)

for DEP in "${RAPIDS_DEPENDENCIES[@]}"; do
  for FILE in dependencies.yaml conda/environments/*.yaml; do
    sed_runner "s/\(${DEP}==\)[0-9]\+\.[0-9]\+/\1${NEXT_SHORT_TAG_PEP440}/" "${FILE}"
  done
  for FILE in python/*/pyproject.toml; do
    sed_runner "s/\(${DEP}==\)[0-9]\+\.[0-9]\+/\1${NEXT_SHORT_TAG_PEP440}/" "${FILE}"
  done
done
```

Can you please consolidate them into just 1? And I think the later form, using a bash array, is preferable to this one using a long string.

ci/release/update-version.sh
```diff
@@ -60,37 +65,36 @@ sed_runner 's/'"\"client_version\": \"[0-9][0-9].[0-9][0-9]\""'/'"\"client_versi
...
 # Update quick-start docs for pip
-sed_runner "s/\(cuopt-cu12==\)[0-9]\+\.[0-9]\+\.\\*/\1${PY_NEXT_SHORT_TAG}.\*/g" docs/cuopt/source/cuopt-python/quick-start.rst
-sed_runner "s/\(libcuopt-cu12==\)[0-9]\+\.[0-9]\+\.\\*/\1${PY_NEXT_SHORT_TAG}.\*/g" docs/cuopt/source/cuopt-c/quick-start.rst
-sed_runner "s/\(cuopt-server-cu12==\)[0-9]\+\.[0-9]\+\.\\*/\1${PY_NEXT_SHORT_TAG}.\*/g" docs/cuopt/source/cuopt-server/quick-start.rst
-sed_runner "s/\(cuopt-sh-client==\)[0-9]\+\.[0-9]\+\.\\*/\1${PY_NEXT_SHORT_TAG}.\*/g" docs/cuopt/source/cuopt-server/quick-start.rst
+sed_runner "s/\(cuopt-cu12==\)[0-9]\+\.[0-9]\+\.\\*/\1${NEXT_SHORT_TAG_PEP440}.\*/g" docs/cuopt/source/cuopt-python/quick-start.rst
+sed_runner "s/\(libcuopt-cu12==\)[0-9]\+\.[0-9]\+\.\\*/\1${NEXT_SHORT_TAG_PEP440}.\*/g" docs/cuopt/source/cuopt-c/quick-start.rst
+sed_runner "s/\(cuopt-server-cu12==\)[0-9]\+\.[0-9]\+\.\\*/\1${NEXT_SHORT_TAG_PEP440}.\*/g" docs/cuopt/source/cuopt-server/quick-start.rst
+sed_runner "s/\(cuopt-sh-client==\)[0-9]\+\.[0-9]\+\.\\*/\1${NEXT_SHORT_TAG_PEP440}.\*/g" docs/cuopt/source/cuopt-server/quick-start.rst
```
jameslamb: This is a lot of very-similar-looking changes to the same 2 files.

Would you please consider consolidating them and looping over a list of library names? Similar to this:

https://github.com/NVIDIA/cuopt/blob/0845a21e0a867babdce73325d1bcf748a32ca9d4/ci/release/update-version-rapids.sh#L70-L87

ci/release/update-version.sh
jameslamb: Also now that these files are being combined, I don't think you need a separate list of "RAPIDS dependencies" vs. "cuopt dependencies".

So the update around line 43 that looks like this:

```shell
dependencies='libcuopt libcuopt-cu12 cuopt cuopt-cu12 cuopt-mps-parser cuopt-server cuopt-server-cu12 cuopt-sh-client'
for FILE in conda/environments/*.yaml dependencies.yaml; do
```

Should be consolidated here as well.

jameslamb (changes requested):
Now that `cuopt` and RAPIDS will have aligned versions, there is a lot of opportunity for consolidating things in this new `update-version.sh` script. I left some specific examples I saw where things should be grouped into a smaller set of updates.

That'll make it run a little faster, reduce the risk of inconsistencies, and reduce the effort to maintain this.

Please also test this. Something like this:

```shell
./ci/release/update-version.sh '25.12.00'

# does the diff look like what you'd expect?
git diff

# did it catch everything?
git grep -E '24\.|25\.'
```

rgsl888prabhu:
wanted to keep them separate in case if we need to be out of sync in future, but I think it can be done quite easily, will update PR as per your review

ci/release/update-version.sh
```diff
@@ -37,60 +38,83 @@ function sed_runner() {
...
+  pylibraft
+  pylibraft-cu12
+  raft-dask
+  raft-dask-cu12
+  rapids-dask-dependency
+)
+
+for DEP in "${DEPENDENCIES[@]}"; do
+  for FILE in dependencies.yaml conda/environments/*.yaml python/*/pyproject.toml; do
+    sed_runner "s/\(${DEP}==\)[0-9]\+\.[0-9]\+/\1${NEXT_SHORT_TAG_PEP440}/" "${FILE}"
```
jameslamb, line 61:
```suggestion
    sed_runner "/-.* ${DEP}\(-cu[[:digit:]]\{2\}\)\{0,1\}\(\[.*\]\)\{0,1\}==/ s/==.*/==${NEXT_SHORT_TAG_PEP440}.*,>=0.0.0a0/g" "${FILE}"
```

Let's please use this pattern used throughout the rest of RAPIDS ([cudf example](https://github.com/rapidsai/cudf/blob/4a309b436685c605404fd82983673c1e94b7cd8c/ci/release/update-version.sh#L66)) which has the following nice properties:

* handles `-cu12` suffixes with an optional group, so no need to hard-code a bunch of `-cu12` entries in `DEPENDENCIES` (which would soon also need `-cu13` entries!)
* adds alpha specs to dependencies, so nightlies can be considered

ci/release/update-version.sh
```diff
@@ -37,60 +38,83 @@ function sed_runner() {
...
+  librmm-cu12
+  libraft-headers
+  rmm
+  rmm-cu12
+  pylibraft
+  pylibraft-cu12
+  raft-dask
+  raft-dask-cu12
+  rapids-dask-dependency
+)
```
jameslamb:
```suggestion
DEPENDENCIES=(
  cudf
  cuvs
  cuopt
  cuopt-mps-parser
  cuopt-server
  cuopt-sh-client
  libcuopt
  libraft-headers
  librmm
  pylibraft
  raft-dask
  rapids-dask-dependency
  rmm
)
```

Let's please alphabetize this (so it's easier to read at a glance) and remove `-cu12` entries (see next comment).

ci/release/update-version.sh
```diff
@@ -37,60 +38,83 @@ function sed_runner() {
...
+  done
+  for FILE in docs/cuopt/source/*/quick-start.rst README.md; do
+    sed_runner "s/\(${DEP}==\)[0-9]\+\.[0-9]\+\.\\*/\1${NEXT_SHORT_TAG_PEP440}.\*/g" "${FILE}"
+  done
+  for FILE in docs/cuopt/source/*/quick-start.rst README.md; do
+    sed_runner "s/\(${DEP}=\)[0-9]\+\.[0-9]\+\(\.[0-9]\+\)\?[^ ]*/\1${NEXT_SHORT_TAG}.*/g" "${FILE}"
+  done
+  for FILE in docs/cuopt/source/*/quick-start.rst README.md; do
+    sed_runner "s/\(${DEP}:\)[0-9]\{2\}\.[0-9]\{1,2\}\.[0-9]\+\(-cuda[0-9]\+\.[0-9]\+-\)\(py[0-9]\+\)/\1${DOCKER_TAG}\2\3/g" docs/cuopt/source/cuopt-python/quick-start.rst
+  done
```
jameslamb:
```suggestion
  for FILE in docs/cuopt/source/*/quick-start.rst README.md; do
    sed_runner "s/\(${DEP}==\)[0-9]\+\.[0-9]\+\.\\*/\1${NEXT_SHORT_TAG_PEP440}.\*/g" "${FILE}"
    sed_runner "s/\(${DEP}=\)[0-9]\+\.[0-9]\+\(\.[0-9]\+\)\?[^ ]*/\1${NEXT_SHORT_TAG}.*/g" "${FILE}"
  done
  sed_runner "s/\(${DEP}:\)[0-9]\{2\}\.[0-9]\{1,2\}\.[0-9]\+\(-cuda[0-9]\+\.[0-9]\+-\)\(py[0-9]\+\)/\1${DOCKER_TAG}\2\3/g" docs/cuopt/source/cuopt-python/quick-start.rst
```

This is currently looping over the same set of files 3 times.

Let's consolidate 2 of those calls so we only loop over the files once, and run that additional line that only applies to  `docs/cuopt/source/cuopt-python/quick-start.rst` outside of the `for FILE` loop, since it only applies to one file.

ci/release/update-version.sh
```diff
@@ -100,29 +124,20 @@ sed_runner "/^set(cuopt_version/ s/[0-9][0-9].[0-9][0-9].[0-9][0-9]/${NEXT_FULL_
...
+  # CI image tags of the form {rapids_version}-{something}
+  sed_runner "s/:[0-9]*\\.[0-9]*-/:${NEXT_SHORT_TAG}-/g" "${FILE}"
 done
+
+# CI for RAPIDS
+sed_runner 's/'"DEPENDENT_PACKAGE_VERSION=\"[0-9][0-9].[0-9][0-9]\""'/'"DEPENDENT_PACKAGE_VERSION=\"${NEXT_SHORT_TAG}\""'/g' ci/build_cpp.sh
+sed_runner 's/'"DEPENDENT_PACKAGE_VERSION=\"[0-9][0-9].[0-9][0-9]\""'/'"DEPENDENT_PACKAGE_VERSION=\"${NEXT_SHORT_TAG}\""'/g' ci/build_python.sh
+
+# PYTHON for RAPIDS
+sed_runner "/DOWNLOAD.*rapids-cmake/ s/branch-[0-9][0-9].[0-9][0-9]/branch-${NEXT_SHORT_TAG}/g" python/cuopt/CMakeLists.txt
```
jameslamb:
```suggestion
sed_runner "/DOWNLOAD.*rapids-cmake/ s/branch-[0-9][0-9].[0-9][0-9]/branch-${NEXT_SHORT_TAG}/g" python/cuopt/CMakeLists.txt

```

Please add a newline here. I think you should just let `pre-commit` handle this.

Remove `files:` from this:

https://github.com/NVIDIA/cuopt/blob/0845a21e0a867babdce73325d1bcf748a32ca9d4/.pre-commit-config.yaml#L20-L22

And only use the `exclude:` pattern to handle excluding files you don't want the check to cover.

That way, by default the hook will cover all files in the repo, so you don't need to remember to keep adding file extensions (in this case, `.sh`).

ci/release/update-version.sh
```diff
@@ -100,29 +124,20 @@ sed_runner "/^set(cuopt_version/ s/[0-9][0-9].[0-9][0-9].[0-9][0-9]/${NEXT_FULL_
...
+for FILE in .github/workflows/*.yaml; do
+  sed_runner "/shared-workflows/ s/@.*/@branch-${NEXT_SHORT_TAG}/g" "${FILE}"
+  # CI image tags of the form {rapids_version}-{something}
+  sed_runner "s/:[0-9]*\\.[0-9]*-/:${NEXT_SHORT_TAG}-/g" "${FILE}"
 done
+
+# CI for RAPIDS
+sed_runner 's/'"DEPENDENT_PACKAGE_VERSION=\"[0-9][0-9].[0-9][0-9]\""'/'"DEPENDENT_PACKAGE_VERSION=\"${NEXT_SHORT_TAG}\""'/g' ci/build_cpp.sh
+sed_runner 's/'"DEPENDENT_PACKAGE_VERSION=\"[0-9][0-9].[0-9][0-9]\""'/'"DEPENDENT_PACKAGE_VERSION=\"${NEXT_SHORT_TAG}\""'/g' ci/build_python.sh
+
```
jameslamb:
```suggestion
```

You already have `RAPIDS_VERSION` in a file in the repo. Instead of doing this:

https://github.com/NVIDIA/cuopt/blob/0845a21e0a867babdce73325d1bcf748a32ca9d4/ci/build_cpp.sh#L35

How about doing this in those scripts?

```shell
DEPENDENT_PACKAGE_VERSION=$(cat ./RAPIDS_VERSION)
export DEPENDENT_PACKAGE_VERSION
```

That way, you don't need to update this in shell scripts... just drive everything through those `VERSION` and `RAPIDS_VERSION` files.

That'd mean fewer updates, less risk of inconsistencies, and an easier-to-understand configuration flow.

ci/release/update-version.sh
```diff
@@ -37,60 +38,83 @@ function sed_runner() {
 
 # Centralized version file update
 echo "${NEXT_FULL_TAG}" > VERSION
+echo "${NEXT_FULL_TAG}" > RAPIDS_VERSION
```
jameslamb:
```suggestion
echo "${NEXT_FULL_TAG}" > ./VERSION
echo "${NEXT_FULL_TAG}" > ./RAPIDS_VERSION
```

Very very very minor comment, ignore if you want... I find it useful to add a leading `./` to stuff like this, to make it clear that we're piping to a file and not a shell variable.

jameslamb (changes requested):
Thanks for the updates. I've left some more suggestions.

ci/release/update-version.sh
ramakrishnap-nv: RAPIDS_VERSION has 25.10.00 instead of 25.10, but let me try if it fixes it.

ci/release/update-version.sh
ramakrishnap-nv: I will make this change in a new PR since it will touch too many files. But make the suggested change manually for now.

ramakrishnap-nv: Created this PR https://github.com/NVIDIA/cuopt/pull/306

ci/release/update-version.sh
jameslamb: Ah ok. Well still, the more important part of this recommendation is using the file, to limit how many places the same version is hard-coded. You could modify that statement to only pull out the major and minor version.

ramakrishnap-nv: Seems like it works as it is, so it should be good.

rgsl888prabhu:
@jameslamb May I get your review on this ?

ci/release/update-version.sh
```diff
@@ -0,0 +1,126 @@
...
+# CMakeLists update
+sed_runner 's/'"VERSION [0-9][0-9].[0-9][0-9].[0-9][0-9]"'/'"VERSION ${NEXT_FULL_TAG}"'/g' cpp/CMakeLists.txt
+sed_runner 's/'"VERSION [0-9][0-9].[0-9][0-9].[0-9][0-9]"'/'"VERSION ${NEXT_FULL_TAG}"'/g' cpp/libmps_parser/CMakeLists.txt
+sed_runner 's/'"DEPENDENT_LIB_MAJOR_VERSION \"[0-9][0-9]\""'/'"DEPENDENT_LIB_MAJOR_VERSION \"${NEXT_MAJOR}\""'/g' cpp/CMakeLists.txt
+sed_runner 's/'"DEPENDENT_LIB_MINOR_VERSION \"[0-9][0-9]\""'/'"DEPENDENT_LIB_MINOR_VERSION \"${NEXT_MINOR}\""'/g' cpp/CMakeLists.txt
+
+# Server version update
+sed_runner 's/'"\"version\": \"[0-9][0-9].[0-9][0-9]\""'/'"\"version\": \"${NEXT_SHORT_TAG}\""'/g' python/cuopt_server/cuopt_server/utils/data_definition.py
+sed_runner 's/'"\"client_version\": \"[0-9][0-9].[0-9][0-9]\""'/'"\"client_version\": \"${NEXT_SHORT_TAG}\""'/g' python/cuopt_server/cuopt_server/utils/routing/data_definition.py
+sed_runner 's/'"\"client_version\": \"[0-9][0-9].[0-9][0-9]\""'/'"\"client_version\": \"${NEXT_SHORT_TAG}\""'/g' python/cuopt_server/cuopt_server/utils/linear_programming/data_definition.py
```
jameslamb, line 82: I think it would be good to calculate this programmatically, in Python code instead.

e.g.

https://github.com/NVIDIA/cuopt/blob/b567ed4c15daf5650cda2ea118cc13e12d5db822/python/cuopt_server/cuopt_server/utils/routing/data_definition.py#L1029

Should rely on this constant:

https://github.com/NVIDIA/cuopt/blob/b567ed4c15daf5650cda2ea118cc13e12d5db822/python/cuopt_server/cuopt_server/_version.py#L18

Same advice goes for all 3 of these lines.

ci/release/update-version.sh
```diff
@@ -0,0 +1,126 @@
...
+
+# Doc update
+sed_runner 's/'"version = \"[0-9][0-9].[0-9][0-9]\""'/'"version = \"${NEXT_SHORT_TAG}\""'/g' docs/cuopt/source/conf.py
```
jameslamb, line 85: This should use `cuopt.__version__` instead.

Example of how `cudf` does this: https://github.com/rapidsai/cudf/blob/21b69cf8c31c42bf981c5c8987b4d0d6ff36b3d5/docs/cudf/source/conf.py#L189-L195

ci/release/update-version.sh
```diff
@@ -0,0 +1,126 @@
...
+sed_runner 's/'"PROJECT_NUMBER         = [0-9][0-9].[0-9][0-9]"'/'"PROJECT_NUMBER         = ${NEXT_SHORT_TAG}"'/g' cpp/doxygen/Doxyfile
```
jameslamb, line 86: `Doxyfile`s can be templated and filled in at runtime with environment variables. This project should consider doing that.

Example from `raft`:

https://github.com/rapidsai/raft/blob/e6efdd1143c988a366619965bcdd891a4474ea7e/cpp/doxygen/Doxyfile#L41

https://github.com/rapidsai/raft/blob/e6efdd1143c988a366619965bcdd891a4474ea7e/ci/build_docs.sh#L13

ci/release/update-version.sh
```diff
@@ -0,0 +1,126 @@
...
+if ! grep -q "\"version\": \"${NEXT_FULL_TAG}\"" "${VERSIONS_FILE}"; then
+  # Remove preferred and latest flags, but keep the version entry and URL
+  sed_runner '/"name": "latest",/d' "${VERSIONS_FILE}"
+  sed_runner '/"preferred": true,\?/d' "${VERSIONS_FILE}"
+  # Remove trailing comma after "url": ... in all version entries
+  sed_runner 's/\("url": "[^"]*"\),/\1/' "${VERSIONS_FILE}"
+  # Add new version entry with both preferred and latest flags
+  NEW_VERSION_ENTRY='    {\n      "version": "'${NEXT_FULL_TAG}'",\n      "url": "../'${NEXT_FULL_TAG}'/",\n      "name": "latest",\n      "preferred": true\n    },'
+  sed_runner "/\[/a\\${NEW_VERSION_ENTRY}" "${VERSIONS_FILE}"
+fi
```
jameslamb, line 104: I think you might find this easier to maintain if it was in Python.

I recommend looking into what I just did over in the `docs` repo:

* https://github.com/rapidsai/docs/pull/669
* https://github.com/rapidsai/docs/pull/670

I think you should have a small Python script that populates `versions1.json`, and a `pre-commit` hook that runs it.

And that script should read from `./VERSION` at the root of the repo.

That'd simplify `update-version.sh`, and make it easier to keep that file up to date.

ci/release/update-version.sh
```diff
@@ -0,0 +1,126 @@
...
+
+# RTD update
+sed_runner "/^set(cuopt_version/ s/[0-9][0-9].[0-9][0-9].[0-9][0-9]/${NEXT_FULL_TAG}/g" python/cuopt/CMakeLists.txt
+sed_runner "/^set(cuopt_version/ s/[0-9][0-9].[0-9][0-9].[0-9][0-9]/${NEXT_FULL_TAG}/g" python/cuopt/cuopt/linear_programming/CMakeLists.txt
+sed_runner "/^set(cuopt_version/ s/[0-9][0-9].[0-9][0-9].[0-9][0-9]/${NEXT_FULL_TAG}/g" python/libcuopt/CMakeLists.txt
```
jameslamb, line 109: It should be possible to derive all of these in CMake, at configure time, from other variables.

Re-using all of these variables:

https://github.com/NVIDIA/cuopt/blob/b567ed4c15daf5650cda2ea118cc13e12d5db822/cmake/rapids_config.cmake#L15-L21

ci/release/update-version.sh
```diff
@@ -0,0 +1,126 @@
...
+sed_runner 's/\(version: \)[0-9][0-9]\.[0-9]\+\.[0-9]\+/\1'${DOCKER_TAG}'/g' helmchart/cuopt-server/Chart.yaml
+
+for FILE in .github/workflows/*.yaml; do
+  sed_runner "/shared-workflows/ s/@.*/@branch-${NEXT_SHORT_TAG}/g" "${FILE}"
+  # CI image tags of the form {rapids_version}-{something}
+  sed_runner "s/:[0-9]*\\.[0-9]*-/:${NEXT_SHORT_TAG}-/g" "${FILE}"
+done
+
+# PYTHON for RAPIDS
+sed_runner "/DOWNLOAD.*rapids-cmake/ s/branch-[0-9][0-9].[0-9][0-9]/branch-${NEXT_SHORT_TAG}/g" python/cuopt/CMakeLists.txt
```
jameslamb, line 126: This should be grouped with the other `CMakeLists.txt` updates further up.

And similar to my comment on those, I think it'd be better to move this logic into CMake code itself.

jameslamb (approved):
Approving this so you can keep moving forward. But if you haven't already, please do test it one more time, in the way I described in https://github.com/NVIDIA/cuopt/pull/286#pullrequestreview-3117686535

I also think you should write up a follow-up issue about consolidating the references to versions in this project. I left a couple of suggestions, but overall there are very very many hard-coded `25.10` around the source code that all need to be kept in sync.

For inspiration, you could look at some of the PRs @KyleFromNVIDIA put up around RAPIDS implementing similar things to what I've suggested here: https://github.com/rapidsai/build-planning/issues/15

PR #303, opened 2025-08-14
Bump libcuopt size for wheel since we have added papilo

rgsl888prabhu:
## Description
Adding papilo and boost to cuOpt wheel increased size of the wheel and needs to bump size restriction for wheel.

## Checklist

- Testing
   - [ ] Added tests
   - [ ] NA
- Documentation
   - [ ] Added new documentation
   - [ ] NA

### DIFF
--- a/python/libcuopt/pyproject.toml
+++ b/python/libcuopt/pyproject.toml
@@ -66,7 +66,7 @@ select = [
 ]
 # PyPI limit is 700 MiB, fail CI before we get too close to that
 # 11.X size is 300M compressed and 12.x size is 600M compressed
-max_allowed_size_compressed = '700M'
+max_allowed_size_compressed = '745M'
 
 [project.scripts]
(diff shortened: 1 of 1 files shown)


python/libcuopt/pyproject.toml
```diff
@@ -66,7 +66,7 @@ select = [
 ]
 # PyPI limit is 700 MiB, fail CI before we get too close to that
 # 11.X size is 300M compressed and 12.x size is 600M compressed
-max_allowed_size_compressed = '700M'
+max_allowed_size_compressed = '775M'
```
jameslamb, line 67: The comment above this is not accurate. There is no "PyPI limit" for this project... `libcuopt-cu12` wheels are not published to `pypi.org`, only to `pypi.nvidia.com` (which does not have size limits).

That comment should be rewritten to something simple and future-proof, like this:

```toml
# detect when package size grows significantly
max_allowed_size_compressed = '500M'
```

PR #305, opened 2025-08-15
[Draft] Simple diving for Branch-and-Bound

nguidotti:
Implementation of a simple diving/plunging method for the branch and bound. 

To-Do list:

[ ] Move diving to another thread, such that it can be used alongside with best first search
[ ] Better integrate the diving settings to the other settings in the branch and bound
[ ] Add unit tests 
[ ] Benchmark all MIPLIB instances and compare against the original code. 
[ ] Undo the removal of the export log to file in the `run_mip.cpp` before merging the PR.

### DIFF
--- a/benchmarks/linear_programming/cuopt/run_mip.cpp
+++ b/benchmarks/linear_programming/cuopt/run_mip.cpp
@@ -17,6 +17,7 @@
 #include "initial_solution_reader.hpp"
 #include "mip_test_instances.hpp"
 
+#include <cstdio>
 #include <cuopt/linear_programming/mip/solver_settings.hpp>
 #include <cuopt/linear_programming/mip/solver_solution.hpp>
 #include <cuopt/linear_programming/optimization_problem.hpp>
@@ -377,7 +378,7 @@ int main(int argc, char* argv[])
   bool write_log_file  = program.get<std::string>("--write-log-file")[0] == 't';
   bool log_to_console  = program.get<std::string>("--log-to-console")[0] == 't';
   bool use_diving      = program.get<std::string>("--diving")[0] == 't';
-
+  
   if (program.is_used("--out-dir")) {
     out_dir     = program.get<std::string>("--out-dir");
     result_file = out_dir + "/final_result.csv";
--- a/benchmarks/linear_programming/cuopt/run_mip.cpp
+++ b/benchmarks/linear_programming/cuopt/run_mip.cpp
@@ -151,7 +151,8 @@ int run_single_file(std::string file_path,
                     int num_cpu_threads,
                     bool write_log_file,
                     bool log_to_console,
-                    double time_limit)
+                    double time_limit,
+                    bool use_diving)
 {
   const raft::handle_t handle_{};
   cuopt::linear_programming::mip_solver_settings_t<int, double> settings;
@@ -168,11 +169,13 @@ int run_single_file(std::string file_path,
     }
   }
 
+  settings.search_strategy.use_diving = use_diving;
+
   constexpr bool input_mps_strict = false;
   cuopt::mps_parser::mps_data_model_t<int, double> mps_data_model;
   bool parsing_failed = false;
   {
-    CUOPT_LOG_INFO("running file %s on gpu : %d", base_filename.c_str(), device);
+    CUOPT_LOG_INFO("Running file %s", base_filename.c_str());
     try {
       mps_data_model = cuopt::mps_parser::parse_mps<int, double>(file_path, input_mps_strict);
     } catch (const std::logic_error& e) {
@@ -232,14 +235,15 @@ int run_single_file(std::string file_path,
   } else {
     CUOPT_LOG_INFO("%s: no solution found", base_filename.c_str());
   }
-  std::stringstream ss;
-  int decimal_places = 2;
-  ss << std::fixed << std::setprecision(decimal_places) << base_filename << "," << sol_found << ","
-     << obj_val << "," << benchmark_info.objective_of_initial_population << ","
-     << benchmark_info.last_improvement_of_best_feasible << ","
-     << benchmark_info.last_improvement_after_recombination << "\n";
-  write_to_output_file(out_dir, base_filename, device, batch_id, ss.str());
-  CUOPT_LOG_INFO("Results written to the file %s", base_filename.c_str());
+  // std::stringstream ss;
+  // int decimal_places = 2;
+  // ss << std::fixed << std::setprecision(decimal_places) << base_filename << "," << sol_found <<
+  // ","
+  //    << obj_val << "," << benchmark_info.objective_of_initial_population << ","
+  //    << benchmark_info.last_improvement_of_best_feasible << ","
+  //    << benchmark_info.last_improvement_after_recombination << "\n";
+  // write_to_output_file(out_dir, base_filename, device, batch_id, ss.str());
+  // CUOPT_LOG_INFO("Results written to the file %s", base_filename.c_str());
   return sol_found;
 }
 
@@ -252,7 +256,8 @@ void run_single_file_mp(std::string file_path,
                         int num_cpu_threads,
                         bool write_log_file,
                         bool log_to_console,
-                        double time_limit)
+                        double time_limit,
+                        bool use_diving)
 {
   std::cout << "running file " << file_path << " on gpu : " << device << std::endl;
   auto memory_resource = make_async();
@@ -266,7 +271,8 @@ void run_single_file_mp(std::string file_path,
                                   num_cpu_threads,
                                   write_log_file,
                                   log_to_console,
-                                  time_limit);
+                                  time_limit,
+                                  use_diving);
   // this is a bad design to communicate the result but better than adding complexity of IPC or
   // pipes
   exit(sol_found);
@@ -340,6 +346,10 @@ int main(int argc, char* argv[])
     .scan<'g', double>()
     .default_value(std::numeric_limits<double>::max());
 
+  program.add_argument("--diving")
+    .help("use diving during branch and bound(t/f)")
+    .default_value(std::string("f"));
+
   // Parse arguments
   try {
     program.parse_args(argc, argv);
@@ -366,6 +376,7 @@ int main(int argc, char* argv[])
   int num_cpu_threads  = program.get<int>("--num-cpu-threads");
   bool write_log_file  = program.get<std::string>("--write-log-file")[0] == 't';
   bool log_to_console  = program.get<std::string>("--log-to-console")[0] == 't';
+  bool use_diving      = program.get<std::string>("--diving")[0] == 't';
 
   if (program.is_used("--out-dir")) {
     out_dir     = program.get<std::string>("--out-dir");
@@ -451,7 +462,8 @@ int main(int argc, char* argv[])
                                num_cpu_threads,
                                write_log_file,
                                log_to_console,
-                               time_limit);
+                               time_limit,
+                               use_diving);
           } else if (sys_pid < 0) {
             std::cerr << "Fork failed!" << std::endl;
             exit(1);
@@ -480,7 +492,8 @@ int main(int argc, char* argv[])
                     num_cpu_threads,
                     write_log_file,
                     log_to_console,
-                    time_limit);
+                    time_limit,
+                    use_diving);
   }
 
   return 0;
--- a/cpp/include/cuopt/linear_programming/mip/solver_settings.hpp
+++ b/cpp/include/cuopt/linear_programming/mip/solver_settings.hpp
@@ -35,6 +35,10 @@ struct benchmark_info_t {
   double objective_of_initial_population      = std::numeric_limits<double>::max();
 };
 
+struct bnb_search_strategy_t {
+  bool use_diving = false;
+};
+
 // Forward declare solver_settings_t for friend class
 template <typename i_t, typename f_t>
 class solver_settings_t;
@@ -78,6 +82,8 @@ class mip_solver_settings_t {
     f_t absolute_mip_gap            = 1.0e-10;
     f_t relative_mip_gap            = 1.0e-4;
   };
+
+  bnb_search_strategy_t search_strategy;
 
   /**
    * @brief Get the tolerance settings as a single structure.
--- a/cpp/src/dual_simplex/branch_and_bound.cpp
+++ b/cpp/src/dual_simplex/branch_and_bound.cpp
@@ -15,6 +15,7 @@
  * limitations under the License.
  */
 
+#include <cmath>
 #include <dual_simplex/branch_and_bound.hpp>
 
 #include <dual_simplex/initial_basis.hpp>
@@ -34,6 +35,7 @@
 #include <queue>
 #include <string>
 #include <thread>
+#include <vector>
 
 namespace cuopt::linear_programming::dual_simplex {
 
@@ -401,7 +403,10 @@ branch_and_bound_t<i_t, f_t>::branch_and_bound_t(
   const user_problem_t<i_t, f_t>& user_problem,
   const simplex_solver_settings_t<i_t, f_t>& solver_settings,
   const bnb_search_strategy_t strategy)
-  : original_problem(user_problem), settings(solver_settings), search_strategy(strategy), original_lp(1, 1, 1)
+  : original_problem(user_problem),
+    settings(solver_settings),
+    search_strategy(strategy),
+    original_lp(1, 1, 1)
 {
   start_time = tic();
   convert_user_problem(original_problem, settings, original_lp, new_slacks);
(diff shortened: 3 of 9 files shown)


gforsyth (approved):
Nothing here touches `cuopt` CI, approving

benchmarks/linear_programming/cuopt/run_mip.cpp
```diff
@@ -168,11 +170,13 @@ int run_single_file(std::string file_path,
     }
   }
 
+  settings.bnb_search_strategy = search_strategy;
+
   constexpr bool input_mps_strict = false;
   cuopt::mps_parser::mps_data_model_t<int, double> mps_data_model;
   bool parsing_failed = false;
   {
-    CUOPT_LOG_INFO("running file %s on gpu : %d", base_filename.c_str(), device);
```
akifcorduk, line 178: Please keep the log structure as we use them in benchmarking.

cpp/src/dual_simplex/mip_node.hpp
```diff
@@ -33,14 +34,15 @@ enum class node_status_t : int {
   INTEGER_FEASIBLE = 2,  // Node has an integer feasible solution
   INFEASIBLE       = 3,  // Node is infeasible
   FATHOMED         = 4,  // Node objective is greater than the upper bound
+  HAS_CHILDREN     = 5,  // Node has children to explore
```
akifcorduk, line 37: Doesn't active mean that the node has childen to explore? The other possible states for leaf nodes are in_progress, integer_feasible or infeasible.

cpp/src/dual_simplex/pseudo_costs.hpp
```diff
@@ -37,6 +37,28 @@ class pseudo_costs_t {
   {
   }
 
+  pseudo_costs_t(const pseudo_costs_t& other)
```
akifcorduk: You can default the copy and move constructor, right?

cpp/src/dual_simplex/mip_node.hpp
```diff
@@ -134,18 +141,23 @@ class mip_node_t {
     }
   }
 
+  node_status_t set_status(node_status_t new_status)
+  {
+    return status.exchange(new_status, std::memory_order_seq_cst);
```
akifcorduk: Sequential consistency might be too strong. Do you think release would also work here?

cpp/src/dual_simplex/branch_and_bound.cpp
```diff
@@ -356,474 +373,685 @@ template <typename i_t, typename f_t>
...
-    if (settings.solution_callback != nullptr) {
-      settings.solution_callback(solution.x, solution.objective);
-    }
-    if (settings.heuristic_preemption_callback != nullptr) {
-      settings.heuristic_preemption_callback();
+  return mip_status_t::UNSET;
+}
+
+template <typename i_t, typename f_t>
+mip_status_t branch_and_bound_t<i_t, f_t>::solve_leaf_lp(mip_node_t<i_t, f_t>* node_ptr,
```
akifcorduk: I am not sure if "leaf" might cause some confusions. Maybe rename it to "node" ?

cpp/src/dual_simplex/branch_and_bound.cpp
```diff
@@ -356,474 +373,685 @@ template <typename i_t, typename f_t>
...
+
+    f_t upper_bound      = get_upper_bound();
+    lower_bound          = get_lower_bound();
+    gap                  = upper_bound - lower_bound;
+    const i_t leaf_depth = node_ptr->depth;
+
+    f_t now            = toc(stats_.start_time);
+    f_t time_since_log = last_log == 0 ? 1.0 : toc(last_log);
+
+    if (settings_.bnb_search_strategy == bnb_search_strategy_t::DEPTH_FIRST) {
```
akifcorduk: We are in the dive function, is there any other possible strategy possible in dive? I guess i am asking if this "if" ever evaluates to false.

benchmarks/linear_programming/cuopt/run_mip.cpp
```diff
@@ -340,6 +346,15 @@ int main(int argc, char* argv[])
...
 
+  program.add_argument("--search-strategy")
+    .help("Search strategy used in B&B (bfs/bfs-diving/dfs)")
+    .default_value(std::string("bfs"));
+
+  program.add_argument("--gpu")
+    .help("id of the GPU to use when running a single test (default: 0)")
+    .scan<'i', int>()
+    .default_value(0);
+
```
aliceb-nv: You can directly set the `CUDA_VISIBLE_DEVICES` environment variable to select the GPU(s) to use :)
e.g. `$ CUDA_VISIBLE_DEVICES=<id> ./solve_MPS_file <...>`

cpp/include/cuopt/linear_programming/mip/solver_settings.hpp
```diff
@@ -79,6 +85,8 @@ class mip_solver_settings_t {
     f_t relative_mip_gap            = 1.0e-4;
   };
 
+  bnb_search_strategy_t bnb_search_strategy = bnb_search_strategy_t::BEST_FIRST;
+
```
aliceb-nv: It's not an immediate priority, but I think it'd be nice to expose the diving parameters in the global solver settings, e.g. what we do here:
https://github.com/NVIDIA/cuopt/blob/36a6a1c0edf42cec2cf07c6be3f16531f33515de/cpp/src/math_optimization/solver_settings.cu#L89-L93

That way the appropriate cmdline arguments are automatically generated :)

cpp/src/dual_simplex/branch_and_bound.cpp
```diff
@@ -207,6 +220,7 @@ f_t relative_gap(f_t obj_value, f_t lower_bound)
   f_t user_mip_gap = obj_value == 0.0
                        ? (lower_bound == 0.0 ? 0.0 : std::numeric_limits<f_t>::infinity())
                        : std::abs(obj_value - lower_bound) / std::abs(obj_value);
+  // Handle NaNs (i.e., NaN != NaN)
   if (user_mip_gap != user_mip_gap) { return std::numeric_limits<f_t>::infinity(); }
```
aliceb-nv: I think it'd be clearer to just use std::isnan

cpp/src/dual_simplex/branch_and_bound.cpp
```diff
@@ -356,474 +373,685 @@ template <typename i_t, typename f_t>
...
+
+template <typename i_t, typename f_t>
+void branch_and_bound_t<i_t, f_t>::branch(mip_node_t<i_t, f_t>* parent_node,
+                                          i_t branch_var,
+                                          f_t branch_var_val,
+                                          const std::vector<variable_status_t>& parent_vstatus)
+{
+  // down child
+  std::unique_ptr<mip_node_t<i_t, f_t>> down_child = std::make_unique<mip_node_t<i_t, f_t>>(
+    original_lp_, parent_node, ++stats_.num_nodes, branch_var, 0, branch_var_val, parent_vstatus);
```
aliceb-nv: nitpick: it's a matter of personal preference, but I find it usually clearer to use 'auto' when declaring a variable w/ std::make_unique: make_unique already makes it explicit that this is a unique_ptr and which type is being pointed to, so this avoids typing the same thing twice

aliceb-nv: Oh this looks like Chris' code! Mea culpa :)

cpp/src/dual_simplex/pseudo_costs.hpp
nguidotti: In this case, no, because the copy constructor from `std::mutex` is deleted.

cpp/src/dual_simplex/branch_and_bound.hpp
```diff
@@ -50,65 +56,113 @@ class branch_and_bound_t {
...
-    f_t total_lp_solve_time;
-    f_t start_time;
-  } stats;
+    f_t start_time                    = 0.0;
+    f_t total_lp_solve_time           = 0.0;
+    std::atomic<i_t> nodes_explored   = 0;
+    std::atomic<i_t> nodes_unexplored = 0;
+    f_t total_lp_iters                = 0;
+    std::atomic<i_t> num_nodes        = 0;
+  } stats_;
```
aliceb-nv, line 124: We'd have to check the generated asm with compiler explorer to be sure, but I think we expose C++20 std::atomic float support in <cuda/std/atomic>: https://nvidia.github.io/cccl/libcudacxx/standard_api.html#libcudacxx-standard-api

That's one of the cool perks of libcu++ :)

aliceb-nv: Hm, although I am unsure if this would require us to switch to cuda compilation for the dual_simplex src files. Regardless, it's a nice thing to know!

cpp/src/dual_simplex/branch_and_bound.hpp
```diff
@@ -50,65 +56,113 @@ class branch_and_bound_t {
...
+  std::vector<std::vector<f_t>> repair_queue_;
+
+  // Variables for the root node in the search tree.
+  std::vector<variable_status_t> root_vstatus_;
+  f_t root_objective_;
+  lp_solution_t<i_t, f_t> root_relax_soln_;
+  std::vector<f_t> edge_norms_;
+
+  // Repairs low-quality solutions from the heuristics, if it is applicable.
+  void repair_heuristic_solutions(const f_t& lower_bound, mip_solution_t<i_t, f_t>& solution);
```
aliceb-nv: nitpick: it's usually better to pass small fundamental types (typically, anything such that `sizeof(T) <= sizeof(void*)`, like integer and float types) by value rather than constant reference

cpp/src/dual_simplex/branch_and_bound.hpp
```diff
@@ -50,65 +56,113 @@ class branch_and_bound_t {
...
+
+  // Explore the search tree using the best-first search strategy.
+  mip_status_t explore_tree(i_t branch_var,
+                            mip_solution_t<i_t, f_t>& solution,
+                            pseudo_costs_t<i_t, f_t> pc);
```
aliceb-nv: Does 'explore_tree' need to copy 'pc'? If not, better to pass it as a const ref since it contains a bunch of vectors

cpp/src/dual_simplex/branch_and_bound.hpp
```diff
@@ -50,65 +56,113 @@ class branch_and_bound_t {
...
+
+  // Explore the search tree using the depth-first search strategy.
+  mip_status_t dive(i_t branch_var,
+                    mip_solution_t<i_t, f_t>& solution,
+                    pseudo_costs_t<i_t, f_t> pc);
```
aliceb-nv: ditto

benchmarks/linear_programming/cuopt/run_mip.cpp
nguidotti: Sorry. I reverted to the previous state.

cpp/src/dual_simplex/mip_node.hpp
nguidotti: `HAS_CHILDREN` means that the nodes has already been processed and has two branches that the code can follow through. `ACTIVE` means that the node is in the tree but not has been processed (maybe a better name would be `PENDING` here)

cpp/src/dual_simplex/mip_node.hpp
```diff
@@ -134,6 +141,8 @@ class mip_node_t {
     }
   }
 
+  node_status_t set_status(node_status_t new_status) { return status; }
```
aliceb-nv: Is std::atomic no longer necessary for new_status?

cpp/src/dual_simplex/mip_node.hpp
nguidotti: Good catch. I forgot to remove the atomic after trying to share the tree between the threads.

cpp/src/dual_simplex/branch_and_bound.cpp
nguidotti: Done.

aliceb-nv (commented):
Thanks for the great work Nicolas, it's so nice getting closer to seeing diving on the main branch! Added a few comments, some are nitpicks :)

cpp/src/dual_simplex/mip_node.hpp
aliceb-nv: Ah nvm, looks like you answered this already!

cpp/src/dual_simplex/branch_and_bound.cpp
nguidotti: This is for enable the logs when using just depth-first search

benchmarks/linear_programming/cuopt/run_mip.cpp
nguidotti: I forgot about the environment variable... I will removed the flag.

cpp/src/dual_simplex/mip_node.hpp
nguidotti: True. I forgot to revert it back.

cpp/include/cuopt/linear_programming/mip/solver_settings.hpp
nguidotti: Done. Although I am not sure about the order of the parameters. It is `{default, min, max}`?

nguidotti:
/ok to test [0017b50](https://github.com/NVIDIA/cuopt/pull/305/commits/0017b5038e04c54fe4cf9bf2a798c4f8f6931c07)

cpp/src/dual_simplex/branch_and_bound.hpp
nguidotti: Since I will be using openmp for multithreading, I will probably replace the `std::atomic` with `omp atomic` which supports floating-points variables.

cpp/include/cuopt/linear_programming/mip/solver_settings.hpp
aliceb-nv: It's `{min, max, default}` :)
https://github.com/NVIDIA/cuopt/blob/36a6a1c0edf42cec2cf07c6be3f16531f33515de/cpp/include/cuopt/linear_programming/utilities/internals.hpp#L79-L88

cpp/src/dual_simplex/pseudo_costs.hpp
nguidotti: I removed these constructors as I use the pseudocost as global variables instead of copying them as parameters.

cpp/src/dual_simplex/branch_and_bound.hpp
nguidotti: Moved the pseudocosts to be a global variable.

cpp/src/dual_simplex/branch_and_bound.hpp
aliceb-nv: Sounds good! You might just want to make sure `#pragma omp atomic` is guaranteed to work even outside of an omp parallel block, or when non-openmp-managed threads exists: the OpenMP programming model might not make this guarantee

nguidotti: As long as you pass the `-fopenmp`, then it [works](https://godbolt.org/z/x9beEz73q) independently if it is inside a `omp parallel` or not. Mixing both `std::thread` and `omp` are often not a very good idea, though...

aliceb-nv: Great! I was mostly worried about this being technically undefined by the openmp specification; but I can think of no real-world/relevant architecture or compiler that could emit anything but hardware atomic instructions in these cases.

But you spawn your threads with `std::async` right? So regardless C++ threads and OpenMP are going to be playing together (even though this should cause no issues)

cpp/src/dual_simplex/branch_and_bound.cpp
```diff
@@ -941,7 +941,7 @@ mip_status_t branch_and_bound_t<i_t, f_t>::solve(mip_solution_t<i_t, f_t>& solut
                          compute_user_objective(original_lp_, root_objective_),
                          toc(stats_.start_time));
     if (settings_.set_simplex_solution_callback != nullptr) {
-      settings_.set_simplex_solution_callback(solution.x, solution.objective);
```
akifcorduk: You should check for if solution call back exists not the simplex call back.

nguidotti:
Sure. Added the flag to submip.

akifcorduk:
/ok to test [73fac2c](https://github.com/NVIDIA/cuopt/pull/305/commits/73fac2c44d8b990571e411fa71c964132e547acf)

cpp/src/dual_simplex/branch_and_bound.cpp
```diff
@@ -356,474 +373,685 @@ template <typename i_t, typename f_t>
...
+mip_status_t branch_and_bound_t<i_t, f_t>::solve_node_lp(mip_node_t<i_t, f_t>* node_ptr,
+                                                         lp_problem_t<i_t, f_t>& leaf_problem,
+                                                         f_t upper_bound,
+                                                         f_t lower_bound,
+                                                         i_t nodes_explored,
+                                                         i_t unexplored_nodes)
+{
+  logger_t log;
+  log.log                                      = false;
+  f_t gap                                      = upper_bound - lower_bound;
```
chris-maes: Where is gap used? Is it necessary to compute it in this function?

chris-maes: I think this gap is not used. We compute gap later in the code. It might be best to set it there.

cpp/src/dual_simplex/pseudo_costs.hpp
```diff
@@ -39,6 +39,14 @@ class pseudo_costs_t {
 
   void update_pseudo_costs(mip_node_t<i_t, f_t>* node_ptr, f_t leaf_objective);
 
+  void initialize(i_t num_variables)
```
chris-maes: It's probably not great to have two methods `initialize` and `initialized` with such similar names. Can you change the name here?

cpp/src/dual_simplex/branch_and_bound.cpp
```diff
@@ -356,474 +373,685 @@ template <typename i_t, typename f_t>
...
+      mutex_upper_.lock();
+      incumbent_.set_incumbent_solution(computed_obj, crushed_guess);
+      upper_bound_ = computed_obj;
+      mutex_upper_.unlock();
+    }
+  }
+
+  root_relax_soln_.resize(original_lp_.num_rows, original_lp_.num_cols);
+  mip_status_t root_status = solve_root_relaxation();
+  if (root_status != mip_status_t::UNSET) { return root_status; }
```
chris-maes: This is a bit strange. You should probably add a comment that says. We only set the MIP status in `solve_root_relaxation` if the problem is infeasible, unbounded, or a time limit occurs

cpp/src/dual_simplex/branch_and_bound.cpp
```diff
@@ -356,474 +373,685 @@ template <typename i_t, typename f_t>
...
+  log.log        = false;
+  i_t branch_var = pc_.variable_selection(
+    fractional, root_relax_soln_.x, original_lp_.lower, original_lp_.upper, log);
+
+  stats_.total_lp_iters   = 0;
+  stats_.nodes_explored   = 0;
+  stats_.nodes_unexplored = 0;
+  stats_.num_nodes        = 1;
+
+  if (settings_.bnb_search_strategy == bnb_search_strategy_t::DEPTH_FIRST) {
```
chris-maes: I'm not sure how much of this we want to expose to the user. 

I'm not sure the user should have the option of choosing the strategy.  I think by default we should be using the MULTI_THREADED_BEST_FIRST_WITH_DIVING strategy unless the user has set the number of CPU threads to 1.

cpp/src/dual_simplex/branch_and_bound.cpp
```diff
@@ -356,474 +373,685 @@ template <typename i_t, typename f_t>
...
+    if (settings_.bnb_search_strategy ==
+        bnb_search_strategy_t::MULTITHREADED_BEST_FIRST_WITH_DIVING) {
+      diving_thread = std::async(std::launch::async, [&]() { return dive(branch_var, solution); });
+    }
+
+    status = explore_tree(branch_var, solution);
+
+    if (settings_.bnb_search_strategy ==
+        bnb_search_strategy_t::MULTITHREADED_BEST_FIRST_WITH_DIVING) {
+      // TODO: Instead of waiting, we should send a signal that we are already done in the main
```
chris-maes: Is this not already implemented? That is, does the diving thread know that we are done now? 

If not, we should probably implement this before merging this PR

cpp/src/dual_simplex/branch_and_bound.cpp
chris-maes: Why would we every use depth first search, by itself, outside of the a sub-MIP call?

cpp/src/dual_simplex/branch_and_bound.cpp
```diff
@@ -356,474 +373,685 @@ template <typename i_t, typename f_t>
...
+    f_t upper_bound      = get_upper_bound();
+    lower_bound          = get_lower_bound();
+    gap                  = upper_bound - lower_bound;
+    const i_t leaf_depth = node_ptr->depth;
+
+    f_t now            = toc(stats_.start_time);
+    f_t time_since_log = last_log == 0 ? 1.0 : toc(last_log);
+
+    if (settings_.bnb_search_strategy == bnb_search_strategy_t::DEPTH_FIRST) {
+      if ((nodes_explored % 1000 == 0 || gap < 10 * settings_.absolute_mip_gap_tol ||
```
chris-maes: I don't think the diving thread should do any logging.

cpp/src/dual_simplex/branch_and_bound.cpp
```diff
@@ -356,474 +373,685 @@ template <typename i_t, typename f_t>
...
-  mutex_branching.lock();
-  currently_branching = false;
-  mutex_branching.unlock();
 
-  if (heap.size() == 0) {
-    mutex_lower.lock();
-    lower_bound = lower_bound_ = root_node.lower_bound;
-    mutex_lower.unlock();
-    gap = get_upper_bound() - lower_bound;
+  if (settings_.bnb_search_strategy == bnb_search_strategy_t::DEPTH_FIRST) {
```
chris-maes: We should think about whether we want to increment the number of nodes explored when are running both best first search and diving.

cpp/src/dual_simplex/branch_and_bound.cpp
chris-maes: It looks like the diving thread understands the gap and the time limit. Is another signal needed to signal that we are done?

cpp/src/dual_simplex/pseudo_costs.hpp
nguidotti: I changed to `resize`

cpp/src/dual_simplex/branch_and_bound.cpp
nguidotti: You are right. I think this is enough.

cpp/src/dual_simplex/branch_and_bound.cpp
nguidotti: I left the logging  in case  only depth-first search is used. However, if the toggle is internal, then I can remove it.

cpp/src/dual_simplex/branch_and_bound.cpp
nguidotti: Maybe we can remove the option from `cuopt_cli` and other tools, but keep as a parameter in the `simplex_settings_t` so subMIP can use.

cpp/src/dual_simplex/branch_and_bound.cpp
nguidotti: In this implementation, the best-first search and the diving are exploring different trees and may explore the same node twice.

nguidotti:
> This PR is a bit bigger than I had envisioned.
> 
> I'm not sure we want to make the strategy (i.e. depth first vs best first) a user accessible parameter. But rather keep it an internal parameter.
> 
> Given the number of changes, I think we need to verify this PR before merging. Let's discuss the verification strategy.

Implement a test for verifying the different search strategies. But I can also add some unit testing. Is there any other verification strategy that you have in mind?

cpp/src/dual_simplex/branch_and_bound.cpp
nguidotti: Moved the root relaxation back to the solve function

nguidotti:
/ok to test [db73ba4](https://github.com/NVIDIA/cuopt/pull/305/commits/db73ba4689ebc667de63140697b4cae3fa4656f0)

nguidotti:
/ok to test [here](https://github.com/NVIDIA/cuopt/pull/305/commits/f9fb5ade7b65fd2daf68ad1f09c41f8765180785)

nguidotti:
/ok to test [dd7e340](https://github.com/NVIDIA/cuopt/pull/305/commits/dd7e340b13c2a90c8f9bdb0d014e656b7dc696b3)

cpp/src/dual_simplex/branch_and_bound.cpp
```diff
@@ -207,7 +247,8 @@ f_t relative_gap(f_t obj_value, f_t lower_bound)
   f_t user_mip_gap = obj_value == 0.0
                        ? (lower_bound == 0.0 ? 0.0 : std::numeric_limits<f_t>::infinity())
                        : std::abs(obj_value - lower_bound) / std::abs(obj_value);
-  if (user_mip_gap != user_mip_gap) { return std::numeric_limits<f_t>::infinity(); }
+  // Handle NaNs (i.e., NaN != NaN)
```
chris-maes: The test `user_mip_gap != user_mip_gap` is equivalent to `std::is_nan`

cpp/src/dual_simplex/branch_and_bound.cpp
```diff
@@ -353,498 +397,664 @@ bool branch_and_bound_t<i_t, f_t>::repair_solution(
...
+
+  csc_matrix_t<i_t, f_t> Arow(1, 1, 1);
+  original_lp_.A.transpose(Arow);
+
+  f_t lower_bound    = get_lower_bound();
+  f_t gap            = get_upper_bound() - lower_bound;
+  i_t nodes_explored = 0;
+
+  while (node_stack.size() > 0) {
+    repair_heuristic_solutions(lower_bound, solution);
```
chris-maes: Why do you need to repair a heuristic solution in the dive? 

I would let the best-first search thread handle repairs.

cpp/src/dual_simplex/branch_and_bound.cpp
```diff
@@ -353,498 +397,664 @@ bool branch_and_bound_t<i_t, f_t>::repair_solution(
...
-  mutex_branching.lock();
-  currently_branching = false;
-  mutex_branching.unlock();
 
-  if (heap.size() == 0) {
-    mutex_lower.lock();
-    lower_bound = lower_bound_ = root_node.lower_bound;
-    mutex_lower.unlock();
-    gap = get_upper_bound() - lower_bound;
+  if (settings_.bnb_search_strategy == search_strategy_t::DEPTH_FIRST) {
```
chris-maes: We are in a dive. Don't we know what the search strategy is?

cpp/src/dual_simplex/branch_and_bound.cpp
```diff
@@ -353,498 +397,664 @@ bool branch_and_bound_t<i_t, f_t>::repair_solution(
...
+  log.log        = false;
+  i_t branch_var = pc_.variable_selection(
+    fractional, root_relax_soln_.x, original_lp_.lower, original_lp_.upper, log);
+
+  stats_.total_lp_iters   = 0;
+  stats_.nodes_explored   = 0;
+  stats_.nodes_unexplored = 0;
+  stats_.num_nodes        = 1;
+
+  if (settings_.bnb_search_strategy == search_strategy_t::DEPTH_FIRST) {
```
chris-maes: Why would the search strategy ever be depth first here?

I think we only want best first or multithreaded best first with diving

cpp/src/dual_simplex/simplex_solver_settings.hpp
```diff
@@ -28,6 +28,12 @@
 
 namespace cuopt::linear_programming::dual_simplex {
 
+enum class search_strategy_t {
+  BEST_FIRST                           = 0,
+  DEPTH_FIRST                          = 1,
```
chris-maes: Remove DEPTH_FIRST?

cpp/src/dual_simplex/branch_and_bound.cpp
```diff
@@ -353,498 +397,664 @@ bool branch_and_bound_t<i_t, f_t>::repair_solution(
...
+      lower_bound = lower_bound_ = root_node.lower_bound;
+      mutex_lower_.unlock();
+      gap = get_upper_bound() - lower_bound;
+
+      mutex_gap_.lock();
+      gap_ = gap;
+      mutex_gap_.unlock();
+    }
+  }
+
```
chris-maes: To confirm: in a multithreaded bfs + dives, the nodes explored and unexplored only comes from the BFS thread?

chris-maes (commented):
Thanks for simplifying the PR. I left a few more comments. These are mostly nitpicks.

The only change I would suggest making before merging is to remove repairing solutions from the diving threads. 

You should make sure to run the MIPLIB benchmarks and put the geomean performance improvement in the commit message. You should also run this on a MIP problem that is infeasible from the LP relaxation and one that is infeasible due to no integer solution and verify the code correctly handles these cases.

cpp/src/dual_simplex/branch_and_bound.cpp
nguidotti: This change was suggested by Alice [here](https://github.com/NVIDIA/cuopt/pull/305#discussion_r2310117659)

cpp/src/dual_simplex/branch_and_bound.cpp
nguidotti: @akifcorduk wanted to have a diving only option for submip (for this reason, I have the checks above). However, for simplicity sake, we have just a single search strategy: bfs + diving. Is that OK?

nguidotti:
/ok to test [413ee24](https://github.com/NVIDIA/cuopt/pull/305/commits/413ee246875000db20f7bc2b279028ae52992c88)

nguidotti:
/ok to test [53127f5](https://github.com/NVIDIA/cuopt/pull/305/commits/53127f51c2f410351dcadfc92889aebecac74390)

nguidotti:
/ok to test [772f49e](https://github.com/NVIDIA/cuopt/pull/305/commits/772f49eb3e564f7e984ea30bc7cd6dcf8d030157)

chris-maes (approved):

PR #315, opened 2025-08-18
Performance tweak for dual_simplex/right_looking_lu

ahehn-nv:
## Description
Small performance optimization of `right_looking_lu()` and `right_looking_lu_row_permutation_only()`.

## Checklist

- Testing
   - [ ] Added tests
   - [x] NA
- Documentation
   - [ ] Added new documentation
   - [x] NA


cpp/src/dual_simplex/right_looking_lu.cpp
```diff
@@ -316,11 +315,13 @@ void update_Cdegree_and_col_count(i_t pivot_i,
...
     assert(cdeg >= 0);
-    for (typename std::list<i_t>::iterator it = col_count[cdeg].begin();
+    for (typename std::vector<i_t>::iterator it = col_count[cdeg].begin();
          it != col_count[cdeg].end();
          it++) {
       if (*it == j) {
-        col_count[cdeg].erase(it);
+        // Remove col j from col_count[cdeg]
+        std::swap(*it, col_count[cdeg].back());
+        col_count[cdeg].pop_back();
```
Kh4ster, line 324: Nice efficient way to remove from the vector! Is it okay in this context to change the order of elements in the vector?

ahehn-nv: Yes, @chris-maes confirmed that the order is not important here.

I had a previous version that preserved the order of elements, but breaking the order with this swap gives another 2x.

Kh4ster: All good then!

PR #316, opened 2025-08-18
Faster engine compile time

Kh4ster:
This PR aims at making the compile time faster by adding several options:

- build lp only (extracted MIP files necessary in LP), also meant we needed a skip c and python adapter option
- put back no fetch rapids
- put back not building test option
- added skip routing
- added skip write of fatbin
- adding ccache to libcuopt target
- added two benchmarking targets, one for LP and for MIP


build.sh
```diff
@@ -78,6 +84,11 @@ INSTALL_TARGET=install
 BUILD_DISABLE_DEPRECATION_WARNING=ON
 BUILD_ALL_GPU_ARCH=0
 BUILD_CI_ONLY=0
+BUILD_LP_ONLY=0
+SKIP_C_PYTHON_ADAPTERS=0
+SKIP_TESTS_BUILD=0
+SKIP_ROUTING_BUILD=0
+WRITE_FATBIN=0
```
ramakrishnap-nv: WRITE FATBIN should be by default enabled, and disabled with option

cpp/CMakeLists.txt
```diff
@@ -52,11 +52,19 @@ option(CMAKE_CUDA_LINEINFO "Enable the -lineinfo option for nvcc useful for cuda
 option(BUILD_TESTS "Configure CMake to build tests" ON)
 option(DISABLE_OPENMP "Disable OpenMP" OFF)
 option(CUDA_STATIC_RUNTIME "Statically link the CUDA toolkit runtime and libraries" OFF)
+option(BUILD_LP_ONLY "Build only linear programming components, exclude routing and MIP-specific files" OFF)
```
ramakrishnap-nv, line 55: Missing BUILD_TESTS
WRITE_FATBIN and BUILD_FATBIN is being used.

build.sh
```diff
@@ -78,6 +84,11 @@ INSTALL_TARGET=install
 BUILD_DISABLE_DEPRECATION_WARNING=ON
 BUILD_ALL_GPU_ARCH=0
 BUILD_CI_ONLY=0
+BUILD_LP_ONLY=0
+SKIP_C_PYTHON_ADAPTERS=0
+SKIP_TESTS_BUILD=0
```
ramakrishnap-nv, line 89: Dow we need to add option to skip BUILD_MIP_BENCHMARKS

cpp/CMakeLists.txt
```diff
@@ -361,44 +371,61 @@ if(Doxygen_FOUND)
 endif()
 
 
-add_executable(cuopt_cli cuopt_cli.cpp)
-target_compile_options(cuopt_cli
-  PRIVATE "$<$<COMPILE_LANGUAGE:CXX>:${CUOPT_CXX_FLAGS}>"
-  "$<$<COMPILE_LANGUAGE:CUDA>:${CUOPT_CUDA_FLAGS}>"
-)
+if(NOT BUILD_LP_ONLY)
```
ramakrishnap-nv, line 374: Should this be 
```suggestion
if(BUILD_LP_ONLY)
```

cpp/tests/CMakeLists.txt
```diff
@@ -13,28 +13,31 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-add_library(cuopttestutils STATIC
-    routing/utilities/check_constraints.cu
-)
+if(BUILD_TESTS)
```
ramakrishnap-nv, line 16: Is there an option where we can skip directly from cpp/CMakefiles.txt

cpp/libmps_parser/CMakeLists.txt
```diff
@@ -84,15 +84,17 @@ target_compile_options(mps_parser
...
-file(WRITE "${MPS_PARSER_BINARY_DIR}/fatbin.ld"
-  [=[
-SECTIONS
-{
-  .nvFatBinSegment : { *(.nvFatBinSegment) }
-  .nv_fatbin : { *(.nv_fatbin) }
-}
-]=])
-target_link_options(mps_parser PRIVATE "${MPS_PARSER_BINARY_DIR}/fatbin.ld")
+if(WRITE_FATBIN)
```
rg20, line 87: Remove this always?

build.sh
Kh4ster: Changed!

cpp/CMakeLists.txt
Kh4ster: Build test is above. We had it already but it was not used and not working in practice

build.sh
Kh4ster: I think we can keep this  for now and see if someone needs it

cpp/CMakeLists.txt
Kh4ster: No I confirm we don't want to have cuopt_cli if we are in lp_only mode since there is a call to the MIP solver

cpp/CMakeLists.txt
Kh4ster: Good catch for the BUILD_FATBIN!

cpp/libmps_parser/CMakeLists.txt
Kh4ster: What do you mean? Remove the fatbin writing completly?

cpp/tests/CMakeLists.txt
Kh4ster: I don't know any. That's the only way I found to truely diable the compilation of tests

cpp/libmps_parser/CMakeLists.txt
ramakrishnap-nv: I think he is asking, by default currently this was disabled, so do we need to disable it by default might be the question. Think I also had asked this question.

ramakrishnap-nv: And you have changed it if I am not wrong.

cpp/tests/CMakeLists.txt
```diff
@@ -82,12 +85,14 @@ if(RAPIDS_DATASET_ROOT_DIR)
...
 
 # ## test sources
-add_subdirectory(routing)
-add_subdirectory(linear_programming)
-add_subdirectory(distance_engine)
-add_subdirectory(mip)
-add_subdirectory(dual_simplex)
-add_subdirectory(examples)
-add_subdirectory(utilities)
+if(BUILD_TESTS)
```
akifcorduk, line 88: I think you need to also do a check here on skip routing. If skip routing is selected, you should not build routing, distance_engine and example.

benchmarks/linear_programming/utils/benchmark_lp_mittelmann.sh
```diff
@@ -30,12 +30,12 @@ echo "Download done"
...
 export CUDA_MODULE_LOADING=EAGER
 
-# Benchmark all instances (cuOpt needs to be compiled first)
+# Benchmark all instances (cuOpt needs to be compiled first, you can compile in LP only mode and you should turn on BUILD_LP_BENCHMARKS)
 for instance in ${CUOPT_HOME}/benchmarks/linear_programming/datasets/*/ ; do
     # Will generate the solver log for each instance. Could addtionally generate the solution file by uncommenting the --solution-path
     instance_name=$(basename $instance)
     echo "Parsing ${instance_name}.mps then solving"
-    ${CUOPT_HOME}/cpp/build/solve_MPS_file --path ${CUOPT_HOME}/benchmarks/linear_programming/datasets/${instance_name}/${instance_name}.mps --time-limit 3600 # --solution-path $CUOPT_HOME/benchmarks/linear_programming/datasets/$instance.sol
+    ${CUOPT_HOME}/cpp/build/solve_LP --path ${CUOPT_HOME}/benchmarks/linear_programming/datasets/${instance_name}/${instance_name}.mps --time-limit 3600 # --solution-path $CUOPT_HOME/benchmarks/linear_programming/datasets/$instance.sol
```
rg20, line 38: Let's start using cuopt_cli here?  That's what Hans would be using as well.

cpp/CMakeLists.txt
```diff
@@ -361,44 +371,63 @@ if(Doxygen_FOUND)
...
+  )
+endif()
 
-option(BUILD_BENCHMARKS "Build benchmarks" ON)
-if(BUILD_BENCHMARKS)
-  add_executable(solve_MPS_file ../benchmarks/linear_programming/cuopt/run_mip.cpp)
-  target_compile_options(solve_MPS_file
+option(BUILD_LP_BENCHMARKS "Build LP benchmarks" OFF)
+if(BUILD_LP_BENCHMARKS)
+  add_executable(solve_LP ../benchmarks/linear_programming/cuopt/run_pdlp.cu)
```
rg20, line 425: For benchmarking, I think we should directly use the cuopt_cli, because that's what everyone will use

Kh4ster: The issue I have is run_pdlp has more options that cuopt_cli

benchmarks/linear_programming/utils/benchmark_lp_mittelmann.sh
Kh4ster: same comment as below

rg20: Here, you are not using any of those additional settings/options from run_pdlp right? We can still use run_pdlp for when benchmarking with those settings.

aliceb-nv (approved):
Approving :) Awesome work Nicolas, this is going to be super useful
Is it ready to merge?

rgsl888prabhu (approved):

PR #330, opened 2025-08-21
Fix logs

akifcorduk:
This PR fixes the solution logging when no external presolve option is provided. Adds a minor log change to population degeneration.


chris-maes:
I think we should only print the final log line from heuristics, if we are in heuristics only mode (which should be phased out), or if B&B fails. All the information in that line is displayed in the log. And its strange to see the same info twice.

PR #337, opened 2025-08-21
Reduce hard-coded version usage in repo

### LINKED ISSUE #329: [FEA] Fix usage of too many version like YY.MM.XX  in repository

Currently we have too many YY.MM.XX usage in repo to show version being used, but most of these can be made to pull from VERSION file so they need not be update every-time for release with script reducing any issues.

More examples and suggestions can be found here

https://github.com/NVIDIA/cuopt/pull/286#pullrequestreview-3133812928

rgsl888prabhu:
## Description
Pr reduces number of hard coded versions in repo and also fixes few issues.

## Issue
closes #329 

## Checklist

- Testing
   - [ ] Added tests
   - [x] NA
- Documentation
   - [ ] Added new documentation
   - [x] NA

### DIFF
--- a/python/cuopt_server/cuopt_server/utils/data_definition.py
+++ b/python/cuopt_server/cuopt_server/utils/data_definition.py
@@ -24,7 +24,6 @@ import jsonref
 from pydantic import BaseModel, Extra, Field
 
 from .._version import __version_major_minor__
-
 from .linear_programming.data_definition import (  # noqa
     IncumbentSolution,
     LPData,
--- a/python/cuopt_server/cuopt_server/__init__.py
+++ b/python/cuopt_server/cuopt_server/__init__.py
@@ -14,4 +14,8 @@
 # limitations under the License.
 
 from cuopt_server import cuopt_service
-from cuopt_server._version import __git_commit__, __version__, __version_major_minor__
+from cuopt_server._version import (
+    __git_commit__,
+    __version__,
+    __version_major_minor__,
+)


ci/utils/update_doc_versions.py
```diff
@@ -0,0 +1,94 @@
...
+        repo_root / "docs" / "cuopt" / "source" / "versions1.json"
+    )
+
+    # Update versions
+    was_updated = update_versions(version_file_path, versions_file_path)
+
+    if was_updated:
+        print("Version update completed successfully!")
+    else:
+        print("No update needed - version already exists")
```
jameslamb: This is fine for debugging but just note... in `pre-commit`, stdout like this will only be printed if the hook fails (script returns a non-0 exit code).

python/cuopt_server/cuopt_server/utils/data_definition.py
```diff
@@ -343,7 +345,12 @@ class LogResponseModel(StrictModel):
             "application/json": {
                 "examples": {
                     "Healthy response": {
-                        "value": {"status": "RUNNING", "version": "25.10"}
+                        "value": {
+                            "status": "RUNNING",
+                            "version": ".".join(
+                                cuopt.__version__.split(".")[:2]
```
jameslamb: Since this is referenced in multiple places and should be cheap to implement, and since I imagine `cuopt-server` is more sensitive to startup time than `cuopt` itself, it might be worthwhile to move this to `cuopt`, like this:

```python
__version_major_minor__ = ".".join(__version__.split(".")[:2])
```

Then all of these places could just do

```python
from cuopt import __version_major_minor__
```

ramakrishnap-nv: Awesome suggestion, will add this.

rgsl888prabhu:
@rg20 @hlinsen may I get your review ?

hlinsen (approved):
lgtm

PR #339, opened 2025-08-22
Fix initial solution accepted index update in generate_initial()

### LINKED ISSUE #338: [BUG] Initial solution ‘accepted’ index updated incorrectly in generate_initial() (Routing solver)
When at least one initial solution has a route count greater than `target_vehicles_`, the code assigns an incorrect value to `next_injection`. For simplicity, let me consider example with a single initial solution by inspecting the first injection iteration. However, the same indexing mismatch can occur with multiple initial solutions whenever the route-count filter’s while-loop is entered.
https://github.com/NVIDIA/cuopt/blob/c76c186124bfde26409d52f9d5377fdd423ae1d3/cpp/src/routing/diversity/diverse_solver.hpp#L863-L873

- At line 865, `temp_pair.first` is set from `injection_info.solutions[0]`.
- Entering the while-loop at lines 868–871, `temp_pair.first` is reassigned (still to `injection_info.solutions[0]` in this example), but the iterator is incremented and, at the end, `next_injection` is set to 1.

As a result, later in the function:
https://github.com/NVIDIA/cuopt/blob/c76c186124bfde26409d52f9d5377fdd423ae1d3/cpp/src/routing/diversity/diverse_solver.hpp#L896-L900

- At line 897, the solver injects `temp_pair.first` (which still refers to solution index 0),
- But at line 900, it updates `injection_info.accepted[next_injection++]` using index 1.

This is incorrect and, when there is only one initial solution (n_sol == 1), it also causes an out-of-bounds write.

Additionally, at the end of the solve, you can get the acceptance flags for initial solutions with `assignment_t::get_accepted()` (see [assignment.cu (L187–L191)](https://github.com/NVIDIA/cuopt/blob/c76c186124bfde26409d52f9d5377fdd423ae1d3/cpp/src/routing/assignment.cu#L187-L191)). Due to the indexing mismatch described above, the injected solution can be evaluated and inserted, yet the corresponding accepted entry (e.g., `accepted[0]`) may remain -1 (“not evaluated”) instead of 1  or 0 (“accepted” or "not accepted"). This is misleading for users and downstream tooling (including the Python API), since it suggests the initial solution wasn’t considered when it actually was.

I will create PR with fix.

SomovMike:
## Description

Fix index handling when selecting initial solutions in `generate_initial()`.

- Adjust while-loop bound from `injection_it < n_sol` to `injection_it < n_sol - 1` to prevent advancing past the last valid index.
- Change candidate selection to use pre-increment (`++injection_it`) when assigning to `temp_pair.first`, ensuring the selected solution and `next_injection` remain aligned.

After the loop, next_injection now points to the same candidate stored in `temp_pair.first`, preventing incorrect updates to accepted.

## Issue
Closes #338

### DIFF
--- a/cpp/src/routing/diversity/diverse_solver.hpp
+++ b/cpp/src/routing/diversity/diverse_solver.hpp
@@ -865,9 +865,9 @@ struct solve {
           temp_pair.first = injection_info.solutions[next_injection];
           if (!p->has_vehicle_fixed_costs()) {
             auto injection_it = next_injection;
-            while (injection_it < injection_info.n_sol &&
+            while (injection_it < injection_info.n_sol-1 &&
                    temp_pair.first.sol.get_n_routes() > target_vehicles_) {
-              temp_pair.first = injection_info.solutions[injection_it++];
+              temp_pair.first = injection_info.solutions[++injection_it];
             }
             next_injection = injection_it;
           }


rgsl888prabhu:
Thank you @SomovMike for the PR, one of us will review your PR shortly.

cpp/src/routing/diversity/diverse_solver.hpp
```diff
@@ -865,9 +865,9 @@ struct solve {
           temp_pair.first = injection_info.solutions[next_injection];
           if (!p->has_vehicle_fixed_costs()) {
             auto injection_it = next_injection;
-            while (injection_it < injection_info.n_sol &&
+            while (injection_it < injection_info.n_sol-1 &&
```
hlinsen, line 868: Thanks for reporting the bug and submitting the PR!
Could you please add the failing test here https://github.com/NVIDIA/cuopt/blob/branch-25.10/python/cuopt/cuopt/tests/routing/test_initial_solutions.py? Assuming you have the code ready.

Regarding the fix, after the change in your example we would be skipping sol[0] and add try to add it while omitting the target vehicle check.
We should set accepted in the for loop:
```
            while (injection_it < injection_info.n_sol &&
                   temp_pair.first.sol.get_n_routes() > target_vehicles_) {
              injection_info.accepted[injection_it] = 0;
              temp_pair.first = injection_info.solutions[injection_it++];
            }
```

And [here](https://github.com/NVIDIA/cuopt/blob/c76c186124bfde26409d52f9d5377fdd423ae1d3/cpp/src/routing/diversity/diverse_solver.hpp#L896) check that `next_injection < injection_info.n_sol` as we could skip all solutions due to vehicle count and try to insert temp.first then write to accepted thus the out of bounds.

rgsl888prabhu:
@SomovMike May I request you to address review comment from @hlinsen

cpp/src/routing/diversity/diverse_solver.hpp
SomovMike: Hi @hlinsen, thank you for your patience, and apologies for the delayed response.

Unfortunately, I don’t have a minimal failing test case available, as I encountered the issue on a large instance while working on my own implementation. I hope that’s not a problem, since the issue seems fairly clear from the code itself.

Regarding the fix, I think it depends on how you envision the logic of the algorithm. If the primary objective is to minimize the number of vehicles, and a solution with fewer vehicles has already been found, then it might make sense to skip all initial solutions with more vehicles.

That said, in my opinion, it could still be beneficial to include those initial solutions with more vehicles, since the crossover algorithm might be able to leverage them to generate improved solutions.

hlinsen: In cuOpt we always assume minimization unless `vehicle_fixed_costs` are set or we use a fixed vehicle count in this case it does not change throughout the solve. Before the vehicle skip loop we are explicitly checking if vehicle fixed costs are set so following that logic we would not want to insert these vehicles at all.

hlinsen: Hi @SomovMike , let me know if my comment makes sense. We should try to merge this asap since the fix is pretty small in any case :)

SomovMike: Hi @hlinsen, okey, I got it. Your comment makes sense.

Edgarruiz8585:
***@***.*** by Edgarruiz8585

Obtener Outlook para Android<https://aka.ms/AAb9ysg>

________________________________
De: github-actions[bot] ***@***.***>
Enviado: sábado, junio 27, 2026 3:46:42 a. m.
Para: NVIDIA/cuopt ***@***.***>
Cc: Subscribed ***@***.***>
Asunto: Re: [NVIDIA/cuopt] Fix initial solution accepted index update in generate_initial() (PR #339)

[https://avatars.githubusercontent.com/in/15368?s=20&v=4]github-actions[bot] left a comment (NVIDIA/cuopt#339)<https://github.com/NVIDIA/cuopt/pull/339#issuecomment-4816484655>

🔔 Hi @anandhkb<https://github.com/anandhkb> @SomovMike<https://github.com/SomovMike>, this pull request has had no activity for 7 days. Please update or let us know if it can be closed. Thank you!

If this is an "epic" issue, then please add the "epic" label to this issue.
If it is a PR and not ready for review, then please convert this to draft.
If you just want to switch off this notification, then use the "skip inactivity reminder" label.

  *   `Epico running by Edgarruiz8585 ***@***.***

Edgarruiz8585:
> 🔔 Hi @anandhkb @hlinsen @SomovMike, this pull request has had no activity for 7 days. Please update or let us know if it can be closed. Thank you!
> 
> If this is an "epic" issue, then please add the "epic" label to this issue.
> If it is a PR and not ready for review, then please convert this to draft.
> If you just want to switch off this notification, then use the "skip inactivity reminder" label.

> Saltar a label <Edgarruiz8585>

cpp/src/routing/diversity/diverse_solver.hpp
Edgarruiz8585, line 1: * } */{Edgar_Manuel_Ruiz_Arias_217596670, title redireccion on ={Edgar Manuel Ruiz Arias mr ed}, journal=default

Edgarruiz8585 (commented):
🆗

Issue #150, opened 2025-06-26
[BUG] Lots of array out-of-bound accesses


legrosbuffle:
**Describe the bug**

There are several out-of-bounds accesses in cuop:. Two easy ones to trigger are:

 - In [`compute_max_variable_violation`](https://github.com/NVIDIA/cuopt/blob/63fbb6b22c8949798fffe8cb34ace85ad203f2bb/cpp/src/mip/solution/solution.cu#L542), it is assumed that `variable_lower_bounds` `variable_upper_bounds` and `assignment` are of size at least `n_variables`. This fails to be the case when there are free variables. to the problem's lower/upper bounds and the solution's assignment arrays.
 - in [`has_variable_bounds_violation`](https://github.com/NVIDIA/cuopt/blob/63fbb6b22c8949798fffe8cb34ace85ad203f2bb/cpp/src/mip/utils.cuh#L344), a similar thing happens.

There are other cases when those types of issues are caught by existing assertions, e.g. in [repair_solution](https://github.com/NVIDIA/cuopt/blob/63fbb6b22c8949798fffe8cb34ace85ad203f2bb/cpp/src/dual_simplex/branch_and_bound.cpp#L327), I'm surprised those were not caught in unit tests by the build bots ?

**Steps/Code to reproduce bug**

A bunch of tests show the issue. For example, run `bound_standardization_test` with assertions on.

**Expected behavior**

No out-of-bound accesses

**Environment details (please complete the following information):**
 - Environment location: Custom
 - Method of cuOpt install: from source
 

**Additional context**

https://github.com/NVIDIA/cuopt/pull/146

akifcorduk:
@legrosbuffle could you please give more details on the environment? I was not able to reproduce this on gcc 13.3.0 and nvcc 12.8.

akifcorduk:
@chris-maes The repair issue is reproducable in one of the CI runs:
https://github.com/NVIDIA/cuopt/actions/runs/16776602702/job/47507156031?pr=258#step:10:2580

PR #346, opened 2025-08-25
Fix out-of-bound access in `clean_up_infeasibilities`.

legrosbuffle:
This OOB access happens when the last element is the one being removed: the second access to `infeasibility_indices[k]` after the `pop_back` is invalid.

This is another example for issue #150.


rgsl888prabhu:
@rg20 @hlinsen may I get your reviews on this PR ?

rg20 (approved):

PR #348, opened 2025-08-25
[FIX] Uninitialized access fixes + improved initcheck error reporting

aliceb-nv:
Up until now, compute-sanitizer's initcheck tool was unusable on the codebase due to the numerous false positives generated by CCCL and cuSparse. Recent CUDA 12.8 improvements now allow code to exclude such false positives more finely by marking memory ranges as treat-initialized.

This PR contains a few uninitialized access fixes throughout the codebase, and silences common false positives.
In a future PR, automated initcheck runs could be performed by CI (as is the case with CUB in CCCL)


tmckayus (approved):

ci/compute-sanitizer-suppressions.xml
```diff
@@ -0,0 +1,225 @@
+<?xml version="1.0" encoding="utf-8"?>
```
akifcorduk, line 1: Do we need this xml?

cpp/src/linear_programming/pdhg.cu
```diff
@@ -122,6 +124,10 @@ void pdhg_solver_t<i_t, f_t>::compute_At_y()
 {
   // A_t @ y
 
+  // cusparse flags a false positive here on the destination tmp buffer, silence it
+  cuopt::mark_span_as_initialized(make_span(current_saddle_point_state_.get_current_AtY()),
```
akifcorduk: Instead of adding this call before all cusparse calls, can't we wrap cusparse calls in a function. I guess we use few of them.

cpp/src/mip/diversity/population.cu
```diff
@@ -745,6 +745,7 @@ void population_t<i_t, f_t>::print()
     if (index.first == 0 && solutions[0].first) {
       CUOPT_LOG_DEBUG(" Best feasible: %f", solutions[index.first].second.get_user_objective());
     }
+    if (index.first == 0 && !solutions[0].first) continue;
```
akifcorduk, line 748: I this ever triggered or is it to prevent some static analysis?

cpp/src/mip/local_search/rounding/constraint_prop.cu
```diff
@@ -816,6 +816,7 @@ bool constraint_prop_t<i_t, f_t>::is_problem_ii(problem_t<i_t, f_t>& problem)
 {
   bounds_update.calculate_activity_on_problem_bounds(problem);
   bounds_update.calculate_infeasible_redundant_constraints(problem);
+  multi_probe.calculate_activity(problem, problem.handle_ptr);
```
akifcorduk, line 819: Why do we need that? Are we using multi_probe activity somewhere without initializing it?

cpp/src/mip/presolve/trivial_presolve.cuh
```diff
@@ -248,6 +248,10 @@ void update_from_csr(problem_t<i_t, f_t>& pb)
   //  create renumbering maps
   rmm::device_uvector<i_t> cnst_renum_ids(pb.n_constraints, handle_ptr->get_stream());
   rmm::device_uvector<i_t> var_renum_ids(pb.n_variables, handle_ptr->get_stream());
+  thrust::sequence(handle_ptr->get_thrust_policy(), cnst_renum_ids.begin(), cnst_renum_ids.end());
```
akifcorduk: Why do we need those? They are initialized in the output of inclusive_scan ?

cpp/src/mip/problem/problem.cu
```diff
@@ -884,6 +889,9 @@ void problem_t<i_t, f_t>::compute_related_variables(double time_limit)
 
     // generate the related var offsets from the prefix sum
     auto offset_it = related_variables_offsets.begin() + 1 + output_offset;
+
+    // avoid initcheck false positive
+    cuopt::mark_span_as_initialized(make_span(offsets), handle_ptr->get_stream());
```
akifcorduk: If something here is uninitialized, there might be a bug right? We are depending on offsets values.

cpp/src/mip/relaxed_lp/lp_state.cuh
```diff
@@ -56,8 +56,24 @@ class lp_state_t {
...
   void resize(problem_t<i_t, f_t>& problem, rmm::cuda_stream_view stream)
   {
+    i_t prev_primal_size = prev_primal.size();
+    i_t prev_dual_size   = prev_dual.size();
     prev_primal.resize(problem.n_variables, stream);
     prev_dual.resize(problem.n_constraints, stream);
+
+    // zero-fill the newly allocated space
+    if (prev_primal_size < problem.n_variables) {
+      thrust::fill(problem.handle_ptr->get_thrust_policy(),
```
akifcorduk, line 66: Here i would clamp_within_bounds for the remaining vars.

cpp/src/mip/solution/solution.cu
```diff
@@ -273,6 +286,7 @@ void solution_t<i_t, f_t>::compute_constraints()
 {
   if (problem_ptr->n_constraints == 0) { return; }
 
+  n_feasible_constraints.set_value_to_zero_async(handle_ptr->get_stream());
```
akifcorduk, line 289: Good catch!

cpp/src/mip/solution/solution.cu
```diff
@@ -476,6 +489,8 @@ template <typename i_t, typename f_t>
 f_t solution_t<i_t, f_t>::get_quality(const rmm::device_uvector<f_t>& cstr_weights,
                                       const rmm::device_scalar<f_t>& objective_weight)
 {
+  compute_constraints();
```
akifcorduk, line 492: I don't think we need that. get_quality should only be called after `compute_feasibility()`, instead of here we should put the `compute_feasibility()` before where it was uninitialized.

cpp/src/utilities/cuda_helpers.cuh
```diff
@@ -208,4 +221,48 @@ DI void sorted_insert(T* array, T item, int curr_size, int max_size)
...
+// previously marked as initialized is now occupied by a new uninitialized object Unlikely to cause
+// issues in practice - but worth noting (RAII? I'm not even sure the API allows to un-mark a range
+// as initialized)
+static inline void mark_memory_as_initialized(const void* ptr, size_t size, cudaStream_t stream = 0)
+{
+#if CUDART_VERSION >= 12080
+
+  if (size == 0 || ptr == nullptr) return;
+
+#if defined(CUDA_API_PER_THREAD_DEFAULT_STREAM)
```
akifcorduk, line 235: Isn't CUDA_API_PER_THREAD_DEFAULT_STREAM always defined on newer cudart versions?

ci/compute-sanitizer-suppressions.xml
aliceb-nv: It is necessary to run compute-sanitizer --initchecks without false positives. It will be useful if we include initcheck runs as part of CI in the future (which I think we should consider)
CCCL has a similar file on their repo: https://github.com/NVIDIA/cccl/blob/main/ci/compute-sanitizer-suppressions.xml

cpp/src/linear_programming/pdhg.cu
aliceb-nv: Good point, I will add a wrapper for cusparsespmv

cpp/src/mip/diversity/population.cu
aliceb-nv: IIRC ff the population is empty, the total_excess value is uninitialized (and thus this appears in the logs as a very long float value). This is functionally harmless, but triggers a positive in initcheck. I decided to keep this check since it also makes the logs a bit cleaner

cpp/src/mip/local_search/rounding/constraint_prop.cu
aliceb-nv: `sort_by_implied_slack_consumption` uses multi_probe.min/max_activity; I seem to remember that constraint_prop::apply_round is sometimes called with an uninitialized multi_probe activity. I haven't been able to reproduce it with the current main branch so maybe this has been fixed
Do you think this is problematic?

cpp/src/mip/presolve/trivial_presolve.cuh
aliceb-nv: You're right, this is a debugging leftover. Removing

cpp/src/mip/solution/solution.cu
aliceb-nv: I am not fully confident in this approach since this puts the responsibility of ensuring an invariant (constraint values are up-to-date) to the caller. `get_quality` shouldn't really be too performance critical 
What do you think?

cpp/src/utilities/cuda_helpers.cuh
aliceb-nv: I'm not sure😅
The documentation isn't crystal clear about it
https://docs.nvidia.com/cuda/cuda-driver-api/stream-sync-behavior.html

Issue #349, opened 2025-08-26
[BUG] High GPU memory usage for MIP solver


nguidotti:
cuOpt consume a large amount of GPU memory when solving MIP problems, even if they are very small (e.g., [markshare_4_0](https://miplib.zib.de/instance_details_markshare_4_0.html)). In fact, solving [markshare_4_0](https://miplib.zib.de/instance_details_markshare_4_0.html) with just 34 variables and 4 constraints consumes the same amount of memory as [savsched1](https://miplib.zib.de/instance_details_savsched1.html) with 328575 variables and 295989 constraints. 

As a result, the solver runs out of memory when running on a RTX ADA 1000 (6 GB of VRAM) and always consume 90 GB of VRAM when running on a GB100.

PR #351, opened 2025-08-26
[DRAFT] Fix high GPU memory usage

aliceb-nv:
Addresses https://github.com/NVIDIA/cuopt/issues/349

compute_related_variables was heuristically allocating memory based on A100/H100 with >=40GB of VRAM.
This is now automatically adjusted based on the total VRAM of the device

A command-line option has also been added to solve_MPS_file to specify device memory allocation limits for ease of testing.

### DIFF
--- a/cpp/src/utilities/cuda_helpers.cuh
+++ b/cpp/src/utilities/cuda_helpers.cuh
@@ -24,6 +24,8 @@
 #include <raft/util/cuda_utils.cuh>
 #include <raft/util/cudart_utils.hpp>
 #include <rmm/device_uvector.hpp>
+#include <rmm/mr/device/cuda_async_memory_resource.hpp>
+#include <rmm/mr/device/limiting_resource_adaptor.hpp>
 
 namespace cuopt {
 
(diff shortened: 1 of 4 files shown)


cpp/src/utilities/cuda_helpers.cuh
```diff
@@ -208,4 +210,27 @@ DI void sorted_insert(T* array, T item, int curr_size, int max_size)
   array[0] = item;
 }
 
+inline size_t get_device_memory_size()
+{
+  // Otherwise, we need to get the free memory from the device
+  size_t free_mem, total_mem;
+  cudaMemGetInfo(&free_mem, &total_mem);
```
rg20, line 217: In pool memory allocator, we would have probably allocated most of the memory, but the allocator itself might have available memory.

aliceb-nv: This is only used to get the total device memory :) `free_mem` is a placeholder here, only total_mem is used

aliceb-nv:
Sure @nguidotti, you can go ahead, thanks!

nguidotti (approved):

PR #361, opened 2025-08-29
Return Infeasible if the user problem contains crossing bounds

### LINKED ISSUE #260: [BUG] Invalid bounds need to return infeasible solution rather than erroring invalid bound
**Describe the bug**
Consider a scenario where lower bound is > upper bound, in this case when you try to set value, you get error invalid bounds. 

```
{"CUOPT_ERROR_TYPE": "ValidationError", "msg": "Variable bounds are invalid"}
```

Test file 

[test.zip](https://github.com/user-attachments/files/21637890/test.zip)

This can't be reproduced through cli or with mps 

**Expected behavior**
Should return infeasible

aliceb-nv:
3rd party APIs expect problems with crossing bounds to be treated as infeasible by the solver. 
This PR replaces the previous behavior (exception on crossing bounds in input problem) to align accordingly.
closes #260

### DIFF
--- a/cpp/libmps_parser/src/mps_parser.cpp
+++ b/cpp/libmps_parser/src/mps_parser.cpp
@@ -505,9 +505,12 @@ void mps_parser_t<i_t, f_t>::parse_string(char* buf)
       variable_lower_bounds[i] = 0;
       variable_upper_bounds[i] = 1;
     }
-    mps_parser_expects(variable_lower_bounds[i] <= variable_upper_bounds[i],
-                       error_type_t::ValidationError,
-                       "MPS Parser Internal Error - Please contact cuOpt team");
+    if (variable_lower_bounds[i] > variable_upper_bounds[i]) {
+      printf("WARNING: Variable %d has crossing bounds: %f > %f\n",
+             i,
+             variable_lower_bounds[i],
+             variable_upper_bounds[i]);
+    }
   }
 }
 
--- a/cpp/src/mip/solve.cu
+++ b/cpp/src/mip/solve.cu
@@ -176,6 +176,13 @@ mip_solution_t<i_t, f_t> solve_mip(optimization_problem_t<i_t, f_t>& op_problem,
     problem_checking_t<i_t, f_t>::check_problem_representation(op_problem);
     problem_checking_t<i_t, f_t>::check_initial_solution_representation(op_problem, settings);
 
+    // Check for crossing bounds. Return infeasible if there are any
+    if (problem_checking_t<i_t, f_t>::has_crossing_bounds(op_problem)) {
+      return mip_solution_t<i_t, f_t>(mip_termination_status_t::Infeasible,
+                                      solver_stats_t<i_t, f_t>{},
+                                      op_problem.get_handle_ptr()->get_stream());
+    }
+
     auto timer = cuopt::timer_t(time_limit);
 
     double presolve_time = 0.0;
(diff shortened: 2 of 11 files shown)


cpp/tests/linear_programming/c_api_tests/c_api_test.c
nguidotti, line 1: A small suggestion: instead of `goto`, is not better to transform 

```
Done:

  cuOptDestroyProblem(&problem);
  cuOptDestroySolverSettings(&settings);
```
into a separated routine (e.g., `cleanup`) and then call `cleanup` + `return` to end the routine?

cpp/libmps_parser/src/mps_parser.cpp
```diff
@@ -505,9 +505,12 @@ void mps_parser_t<i_t, f_t>::parse_string(char* buf)
       variable_lower_bounds[i] = 0;
       variable_upper_bounds[i] = 1;
     }
-    mps_parser_expects(variable_lower_bounds[i] <= variable_upper_bounds[i],
-                       error_type_t::ValidationError,
-                       "MPS Parser Internal Error - Please contact cuOpt team");
+    if (variable_lower_bounds[i] > variable_upper_bounds[i]) {
+      printf("WARNING: Variable %d has crossing bounds: %f > %f\n",
```
akifcorduk, line 788: Is logger not suitable here?

cpp/src/mip/solve.cu
```diff
@@ -176,6 +176,13 @@ mip_solution_t<i_t, f_t> solve_mip(optimization_problem_t<i_t, f_t>& op_problem,
     problem_checking_t<i_t, f_t>::check_problem_representation(op_problem);
     problem_checking_t<i_t, f_t>::check_initial_solution_representation(op_problem, settings);
 
+    // Check for crossing bounds. Return infeasible if there are any
+    if (problem_checking_t<i_t, f_t>::has_crossing_bounds(op_problem)) {
```
akifcorduk, line 183: There may be some unset vars which we set in problem constructor. I think this should come after the problem cosntructor.

rg20, line 183: We have to do this for LP as well.

cpp/tests/mip/unit_test.cu
```diff
@@ -159,7 +159,7 @@ TEST(ErrorTest, TestError)
 
   // Set constraint bounds
   std::vector<double> lower_bounds = {1.0};
-  std::vector<double> upper_bounds = {0.0};
+  std::vector<double> upper_bounds = {1.0, 1.0};
```
rg20, line 162: The sizes are different for lower and upper bounds?

aliceb-nv: This is a test that is meant to catch is a problem is ill-formed (e.g., contains bounds of different sizes)

cpp/tests/linear_programming/c_api_tests/c_api_test.c
aliceb-nv: That part of the code wasn't written by me, but to be fair it's more of a flavour / personal prerefence kind of thing :) It is a common pattern in C for resource cleanup; and it lets the code have a single exit point

cpp/libmps_parser/src/mps_parser.cpp
aliceb-nv: We don't link against rapids_logger in libmps_parser iirc

cpp/tests/linear_programming/c_api_tests/c_api_test.c
nguidotti: Fair enough

nguidotti (approved):

PR #363, opened 2025-09-01
Fix FJ numerical issue turning a solution infeasible 

### LINKED ISSUE #360: [BUG] Feasibility Jump makes some solutions infeasible
This is a rare occurance where FJ causes feasible input solutions to become infesible. The assert in FJ is triggerred. It might be reproduced with 50v-10 by repeatedly running it.

aliceb-nv:
closes #360

On some numerically tricky problems, the infeasibility excess may be close enough to the tolerance that FJ considers it feasible, but solution_t does not. 
This PR ensures that FJ doesn't return this new solution if solution_t::compute_feasibility() considers it infeasible.
The refresh period for the computation of constraint LHS values is also reduced to minimize error.


aliceb-nv:
Will be fixed by the CPUFJ PR

PR #368, opened 2025-09-02
Implement node presolve

### LINKED ISSUE #276: [FEA] Node Presolve
* MIP Node Presolve
** Bound strengthening at the nodes (S) P0

rg20:
## Description
This PR implements node presolve using the bounds strengthening algorithm. At each node in branch-and-bound tree,  host presolve is run to (i) improve the bounds on the variables (ii) to detect infeasibility and fathom nodes accordingly

**Details**

- The bounds strengthening takes into account of integrality constraints. This leads to more infeasibility detection than doing the dual simplex. 
- Bounds strengthening is done only starting with the constraints that are associated with the branched variables (all up to the root node) only.  For examples, if the current node is depth 7, only constraints relating to these 7 variables are considered changed for the first iteration of bounds strengthening. This greatly improves the performance of the node presolve
- This bounds strengthening is performed on the cpu

## Issue
Closes #276 

## Checklist

- Testing
   - [ ] Added tests
   - [x] NA
- Documentation
   - [ ] Added new documentation
   - [x] NA

### DIFF
--- a/cpp/src/dual_simplex/branch_and_bound.cpp
+++ b/cpp/src/dual_simplex/branch_and_bound.cpp
@@ -676,6 +676,10 @@ mip_status_t branch_and_bound_t<i_t, f_t>::solve(mip_solution_t<i_t, f_t>& solut
 
     nodes_explored++;
     if (lp_status == dual::status_t::DUAL_UNBOUNDED) {
+      if (!feasible) {
+        settings.log.printf("Infeasible after bounds strengthening. Fathoming node %d.\n",
+                            nodes_explored);
+      }
       node_ptr->lower_bound = inf;
       std::vector<mip_node_t<i_t, f_t>*> stack;
       node_ptr->set_status(node_status_t::INFEASIBLE, stack);
--- a/cpp/src/dual_simplex/branch_and_bound.cpp
+++ b/cpp/src/dual_simplex/branch_and_bound.cpp
@@ -529,6 +529,8 @@ mip_status_t branch_and_bound_t<i_t, f_t>::solve(mip_solution_t<i_t, f_t>& solut
                          std::move(up_child));  // child pointers moved into the tree
   lp_problem_t leaf_problem =
     original_lp;  // Make a copy of the original LP. We will modify its bounds at each leaf
+  csc_matrix_t<i_t, f_t> Arow(1, 1, 1);
+  original_lp.A.transpose(Arow);
   f_t gap            = get_upper_bound() - lower_bound;
   i_t nodes_explored = 0;
   settings.log.printf(
@@ -631,8 +633,11 @@ mip_status_t branch_and_bound_t<i_t, f_t>::solve(mip_solution_t<i_t, f_t>& solut
     // Set the correct bounds for the leaf problem
     leaf_problem.lower = original_lp.lower;
     leaf_problem.upper = original_lp.upper;
-    // FIXME:: should we get the bounds from the node/parent instead of the original problem?
-    node_ptr->get_variable_bounds(leaf_problem.lower, leaf_problem.upper);
+    std::vector<bool> bounds_changed(original_lp.num_cols, false);
+    // Technically, we can get the already strengthened bounds from the node/parent instead of
+    // getting it from the original problem and re-strengthening. But this requires storing
+    // two vectors at each node and potentially cause memory issues
+    node_ptr->get_variable_bounds(leaf_problem.lower, leaf_problem.upper, bounds_changed);
 
     std::vector<variable_status_t>& leaf_vstatus = node_ptr->vstatus;
     lp_solution_t<i_t, f_t> leaf_solution(leaf_problem.num_rows, leaf_problem.num_cols);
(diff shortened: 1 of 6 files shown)


cpp/src/dual_simplex/presolve.cpp
```diff
@@ -18,131 +18,263 @@
...
+      for (i_t p = row_start; p < row_end; ++p) {
         const i_t i = problem.A.i[p];
-        less_rows.push_back(i);
+
+        if (!constraint_changed[i]) { continue; }
+        const f_t a_ik = problem.A.x[p];
+        f_t min_a      = min_activity[i];
+        f_t max_a      = max_activity[i];
+        f_t cnst_lb    = constraint_lb[i];
+        f_t cnst_ub    = constraint_ub[i];
```
kaatish: If you store the delta_max_act and delta_min_act in constraint activity calculation phase (prior to the activity correction in the lines below) then you will be cutting down the memory loads by half in this for loop.

cpp/src/dual_simplex/branch_and_bound.cpp
```diff
@@ -631,6 +631,7 @@ mip_status_t branch_and_bound_t<i_t, f_t>::solve(mip_solution_t<i_t, f_t>& solut
     // Set the correct bounds for the leaf problem
     leaf_problem.lower = original_lp.lower;
     leaf_problem.upper = original_lp.upper;
+    // FIXME:: should we get the bounds from the node/parent instead of the original problem?
```
chris-maes: As we discussed offline, it would be possible to store used the strengthened bounds on all variables. But doing so would require storing 2 x n  x 64 bits of memory per node. The hunch is that it is better to just run node presolve at each node to get the tighter bounds.

cpp/src/dual_simplex/branch_and_bound.cpp
```diff
@@ -642,28 +643,42 @@ mip_status_t branch_and_bound_t<i_t, f_t>::solve(mip_solution_t<i_t, f_t>& solut
...
-        leaf_problem, lp_start_time, lp_settings, leaf_solution, leaf_vstatus, leaf_edge_norms);
-      lp_status = convert_lp_status_to_dual_status(second_status);
+    lp_settings.cut_off    = upper_bound + settings.dual_tol;
+    lp_settings.inside_mip = 2;
+
+    // in B&B we only have equality constraints
+    std::vector<char> row_sense(leaf_problem.num_rows, 'E');
+    bool feasible = bound_strengthening(row_sense, lp_settings, leaf_problem, var_types);
+
+    dual::status_t lp_status = dual::status_t::DUAL_UNBOUNDED;
```
chris-maes, line 659: remove this line?

cpp/src/dual_simplex/branch_and_bound.cpp
```diff
@@ -642,28 +643,42 @@ mip_status_t branch_and_bound_t<i_t, f_t>::solve(mip_solution_t<i_t, f_t>& solut
...
-    if (lp_status == dual::status_t::NUMERICAL) {
-      settings.log.printf("Numerical issue node %d. Resolving from scratch.\n", nodes_explored);
-      lp_status_t second_status = solve_linear_program_advanced(
-        leaf_problem, lp_start_time, lp_settings, leaf_solution, leaf_vstatus, leaf_edge_norms);
-      lp_status = convert_lp_status_to_dual_status(second_status);
+    lp_settings.cut_off    = upper_bound + settings.dual_tol;
+    lp_settings.inside_mip = 2;
+
+    // in B&B we only have equality constraints
+    std::vector<char> row_sense(leaf_problem.num_rows, 'E');
```
chris-maes: Maybe move this up, so you aren't allocating and deallocating row sense before each node

cpp/src/dual_simplex/presolve.cpp
```diff
@@ -18,131 +18,263 @@
...
 
-  std::vector<i_t> less_rows;
-  less_rows.reserve(m);
+  std::vector<f_t> min_activity(m);
+  std::vector<f_t> max_activity(m);
+  std::vector<f_t> constraint_lb(m);
+  std::vector<f_t> constraint_ub(m);
+
+  // FIXME:: Instead of initializing constraint_changed to true, we can only look 
+  // at the constraints corresponding to branched variable in branch and bound
```
chris-maes, line 86: This is an important optimization, you might want to have a specific bound strengthening function just for this maybe

```
bool bound_strengthening(const std::vector<char>& row_sense, 
                                              const lp_problem_t<i_t, f_t>& problem, 
                                              i_t branched_var, 
                                              f_t branched_var_lower_bound,
                                              f_t  branched_var_upper_bound)
```

cpp/src/dual_simplex/branch_and_bound.cpp
rg20: I made a change to look for empty vector and treating it as equality in that case. So no more allocations here.

cpp/src/dual_simplex/presolve.cpp
rg20: I actually tried this, but it does not do a good job. I think it might be related to not using the strengthened bounds from the parent.

kaatish (approved):
Looks good to me!

Issue #369, opened 2025-09-03
[FEA] Report solver version / hash in the log


aliceb-nv:
Most other solvers like SCIP/Highs/Gurobi report the version of the solver used in their logs, along with some extra information such as build date or commit hash.

Furthermore, so far users have no way of printing the version of cuOpt they are using. Adding such information to the logs is also important for benchmarking and result reporting purposes.

Possible output format:
```
cuOpt optimization solver, version 25.10 (build hash <hash>)
CPU Model: <model>, Host memory: <RAM>
Thread count: <threads>
GPU Model: <model>, Device memory: <VRAM>
```

PR #370, opened 2025-09-03
Print cuOpt version / machine info before solving

aliceb-nv:
closes https://github.com/NVIDIA/cuopt/issues/369

Most solvers report the version in their logs, along with some extra information such as build date or commit hash.

This PR adds code to cuopt_cli (and solve_MPS_problem, until we deprecate it) to print out the cuOpt version, build hash, and target architectures, along with host/device details when starting the solver.

Example output:

```
cuOpt version: 25.10.0, git hash: 9006ad7, host arch: x86_64, device archs: 90-real
CPU: AMD EPYC 7742 64-Core Processor, threads (physical/logical): 128/256, RAM: 726.66 GiB
CUDA 12.9, device: NVIDIA H100 PCIe (ID 0), VRAM: 79.18 GiB
CUDA device UUID: 1effffffa4ffffff9937-fffffff928-ffff
```


cpp/cuopt_cli.cpp
```diff
@@ -67,6 +67,148 @@
...
+      if (line.find("cpu cores") != std::string::npos) {
+        return std::stoi(line.substr(line.find(":") + 1));
+      }
+    }
+    return 1;
+  }
+  return cores.size();
+}
+
+static std::string get_cpu_model_from_proc()
```
rg20: Move these functions to utilities.  We would want to use these for all the solvers.

cpp/cuopt_cli.cpp
```diff
@@ -96,6 +238,55 @@ int run_single_file(const std::string& file_path,
     return -1;
   }
 
+  int device_id = 0;
```
rg20: Write a function, may be "PrintHardwareCudaCuOptInfos". This helps us calling from each solver easily.

rgsl888prabhu (approved):

cpp/cuopt_cli.cpp
aliceb-nv: I've kept them in the cuopt_cli source file to keep them away from the main solver since this seems more like a CLI tool responsibility rather than a libcuopt one (and the duplication is a consequence of the legacy solve_MPS_file still being in the repo). Do you think this should be moved to the common solver code?

BTW we should probably discuss removing solve_MPS_file for good :) I've dragged my feet about it a little too long but the repo will benefit from lessened duplication

cpp/cuopt_cli.cpp
aliceb-nv: see above :)

cpp/cuopt_cli.cpp
rg20: Log files can be generated from all other APIs (C, Python, service) as well right?  If we restrict this to cli, the log files generated will be different. 

With regards to removing the solve_MPS_file, yes I agree.  We have a script to run the benchmarks (https://github.com/NVIDIA/cuopt/blob/branch-25.10/benchmarks/linear_programming/run_mps_files.sh). Its missing on gathering results, but now that you have the log file readers, that should be more than sufficient.

aliceb-nv: Oh you're completely right, mea culpa :)
This makes sense. I will move this to solver side

PR #372, opened 2025-09-03
Combined variable bounds

kaatish:
## Description
This PR combined the buffers problem_t::variable_lower_bounds and problem_t::variable_upper_bounds into a single buffer. This is done so as to improve the cache utilization when randomly accessing variable bounds.

## Checklist

- Testing
   - [ ] Added tests
   - [ ] NA
- Documentation
   - [ ] Added new documentation
   - [ ] NA


cpp/src/mip/problem/problem.cuh
```diff
@@ -187,8 +187,9 @@ class problem_t {
     raft::device_span<i_t> variables;
     raft::device_span<i_t> offsets;
     raft::device_span<f_t> objective_coefficients;
-    raft::device_span<f_t> variable_lower_bounds;
-    raft::device_span<f_t> variable_upper_bounds;
+    raft::device_span<f_t2> variable_bounds;
+    // raft::device_span<f_t> variable_lower_bounds;
```
hlinsen: Dead code

cpp/src/mip/problem/problem_helpers.cuh
```diff
@@ -90,6 +122,7 @@ static void set_bounds_if_not_set(detail::problem_t<i_t, f_t>& op_problem)
                       transform_bounds_functor<f_t>());
   }
 
+#if 0
```
hlinsen: Can we remove this codeblock?

cpp/src/mip/solution/solution.cu
```diff
@@ -35,11 +35,25 @@
 
 namespace cuopt::linear_programming::detail {
 
+template <typename f_t>
+rmm::device_uvector<f_t> get_lower_bounds(rmm::device_uvector<typename type_2<f_t>::type>& bounds,
```
hlinsen:
```suggestion
rmm::device_uvector<f_t> get_lower_bounds(rmm::device_uvector<typename type_2<f_t>::type> const& bounds,
```

hlinsen (approved):

cpp/src/mip/diversity/diversity_manager.cu
```diff
@@ -320,7 +320,6 @@ bool diversity_manager_t<i_t, f_t>::run_presolve(f_t time_limit)
   }
   stats.presolve_time = presolve_timer.elapsed_time();
   lp_optimal_solution.resize(problem_ptr->n_variables, problem_ptr->handle_ptr->get_stream());
-  problem_ptr->handle_ptr->sync_stream();
```
hlinsen: Why was this removed?

cpp/src/mip/problem/problem_helpers.cuh
```diff
@@ -58,6 +58,38 @@ struct transform_bounds_functor {
   }
 };
 
+#if 1
```
hlinsen: We can remove the macro

cpp/src/mip/local_search/rounding/bounds_repair.cu
```diff
@@ -356,28 +359,28 @@ void bounds_repair_t<i_t, f_t>::apply_move(problem_t<i_t, f_t>& problem,
...
         *candidates.at_least_one_singleton_moved = 1;
       }
 
-      problem.variable_lower_bounds[var_idx] += shift_value;
-      problem.variable_upper_bounds[var_idx] += shift_value;
+      bounds.x += shift_value;
+      bounds.y += shift_value;
+      problem.variable_bounds[var_idx] = bounds;
+      cuopt_assert(original_problem.variable_bounds[var_idx].x <=
+                     bounds.y + problem.tolerances.integrality_tolerance,
```
aliceb-nv: Shouldn't that be `bounds.x`?

cpp/src/utilities/copy_helpers.hpp
```diff
@@ -27,6 +28,66 @@
...
+
+template <>
+struct scalar_type<double2> {
+  using type = double;
+};
+
+template <typename T>
+raft::device_span<typename type_2<T>::type> make_span_2(rmm::device_uvector<T>& container)
+{
+  // TODO : ceildiv or throw assert
```
aliceb-nv: Let's add a cuopt_assert here
(We could also check for the invariant that `static_assert(sizeof(T2) == 2*sizeof(T))` )

aliceb-nv (approved):
Thanks a lot for the great work Aatish!
Since this is a large refactoring PR and the potential for typos is fairly high (x and y instead of lb and ub), let's make sure no asserts get triggered when running on the whole miplib test set

cpp/src/mip/diversity/diversity_manager.cu
kaatish: that was an oversight. Will fix it.

cpp/src/mip/local_search/rounding/bounds_repair.cu
kaatish: Very good catch. Thanks!

cpp/src/linear_programming/utils.cuh
```diff
@@ -77,13 +77,17 @@ struct a_sub_scalar_times_b {
   const f_t* scalar_;
 };
 
-template <typename f_t>
+template <typename f_t, typename f_t2>
```
hlinsen, line 80: Should we move all f_t2 to bound_t? It would help avoiding confusions or typos
```suggestion
template <typename f_t, typename bound_t>
```

cpp/src/linear_programming/utilities/problem_checking.cu
```diff
@@ -313,26 +300,21 @@ template <typename i_t, typename f_t>
 void problem_checking_t<i_t, f_t>::check_unscaled_solution(
   detail::problem_t<i_t, f_t>& op_problem, rmm::device_uvector<f_t> const& assignment)
 {
-  auto& d_variable_upper_bounds = op_problem.variable_upper_bounds;
-  auto& d_variable_lower_bounds = op_problem.variable_lower_bounds;
-  std::vector<f_t> variable_upper_bounds(d_variable_upper_bounds.size());
-  std::vector<f_t> variable_lower_bounds(d_variable_lower_bounds.size());
+  using f_t2              = typename type_2<f_t>::type;
```
hlinsen, line 305: Can we move this to a single header?

cpp/src/linear_programming/utils.cuh
aliceb-nv: I wonder if we could declare `bound_t` as a union of `f_t2` and `{ f_t lower; f_t upper; }` (using an anonymous struct member). Might not play too well with C++ metaprogramming but if it works it could make the code clearer at no cost

kaatish: f_t2 will be used in load balancing not just for bounds but for any values that comes in pairs. Such as :
constraint lower/upper bounds
min/max activity
variable lower/upper bounds.

What we want is to have a double2 vector. I do not think you can create a device_uvector of a union because device_uvector element needs to be trivially_copyable as seen [here](https://github.com/rapidsai/rmm/blob/7cef2f5f30e962e9f3b27a3a3f2753a40277c093/cpp/include/rmm/device_uvector.hpp#L79-L80).

We can pick a better name but I want the name to clearly identify that it is a vector type of an underlying scalar. This will help in clarifying the purpose of using this type :

- let's say you want to randomly access property_0[i] and you know you need property_1[i].
There is no guarantee that property_0[i+1] and property_1[i+1] will be used but they will definitely be loaded in to the cache if you access property_0[i] and property_1[i].
So packing both of them into a double2 would improve cache utilization because you reduce the chances of loading things that you don't need.

aliceb-nv: unless i'm mistaken, unions can be trivially copyiable so long as their members are trivially copyable themselves, right?
On the other hand, we could also typedef bound_t as being f_t2 in the context of constraint/variable bounds
But I think it is no big deal. f_t2 alone is a fine choice

hlinsen: Are we reading max_activity randomly? For constraint bounds the same type `bound_t` would work. I like Alice's suggestion if we can make it work but otherwise we can keep it as is

kaatish: Yes, in the update bounds process (and others like it for eg compute damages kernels in bounds_repair.cu) a block works on a variable and reads the constraints (and properties) associated with it. These reads are random since the constraint ids are not guaranteed to be contiguous.

kaatish: What do you think about having a function get_lower(f_t2) and get_upper(f_t2) for clarity

cpp/src/linear_programming/utilities/problem_checking.cu
kaatish: Do you mean to separate it out from copy_helpers and make a separate header for vector_types?

cpp/src/linear_programming/utils.cuh
aliceb-nv: That sounds good! We don't have to refactor the entire PR to switch to these functions but it'll be good to have for clarity. Maybe they could be renamed shorter like 'lower' and 'upper', and stored in a specific namespace so that they don't contaminate the global namespace and can be brought up via `using`  
I will try converting the code to see if using `union` is possible, but it might be tricky. Let's focus on getting the PR tested and merged for now

cpp/src/mip/local_search/rounding/constraint_prop.cu
```diff
@@ -715,21 +741,19 @@ void constraint_prop_t<i_t, f_t>::update_host_assignment(const solution_t<i_t, f
              sol.handle_ptr->get_stream());
 }
 
+template <typename f_t, typename f_t2>
```
akifcorduk: I can't see this being used. We can remove it i think.

cpp/src/mip/presolve/lb_bounds_udpate_data.cuh
```diff
@@ -0,0 +1,69 @@
+/*
```
akifcorduk: Typo in the file name.

cpp/src/mip/problem/host_helper.cuh
```diff
@@ -52,7 +53,9 @@ struct constraints_delta_t {
 
 template <typename i_t, typename f_t>
 struct variables_delta_t {
+  using f_t2 = typename type_2<f_t>::type;
   std::vector<f_t> objective_coefficients;
+  std::vector<f_t2> variable_bounds;
   std::vector<f_t> lower_bounds;
```
akifcorduk: I think we can remove `lower_bounds` and `upper_bounds` now.

akifcorduk (approved):
LGTM!
One general comment: I wonder if it is a bad idea to define f_t2 in a global header (i.e problem.cuh). I know if it is generally a bad practice, but it might provide ease of use throughout the repo.

cpp/src/linear_programming/utils.cuh
```diff
@@ -129,13 +133,21 @@ struct a_divides_sqrt_b_bounded {
 };
 
 template <typename f_t>
-struct clamp {
+struct constraint_clamp {
```
ahehn-nv, line 136: I'm not reviewing, but glanced at the changes and have a comment here: I would leave the original `clamp` untouched and introduce a new `clamp_bounds` or similar instead. This would a) be a less surprising change in the code, and b) be more consistent with C++ since the original `clamp` roughly matched the [std::clamp](https://en.cppreference.com/w/cpp/algorithm/clamp.html) syntax.

rgsl888prabhu (approved):

cpp/src/linear_programming/utils.cuh
ramakrishnap-nv: @aliceb-nv may be we can add a follow-up PR for this

PR #373, opened 2025-09-04
[DRAFT] cudss with multi threading 

rgsl888prabhu:
## Checklist

- Testing
   - [ ] Added tests
   - [ ] NA
- Documentation
   - [ ] Added new documentation
   - [ ] NA


ci/utils/install_cudss.sh
```diff
@@ -0,0 +1,31 @@
...
+# Clean metadata & install cudss
+if command -v dnf &> /dev/null; then
+    dnf clean all
+    # Adding static library just to please CMAKE requirements
+    dnf -y install libcudss0-static-cuda-12 libcudss0-devel-cuda-12 libcudss0-cuda-12
+elif command -v apt-get &> /dev/null; then
+    apt-get update
+    apt-get install -y libcudss-devel
+else
+    echo "Neither dnf nor apt-get found. Cannot install cudss dependencies."
```
chris-maes, line 29: Why are we installing cudss via apt-get or dnf?  What about just downloading the tarball directly? And saving it somewhere in the cuopt directory?

ramakrishnap-nv: This will help us be upto date with new releases and keeping packages inside repo will cause other dependency issues and also you would need to maintain for different architectures.

chris-maes:
Thanks for doing this!  I merged this into my barrier PR. You should be able to close this PR now.

PR #381, opened 2025-09-09
Loosen presolve tolerance and update timers to report cumulative presolve/solve time

hlinsen:
This PR updates:
- Presolve tolerance setting from 9e-7 to 1e-5 as some models were incorrectly detected as infeasible.
- Usage of `timer_t` class to propagate and report more precisely the cumulative time of presolve + solve time (concurrent, dual, pldp).

### DIFF
--- a/cpp/src/linear_programming/pdlp.cu
+++ b/cpp/src/linear_programming/pdlp.cu
@@ -253,22 +253,17 @@ void pdlp_solver_t<i_t, f_t>::set_initial_dual_solution(
     initial_dual_.data(), initial_dual_solution.data(), initial_dual_solution.size(), stream_view_);
 }
 
-static bool time_limit_reached(const std::chrono::high_resolution_clock::time_point& start_time,
-                               double seconds)
+static bool time_limit_reached(const timer_t& timer, double seconds)
 {
-  auto current_time = std::chrono::high_resolution_clock::now();
-  auto elapsed =
-    std::chrono::duration_cast<std::chrono::milliseconds>(current_time - start_time).count();
-
-  return elapsed >= (seconds * 1000.0);
+  return timer.check_time_limit();
 }
 
 template <typename i_t, typename f_t>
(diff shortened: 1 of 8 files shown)


cpp/src/linear_programming/pdlp.cu
```diff
@@ -253,22 +253,17 @@ void pdlp_solver_t<i_t, f_t>::set_initial_dual_solution(
     initial_dual_.data(), initial_dual_solution.data(), initial_dual_solution.size(), stream_view_);
 }
 
-static bool time_limit_reached(const std::chrono::high_resolution_clock::time_point& start_time,
-                               double seconds)
+static bool time_limit_reached(const timer_t& timer, double seconds)
```
rg20:
```suggestion
static bool time_limit_reached(const timer_t& timer)
```

cpp/src/linear_programming/solve.cu
```diff
@@ -318,10 +316,7 @@ run_dual_simplex(dual_simplex::user_problem_t<i_t, f_t>& user_problem,
   auto status =
     dual_simplex::solve_linear_program<i_t, f_t>(user_problem, dual_simplex_settings, solution);
 
-  auto end      = std::chrono::high_resolution_clock::now();
-  auto duration = std::chrono::duration_cast<std::chrono::milliseconds>(end - start_solver);
-
-  CUOPT_LOG_INFO("Dual simplex finished in %.2f seconds", duration.count() / 1000.0);
+  CUOPT_LOG_INFO("Dual simplex finished in %.2f seconds", timer.elapsed_time());
```
rg20:
```suggestion
  CUOPT_LOG_INFO("Dual simplex finished in %.2f seconds, total time %.2f", dual_simplex_time, timer.elapsed_time());
```
something like this?

hlinsen: Makes sense, I will do the same for PDLP to be consistent.

PR #382, opened 2025-09-09
Heuristic Improvements: balance between generation and improvement heuristics

### LINKED ISSUE #142: [FEA] Remove relative tolerance for MILP
Remove relative tolerance for MILP and lower absolute tolerance to 1e-6

### LINKED ISSUE #218: [BUG] Lower number of feasible count with B&B on benchmarks
The benchmarks show consistently lower number of feasible solutions compared to heuristics only runs. The difference is consistenly 2-3 solutions.

Expected behaviour: On average, same number of feasible solutions as the heuristic only run.

### LINKED ISSUE #374: [BUG] Solver doesn't stop immediately when optimal is found
The B&B finds a solution but the heuristics continues executing despite the stopping signal from B&B. This can happen on long running functions such as probing cache. 

Solution: propagate the stopping signal deep enough into the functions of heuristics.

akifcorduk:
## Description
This PR changes the heuristic structure by creating a natural balance between generation and improvement.
The FP/FJ loop now adds solution to the population and only if we have enough diverse solutions we exit the loop and execute the population improvement. The diversity is increased to `sqrt(n_integers)`. The recombiners are run between the current best and all other solutions in the current population, if stagnation is detected in FP/FJ loop and then the loop continues. The bounds prop rounding in the context of FP is also improved. When the dual simplex solution is set, the pdlp is warm started now with both primal and dual solutions.

The default tolerance is now 1e-6 absolute tolerance and 1e-12 relative tolerance.

This PR includes bug fixes on:
- Apperance of inf/nan on `z` vector dual simplex phase2.
- Invalid launch dimensions on FJ and hash kernels.
- Timer diff and function time limit issues when the solver is run with unlimited time limit.

Benchmark results in 10 mins run on H100:
- Main branch: 207 feasible solutions and average gap: '28.54', 3 unfinished/crashed
- This PR: 213 feasible and average gap: '23.11', 1 unfinished/crushed. (The PR didn't have any crash before merge with main branch)

closes #142  
closes #374 
closes #218


cpp/src/dual_simplex/phase2.cpp
```diff
@@ -1516,8 +1516,11 @@ void update_dual_variables(const sparse_vector_t<i_t, f_t>& delta_y_sparse,
   for (i_t k = 0; k < delta_z_nz; ++k) {
     const i_t j = delta_z_indices[k];
     z[j] += step_length * delta_z[j];
+    if (std::isnan(z[j]) || std::isinf(z[j])) { return -1; }
```
aliceb-nv: We could probably try to run our code with x86 floating-point exceptions turned on at some point; do you think being able to catch operations producing NaNs could be helpful? (at least host side)

akifcorduk: Yes that would be super helpful! That might reveal the place of the error. I will try a benchmark with that thanks!

akifcorduk: Since I am limited on bandwidth these days, I added this to to the issue: https://github.com/NVIDIA/cuopt/issues/397

cpp/include/cuopt/linear_programming/mip/solver_settings.hpp
```diff
@@ -72,8 +72,8 @@ class mip_solver_settings_t {
 
   struct tolerances_t {
     f_t presolve_absolute_tolerance = 1.0e-6;
-    f_t absolute_tolerance          = 1.0e-4;
-    f_t relative_tolerance          = 1.0e-6;
+    f_t absolute_tolerance          = 1.0e-6;
```
hlinsen, line 75: We need to update the docs: https://github.com/NVIDIA/cuopt/blob/branch-25.10/docs/cuopt/source/lp-milp-settings.rst

cpp/src/mip/presolve/load_balanced_bounds_presolve.cu
```diff
@@ -86,7 +86,7 @@ load_balanced_bounds_presolve_t<i_t, f_t>::load_balanced_bounds_presolve_t(
     warp_vars_id_offsets(0, problem_.handle_ptr->get_stream()),
     context(context_)
 {
-  setup(problem_);
+  // setup(problem_);
```
hlinsen: Should we remove `setup` completely?

benchmarks/linear_programming/cuopt/run_mip.cpp
```diff
@@ -210,6 +210,8 @@ int run_single_file(std::string file_path,
   settings.log_to_console                = log_to_console;
   settings.tolerances.relative_tolerance = 1e-12;
   settings.tolerances.absolute_tolerance = 1e-6;
+  settings.presolve                      = true;
+  settings.num_cpu_threads               = 8;
```
nguidotti: The `num_cpu_threads` is already set  to the command line option in some lines prior . It is better than to be a hardcoded value.

cpp/src/mip/presolve/probing_cache.cu
nguidotti, line 1: Is there a reason why we are limiting the number of threads to 8? It is not better to let be a user controller setting?

cpp/src/dual_simplex/phase2.cpp
chris-maes: I sometimes use initialize variables with `std::numeric_limits<double>::quiet_NaN()`, I'm guessing since it is quiet it would not rigger a floating-point exception?

cpp/src/dual_simplex/branch_and_bound.cpp
```diff
@@ -409,9 +409,12 @@ mip_status_t branch_and_bound_t<i_t, f_t>::solve(mip_solution_t<i_t, f_t>& solut
   assert(root_vstatus.size() == original_lp.num_cols);
   if (root_status == lp_status_t::INFEASIBLE) {
     settings.log.printf("MIP Infeasible\n");
-    if (settings.heuristic_preemption_callback != nullptr) {
-      settings.heuristic_preemption_callback();
-    }
+    // FIXME: rarely dual simplex detects infeasible whereas it is feasible.
```
chris-maes, line 412: So if heuristics has a feasible solution, but branch and bound says it is infeasible, you take the feasible solution?

cpp/src/dual_simplex/phase2.cpp
```diff
@@ -1515,8 +1515,11 @@ void update_dual_variables(const sparse_vector_t<i_t, f_t>& delta_y_sparse,
...
```
chris-maes: This is a performance critical loop. Hopefully the compiler is smart about this, but would you mind doing something like:
```
f_t zj = z[j] + step_length * delta_z[j];
z[j] = zj;
if (zj != zj || std::isinf(zj)) {
 return -1;
}
```

Also, can you file an issue so that we can track that this should be removed?

cpp/include/cuopt/logger.hpp
```diff
@@ -78,7 +78,7 @@ inline rapids_logger::logger& default_logger()
     logger_.set_pattern(default_pattern());
 #endif
     logger_.set_level(default_level());
-    logger_.flush_on(rapids_logger::level_enum::info);
+    logger_.flush_on(rapids_logger::level_enum::debug);
```
nguidotti, line 81: Maybe we should keep the flush in the `info` option unless there is a performance penalty.

cpp/CMakeLists.txt
```diff
@@ -422,6 +422,7 @@ endif()
 
 
 option(BUILD_MIP_BENCHMARKS "Build MIP benchmarks" OFF)
+set(BUILD_MIP_BENCHMARKS ON)
```
aliceb-nv: Debugging leftover I assume :)
On my setup I add `--cmake-args=\"-DBUILD_MIP_BENCHMARKS=ON\"` to my build.sh command which does the trick

cpp/src/dual_simplex/bound_flipping_ratio_test.cpp
```diff
@@ -93,6 +93,11 @@ i_t bound_flipping_ratio_test_t<i_t, f_t>::single_pass(i_t start,
   }
   step_length       = min_val;
   nonbasic_entering = candidate;
+  // this should be temporary, find root causes where the candidate is not filled
+  if (nonbasic_entering == -1) {
+    // -1,-2 and -3 are reserved for other things
+    return -4;
+  }
```
aliceb-nv, line 100: This issue is tracked, right?

cpp/src/mip/diversity/population.cu
```diff
@@ -38,23 +39,42 @@ constexpr double halving_skip_ratio          = 0.75;
...
+    population_hash_map(*problem_ptr),
     timer(0)
 {
   best_feasible_objective = std::numeric_limits<f_t>::max();
 }
 
+template <typename i_t>
+i_t get_max_var_threshold(i_t n_vars)
+{
+  return n_vars - sqrt(n_vars);
```
aliceb-nv, line 65: This early return is intended, right? If so you might want to remove the following dead code

aliceb-nv (approved):
LGTM engine side, approving; thanks a lot for the great work Akif!

cpp/src/dual_simplex/branch_and_bound.cpp
akifcorduk: Yes, that's what i wanted to do for now.

cpp/src/dual_simplex/phase2.cpp
akifcorduk: I think isfinite is probably better suited as a single instruction, let me know if it is better.

akifcorduk: Oh you are worried about double access, I am 90% sure compiler will optimize this but sure i can fix this to have single access.

cpp/include/cuopt/logger.hpp
akifcorduk: Debug is disabled in production, so there will be no performance penalty.

cpp/src/mip/presolve/probing_cache.cu
akifcorduk: Actually yes, but since we don't have a thread pool now, it is separate for now. Once we have a thread pool we should enable more dynamic strategy.

cpp/CMakeLists.txt
akifcorduk: Makes sense, thanks will do that. I think we should add a build.sh command for this.

cpp/src/dual_simplex/bound_flipping_ratio_test.cpp
akifcorduk: I will create an issue thanks!

cpp/src/mip/presolve/load_balanced_bounds_presolve.cu
akifcorduk: There is a bug in load balanced setup, @kaatish is fixing that. This should normally be enabled, I will try if it is resolved with Aatish's recommendation.

cpp/src/dual_simplex/bound_flipping_ratio_test.cpp
akifcorduk: https://github.com/NVIDIA/cuopt/issues/399

cpp/src/dual_simplex/phase2.cpp
akifcorduk: We might not need this at all because the bound_flipping_ratio test has solved the numerical issue.

akifcorduk: Indeed, we don't need this. Removed.

PR #385, opened 2025-09-10
Add sanitizer build option

akifcorduk:
This PR adds sanitizer build option.
To use sanitizer with cuda runtime, one must follow a few steps:
1. Find libasan.so with "gcc -print-file-name=libasan.so"
2. Run the binary with env var set: LD_PRELOAD=$PATH_TO_LIBASAN ASAN_OPTIONS='protect_shadow_gap=0:replace_intrin=0'
3. (Optional) To run with a debugger (gdb or cuda-gdb) use the additional ASAN option alloc_dealloc_mismatch=0

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,7 +73,6 @@ Development wheels are available as nightlies, please update `--extra-index-url`
 pip install --pre \
   --extra-index-url=https://pypi.nvidia.com \
   --extra-index-url=https://pypi.anaconda.org/rapidsai-wheels-nightly/simple/ \
-  nvidia-cuda-runtime-cu12=12.9.* \
   cuopt-server-cu12==25.10.* cuopt-sh-client==25.10.*
 ```
 
@@ -82,7 +81,6 @@ For CUDA 13.x:
 ```bash
 pip install \
   --extra-index-url=https://pypi.nvidia.com \
-  nvidia-cuda-runtime==13.0.* \
   cuopt-server-cu13==25.10.* cuopt-sh-client==25.10.*
 ```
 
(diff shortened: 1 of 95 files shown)


build.sh
```diff
@@ -44,6 +44,7 @@ HELP="$0 [<target> ...] [<flag> ...]
    -g               - build for debug
    -a               - Enable assertion (by default in debug mode)
    -b               - Build with benchmark settings
+   -s               - Build with sanitizer
```
hlinsen: More explicit naming?
```suggestion
   -fsanitize               - Build with sanitizer
```

cpp/CMakeLists.txt
```diff
@@ -84,6 +84,15 @@ if(CMAKE_COMPILER_IS_GNUCXX)
   list(APPEND CUOPT_CXX_FLAGS -Werror -Wno-error=deprecated-declarations)
 endif(CMAKE_COMPILER_IS_GNUCXX)
 
+# To use sanitizer with cuda runtime, one must follow a few steps:
+# 1. Find libasan.so with "gcc -print-file-name=libasan.so"
+# 2. Run the binary with env var set: LD_PRELOAD=$PATH_TO_LIBASAN ASAN_OPTIONS='protect_shadow_gap=0:replace_intrin=0'
+# 3. (Optional) To run with a debugger (gdb or cuda-gdb) use the additional ASAN option alloc_dealloc_mismatch=0
+if(BUILD_SANITIZER)
+  list(APPEND CUOPT_CXX_FLAGS -fsanitize=address,undefined -fno-omit-frame-pointer -g -Wno-error=maybe-uninitialized)
```
hlinsen, line 92: I am curious to know, what is the workaround you found for the rmm failures?

akifcorduk: It is in the comments in the PR description and above this line in `CMakeLists`

build.sh
aliceb-nv: I agree, might be better to have it as a long-form argument (like --fsanitize) since it's not supposed to be a commonly used option (and -s has different meanings depending on tools: 'optimize for size', 'strip symbols', etc)
Awesome addition by the way!! This is going to be really helpful

ramakrishnap-nv: @akifcorduk for viz

PR #388, opened 2025-09-10
cuPDLPx

Kh4ster:
This PR add a new PDLP Solver mode code Stable3, the new default.

This mode is designed to behave like cuPDLP+ from Haihao et al:

@article{lu2025cupdlpx,
  title={cuPDLPx: A Further Enhanced GPU-Based First-Order Solver for Linear Programming},
  author={Lu, Haihao and Peng, Zedong and Yang, Jinwen},
  journal={arXiv preprint arXiv:2507.14051},
  year={2025}
}

@article{lu2024restarted,
  title={Restarted Halpern PDHG for linear programming},
  author={Lu, Haihao and Yang, Jinwen},
  journal={arXiv preprint arXiv:2407.16144},
  year={2024}
}

Some features are not supported:
- warm start
- infeasiblity detection
- some combination of settings that might be used in MIP like per constraint residual


cpp/include/cuopt/linear_programming/constants.h
```diff
@@ -99,8 +99,9 @@
 
 #define CUOPT_PDLP_SOLVER_MODE_STABLE1     0
 #define CUOPT_PDLP_SOLVER_MODE_STABLE2     1
-#define CUOPT_PDLP_SOLVER_MODE_METHODICAL1 2
```
tmckayus, line 102: would it be better here to keep the original numbering and just add Stable3 as the new value 4?  We may have integrations with driver programs / apis etc that expect the values to be stable, or map strings to enum values or something.
so have stable1=0, stable2=1, methodical1=2, fast1=3, stable3=4

aliceb-nv: Agreed, this makes sense to me. I will make the change on Nicolas' behalf

tmckayus: thanks!

benchmarks/linear_programming/cuopt/run_pdlp.cu
```diff
@@ -90,6 +91,8 @@ static cuopt::linear_programming::pdlp_solver_mode_t string_to_pdlp_solver_mode(
     return cuopt::linear_programming::pdlp_solver_mode_t::Methodical1;
   else if (mode == "Fast1")
     return cuopt::linear_programming::pdlp_solver_mode_t::Fast1;
+  else if (mode == "Stable3")
+    return cuopt::linear_programming::pdlp_solver_mode_t::Stable3;
   return cuopt::linear_programming::pdlp_solver_mode_t::Stable2;
```
tmckayus: @aliceb-nv does this last return need to change? If I am reading this correctly, Stable2 is the value if no other is specified, but should that be changed to Stable3?  Or is this effectively just an "if ... else if ... else" that covers all cases?

python/cuopt_server/cuopt_server/utils/linear_programming/data_definition.py
```diff
@@ -498,14 +498,16 @@ class SolverConfig(StrictModel):
         "<br>"
```
tmckayus, line 498: @aliceb-nv can we change the default here to 4 to map to the new Stable3? (line 496 in current file)

benchmarks/linear_programming/cuopt/run_pdlp.cu
aliceb-nv: Good catch, will do :)

aliceb-nv (approved):

tmckayus (approved):

PR #391, opened 2025-09-15
Multithreaded Branch & Bound using Best First Search With Diving

nguidotti:
This (large) PR introduces the following changes:
- Explore the tree in parallel using the best first search with diving and backtracking, i.e., each thread pick the node from a global heap and then perform a dive starting on that node. Each node maintains a local stack that contains the pending nodes until a certain depth, allowing the thread to backtrack to a early point when the dive ends due to the integer (in)feasibility.
- Refactor the branch-and-bound code such that some steps are now moved to dedicated methods. This is consequence of the multithreaded framework, in addition to trying to improve readability and reduce code duplication.
- Transform most operation with mutexes into atomic ones. 
- Modified strong branching to use `openmp` threads instead of `std::thread` to use dynamic scheduling and improve CPU usage.

This PR is the continued work from #305.

### DIFF
--- a/cpp/src/dual_simplex/branch_and_bound.cpp
+++ b/cpp/src/dual_simplex/branch_and_bound.cpp
@@ -186,81 +186,6 @@ dual::status_t convert_lp_status_to_dual_status(lp_status_t status)
   }
 }
 
-}  // namespace
-
-template <typename i_t, typename f_t>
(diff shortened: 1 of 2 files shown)


nguidotti:
/ok to test [78ab1bf](https://github.com/NVIDIA/cuopt/pull/391/commits/78ab1bfe1ba1978d0baa3a178ca8de856d2946e7)

PR #392, opened 2025-09-15
Add optional timestamps enabled by env CUOPT_EXTRA_TIMESTAMPS

tmckayus:
This change adds optional internal timestamps which can be enabled with env var CUOPT_EXTRA_TIMESTAMPS=True.

The timestamps are to allow benchmarking scripts to key off of the events in order to calculate problem setup and teardown time.

### DIFF
--- a/cpp/src/linear_programming/cuopt_c.cpp
+++ b/cpp/src/linear_programming/cuopt_c.cpp
@@ -27,7 +27,9 @@
 #include <cuopt/version_config.hpp>
 
 #include <memory>
+#include <cstdlib>
 #include <string>
+
 
 using namespace cuopt::mps_parser;
 using namespace cuopt::linear_programming;
@@ -74,6 +76,29 @@ cuopt_int_t cuOptGetVersion(cuopt_int_t* version_major,
   return CUOPT_SUCCESS;
 }
 
+bool extraTimestamps() {
+    const char* envValue = std::getenv("CUOPT_EXTRA_TIMESTAMPS");
+    if (envValue == nullptr) {
+        return false;
+    }
+    
+    std::string value(envValue);
+    return (value == "True" || value == "true" || value == "1");
+}
+
+
+double getCurrentTimestamp() {
+    auto now = std::chrono::system_clock::now();
+    return std::chrono::duration<double>(now.time_since_epoch()).count();
+}
+
+void printTimestamp(const std::string& label) {
+    if (extraTimestamps()) {    
+        std::cout << std::fixed << std::setprecision(6);
+        std::cout << std::endl << label << ": " << getCurrentTimestamp() << std::endl;
+    }
+}
+
 cuopt_int_t cuOptReadProblem(const char* filename, cuOptOptimizationProblem* problem_ptr)
 {
   problem_and_stream_view_t* problem_and_stream = new problem_and_stream_view_t();
@@ -115,6 +140,9 @@ cuopt_int_t cuOptCreateProblem(cuopt_int_t num_constraints,
                                const char* variable_types,
                                cuOptOptimizationProblem* problem_ptr)
 {
+
+  printTimestamp("CUOPT_CREATE_PROBLEM");
+  
   if (problem_ptr == nullptr || objective_coefficients == nullptr ||
       constraint_matrix_row_offsets == nullptr || constraint_matrix_column_indices == nullptr ||
       constraint_matrix_coefficent_values == nullptr || constraint_sense == nullptr ||
@@ -170,6 +198,9 @@ cuopt_int_t cuOptCreateRangedProblem(cuopt_int_t num_constraints,
                                      const char* variable_types,
                                      cuOptOptimizationProblem* problem_ptr)
 {
+
+  printTimestamp("CUOPT_CREATE_PROBLEM");
+  
   if (problem_ptr == nullptr || objective_coefficients == nullptr ||
       constraint_matrix_row_offsets == nullptr || constraint_matrix_column_indices == nullptr ||
       constraint_matrix_coefficent_values == nullptr || constraint_lower_bounds == nullptr ||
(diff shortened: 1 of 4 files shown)


cpp/src/linear_programming/cuopt_c.cpp
```diff
@@ -74,6 +76,29 @@ cuopt_int_t cuOptGetVersion(cuopt_int_t* version_major,
   return CUOPT_SUCCESS;
 }
 
+bool extraTimestamps() {
+    const char* envValue = std::getenv("CUOPT_EXTRA_TIMESTAMPS");
```
ramakrishnap-nv: If we can get it once and store and reuse this, we don't have to request env name every-time, but this also depends on how many times this function is going to be called. 

And would it be helpful if we move it out and make it common for routing and linear programming ?

python/cuopt/cuopt/linear_programming/solver/solver.py
```diff
@@ -78,6 +81,11 @@ def Solve(data_model, solver_settings=None):
     >>> # Print the value of one specific variable
     >>> print(solution.get_vars()["var_name"])
     """
+
+    emit_stamps = os.environ.get("CUOPT_EXTRA_TIMESTAMPS", False) in (True, "True", "true")
+    if emit_stamps:
+        print(f"CUOPT_SOLVE_START: {time.time()}")
```
ramakrishnap-nv, line 91:
```suggestion
        print(f"PY_CUOPT_SOLVE_START: {time.time()}")
```

python/cuopt/cuopt/linear_programming/solver/solver.py
```diff
@@ -95,11 +103,14 @@ def is_mip(var_types):
...
             return any(vt == "I" or vt == b"I" for vt in var_types)
 
-    return solver_wrapper.Solve(
+    s = solver_wrapper.Solve(
         data_model,
         solver_settings,
         mip=is_mip(data_model.get_variable_types()),
     )
+    if emit_stamps:
+        print(f"CUOPT_SOLVE_RETURN: {time.time()}")
```
ramakrishnap-nv, line 116:
```suggestion
        print(f"PY_CUOPT_SOLVE_RETURN: {time.time()}")
```

ramakrishnap-nv: To distinguish between python and cpp time stamp.

cpp/src/linear_programming/cuopt_c.cpp
```diff
@@ -115,6 +140,9 @@ cuopt_int_t cuOptCreateProblem(cuopt_int_t num_constraints,
                                const char* variable_types,
                                cuOptOptimizationProblem* problem_ptr)
 {
+
+  printTimestamp("CUOPT_CREATE_PROBLEM");
```
ramakrishnap-nv:
```suggestion
  printTimestamp("CPP_CUOPT_CREATE_PROBLEM");
```

cpp/src/linear_programming/cuopt_c.cpp
```diff
@@ -170,6 +198,9 @@ cuopt_int_t cuOptCreateRangedProblem(cuopt_int_t num_constraints,
                                      const char* variable_types,
                                      cuOptOptimizationProblem* problem_ptr)
 {
+
+  printTimestamp("CUOPT_CREATE_PROBLEM");
```
ramakrishnap-nv:
```suggestion
  printTimestamp("CPP_CUOPT_CREATE_PROBLEM");
```

cpp/src/linear_programming/cuopt_c.cpp
```diff
@@ -596,6 +627,8 @@ cuopt_int_t cuOptSolve(cuOptOptimizationProblem problem,
                        cuOptSolverSettings settings,
                        cuOptSolution* solution_ptr)
 {
+  printTimestamp("CUOPT_SOLVE_START");
```
ramakrishnap-nv:
```suggestion
  printTimestamp("CPP_CUOPT_SOLVE_START");
```

cpp/src/linear_programming/cuopt_c.cpp
```diff
@@ -614,6 +647,9 @@ cuopt_int_t cuOptSolve(cuOptOptimizationProblem problem,
     solution_and_stream_view->mip_solution_ptr = new mip_solution_t<cuopt_int_t, cuopt_float_t>(
       solve_mip<cuopt_int_t, cuopt_float_t>(*op_problem, mip_settings));
     *solution_ptr = static_cast<cuOptSolution>(solution_and_stream_view);
+
+    printTimestamp("CUOPT_SOLVE_RETURN");
```
ramakrishnap-nv:
```suggestion
    printTimestamp("CPP_CUOPT_SOLVE_RETURN");
```

cpp/src/linear_programming/cuopt_c.cpp
```diff
@@ -629,6 +665,9 @@ cuopt_int_t cuOptSolve(cuOptOptimizationProblem problem,
       new optimization_problem_solution_t<cuopt_int_t, cuopt_float_t>(
         solve_lp<cuopt_int_t, cuopt_float_t>(*op_problem, pdlp_settings));
     *solution_ptr = static_cast<cuOptSolution>(solution_and_stream_view);
+
+    printTimestamp("CUOPT_SOLVE_RETURN");
```
ramakrishnap-nv:
```suggestion
    printTimestamp("CPP_CUOPT_SOLVE_RETURN");
```

cpp/src/linear_programming/cuopt_c.cpp
tmckayus: yes this may be a naive implementation. I asked AI to generate it :)

This would be called a handful of times per solve from the C api

python/cuopt/cuopt/linear_programming/solver/solver.py
tmckayus: hmm that may mess up the benchmark script. There are two different paths to the solver, via Python or via the C api. You won't get both for the same problem, you will get either the Python strings emitted or the C strings emitted depending on the API, but never both.  I would be inclined to leave them the same unless you feel strongly about it.

tmckayus: Same for all the other cases below. Two mutually exclusive paths, same timestamp. The benchmark script just looks for the single string. It could be smart enough to look for pairs/groups though.

KyleFromNVIDIA (approved):

rgsl888prabhu (approved):

PR #394, opened 2025-09-15
Add jump testing to nightly testing

### LINKED ISSUE #380: [FEA] Add jump testing to nightly for exhaustive testing
Jump has several tests which would be a good set of unite tests to run on cuOpt to catch any corner cases

This will be enabled only for nightly so it will not bother PRs.

rgsl888prabhu:
## Description
This PR adds jump testing to take advantage of several unit tests to bolster test suite

## Issue
closes #380 

## Checklist

- Testing
   - [x] Added tests
   - [ ] NA
- Documentation
   - [ ] Added new documentation
   - [x] NA


rgsl888prabhu:
@gforsyth May I get your review on this PR ?

ci/thirdparty-testing/run_jump_tests.sh
```diff
@@ -0,0 +1,69 @@
...
+    rapids-logger "Installing Julia using official installer..."
+    # Pass 'yes' to the installer to accept any prompts automatically
+    curl -fsSL https://install.julialang.org | sh -s -- --yes
+    # Add Julia to PATH for current session
+    export PATH="$HOME/.juliaup/bin:$PATH"
+    # Also add to .bashrc for future sessions
+    rapids-logger 'export PATH="$HOME/.juliaup/bin:$PATH"' >> ~/.bashrc
+else
+    rapids-logger "Julia is already installed."
+fi
```
gforsyth, line 32: I'm not sure we should be `curl` installing Julia in CI like this.

Maybe instead a separate actions workflow that uses https://github.com/julia-actions/setup-julia

ramakrishnap-nv: The curl command was standard way which was shown in the documentation, but I am open to switch to actions workflow, is there an example on how I can use this along with existing setup or should I add it with custom job ?

gforsyth: Hey @rgsl888prabhu -- I don't want to make a bunch of extra work for you.  My initial reaction to `curl ... | sh` is a little wary, but that _does_ seem like the recommended install path.

Bradley pointed out that there _are_ `conda` packages for installing Julia -- can we try that quickly to see if it works?  And if it doesn't, we can stick wtih the `curl`-based install.

https://anaconda.org/conda-forge/julia

bdice: This seems like that’s really what they recommend. They don’t accept bug reports except from these official packages.

There are no Ubuntu apt packages: https://packages.ubuntu.com/search?keywords=Julia&searchon=names&suite=plucky&section=all

There are conda packages: https://anaconda.org/conda-forge/julia

Using a reusable GitHub Action here would make it harder to reproduce this workflow locally, so I am in favor of what is implemented here.

ramakrishnap-nv: @gforsyth Since I am building and running this in a wheel based set-up, conda based install might not be feasible.

gforsyth (approved):
Sounds good

PR #400, opened 2025-09-17
[DONOTMERGE]  Allow different presolve behavior for different solver methods in LP

rg20:
## Description

This PR enhances the presolve options for LP problems. Notable changes are:

1.  Provides new option `dual_postsolve`  and is enabled by default.  With this change, the presolve can be enabled by default without breaking the interfaces
2.  Presolve option is changed from boolean to integer with options for `default`, `on`, `off`
4. default has different meaning depending on solver method: (i) for method PDLP, `default == off` (ii) for method dual simplex and barrier,` default == on` (iii) for concurrent, PDLP is run without the presolve while barrier and dual simplex are run with presolve 
5. In concurrent mode, default presolve means that PDLP is launched without presolve while barrier and dual simplex wait for presolve to complete. The wait is implemented similar to concurrent_halt mechanism

Need to update the python API, documentation accordingly 

## Checklist

- Testing
   - [ ] Added tests
   - [ ] NA
- Documentation
   - [ ] Added new documentation
   - [ ] NA


rg20:
Closing this as we are switching to simpler paradigm where presolve is just on or off for all the solvers.

PR #404, opened 2025-09-19
add http timeout values for general, send, and receive to client

tmckayus:
Some large datasets, or even smaller datasets on a slow network, may take more than 30s to transfer to the server. Add timeout parameters to the cuopt service client to control the timeouts, defaulted to 30s.

Timeout parameters added: http_general_timeout (default 30), result_receive_timeout (default None), and data_send_timeout (default None).  The latter two inherit the value of http_general_timeout if they are None.

All 3 are settable in the client class, however only the general timeout is exposed in the CLI (which should be enough control).

### DIFF
--- a/python/cuopt_self_hosted/cuopt_sh_client/cuopt_self_host_client.py
+++ b/python/cuopt_self_hosted/cuopt_sh_client/cuopt_self_host_client.py
@@ -291,6 +291,17 @@ class CuOptServiceSelfHostClient:
             wildcard is used, the result mime_type will be set to the
             content_type mime_type of the original request.
             If not provided, result_type defaults to mime_type.MSGPACK
+    http_general_timeout: int
+            The time in seconds that http will wait before timing out
+            on a general request such as a job status check. Default is 30s.
+    data_send_timeout: int
+            The time in seconds that http will wait before timing out
+            on a problem submission to the server. If set to None,
+            the http_general_timeout value will be used. Default is None.
+    result_receive_timeout: int
+            The time in seconds that http will wait before timing out
+            on receiving a result from the server. If set to None,
+            the http_general_timeout value will be used. Default is None.
     """
 
     # Initialize the CuOptServiceSelfHostClient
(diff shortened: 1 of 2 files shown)


python/cuopt_self_hosted/cuopt_sh_client/cuopt_self_host_client.py
```diff
@@ -306,13 +317,30 @@ def __init__(
         polling_timeout=600,
         timeout_exception=True,
         result_type=mime_type.MSGPACK,
+        http_general_timeout=30,
```
ramakrishnap-nv, line 324: I think we should enforce the type that we will accept in the parameter value, but we can tackle it in a follow-up PR.

python/cuopt_self_hosted/cuopt_sh_client/cuopt_self_host_client.py
```diff
@@ -306,13 +317,30 @@ def __init__(
...
         self.timeout_exception = timeout_exception
         self.ip = ip
         self.port = port
         self.only_validate = only_validate
         self.accept_type = result_type
 
+        if not isinstance(http_general_timeout, (int, float)):
+            http_general_timeout = 30
+        self.http_general_timeout = http_general_timeout
+        self.data_send_timeout = (
```
ramakrishnap-nv, line 338: Is it possible to set None or -1 to make it indefinite rather than setting arbitrary value ?

python/cuopt_self_hosted/cuopt_sh_client/cuopt_sh.py
```diff
@@ -587,6 +592,15 @@ def main():
         help="Upload a solution to be cached on the server. The reqId "
         "returned may be used as an initial solution for VRP.",
     )
+    parser.add_argument(
+        "-ht",
+        "--http-timeout",
+        type=int,
+        default=30,
```
ramakrishnap-nv, line 599: Think it is better to set None, since the possibility of hitting this issue might be higher

python/cuopt_self_hosted/cuopt_sh_client/cuopt_self_host_client.py
tmckayus: yes we could add type hints for everything. Although I'm not sure those are binding.

python/cuopt_self_hosted/cuopt_sh_client/cuopt_self_host_client.py
tmckayus: done

python/cuopt_self_hosted/cuopt_sh_client/cuopt_sh.py
tmckayus: Leaving this at 30 for backward compat so the behavior doesn't change (the 30 was default in the library before this option existed)

Issue #431, opened 2025-10-02
[BUG] GF2 presolve doesn't work on ARM machines


akifcorduk:
GF2 presolve which is used to reduce enlight_hard does not work on ARM machines.

PR #437, opened 2025-10-03
GF2 presolve infeasibility fix + CI occasional failure fix

aliceb-nv:
This PR fixes an UB situation in GF2 caused by a cast of a negative floating-point value to an unsigned integer, which only manifested on ARM.
Also included is a fix to improve CI success rates on a PDLP python test.
closes https://github.com/NVIDIA/cuopt/issues/431


rg20 (approved):
Thanks for the quick fix!

Iroy30 (approved):

PR #448, opened 2025-10-06
Enable green context in Barrier

hlinsen:
This PR enables stream concurrency by sharing SM resources between PDLP and Barrier. It fixes compile errors that occured while merging https://github.com/NVIDIA/cuopt/pull/272

### DIFF
--- a/cpp/CMakeLists.txt
+++ b/cpp/CMakeLists.txt
@@ -283,7 +283,6 @@ set(CUOPT_PRIVATE_CUDA_LIBS
   CUDA::curand
   CUDA::cusolver
   TBB::tbb
-  CUDA::cuda_driver
   OpenMP::OpenMP_CXX)
 
 list(PREPEND CUOPT_PRIVATE_CUDA_LIBS CUDA::cublasLt)
--- a/cpp/src/dual_simplex/sparse_cholesky.cuh
+++ b/cpp/src/dual_simplex/sparse_cholesky.cuh
@@ -29,6 +29,8 @@
 #include "cuda.h"
 #include "cudss.h"
 
+#define SPLIT_SM_FOR_BARRIER
+
 namespace cuopt::linear_programming::dual_simplex {
 
 template <typename i_t, typename f_t>
--- a/cpp/src/dual_simplex/sparse_cholesky.cuh
+++ b/cpp/src/dual_simplex/sparse_cholesky.cuh
@@ -23,13 +23,11 @@
 #include "dual_simplex/tic_toc.hpp"
 
 #include <cuda_runtime.h>
+#include <utilities/driver_helpers.cuh>
 
 #include <raft/common/nvtx.hpp>
 
-#include "cuda.h"
 #include "cudss.h"
-
-#define SPLIT_SM_FOR_BARRIER
 
 namespace cuopt::linear_programming::dual_simplex {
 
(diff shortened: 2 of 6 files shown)


cpp/src/dual_simplex/sparse_cholesky.cuh
chris-maes: For each of the cu* functions we use, I think we are supposed to call cudaDriverGetEntryPoint or cudaGetDriverEntryPointByVersion to get a function pointer and then make the call using a function pointer. 
CU_CHECK(cuDriverGetVersion(&driver_version));
CU_CHECK(cuDeviceGetDevResource(...));
CU_CHECK(cuDevSmResourceSplitByCount(...));
CU_CHECK(cuDevResourceGenerateDesc(...));
CU_CHECK(cuGreenCtxCreate(...));
CU_CHECK(cuGreenCtxStreamCreate(...));

I think this is because we aren't supposed to link directly against the driver. But I don't understand the details. 

We can probably merge as is, but we probably need a follow up PR to use the `EntryPoint` functions

hlinsen: I am looking into doing it this way as well

cpp/src/dual_simplex/sparse_cholesky.cuh
```diff
@@ -157,37 +157,46 @@ class sparse_cholesky_cudss_t : public sparse_cholesky_base_t<i_t, f_t> {
...
+      auto cuDeviceGetDevResource_func = cuopt::get_driver_entry_point("cuDeviceGetDevResource");
+      CU_CHECK(reinterpret_cast<decltype(::cuDeviceGetDevResource)*>(cuDeviceGetDevResource_func)(
+                 handle_ptr_->get_device(), &initial_device_GPU_resources, CU_DEV_RESOURCE_TYPE_SM),
+               reinterpret_cast<decltype(::cuGetErrorString)*>(cuGetErrorString_func));
+
 #ifdef DEBUG
-      std::cout << "Initial GPU resources retrieved via cuDeviceGetDevResource() have type "
-                << initial_device_GPU_resources.type << " and SM count "
-                << initial_device_GPU_resources.sm.smCount << std::endl;
+      printf(
```
chris-maes: Do you want these to appear in the log if DEBUG is enabled? Or just go to the console. As you have it now it goes to the console. If you want it to go the the console and the log use `settings_.log.printf`

chris-maes (approved):
LGTM. But perhaps someone with more experience with CUDA driver calls should check?

cpp/src/dual_simplex/sparse_cholesky.cuh
hlinsen: I'll use the logger to be consistent

cpp/src/utilities/driver_helpers.cuh
```diff
@@ -0,0 +1,35 @@
...
+ * limitations under the License.
+ */
+
+#pragma once
+
+#include "cuda.h"
+
+namespace cuopt {
+
+inline auto get_driver_entry_point(const char* name)
```
bdice, line 26: Do you have a `detail` namespace? I know this is shielded from public usage by being defined in `src/` rather than a public header but it might be clearer if this was marked as `detail`.

cpp/CMakeLists.txt
```diff
@@ -283,7 +283,6 @@ set(CUOPT_PRIVATE_CUDA_LIBS
   CUDA::curand
   CUDA::cusolver
   TBB::tbb
-  CUDA::cuda_driver
```
bdice, line 286: I guess this does a dlopen now, rather than dynamically linking, which is why this is removed?

hlinsen: I haven't checked the trace but I think so. I wanted to avoid linking to the driver with these changes.

rgsl888prabhu (approved):

PR #450, opened 2025-10-06
Fix compile and log

akifcorduk:
This PR fixes a log level and a compilation error that happens on non-CI envs.


rgsl888prabhu:
@akifcorduk not sure why it is failing or even where it is failing, but this test run has been failing several times, I think it has been run atleast 4 times. Please take a look in earlier run if you can find any clue https://github.com/NVIDIA/cuopt/actions/runs/18288323659/job/52087939557?pr=450

On the top you can see attempts, where you can switch to previous  runs.

nguidotti:
The latest run fails due to a SIGSEGV in this test

```
 [ RUN      ] level0_ges/double_test_vrp.GES_VRP/0
double free or corruption (out)
timeout: the monitored command dumped core
ci/test_cpp.sh: line 61:  1634 Aborted                 timeout 20m "${gt}" --gtest_output=xml:"{RAPIDS_TESTS_DIR}"

```

rg20 (approved):

PR #451, opened 2025-10-06
Fixed delayed termination after timeout

nguidotti:
This PR fixes the delayed termination of the branch-and-bound algorithm after reaching the time limit.


chris-maes (approved):
LGTM

PR #453, opened 2025-10-07
Explicitely create new stream for Barrier concurrent

hlinsen:
In C++ when spawning a new thread the cuda driver by default creates a legacy default stream for each thread with unique ids.
When calling concurrent mode from Python, PDLP and Barrier were running on the same stream.
Now we explicitly create a new handle for Barrier with a new stream.

### DIFF
--- a/cpp/src/dual_simplex/branch_and_bound.cpp
+++ b/cpp/src/dual_simplex/branch_and_bound.cpp
@@ -584,6 +584,7 @@ node_status_t branch_and_bound_t<i_t, f_t>::solve_node(search_tree_t<i_t, f_t>& 
   lp_settings.set_log(false);
   lp_settings.cut_off    = upper_bound + settings_.dual_tol;
   lp_settings.inside_mip = 2;
+  lp_settings.time_limit = settings_.time_limit - toc(stats_.start_time);
 
   // in B&B we only have equality constraints, leave it empty for default
   std::vector<char> row_sense;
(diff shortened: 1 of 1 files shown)


cpp/src/linear_programming/solve.cu
```diff
@@ -670,8 +670,16 @@ optimization_problem_solution_t<i_t, f_t> run_concurrent(
   // Initialize the dual simplex structures before we run PDLP.
   // Otherwise, CUDA API calls to the problem stream may occur in both threads and throw graph
   // capture off
+  auto barrier_handle = raft::handle_t(*op_problem.get_handle_ptr());
+  detail::problem_t<i_t, f_t> d_barrier_problem(problem);
```
rg20, line 674:
```suggestion
  auto barrier_straem = rmm::cuda_stream_per_thread;
  raft::resource::set_cuda_stream(barrier_handle, barrier_stream);
  detail::problem_t<i_t, f_t> d_barrier_problem(problem, barrier_handle);
```

hlinsen: I considered that overload but I don't think it is needed. According to [rmm doc](https://docs.rapids.ai/api/librmm/25.10/classrmm_1_1device__uvector#addeb159f22d1c1cd7203b040beda393e) and current [problem_t](https://github.com/NVIDIA/cuopt/blob/branch-25.10/cpp/src/mip/problem/problem.cu#L225) passing a new stream will create a deep copy on that stream.

rgsl888prabhu (approved):

PR #449, opened 2025-10-06
Add  docs and example tests to barrier

rgsl888prabhu:
## Description
This PR is a follow-up for PR #272 to add documentation and additional tests

## Checklist

- [x] I am familiar with the [Contributing Guidelines](https://github.com/NVIDIA/cuopt/blob/HEAD/CONTRIBUTING.md).
- Testing
   - [ ] New or existing tests cover these changes
   - [x] Added tests
   - [ ] Created an issue to follow-up
   - [ ] NA
- Documentation
   - [ ] The documentation is up to date with these changes
   - [x] Added new documentation
   - [ ] NA

PR #468, opened 2025-10-08
Add setting to change the dual initial point used by barrier

chris-maes:
Allow users to select the initial point used for barrier.

### DIFF
--- a/cpp/include/cuopt/linear_programming/constants.h
+++ b/cpp/include/cuopt/linear_programming/constants.h
@@ -54,6 +54,7 @@
 #define CUOPT_AUGMENTED                   "augmented"
 #define CUOPT_DUALIZE                     "dualize"
 #define CUOPT_ORDERING                    "ordering"
+#define CUOPT_BARRIER_DUAL_INITIAL_POINT  "barrier_dual_initial_point"
 #define CUOPT_ELIMINATE_DENSE_COLUMNS     "eliminate_dense_columns"
 #define CUOPT_CUDSS_DETERMINISTIC         "cudss_deterministic"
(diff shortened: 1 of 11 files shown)


cpp/include/cuopt/linear_programming/constants.h
```diff
@@ -54,6 +54,7 @@
 #define CUOPT_AUGMENTED                   "augmented"
 #define CUOPT_DUALIZE                     "dualize"
 #define CUOPT_ORDERING                    "ordering"
+#define CUOPT_BARRIER_DUAL_INITIAL_POINT  "barrier_dual_initial_point"
```
ramakrishnap-nv, line 57: Lets also add documentation for this

python/cuopt_server/cuopt_server/utils/linear_programming/solver.py
```diff
@@ -451,6 +452,11 @@ def is_mip(var_types):
             solver_settings.set_parameter(
                 CUOPT_ORDERING, solver_config.ordering
             )
+        if solver_config.barrier_dual_initial_point != "":
```
ramakrishnap-nv:
```suggestion
        if solver_config.barrier_dual_initial_point is not None:
```

cpp/include/cuopt/linear_programming/constants.h
ramakrishnap-nv: Else let me add these in my PR #449

chris-maes: Let's add in your PR, if you don't mind.

PR #470, opened 2025-10-08
Fix an UMA on ns1952667 due to ignoring rank in factorize_basis when no singletons present

chris-maes:
When a basis had no singletons we were incorrectly setting the rank to m, even if the right-looking LU reported that the basis was rank deficient. This caused the column pointers and row indices in the L factor to be invalid. (Since the L factor itself was invalid).

We now correctly handle rank deficiency in the basis when there are no singletons.

### DIFF
--- a/cpp/CMakeLists.txt
+++ b/cpp/CMakeLists.txt
@@ -190,9 +190,9 @@ FetchContent_Declare(
   # does not have some of the presolvers and settings that we need
   # Mainly, probing and clique merging.
   # This is the reason we are using the development branch
-  # commit from Sep 26, 2025. Once these changes are merged into the main branch,
+  # commit from Oct 8, 2025. Once these changes are merged into the main branch,
   #we can switch to the main branch.
-  GIT_TAG "34a40781fa14f8870cb6368cffb6c0eda2f47511"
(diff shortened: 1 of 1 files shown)


cpp/src/dual_simplex/sparse_matrix.cpp
```diff
@@ -504,18 +504,24 @@ i_t scatter(const csc_matrix_t<i_t, f_t>& A,
...
     const i_t col_start = this->col_start[j];
     const i_t col_end   = this->col_start[j + 1];
+    if (col_start > col_end || col_start > this->col_start[this->n]) {
+      printf("CSC error: column pointers. col start %d col end %d nz %d\n", col_start, col_end, this->col_start[this->n]);
+      return -1;
+    }
     for (i_t p = col_start; p < col_end; ++p) {
       const i_t i = this->i[p];
-      if (row_marker[i] == j) { printf("CSC error: repeated row index %d in column %d\n", i, j); }
+      if (i < 0 || i >= this->m) { printf("CSC error: row index %d not in range [0, %d]\n", i, this->m-1); }
```
hlinsen: We should probably use cuopt_assert for these checks

PR #473, opened 2025-10-09
Propagate the bounds from the parent to the child nodes

nguidotti:
With this PR, the solver propagates the bounds from the parent to the child nodes. If the solver reaches the end of a branch (i.e., when the node becomes infeasible or an integer feasible solution is found), the bounds are recomputed from scratch. This allows the solver to reuse the tightened bounds from the parent and re-apply the bounds strengthening on the branched variable.

Average (Primal) Gap over the MIPLIB2017 dataset: 
`main` (10f116b): `14.472358` with `224` feasible solutions
This PR: `14.058718` with `226` feasible solutions

On average, the `main` branch explored `1012412` nodes. This PR increases to `1607821`  nodes (i.e., `58%` more nodes) for the same time frame.

## Checklist

- [ ] I am familiar with the [Contributing Guidelines](https://github.com/NVIDIA/cuopt/blob/HEAD/CONTRIBUTING.md).
- Testing
   - [x] New or existing tests cover these changes
   - [ ] Added tests
   - [ ] Created an issue to follow-up
   - [ ] NA
- Documentation
   - [x] The documentation is up to date with these changes
   - [ ] Added new documentation
   - [ ] NA

## Release Notes

* **New Features**
  * Integrated bounds strengthening into branch-and-bound presolver for tighter variable bound management.
  * Enhanced logging configuration with customizable file output modes.

* **Improvements**
  * Improved solver status tracking and internal thread-type handling for better diagnostics.
  * Optimized basis update threshold tolerances for numerical stability.

* **Refactor**
  * Restructured internal solver APIs for cleaner architecture and improved maintainability.
  * Migrated constraint bound management to unified presolver framework.

PR #474, opened 2025-10-09
[BUG] Fixed starting variable bounds for diving

nguidotti:
This PR also fixes the starting bounds for the diving subtrees. As these nodes are detached from the main B&B tree, the bounds from the parent were lost. Now, the bounds are also kept when inserting the nodes in the diving queue.

## Checklist

- Testing
   - [ ] Added tests
   - [ ] NA
- Documentation
   - [ ] Added new documentation
   - [ ] NA

### DIFF
--- a/benchmarks/linear_programming/cuopt/run_mip.cpp
+++ b/benchmarks/linear_programming/cuopt/run_mip.cpp
@@ -80,10 +80,11 @@ void merge_result_files(const std::string& out_dir,
 void write_to_output_file(const std::string& out_dir,
                           const std::string& base_filename,
                           int gpu_id,
+                          int n_gpus,
                           int batch_id,
                           const std::string& data)
 {
-  int output_id        = batch_id * 8 + gpu_id;
+  int output_id        = batch_id * n_gpus + gpu_id;
   std::string filename = out_dir + "/result_" + std::to_string(output_id) + ".txt";
   std::ofstream outfile(filename, std::ios_base::app);
   if (outfile.is_open()) {
@@ -149,6 +150,7 @@ std::vector<std::vector<double>> read_solution_from_dir(const std::string file_p
 int run_single_file(std::string file_path,
                     int device,
                     int batch_id,
+                    int n_gpus,
                     std::string out_dir,
                     std::optional<std::string> initial_solution_dir,
                     bool heuristics_only,
@@ -243,7 +245,7 @@ int run_single_file(std::string file_path,
      << obj_val << "," << benchmark_info.objective_of_initial_population << ","
      << benchmark_info.last_improvement_of_best_feasible << ","
      << benchmark_info.last_improvement_after_recombination << "\n";
-  write_to_output_file(out_dir, base_filename, device, batch_id, ss.str());
+  write_to_output_file(out_dir, base_filename, device, n_gpus, batch_id, ss.str());
   CUOPT_LOG_INFO("Results written to the file %s", base_filename.c_str());
   return sol_found;
 }
@@ -251,6 +253,7 @@ int run_single_file(std::string file_path,
 void run_single_file_mp(std::string file_path,
                         int device,
                         int batch_id,
+                        int n_gpus,
                         std::string out_dir,
                         std::optional<std::string> input_file_dir,
                         bool heuristics_only,
@@ -265,6 +268,7 @@ void run_single_file_mp(std::string file_path,
   int sol_found = run_single_file(file_path,
                                   device,
                                   batch_id,
+                                  n_gpus,
                                   out_dir,
                                   input_file_dir,
                                   heuristics_only,
@@ -462,6 +466,7 @@ int main(int argc, char* argv[])
             run_single_file_mp(file_name,
                                gpu_id,
                                batch_num,
+                               n_gpus,
                                out_dir,
                                initial_solution_file,
                                heuristics_only,
@@ -501,6 +506,7 @@ int main(int argc, char* argv[])
     run_single_file(path,
                     0,
                     0,
+                    n_gpus,
                     out_dir,
                     initial_solution_file,
                     heuristics_only,
--- a/ci/test_python.sh
+++ b/ci/test_python.sh
@@ -60,6 +60,9 @@ EXITCODE=0
 trap "EXITCODE=1" ERR
 set +e
 
+# Due to race condition in certain cases UCX might not be able to cleanup properly, so we set the number of threads to 1
+export OMP_NUM_THREADS=1
+
 rapids-logger "Test cuopt_cli"
 timeout 10m bash ./python/libcuopt/libcuopt/tests/test_cli.sh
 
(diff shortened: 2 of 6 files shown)


cpp/src/dual_simplex/branch_and_bound.cpp
```diff
@@ -695,17 +691,18 @@ void branch_and_bound_t<i_t, f_t>::exploration_ramp_up(search_tree_t<i_t, f_t>*
                                                        i_t initial_heap_size)
 {
   if (status_ != mip_exploration_status_t::RUNNING) { return; }
-  if (omp_get_thread_num() == 0) { repair_heuristic_solutions(); }
+
+  // Note that we do not know which thread will execute the
+  // `exploration_ramp_up` task, so we allow to any thread
+  // to repair the heuristic solution.
+  repair_heuristic_solutions();
```
chris-maes, line 698: The original intent was just to have a single thread attempt to repair the heuristic solution. We don't want multiple threads to try to repair the same heuristic solution. Is there a way to ensure that only a single thread repairs?

cpp/src/dual_simplex/branch_and_bound.cpp
```diff
@@ -566,20 +566,13 @@ node_status_t branch_and_bound_t<i_t, f_t>::solve_node(search_tree_t<i_t, f_t>&
   lp_solution_t<i_t, f_t> leaf_solution(leaf_problem.num_rows, leaf_problem.num_cols);
   assert(leaf_vstatus.size() == leaf_problem.num_cols);
 
-  // Set the correct bounds for the leaf problem
-  leaf_problem.lower = original_lp_.lower;
-  leaf_problem.upper = original_lp_.upper;
```
chris-maes: Don't you need these lines to restore the bounds to that of the original problem? Since you may modified the bounds of the leaf_problem previously?

cpp/src/dual_simplex/branch_and_bound.cpp
```diff
@@ -714,14 +711,18 @@ void branch_and_bound_t<i_t, f_t>::exploration_ramp_up(search_tree_t<i_t, f_t>*
     return;
   }
 
-  f_t now = toc(stats_.start_time);
+  f_t now                 = toc(stats_.start_time);
+  f_t time_since_last_log = stats_.last_log == 0 ? 1.0 : toc(stats_.last_log);
 
-  if (omp_get_thread_num() == 0) {
```
chris-maes: Don't we only want one thread to print?

cpp/src/dual_simplex/branch_and_bound.cpp
```diff
@@ -735,15 +736,18 @@ void branch_and_bound_t<i_t, f_t>::exploration_ramp_up(search_tree_t<i_t, f_t>*
...
-      stats_.nodes_since_last_log = 0;
     }
   }
 
   if (now > settings_.time_limit) {
     status_ = mip_exploration_status_t::TIME_LIMIT;
     return;
   }
+
+  // Set the correct bounds for the leaf problem
```
chris-maes, line 747: I see. You are restoring here the bounds here. Is there a reason to move it from the original spot?

cpp/src/dual_simplex/branch_and_bound.cpp
nguidotti: There is a mutex for retrieving the solutions in the repair queue. Once all solutions are copied, the repair queue is cleared and the mutex is released (line 390 to 396):

```
  std::vector<std::vector<f_t>> to_repair;
  mutex_repair_.lock();
  if (repair_queue_.size() > 0) {
    to_repair = repair_queue_;
    repair_queue_.clear();
  }
  mutex_repair_.unlock();
```

The other threads will see the repair queue is empty and then move on.

cpp/src/dual_simplex/branch_and_bound.cpp
nguidotti: In the tasking model, one thread spawns a task, it goes to a pool and then it is distributed to any available thread. In the ramp-up phase, it will have more threads than tasks, such that some will execute the corresponding task, while others are idling. We do not know which thread will be in each group.

Hence, if we restrict logging to `T0` (or any other thread), then the log will only appear when it is active, which may only happen near the end.

nguidotti: When it is time to report, the code is atomically check if the node is the last one explored. If it is the case, then the thread will print the log. Otherwise, it will skip the report and allow another thread to do it instead.

cpp/src/dual_simplex/branch_and_bound.cpp
nguidotti: There are two reasons:
1. The initial bounds for the best first and diving are different. Best-first uses the bounds of the original problem, while diving uses the bounds of the starting node (we need to keep the bounds separated since the node was detached from the main tree).
2. In #473 , we only need to change the bounds after reaching the bottom of the branch.

cpp/src/dual_simplex/branch_and_bound.cpp
```diff
@@ -566,20 +566,13 @@ node_status_t branch_and_bound_t<i_t, f_t>::solve_node(search_tree_t<i_t, f_t>&
...
   // Technically, we can get the already strengthened bounds from the node/parent instead of
   // getting it from the original problem and re-strengthening. But this requires storing
   // two vectors at each node and potentially cause memory issues
   node_ptr->get_variable_bounds(leaf_problem.lower, leaf_problem.upper, bounds_changed);
 
-  i_t node_iter                    = 0;
-  f_t lp_start_time                = tic();
-  std::vector<f_t> leaf_edge_norms = edge_norms_;  // = node.steepest_edge_norms;
-
+  std::vector<f_t> leaf_edge_norms      = edge_norms_;  // = node.steepest_edge_norms;
```
akifcorduk: You can copy this in the `if(feasible)` block

cpp/src/dual_simplex/branch_and_bound.cpp
akifcorduk: Note that: the repair will probably be needed much less since now the tolerances are compatible.

cpp/src/dual_simplex/branch_and_bound.cpp
```diff
@@ -784,7 +788,7 @@ void branch_and_bound_t<i_t, f_t>::explore_subtree(i_t id,
   stack.push_front(start_node);
 
   while (stack.size() > 0 && status_ == mip_exploration_status_t::RUNNING) {
-    if (omp_get_thread_num() == 0) { repair_heuristic_solutions(); }
+    if (id == 0) { repair_heuristic_solutions(); }
```
akifcorduk: Instead of id, could we use more descriptive name? like `worker_id` , `thread_id` etc.

cpp/src/dual_simplex/branch_and_bound.cpp
nguidotti: Done

cpp/src/dual_simplex/branch_and_bound.cpp
nguidotti: Changed to `task_id`

cpp/src/dual_simplex/branch_and_bound.cpp
nguidotti: When the heuristic thread injects the solution to the B&B, it would not be easier for it to attempt to repair immediately instead of adding to a queue? The only caveat is the heuristic thread can only repair it after the root node solution.

PR #478, opened 2025-10-09
Barrier log fixes

chris-maes:
Minor tweaks to the barrier log.

### DIFF
--- a/conda/environments/all_cuda-129_arch-aarch64.yaml
+++ b/conda/environments/all_cuda-129_arch-aarch64.yaml
@@ -59,7 +59,6 @@ dependencies:
 - pytest-cov
 - pytest<8
 - python>=3.10,<3.14
-- raft-dask==25.10.*,>=0.0.0a0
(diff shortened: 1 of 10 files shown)


cpp/src/dual_simplex/sparse_cholesky.cuh
```diff
@@ -703,7 +703,7 @@ class sparse_cholesky_cudss_t : public sparse_cholesky_base_t<i_t, f_t> {
     }
 
     if (first_factor) {
-      settings_.log.printf("Factor time %.2fs\n", numeric_time);
+      settings_.log.debug("Factorization time          : %.2fs\n", numeric_time);
```
hlinsen, line 706: Not the same indentation as Symbolic factorization time

hlinsen: Not sure if it is intended

chris-maes: Thanks. This is ok, because it is in debug mode.

PR #483, opened 2025-10-10
Gracefully handle OOM in Barrier

hlinsen:
CUDSS does not throw on OOM and even though we check for the return status. There was a sticky cuda error on any successive CUDA API call. A workaround is to throw if `cudaMallocAsync` fails in the device mem handler and catch that error in the solve method.

### DIFF
--- a/cpp/src/dual_simplex/sparse_cholesky.cuh
+++ b/cpp/src/dual_simplex/sparse_cholesky.cuh
@@ -100,13 +100,17 @@ class sparse_cholesky_base_t {
 template <typename mem_pool_t>
 int cudss_device_alloc(void* ctx, void** ptr, size_t size, cudaStream_t stream)
 {
-  return cudaMallocAsync(ptr, size, stream);
+  int status = cudaMallocAsync(ptr, size, stream);
+  if (status != cudaSuccess) { throw raft::cuda_error("Cuda error in cudss_device_alloc"); }
+  return status;
 }
 
 template <typename mem_pool_t>
 int cudss_device_dealloc(void* ctx, void* ptr, size_t size, cudaStream_t stream)
 {
-  return cudaFreeAsync(ptr, stream);
+  int status = cudaFreeAsync(ptr, stream);
+  if (status != cudaSuccess) { throw raft::cuda_error("Cuda error in cudss_device_dealloc"); }
+  return status;
 }
 
 template <class T>
--- a/cpp/src/dual_simplex/barrier.cu
+++ b/cpp/src/dual_simplex/barrier.cu
@@ -3242,331 +3242,339 @@ lp_status_t barrier_solver_t<i_t, f_t>::solve(f_t start_time,
                                               const barrier_solver_settings_t<i_t, f_t>& options,
                                               lp_solution_t<i_t, f_t>& solution)
 {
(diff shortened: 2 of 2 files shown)


cpp/src/dual_simplex/sparse_cholesky.cuh
```diff
@@ -760,7 +778,10 @@ class sparse_cholesky_cudss_t : public sparse_cholesky_base_t<i_t, f_t> {
       status,
       "cudssMatrixCreateDn for x");
 
-    status = cudssExecute(handle, CUDSS_PHASE_SOLVE, solverConfig, solverData, A, cudss_x, cudss_b);
+    CUDSS_CALL_AND_CHECK(
```
chris-maes: If we get a user interrupt during a cuDSS solve, cuDSS will stop and return a bad status code. And then the CUDSS_CALL_AND_CHECK macro, will print out a big warning. 

Since this can happen normally during concurrent solves, I avoid that warning by just getting the status code directly. And then checking the concurrent halt status.

cpp/src/dual_simplex/sparse_cholesky.cuh
```diff
@@ -501,8 +516,11 @@ class sparse_cholesky_cudss_t : public sparse_cholesky_base_t<i_t, f_t> {
       cudssMatrixSetValues(A, Arow.x.data()), status, "cudssMatrixSetValues for A");
 
     f_t start_numeric = tic();
-    status            = cudssExecute(
-      handle, CUDSS_PHASE_FACTORIZATION, solverConfig, solverData, A, cudss_x, cudss_b);
+    CUDSS_CALL_AND_CHECK(
```
chris-maes: See comment below. I think we don't want to use the `CUDSS_CALL_AND_CHECK_MACRO` here.

cpp/src/dual_simplex/sparse_cholesky.cuh
```diff
@@ -445,8 +451,16 @@ class sparse_cholesky_cudss_t : public sparse_cholesky_base_t<i_t, f_t> {
       settings_.log.printf("Reordering time             : %.2fs\n", reordering_time);
       start_symbolic_factor = tic();
 
-      status = cudssExecute(
-        handle, CUDSS_PHASE_SYMBOLIC_FACTORIZATION, solverConfig, solverData, A, cudss_x, cudss_b);
+      CUDSS_CALL_AND_CHECK(cudssExecute(handle,
```
chris-maes: See comment below.

cpp/src/dual_simplex/sparse_cholesky.cuh
```diff
@@ -431,8 +435,10 @@ class sparse_cholesky_cudss_t : public sparse_cholesky_base_t<i_t, f_t> {
 
     {
       raft::common::nvtx::range fun_scope("Barrier: cuDSS Analyze : CUDSS_PHASE_ANALYSIS");
-      status =
-        cudssExecute(handle, CUDSS_PHASE_REORDERING, solverConfig, solverData, A, cudss_x, cudss_b);
+      CUDSS_CALL_AND_CHECK(
```
chris-maes: See comment below.

PR #488, opened 2025-10-13
Print scaling information about the user model

chris-maes:
Print scaling information and issue a warning if the range of coefficients is greater than 10^6. 

This PR also prints out problem size information before scaling information. 

It also changes 'Running' to 'Reading' so that users understand that we are still reading the file.

### DIFF
--- a/cpp/src/mip/presolve/third_party_presolve.cpp
+++ b/cpp/src/mip/presolve/third_party_presolve.cpp
@@ -402,7 +402,7 @@ std::pair<optimization_problem_t<i_t, f_t>, bool> third_party_presolve_t<i_t, f_
 {
   papilo::Problem<f_t> papilo_problem = build_papilo_problem(op_problem, category);
 
-  CUOPT_LOG_INFO("Unpresolved problem: %d constraints, %d variables, %d nonzeros",
+  CUOPT_LOG_INFO("Original problem: %d constraints, %d variables, %d nonzeros",
                  papilo_problem.getNRows(),
                  papilo_problem.getNCols(),
                  papilo_problem.getConstraintMatrix().getNnz());


cpp/src/linear_programming/optimization_problem.cu
```diff
@@ -273,6 +274,20 @@ i_t optimization_problem_t<i_t, f_t>::get_nnz() const
   return A_.size();
 }
 
+template <typename i_t, typename f_t>
+i_t optimization_problem_t<i_t, f_t>::get_n_integers() const
```
hlinsen, line 278: If you compute the scaling info on `problem_t` we already have a method to compute integer info

cpp/src/linear_programming/optimization_problem.cu
```diff
@@ -583,6 +598,84 @@ void optimization_problem_t<i_t, f_t>::write_to_mps(const std::string& mps_file_
   cuopt::mps_parser::write_mps(data_model_view, mps_file_path);
 }
 
+template <typename i_t, typename f_t>
+void optimization_problem_t<i_t, f_t>::print_scaling_information() const
```
hlinsen, line 602: Should it be printed by default? I am asking because for large problems since we are computing abs values on the host for many buffers and the coefficient matrix it could add some overhead.

chris-maes: Well, I think we always need to check the scaling. In case it turns out to be bad. If we are worried about speed we could move it to the GPU.

cpp/src/linear_programming/optimization_problem.cu
chris-maes: What's the difference between `optimization_problem_t` and `problem_t`? The reason why I did it on `optimization_problem_t` is because it only make sense to warn the user about poor scaling on their input. We don't want to warn if we transformed the problem and made it worse.

hlinsen: problem_t is adapted to store more info and the modified problem. We could check before any transformation on problem_t here for lp: https://github.com/NVIDIA/cuopt/blob/branch-25.10/cpp/src/linear_programming/solve.cu#L823 and here for mip: https://github.com/NVIDIA/cuopt/blob/branch-25.10/cpp/src/mip/solve.cu#L195. This also work so it's not a blocker.

PR #489, opened 2025-10-13
Fix return on size mismatch in Barrier

hlinsen:
TODO: Check reason of mismtach


cpp/src/dual_simplex/sparse_cholesky.cuh
```diff
@@ -568,7 +568,7 @@ class sparse_cholesky_cudss_t : public sparse_cholesky_base_t<i_t, f_t> {
 #endif
     if (A_in.n != n) {
       printf("Analyze input does not match size %d != %d\n", A_in.n, n);
```
chris-maes, line 570: Maybe we should do `settings.log.printf("Analyze ..")` here? So it goes to the log.

PR #503, opened 2025-10-14
Fix TSP when order locations are set

hlinsen:
This PR fixes uninitialized depot pred/succ when order locations are set.
There was also an issue when computing the distance cost with the exclusive scan for the order location case. We are now using the correct number of nodes for both cases.

## Summary by CodeRabbit

* **Refactor**
  * Simplified internal route copying mechanism by removing explicit parameters.
  * Adjusted sliding TSP local search algorithm to consider a different set of orders during optimization.
  * Modified memory allocation strategy for routing computations.


cpp/src/routing/local_search/sliding_tsp.cu
```diff
@@ -275,7 +275,7 @@ __global__ void execute_sliding_moves_tsp(
   s_route.copy_from(route);
   __syncthreads();
 
-  s_route.copy_to_tsp_route(sol.problem.order_info.depot_included);
```
rg20, line 278: Why is it handled differently for TSP?

hlinsen: Since we have move recycling this representation provides ~30% faster move execution. We only update pred/succ pointers and copying to route_t at the end rather than shifting route_t to eject/insert nodes for each non overlapping move.

rg20: Sorry, I meant the depot included logic

hlinsen: I introduced it by mistake previously, now we set the depot to start/end node always. It is the same for `route_t` right?

rgsl888prabhu:
Merging this manually since it is a critical PR

PR #504, opened 2025-10-15
Crossover dualize fix

chris-maes:
When we solve the dual, instead of the primal problem, we weren't populating the primal slack variables. As a result crossover had a large initial primal residual and could fail. 

This issue appeared when solving the LP relaxation of graph20-80-1rand.

This issue was not present if crossover was disabled.

This PR also removes an unnecessary transpose of the constraint matrix in `convert_less_than_to_equal`.


chris-maes:
This isn't critical for the 25.12 release. I'm moving it to 26.02.

chris-maes:
This PR included as part of #814

PR #527, opened 2025-10-22
RINS heuristic

### LINKED ISSUE #531: [FEA] Recombiner with RINS

aliceb-nv:
This PR adds support for the RINS heuristic, which is run every N nodes to search for feasible solutions in a neighborhood computed from the equal integers between the best integer incumbent and the node's LP optimal.
RINS provides a ~1% improvement over the main branch: `~15.5%` vs `~14.3-14.7%`.

This PR also fixes a bug in the CPUFJ external solution bookkeeping.

closes #531 

## Summary by CodeRabbit

* **New Features**
  * Integrated multi-threaded RINS local search and CPU worker scaffold for stronger MIP diversification.
  * Configurable per-run node limit and node-processed callback for live node-level reporting.
  * Log-prefix propagation for clearer, consistent logs.

* **Bug Fixes**
  * Safer solution copying and presolve bounds/flag consistency checks.
  * Improved thread-safety and synchronization across population/diversity components.
  * More robust incumbent handling and explicit node-limit status reporting.

### DIFF
--- a/cpp/src/mip/diversity/diversity_manager.cu
+++ b/cpp/src/mip/diversity/diversity_manager.cu
@@ -320,10 +320,10 @@ solution_t<i_t, f_t> diversity_manager_t<i_t, f_t>::run_solver()
     ls.constraint_prop.bounds_update.calculate_infeasible_redundant_constraints(*problem_ptr),
     "The problem must not be ii");
   population.initialize_population();
+  population.allocate_solutions();
   if (check_b_b_preemption()) { return population.best_feasible(); }
   add_user_given_solutions(initial_sol_vector);
   // Run CPUFJ early to find quick initial solutions
-  population.allocate_solutions();
   ls_cpufj_raii_guard_t ls_cpufj_raii_guard(ls);  // RAII to stop cpufj threads on solve stop
   ls.start_cpufj_scratch_threads(population);
 


cpp/src/mip/diversity/lns/rins.cu
```diff
@@ -0,0 +1,396 @@
...
+
+  if (node_count - node_count_at_last_rins > settings.node_freq) {
+    std::lock_guard<std::mutex> lock(rins_mutex);
+    if (!rins_thread->cpu_thread_should_start && dm.population.current_size() > 0 &&
+        dm.population.is_feasible()) {
+      lp_optimal_solution = solution;
+      rins_thread->start_cpu_solver();
+    }
+  }
+}
```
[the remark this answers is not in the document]
aliceb-nv: already atomic, plenty of implicit memory barriers

cpp/src/mip/diversity/lns/rins.cu
```diff
@@ -0,0 +1,396 @@
...
+                 "Assignment size mismatch");
+
+    rmm::device_uvector<i_t> vars_to_fix(problem_ptr->n_integer_vars,
+                                         problem_ptr->handle_ptr->get_stream());
+    auto end = thrust::copy_if(problem_ptr->handle_ptr->get_thrust_policy(),
+                               thrust::make_counting_iterator(i_t(0)),
+                               thrust::make_counting_iterator(problem_ptr->n_variables),
+                               vars_to_fix.begin(),
+                               [lpopt     = lp_opt_device.data(),
+                                incumbent = best_sol.assignment.data(),
```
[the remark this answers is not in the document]
aliceb-nv: n_integer_vars == 0 is impossible for MIP problems. adding assert nonetheless

cpp/src/mip/diversity/lns/rins.cu
```diff
@@ -0,0 +1,396 @@
...
+  if (node_count - node_count_at_last_improvement < settings.nodes_after_later_improvement) return;
+
+  if (node_count - node_count_at_last_rins > settings.node_freq) {
+    std::lock_guard<std::mutex> lock(rins_mutex);
+    if (!rins_thread->cpu_thread_should_start && dm.population.current_size() > 0 &&
+        dm.population.is_feasible()) {
+      lp_optimal_solution = solution;
+      rins_thread->start_cpu_solver();
+    }
+  }
```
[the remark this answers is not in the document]
aliceb-nv: Addressed, thanks

rgsl888prabhu (approved):

cpp/src/dual_simplex/presolve.cpp
```diff
@@ -1381,7 +1381,10 @@ void uncrush_primal_solution(const user_problem_t<i_t, f_t>& user_problem,
 {
   user_solution.resize(user_problem.num_cols);
   assert(problem.num_cols >= user_problem.num_cols);
-  std::copy(solution.begin(), solution.begin() + user_problem.num_cols, user_solution.data());
+  assert(solution.size() >= user_problem.num_cols);
+  std::copy(solution.begin(),
+            solution.begin() + std::min((i_t)solution.size(), user_problem.num_cols),
```
akifcorduk, line 1386: There is no need for min right? It can be directly `user_problem.num_cols`.

cpp/src/mip/diversity/lns/rins.cu
```diff
@@ -0,0 +1,433 @@
...
+  : context(context_), problem_ptr(context.problem_ptr), dm(dm_), settings(settings_)
+{
+  fixrate    = settings.default_fixrate;
+  time_limit = settings.default_time_limit;
+}
+
+template <typename i_t, typename f_t>
+rins_thread_t<i_t, f_t>::rins_thread_t()
+{
+  cpu_worker = std::thread(&rins_thread_t<i_t, f_t>::cpu_worker_thread, this);
```
akifcorduk: I think we should unify our threading methods soon. It might be performance degradation when we use both OMP and std::threads(pthreads?).

cpp/src/mip/diversity/lns/rins.cu
```diff
@@ -0,0 +1,433 @@
...
+
+    {
+      std::lock_guard<std::mutex> lock(cpu_mutex);
+      cpu_thread_done = true;
+    }
+  }
+}
+
+template <typename i_t, typename f_t>
+void rins_thread_t<i_t, f_t>::kill_cpu_solver()
```
akifcorduk: I think we can rename it as `kill_cpu_worker`, it currently sounds like we are terminating B&B.

cpp/src/mip/diversity/lns/rins.cu
```diff
@@ -0,0 +1,433 @@
...
+
+    // Run RINS
+    {
+      raft::common::nvtx::range fun_scope("Running RINS");
+      rins_ptr->run_rins();
+    }
+
+    {
+      std::lock_guard<std::mutex> lock(cpu_mutex);
+      cpu_thread_done = true;
```
akifcorduk: Do we need the mutex here, since the var is already atomic?

cpp/src/mip/diversity/lns/rins.cu
```diff
@@ -0,0 +1,433 @@
...
+  {
+    std::lock_guard<std::mutex> lock(cpu_mutex);
+    cpu_thread_done         = false;
+    cpu_thread_should_start = true;
+  }
+  cpu_cv.notify_one();
+}
+
+template <typename i_t, typename f_t>
+void rins_thread_t<i_t, f_t>::stop_cpu_solver()
```
akifcorduk: Should this be removed?

cpp/src/mip/diversity/lns/rins.cu
```diff
@@ -0,0 +1,433 @@
...
+{
+  while (!cpu_thread_done && !cpu_thread_terminate) {
+    std::this_thread::sleep_for(std::chrono::milliseconds(1));
+  }
+
+  return true;
+}
+
+template <typename i_t, typename f_t>
+void rins_t<i_t, f_t>::new_best_incumbent_callback(const std::vector<f_t>& solution)
```
akifcorduk, line 51: `solution` is not used in the callback. If it is intended, we should also change the function name.

cpp/src/mip/diversity/lns/rins.cu
```diff
@@ -0,0 +1,433 @@
...
+  seed                  = cuopt::seed_generator::get_seed();
+  problem_copy          = std::make_unique<problem_t<i_t, f_t>>(*problem_ptr, /*deep_copy=*/true);
+  problem_copy->handle_ptr = &rins_handle;
+  enabled                  = true;
+}
+
+template <typename i_t, typename f_t>
+void rins_t<i_t, f_t>::run_rins()
+{
+  if (total_calls == 0) cudaSetDevice(context.handle_ptr->get_device());
```
akifcorduk: Please wrap this in error checking.

cpp/src/mip/diversity/lns/rins.cu
```diff
@@ -0,0 +1,433 @@
...
+
+  if (!dm.population.is_feasible()) return;
```
akifcorduk, line 93: I think here you can check if there are any external solutinos and add them to the population. Sometimes, FP loop can run longer without adding those external solutions coming from CPUFJ and B&B.

cpp/src/mip/diversity/lns/rins.cu
```diff
@@ -0,0 +1,433 @@
...
+
+  cuopt_assert(lp_optimal_solution.size() == problem_ptr->n_variables, "Assignment size mismatch");
+  cuopt_assert(problem_copy->handle_ptr == &rins_handle, "Handle mismatch");
+  cuopt_assert(problem_copy->n_variables == problem_ptr->n_variables, "Problem size mismatch");
+  cuopt_assert(problem_copy->n_constraints == problem_ptr->n_constraints, "Problem size mismatch");
+  cuopt_assert(problem_copy->n_integer_vars == problem_ptr->n_integer_vars,
+               "Problem size mismatch");
+  cuopt_assert(problem_copy->n_binary_vars == problem_ptr->n_binary_vars, "Problem size mismatch");
+
+  if (dm.population.current_size() > 0) {
```
akifcorduk: I could just check if population is feasible. There may be cases, population might be infeasible with non-zero solutino count.

cpp/src/mip/diversity/lns/rins.cu
```diff
@@ -0,0 +1,433 @@
...
+      cuopt_assert(best_feasible_ref.get_feasible(), "Best feasible is not feasible");
+      expand_device_copy(
+        best_sol.assignment, best_feasible_ref.assignment, rins_handle.get_stream());
+      best_sol.handle_ptr  = &rins_handle;
+      best_sol.problem_ptr = problem_copy.get();
+      best_sol.compute_feasibility();
+    }
+    // also scour through the external solution queue to potentially
+    // nab better solutions early
+    {
```
akifcorduk: I think if you add all external solutions before in the first if check, you might not need this at all.

cpp/src/mip/diversity/lns/rins.cu
```diff
@@ -0,0 +1,433 @@
...
+          best_sol.compute_feasibility();
+          printf("RINS External solution is feas? %d, excess %g\n",
+                 best_sol.get_feasible(),
+                 best_sol.get_total_excess());
+        }
+      }
+    }
+    cuopt_assert(best_sol.handle_ptr == &rins_handle, "Handle mismatch");
+
+    if (!best_sol.get_feasible()) { return; }
```
akifcorduk: I think you don't need this if you add external solutions in the beginning

cpp/src/mip/diversity/lns/rins.cu
```diff
@@ -0,0 +1,433 @@
...
+    i_t sol_size_before_rins = best_sol.assignment.size();
+    auto lp_opt_device = cuopt::device_copy(this->lp_optimal_solution, rins_handle.get_stream());
+    cuopt_assert(lp_opt_device.size() == problem_ptr->n_variables, "Assignment size mismatch");
+    cuopt_assert(best_sol.assignment.size() == problem_ptr->n_variables,
+                 "Assignment size mismatch");
+
+    rmm::device_uvector<i_t> vars_to_fix(problem_ptr->n_integer_vars, rins_handle.get_stream());
+    auto end = thrust::copy_if(rins_handle.get_thrust_policy(),
```
akifcorduk: For performance, you could just iterate on integer_indices and remove `if (!pb.is_integer_var(var_idx)) return false;`

cpp/src/mip/diversity/lns/rins.cu
```diff
@@ -0,0 +1,433 @@
...
+    }
+
+    thrust::default_random_engine g(seed + node_count);
+
+    // shuffle fixing order
+    thrust::shuffle(rins_handle.get_thrust_policy(), vars_to_fix.begin(), vars_to_fix.end(), g);
+
+    // fix n first according to fractional ratio
+    f_t rins_ratio = fixrate;
+    i_t n_to_fix   = std::max((int)(vars_to_fix.size() * rins_ratio), 0);
```
akifcorduk: Don't you think there should be some min ? I think we could use fixrate if the number of vars are too large. Fixing too few is also an issue i think.

cpp/src/mip/diversity/lns/rins.cu
```diff
@@ -0,0 +1,433 @@
...
+    namespace dual_simplex = cuopt::linear_programming::dual_simplex;
+    dual_simplex::user_problem_t<i_t, f_t> branch_and_bound_problem(&rins_handle);
+    dual_simplex::simplex_solver_settings_t<i_t, f_t> branch_and_bound_settings;
+    dual_simplex::mip_solution_t<i_t, f_t> branch_and_bound_solution(1);
+    dual_simplex::mip_status_t branch_and_bound_status = dual_simplex::mip_status_t::UNSET;
+    fixed_problem.get_host_user_problem(branch_and_bound_problem);
+    branch_and_bound_solution.resize(branch_and_bound_problem.num_cols);
+    // Fill in the settings for branch and bound
+    branch_and_bound_settings.time_limit = time_limit;
+    // branch_and_bound_settings.node_limit = 5000 + node_count / 100;  // try harder as time goes
```
akifcorduk: Stale code

cpp/src/mip/diversity/lns/rins.cu
```diff
@@ -0,0 +1,433 @@
...
+    // on
+    branch_and_bound_settings.print_presolve_stats = false;
+    branch_and_bound_settings.absolute_mip_gap_tol = context.settings.tolerances.absolute_mip_gap;
+    branch_and_bound_settings.relative_mip_gap_tol = 0.03;  // 3%
```
akifcorduk: I think there should be some min. If the mip_gap is already lower than 3%, it will not run at all.

cpp/src/mip/diversity/lns/rins.cu
```diff
@@ -0,0 +1,433 @@
...
+    cpu_fj_thread.stop_cpu_solver();
+    bool fj_solution_found = cpu_fj_thread.wait_for_cpu_solver();
+    CUOPT_LOG_DEBUG("RINS FJ ran for %d iterations", cpu_fj_thread.fj_cpu->iterations);
+    if (fj_solution_found) {
+      CUOPT_LOG_DEBUG("RINS FJ solution found. Objective %.16e",
+                      cpu_fj_thread.fj_cpu->h_best_objective);
+      if (cpu_fj_thread.fj_cpu->h_best_objective < branch_and_bound_solution.objective ||
+          std::isnan(branch_and_bound_solution.objective)) {
+        // first post process the trivial presolve on a device vector
+        rmm::device_uvector<f_t> post_processed_solution(branch_and_bound_solution.x.size(),
```
akifcorduk: I would also add intermediate B&B solutions to population, so that we don't waste any diverse solutions coming out. You can check in submip how it is already implemented.

cpp/src/mip/diversity/lns/rins.cu
```diff
@@ -0,0 +1,433 @@
...
+      if (best_sol.get_user_objective() < prev_obj) {
+        CUOPT_LOG_DEBUG("RINS solution improved objective from %g to %g",
+                        prev_obj,
+                        best_sol.get_user_objective());
+        total_success++;
+      }
+      cuopt_assert(best_sol.assignment.size() == sol_size_before_rins, "Assignment size mismatch");
+      cuopt_assert(best_sol.assignment.size() == problem_ptr->n_variables,
+                   "Assignment size mismatch");
+      if (best_sol.get_objective() < dm.population.best_feasible_objective) {
```
akifcorduk: I think you can just add the solution found and delegate the objective check to population.

cpp/src/mip/diversity/lns/rins.cuh
```diff
@@ -0,0 +1,118 @@
...
+struct rins_settings_t {
+  int node_freq                     = 100;
+  int nodes_after_later_improvement = 200;
+  double min_fixrate                = 0.3;
+  double max_fixrate                = 0.8;
+  double default_fixrate            = 0.5;
+  double min_fractional_ratio       = 0.3;
+  double min_time_limit             = 10.;
+  double max_time_limit             = 20.;
+  double default_time_limit         = 6.;
```
akifcorduk: 6 seconds is too much i think, did you have good results with that?. also it is out of `min_time_limit` and `max_time_limit`.

cpp/src/mip/diversity/population.cu
```diff
@@ -391,6 +388,7 @@ void population_t<i_t, f_t>::adjust_weights_according_to_best_feasible()
 template <typename i_t, typename f_t>
 std::pair<i_t, bool> population_t<i_t, f_t>::add_solution(solution_t<i_t, f_t>&& sol)
 {
+  std::lock_guard<std::recursive_mutex> lock(write_mutex);
```
akifcorduk, line 401: Why do we need those mutex's ? We are adding all other solution through add_external_solution right? it handles the mutex and proper addition to the population. Otherwise, population is always used by a single thread.

cpp/src/mip/diversity/population.cuh
```diff
@@ -35,13 +35,15 @@ namespace cuopt::linear_programming::detail {
 template <typename i_t, typename f_t>
 class diversity_manager_t;
 
-enum class solution_origin_t { BRANCH_AND_BOUND, CPUFJ, EXTERNAL };
+enum class solution_origin_t { BRANCH_AND_BOUND, CPUFJ, RENS, RINS, EXTERNAL };
```
akifcorduk: rens is not here yet right?

cpp/src/mip/feasibility_jump/fj_cpu.cu
```diff
@@ -1054,12 +1054,91 @@ bool fj_t<i_t, f_t>::cpu_solve(fj_cpu_climber_t<i_t, f_t>& fj_cpu, f_t in_time_l
   return fj_cpu.feasible_found;
 }
 
+template <typename i_t, typename f_t>
+cpu_fj_thread_t<i_t, f_t>::cpu_fj_thread_t()
```
akifcorduk: Maybe we can templetize it and create a common thread worker between rins and cpufj ?

cpp/src/mip/problem/problem.cu
```diff
@@ -198,7 +198,7 @@ problem_t<i_t, f_t>::problem_t(const problem_t<i_t, f_t>& problem_)
 }
 
 template <typename i_t, typename f_t>
-problem_t<i_t, f_t>::problem_t(const problem_t<i_t, f_t>& problem_, bool no_deep_copy)
+problem_t<i_t, f_t>::problem_t(const problem_t<i_t, f_t>& problem_, bool deep_copy)
```
akifcorduk: It seems the constructor is completely changed. Now all of them does a deep copy. Why do we chage the names and the way constructor works? Did you test it with pdlp ?

cpp/src/mip/problem/problem.cu
```diff
@@ -213,70 +213,66 @@ problem_t<i_t, f_t>::problem_t(const problem_t<i_t, f_t>& problem_, bool no_deep
     empty(problem_.empty),
     is_binary_pb(problem_.is_binary_pb),
     // Copy constructor used by PDLP and MIP
-    // PDLP uses the version with no_deep_copy = false which deep copy some fields but doesn't
+    // PDLP uses the version with deep_copy = false which deep copy some fields but doesn't
```
akifcorduk: Change comment?

cpp/src/mip/relaxed_lp/relaxed_lp.cu
```diff
@@ -62,7 +62,7 @@ optimization_problem_solution_t<i_t, f_t> get_relaxed_lp_solution(
   pdlp_settings.concurrent_halt                      = settings.concurrent_halt;
   pdlp_settings.per_constraint_residual              = settings.per_constraint_residual;
   pdlp_settings.first_primal_feasible                = settings.return_first_feasible;
-  pdlp_settings.pdlp_solver_mode                     = pdlp_solver_mode_t::Stable2;
+  pdlp_settings.pdlp_solver_mode                     = pdlp_solver_mode_t::Stable3;  // 3
```
akifcorduk: Stable 3 does not support some features that we use in MIP like `per_constraint_residual`,`return_first_feasible` etc. Also it is not tested widely yet.

cpp/src/dual_simplex/presolve.cpp
aliceb-nv: That's a safeguard in case we compile with asserts off. I've had the issue pop up a few times during testing (probably because I was using an outdated version of main branch w/o some of nicolas' bugfixes)

cpp/src/mip/diversity/lns/rins.cu
aliceb-nv: Definitely :) We should schedule a sync on this soon

cpp/src/mip/diversity/lns/rins.cu
aliceb-nv: We don't :) Oversight, fixed

cpp/src/mip/diversity/lns/rins.cu
aliceb-nv: This is future proofing; it might prove useful to some later heuristics (but is harmless as it is since this is a reference)

cpp/src/mip/diversity/lns/rins.cu
aliceb-nv: We perform an abort earlier w/ min_fractional_ratio if this is the case; do you think we should add a min here?

cpp/src/mip/diversity/lns/rins.cu
aliceb-nv: I'd prefer to keep it for future reference (iirc this is the strategy used by SCIP and might be worth trying again sometime in the future)

cpp/src/mip/diversity/lns/rins.cuh
aliceb-nv: I will run some benches again w/ the latest changes, but these were settings yielding good results. RINS usually requires a fair amount of runtime to explore the space thoroughly

aliceb-nv: The default_time_limit is an oversight :) Will fix

cpp/src/mip/diversity/population.cu
aliceb-nv: I added this to ensure the RINS thread wouldn't try to fetch the population vector while it is in the middle of an update by the main GPU heuristic thread

cpp/src/mip/diversity/population.cuh
aliceb-nv: Overly-optimistic future-proofing :) I will remove it

cpp/src/mip/feasibility_jump/fj_cpu.cu
aliceb-nv: Yes. Overall this part of the code is very ugly, I will refactor before merge

cpp/src/mip/relaxed_lp/relaxed_lp.cu
aliceb-nv: Oops, I accidentally pushed an benchmark experiment commit to the PR 
You're right, it should be Stable2

cpp/src/mip/problem/problem.cu
aliceb-nv: I initially made those changes because PDLP used no_deep_copy = false to do shallow copies which I found confusing, but it seems I messed the whole constructor up in the process. Let me revert

cpp/src/mip/diversity/lns/rins.cu
akifcorduk: I didn't pay attention to min_fractional_ratio, I think that's okay. We could probably experiment on some discrete min number too.

cpp/src/mip/problem/problem.cu
```diff
@@ -213,70 +213,66 @@ problem_t<i_t, f_t>::problem_t(const problem_t<i_t, f_t>& problem_, bool no_deep
...
     // allocate others that are not needed in PDLP
     presolve_data(
-      (!no_deep_copy)
+      (!deep_copy)
```
nguidotti: Both branches of the `if` statement are the same... Is it intentional?

cpp/src/dual_simplex/branch_and_bound.cpp
```diff
@@ -1018,13 +1041,15 @@ void branch_and_bound_t<i_t, f_t>::diving_thread(lp_problem_t<i_t, f_t>& leaf_pr
 }
 
 template <typename i_t, typename f_t>
-mip_status_t branch_and_bound_t<i_t, f_t>::solve(mip_solution_t<i_t, f_t>& solution)
+mip_status_t branch_and_bound_t<i_t, f_t>::solve(mip_solution_t<i_t, f_t>& solution,
```
nguidotti: Should be better to set the `log_prefix` in the settings instead of here?

aliceb-nv: You make a good point :) Will change

cpp/src/mip/problem/problem.cu
aliceb-nv: Outdated; this was an oversight :)

PR #534, opened 2025-10-23
Carry over implied integers from papilo presolve reductions

aliceb-nv:
Papilo computes implied integer (continuous variables that are proven to take only integer values when integers are fixed) variables during its presolving process, via `ImplIntDetection`. 
We do not make use of these yet, but implied integers will prove useful when cuts are implemented.
This PR exposes this information as a variable flag in `problem_t`.
Additionally, a flag is set if the objective function is guaranteed to be integer as well for feasible solutions. 

## Summary by CodeRabbit

* **New Features**
  * Presolve now returns implied-integer info and applies it to problems.
  * Problems can report whether the objective is integral.
  * New utility to convert raw buffers into cuDF columns.

* **Improvements**
  * Per-variable flags added to preserve and propagate implied-integer status.
  * Logging reports implied-integer counts and objective integrality.

* **Tests**
  * Added a unit test validating presolve produces implied-integer flags.

### DIFF
--- a/cpp/src/mip/presolve/third_party_presolve.hpp
+++ b/cpp/src/mip/presolve/third_party_presolve.hpp
@@ -17,12 +17,15 @@
 
 #pragma once
 
+#include <optional>
+
 #include <cuopt/linear_programming/optimization_problem.hpp>
 
 namespace cuopt::linear_programming::detail {
 
 template <typename i_t, typename f_t>
-struct presolve_information_t {
+struct third_party_presolve_result_t {
+  optimization_problem_t<i_t, f_t> reduced_problem;
   std::vector<i_t> implied_integer_indices;
   // clique info, etc...
 };
@@ -32,7 +35,7 @@ class third_party_presolve_t {
  public:
   third_party_presolve_t() = default;
 
-  std::tuple<optimization_problem_t<i_t, f_t>, bool, presolve_information_t<i_t, f_t>> apply(
+  std::optional<third_party_presolve_result_t<i_t, f_t>> apply(
     optimization_problem_t<i_t, f_t> const& op_problem,
     problem_category_t category,
     bool dual_postsolve,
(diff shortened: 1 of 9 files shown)


cpp/src/mip/presolve/third_party_presolve.hpp
```diff
@@ -21,12 +21,18 @@
 
 namespace cuopt::linear_programming::detail {
 
+template <typename i_t, typename f_t>
+struct presolve_information_t {
```
rg20:
```suggestion
struct presolve_result_t {
```

You can move all the returned objects including `optimization_problem_t`, `feasible` into this one.

cpp/src/mip/problem/problem.cu
```diff
@@ -89,6 +89,8 @@ void problem_t<i_t, f_t>::op_problem_cstr_body(const optimization_problem_t<i_t,
   // If maximization problem, convert the problem
   if (maximize) convert_to_maximization_problem(*this);
   if (is_mip) {
+    var_flags.resize(n_variables, handle_ptr->get_stream());
```
rg20: you can resize only if needed right?

cpp/src/mip/problem/problem.cu
```diff
@@ -953,6 +963,7 @@ void problem_t<i_t, f_t>::resize_variables(size_t size)
   variable_types.resize(size, handle_ptr->get_stream());
   objective_coefficients.resize(size, handle_ptr->get_stream());
   is_binary_variable.resize(size, handle_ptr->get_stream());
+  var_flags.resize(size, handle_ptr->get_stream());  // 0 is default - no flag
```
rg20: resize only if the previous one was not empty?

cpp/src/mip/presolve/third_party_presolve.hpp
hlinsen: I would try to reuse `presolve_data_t` if we can https://github.com/NVIDIA/cuopt/blob/main/cpp/src/mip/problem/presolve_data.cuh#L33

cpp/src/mip/problem/problem.cu
aliceb-nv: This is the constructor for MIP, var_flags is otherwise empty (e.g. for PDLP when memory footprint matters)

cpp/src/mip/problem/problem.cu
aliceb-nv: I prefer to keep consistency with the rest of the resize operations for now, I can do a pass for memory footprint optimization later if the need arises for MIP

PR #541, opened 2025-10-24
Handle concurrent mode oom and avoid unnecessary deep copy of problem_t

hlinsen:
`user_problem_t` holds the `raft::handle_t` for the whole Barrier solver run. Setting the handle correctly and calling run_barrier_thread allows running Barrier in its own stream without deep copying the `problem_t`.
We are now able to solve large scale instances and any OOM in Barrier is caught and PDLP keeps iterating.


chris-maes (approved):
LGTM. Thanks Hugo!

Issue #537, opened 2025-10-24
[BUG] cuOpt Crash with Invalid __global__ Write and cudaErrorLaunchFailure


upsilon-patrik:
**Describe the bug**
When solving a MIP problem using `pulp.CUOPT` in Python with cuOpt version 25.10.0, the solver crashes with a CUDA launch failure. The error is detected by NVIDIA Compute Sanitizer as an invalid `__global__` write, followed by repeated `cudaErrorLaunchFailure` (error 719) on various CUDA API calls. The crash is ultimately due to an internal cuSparse error during matrix transposition.
Note that solver works for problem size T=10, I=9, and for 'any' size using solver PULP_CBC_CMD.

**Steps/Code to reproduce bug**
```python
import json
import pandas as pd
import numpy as np
import pulp

# Load data
with open('solver_exports_right.json', 'r') as f:
    data_a = json.load(f)
with open('solver_exports_left.json', 'r') as f:
    data_b = json.load(f)

# Add side labels
for rec in data_a:
    rec['cat'] = 'a'
for rec in data_b:
    rec['cat'] = 'b'

# Combine and make DataFrame
dataset = data_a + data_b
df_anon = pd.DataFrame(dataset)

max_metric = df_anon['secSolver'].max()
row_max = df_anon[df_anon['secSolver'] == max_metric]

# Prepare data
rec = row_max.iloc[0]

arr1 = np.array(rec['aoiArr'], dtype=float)    # shape: [dim1, dim2]
T_ = arr1.shape[1]                 # Number of time periods
I_ = arr1.shape[0]                 # Number of options

time_unit = 10                     # seconds per period

# Parameters
PARAM_A = float(60)
PARAM_B = float(200)
PARAM_C = float(1400)
PARAM_D = float(100)
PARAM_E = float(4000)
PARAM_F = float(40)
param_slots = PARAM_F // time_unit

print("Periods:", T_)
print("Options:", I_)

# Model & Variables
prob = pulp.LpProblem("Opt_Model", pulp.LpMaximize)

# Decision variables
var1 = pulp.LpVariable.dicts("opt", ((t, i) for t in range(T_) for i in range(I_)), cat='Binary')
var2 = pulp.LpVariable.dicts("flag", (t for t in range(T_)), cat='Binary')
var3 = pulp.LpVariable.dicts("resource", (t for t in range(T_ + 1)), lowBound=float(0))
[...]

PR #542, opened 2025-10-25
Fix papilo assertion mip

hlinsen:
This PR fixes: https://github.com/NVIDIA/cuopt/issues/537
The root cause was an assertion triggered in papilo in assert mode. Papilo expects the CSR to be sorted if sorted flag is enabled.
I added a method to sort in on the GPU when presolve is run. After the fix the transpose happening after presolve is successful and there are no cusparse error.
```
python: /home/hugo/rapids/cuopt/cpp/build/_deps/papilo-src/src/papilo/core/SparseStorage.hpp:430: papilo::SparseStorage<REAL>::SparseStorage(papilo::Vec<std::tuple<int, int, REAL> >, int, int, bool, double, int) [with REAL = double; papilo::Vec<std::tuple<int, int, REAL> > = std::vector<std::tuple<int, int, double>, std::allocator<std::tuple<int, int, double> > >]: Assertion `!sorted || std::is_sorted( entries.begin(), entries.end() )' failed.
```

## Summary by CodeRabbit

* **Performance**
  * Enhanced constraint data optimization in preprocessing stages of linear programming and mixed-integer programming solvers.


cpp/src/linear_programming/solve.cu
```diff
@@ -837,19 +838,22 @@ optimization_problem_solution_t<i_t, f_t> solve_lp(optimization_problem_t<i_t, f
     if (!run_presolve) { CUOPT_LOG_INFO("Third-party presolve is disabled, skipping"); }
 
     if (run_presolve) {
+      detail::sort_csr(problem);
```
aliceb-nv: Couldn't we just sort op_problem directly instead? To avoid the tuple parameter passing to apply()

cpp/src/mip/utilities/sort_csr.cuh
```diff
@@ -0,0 +1,66 @@
...
+                                      problem.variables.data(),
+                                      problem.variables.data(),
+                                      problem.coefficients.data(),
+                                      problem.coefficients.data(),
+                                      num_items,
+                                      num_segments,
+                                      problem.offsets.data(),
+                                      problem.offsets.data() + 1,
+                                      stream_view);
+  RAFT_CHECK_CUDA(stream_view);
```
aliceb-nv: Couldn't we just use 'cusparseXcsrsort'? Or is the setup too much of a pain compared to just using CUB

aliceb-nv (approved):
Thanks Hugo!

cpp/src/mip/presolve/third_party_presolve.cpp
```diff
@@ -393,14 +399,16 @@ void set_presolve_parameters(papilo::Presolve<f_t>& presolver,
 template <typename i_t, typename f_t>
 std::pair<optimization_problem_t<i_t, f_t>, bool> third_party_presolve_t<i_t, f_t>::apply(
   optimization_problem_t<i_t, f_t> const& op_problem,
+  std::tuple<rmm::device_uvector<i_t>&, rmm::device_uvector<i_t>&, rmm::device_uvector<f_t>&>
```
rg20: I think we can just sort the op_problem.   Or just figure out if it is already sorted or not, if it is not, you can set the flag to false for papilo

hlinsen: Papilo will sort if it is not. Yes we can look to figure out when it's sorted. If MPS parser guarantees sorting and model building does not we can insert a flag.

cpp/src/linear_programming/solve.cu
hlinsen: Thanks yes, for some reason I thought it was const. It's updated.

cpp/src/mip/utilities/sort_csr.cuh
hlinsen: In terms of line of code it shouldn't change much. Is there better load balancing and memory usage using cusparse?

aliceb-nv: Likely, but I really doubt it's critical. On second thought we might as well just stick with CUB for ease and clarity since it seems raft lacks the appropriate cusparse wrapper

PR #492, opened 2025-10-14
Reuse of the basis factorization from the parent node

nguidotti:
With this PR, the child node can reuse the basis factorisation from the parent, reducing the time required for solving the LP relaxation of the child. The solver recomputes the basis factorisation when the branch reaches the end. 
This is a follow-up of #383 and #473. 

Average (Primal) Gap over the MIPLIB2017 dataset: 
`main` (10f116b): `14.472358` with `224` feasible solutions
This PR: `~14.26` with `225` feasible solutions

On average, the `main` branch explored `1012412` nodes. This PR increases to `1850797`  nodes (i.e., `82%` more nodes) for the same time frame. Compared to #473, it explores `15%` more nodes.

This also fixes an incorrect report of an infeasible solution after a timeout and there are no nodes in the heap. 

## Checklist

- [x] I am familiar with the [Contributing Guidelines](https://github.com/NVIDIA/cuopt/blob/HEAD/CONTRIBUTING.md).
- Testing
   - [x] New or existing tests cover these changes
   - [ ] Added tests
   - [ ] Created an issue to follow-up
   - [ ] NA
- Documentation
   - [x] The documentation is up to date with these changes
   - [ ] Added new documentation
   - [ ] NA

## Summary by CodeRabbit

* **New Features**
  * Branch-and-bound now propagates advanced basis data between LP solves for more reliable re-use across nodes.

* **Improvements**
  * Wider adoption of row-oriented matrix handling for steadier solves.
  * Numerical fallback paths and bound updates are now basis-aware, improving stability on difficult instances.
  * Multithreaded exploration and logging reflect basis-aware execution.

* **Refactor**
  * Public solve and exploration interfaces updated; logging made atomic for safer concurrent runs.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

PR #564, opened 2025-11-04
Improve sorting for the diving queue via pseudocost estimate and randomness

nguidotti:
This PR improves the sorting and selection of the nodes for diving. More specifically, it sorts the nodes on the diving queue based on the pseudocost estimation of the objective value (see [1, 2]). Every time a thread consume a node from the diving queue, it has a `0.1` chance in selecting a random node. The randomness is generated via PCG (see [3], same as RAFT). This also opens the doors for implementing the hybrid best-estimate/best-first search for the main thread in the future. 

Some fixes for the incumbent in B&B is also included and was taken from #527.
This PR includes the changes from #492 and #473 and may only be merge after those PRs. 

Performance (MIPLIB2017):
`reuse-basis-factorization` (#492): `~14.26` with `225` feasible solutions
this PR: `~14.09` with `225` feasible solutions. Improves the consistency in solving the `cmflsp50-24-8-8` instance.

References:
[1] J. J. H. Forrest, J. P. H. Hirst, and J. A. Tomlin, “Practical Solution of Large Mixed Integer Programming Problems with Umpire,” Management Science, vol. 20, no. 5, pp. 736–773, Jan. 1974, doi: [10.1287/mnsc.20.5.736](https://doi.org/10.1287/mnsc.20.5.736).

[2] T. Achterberg, “Constraint Integer Programming,” PhD, Technischen Universität Berlin, Berlin, 2007. doi: [10.14279/depositonce-1634](https://doi.org/10.14279/depositonce-1634). (Section 6.4)

[3] https://www.pcg-random.org/index.html

## Checklist

- Testing
   - [ ] Added tests
   - [ ] NA
- Documentation
   - [ ] Added new documentation
   - [ ] NA

## Summary by CodeRabbit

* **New Features**
  * Integrated bounds strengthening to tighten LP relaxations during solving
  * New diving queue with randomized node selection for more diverse starts
  * Enhanced pseudo-cost objective estimation to improve branching choices

* **Optimization**
  * Refactored branch-and-bound exploration and threading for improved solver throughput
  * Adjusted default solver tuning (more frequent refactoring) for better performance

### DIFF
--- a/cpp/src/dual_simplex/branch_and_bound.cpp
+++ b/cpp/src/dual_simplex/branch_and_bound.cpp
@@ -192,11 +192,11 @@ std::string user_mip_gap(f_t obj_value, f_t lower_bound)
 {
   const f_t user_mip_gap = relative_gap(obj_value, lower_bound);
   if (user_mip_gap == std::numeric_limits<f_t>::infinity()) {
-    return "   -  ";
+    return "  -  ";
   } else {
     constexpr int BUFFER_LEN = 32;
     char buffer[BUFFER_LEN];
-    snprintf(buffer, BUFFER_LEN - 1, "%5.1f%%", user_mip_gap * 100);
+    snprintf(buffer, BUFFER_LEN - 1, "%4.1f%%", user_mip_gap * 100);
     return std::string(buffer);
   }
 }
@@ -204,9 +204,9 @@ std::string user_mip_gap(f_t obj_value, f_t lower_bound)
 inline const char* thread_type_symbol(thread_type_t type)
 {
   switch (type) {
-    case thread_type_t::EXPLORATION: return "B ";
-    case thread_type_t::DIVING: return "D ";
-    default: return "U ";
+    case thread_type_t::EXPLORATION: return "B";
+    case thread_type_t::DIVING: return "D";
+    default: return "U";
   }
 }
 
@@ -312,7 +312,7 @@ void branch_and_bound_t<i_t, f_t>::set_new_solution(const std::vector<f_t>& solu
       std::string gap = user_mip_gap<f_t>(user_obj, user_lower);
 
       settings_.log.printf(
-        "H                            %+13.6e    %+10.6e                        %s %9.2f\n",
+        "H                           %+13.6e    %+10.6e                        %s %9.2f\n",
         user_obj,
         user_lower,
         gap.c_str(),
@@ -425,7 +425,7 @@ void branch_and_bound_t<i_t, f_t>::repair_heuristic_solutions()
           std::string user_gap = user_mip_gap<f_t>(obj, lower);
 
           settings_.log.printf(
-            "H                            %+13.6e    %+10.6e                        %s %9.2f\n",
+            "H                           %+13.6e    %+10.6e                        %s %9.2f\n",
             obj,
             lower,
             user_gap.c_str(),
@@ -458,11 +458,6 @@ mip_status_t branch_and_bound_t<i_t, f_t>::set_final_solution(mip_solution_t<i_t
   if (status_ == mip_exploration_status_t::TIME_LIMIT) {
     settings_.log.printf("Time limit reached. Stopping the solver...\n");
     mip_status = mip_status_t::TIME_LIMIT;
-  }
-
-  if (status_ == mip_exploration_status_t::NODE_LIMIT) {
-    settings_.log.printf("Node limit reached. Stopping the solver...\n");
-    mip_status = mip_status_t::NODE_LIMIT;
   }
 
   f_t upper_bound      = get_upper_bound();
@@ -663,6 +658,7 @@ node_status_t branch_and_bound_t<i_t, f_t>::solve_node(mip_node_t<i_t, f_t>* nod
 
       lp_status = convert_lp_status_to_dual_status(second_status);
     }
+
     stats.total_lp_solve_time += toc(lp_start_time);
     stats.total_lp_iters += node_iter;
   }
@@ -721,7 +717,6 @@ node_status_t branch_and_bound_t<i_t, f_t>::solve_node(mip_node_t<i_t, f_t>* nod
       search_tree.update(node_ptr, node_status_t::FATHOMED);
       return node_status_t::FATHOMED;
     }
-
   } else if (lp_status == dual::status_t::TIME_LIMIT) {
     return node_status_t::TIME_LIMIT;
 
@@ -789,7 +784,7 @@ void branch_and_bound_t<i_t, f_t>::exploration_ramp_up(mip_node_t<i_t, f_t>* nod
       f_t user_lower       = compute_user_objective(original_lp_, root_objective_);
       std::string gap_user = user_mip_gap<f_t>(obj, user_lower);
 
-      settings_.log.printf("  %10d   %10lu    %+13.6e    %+10.6e   %6d   %7.1e     %s %9.2f\n",
+      settings_.log.printf(" %10d   %10lu    %+13.6e    %+10.6e   %6d   %7.1e     %s %9.2f\n",
                            nodes_explored,
                            nodes_unexplored,
                            obj,
@@ -910,7 +905,7 @@ void branch_and_bound_t<i_t, f_t>::explore_subtree(i_t task_id,
         f_t obj              = compute_user_objective(original_lp_, upper_bound);
         f_t user_lower       = compute_user_objective(original_lp_, get_lower_bound());
         std::string gap_user = user_mip_gap<f_t>(obj, user_lower);
-        settings_.log.printf("  %10d   %10lu    %+13.6e    %+10.6e   %6d   %7.1e     %s %9.2f\n",
+        settings_.log.printf(" %10d   %10lu    %+13.6e    %+10.6e   %6d   %7.1e     %s %9.2f\n",
                              nodes_explored,
                              nodes_unexplored,
                              obj,
@@ -1294,7 +1289,7 @@ mip_status_t branch_and_bound_t<i_t, f_t>::solve(mip_solution_t<i_t, f_t>& solut
                        settings_.num_diving_threads);
 
   settings_.log.printf(
-    "  | Explored | Unexplored |    Objective    |     Bound     | Depth | Iter/Node |   Gap    "
+    " | Explored | Unexplored |    Objective    |     Bound     | Depth | Iter/Node |   Gap    "
     "|  Time  |\n");
 
   stats_.nodes_explored       = 0;


cpp/src/dual_simplex/diving_queue.hpp
```diff
@@ -50,9 +53,11 @@ class diving_queue_t {
  private:
   std::vector<diving_root_t<i_t, f_t>> buffer;
   static constexpr i_t max_size_ = INT16_MAX;
+  PCG rng;
```
aliceb-nv, line 54: Let's use `simplex_solver_settings_t`'s `random_seed` to initialize the PCG

aliceb-nv: Additional silly question: Why did we not simply use raft's PCG? iirc it is `__host__` as well
Is it from a concern of separating external library code from the dual simplex codebase?

cpp/src/dual_simplex/diving_queue.hpp
```diff
@@ -50,9 +53,11 @@ class diving_queue_t {
...
+  const double epsilon = 0.1;  // Probability to grab a random node
 
  public:
-  diving_queue_t() { buffer.reserve(max_size_); }
```
aliceb-nv: Why did we remove that reserve call?

cpp/src/utilities/pcg.hpp
```diff
@@ -0,0 +1,165 @@
...
+
+  void next(uint32_t& ret) { ret = next_u32(); }
+  void next(uint64_t& ret) { ret = next_u64(); }
+  void next(int32_t& ret) { ret = next_i32(); }
+  void next(int64_t& ret) { ret = next_i64(); }
+  void next(float& ret) { ret = next_float(); }
+  void next(double& ret) { ret = next_double(); }
+
+  // Generate a random integer uniformly distributed in [low, high].
+  // FIXME: When C++20 is enabled, switch to `std::integer`.
```
aliceb-nv: Out of personal curiosity :) What's std::integer?

cpp/src/dual_simplex/pseudo_costs.cpp
```diff
@@ -263,63 +263,75 @@ i_t pseudo_costs_t<i_t, f_t>::variable_selection(const std::vector<i_t>& fractio
...
   return branch_var;
 }
 
+template <typename i_t, typename f_t>
+f_t pseudo_costs_t<i_t, f_t>::objective_estimate(const std::vector<i_t>& fractional,
+                                                 const std::vector<f_t>& solution,
+                                                 f_t lower_bound,
+                                                 logger_t& log)
+{
+  mutex.lock();
```
aliceb-nv: Let's use `std::lock_guard` for these locking scenarios

cpp/src/dual_simplex/branch_and_bound.cpp
```diff
@@ -1309,7 +1315,7 @@ mip_status_t branch_and_bound_t<i_t, f_t>::solve(mip_solution_t<i_t, f_t>& solut
 
       for (i_t i = 0; i < settings_.num_diving_threads; i++) {
 #pragma omp task
-        diving_thread(Arow);
+        diving_thread(5, Arow);
```
aliceb-nv: This should probably be stored as a constant somewhere

aliceb-nv (commented):
Thanks for the great work Nicolas!

cpp/src/utilities/pcg.hpp
nguidotti: It is actually `std::integral` instead of `std::integer` (sorry for the typo). C++20 introduced concepts that enforces a set of rules over the template parameters. For instance, `std::integral` (https://en.cppreference.com/w/cpp/concepts/integral.html) enforces that the class/method only accepts integer types:

```cpp
template <std::integral I>
void foo();

foo<int>(); // It is fine
// foo<double>(); // Compilation error
```

You can also define the concepts yourself to enforce how the template substitution should be done.

nguidotti: This is more elegant that the `std::enable_if` everywhere and should result in faster compilation times as the concepts are evaluated only once.

aliceb-nv: Thanks for the info, can't wait for concepts!
In the meantime, I think we can simply add a safeguard via `static_assert(std::is_integral_v<T>, "blah blah")`; that shouldn't affect compile times

cpp/src/dual_simplex/diving_queue.hpp
nguidotti: The `PCGenerator` header from raft includes CUDA code, such that the compilation will fail when called in a cpp file, like `branch_and_bound.cpp`

nguidotti: We can change the extension of the file, and `PCGenerator` should work in B&B.

aliceb-nv: Makes sense! We can leave things this way then, lets keep B&B as .cpp

cpp/src/dual_simplex/diving_queue.hpp
nguidotti: Ops, my mistake. I reverted this change.

cpp/src/utilities/pcg.hpp
nguidotti: This should also work for floating point numbers, so I am lifting the restriction.

cpp/src/utilities/pcg.hpp
```diff
@@ -0,0 +1,165 @@
...
+  template <typename i_t>
+  i_t uniform(i_t low, i_t high)
+  {
+    // Fractional scaling may exhibit slightly bias, but should be
+    // fine for our use case.
+    double val = next_double();
+    i_t dist   = high - low;
+    return low + static_cast<i_t>(val * dist);
+  }
```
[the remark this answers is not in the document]
nguidotti: Fixed the documentation. It should only generate a random number in a semi-closed interval.

cpp/src/dual_simplex/branch_and_bound.cpp
```diff
@@ -873,8 +958,12 @@ void branch_and_bound_t<i_t, f_t>::explore_subtree(i_t task_id,
...
         if (get_heap_size() > settings_.num_bfs_threads) {
+          std::vector<f_t> lower = original_lp_.lower;
+          std::vector<f_t> upper = original_lp_.upper;
+          node->get_variable_bounds(lower, upper, presolver.bounds_changed);
+
           mutex_dive_queue_.lock();
-          dive_queue_.emplace(node->detach_copy(), leaf_problem.lower, leaf_problem.upper);
+          diving_queue_.emplace(node->detach_copy(), std::move(lower), std::move(upper));
           mutex_dive_queue_.unlock();
         }
```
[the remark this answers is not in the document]
nguidotti: The bounds_changed will the reset in the next iteration anyway.

cpp/src/utilities/pcg.hpp
```diff
@@ -0,0 +1,155 @@
...
+  void next(float& ret) { ret = next_float(); }
+  void next(double& ret) { ret = next_double(); }
+
+  /// Draws a sample from a uniform distribution. The samples are uniformly distributed over
+  /// the semi-closed interval `[low, high)`. This routine may have a **slight bias** toward
+  /// some numbers in the range (scaling by floating-point).
+  template <typename T>
+  T uniform(T low, T high)
+  {
+    double val = next_double();
```
rg20, line 146:
```suggestion
    T val; next(val); 
```

cpp/src/utilities/pcg.hpp
```diff
@@ -0,0 +1,155 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#pragma once
+
+#include <stdint.h>
+
+// Copied from raft/PCGenerator (rng_device.cuh).
```
rg20, line 22: Why do we need to copy?  we can use raft directly right? 

Also, there is random generator within cuopt source code (https://github.com/NVIDIA/cuopt/blob/a7df1219fa8f96f1f749efb36fd073bd47090615/cpp/src/utilities/seed_generator.cuh)

cpp/src/dual_simplex/solution.hpp
```diff
@@ -79,6 +80,7 @@ class mip_solution_t {
   f_t lower_bound;
   i_t nodes_explored;
   i_t simplex_iterations;
+  bool has_incumbent;
```
rg20, line 87:
```suggestion
  bool has_incumbent{false};
```

cpp/src/utilities/pcg.hpp
nguidotti: The `PCGenerator` header from raft includes CUDA code, such that the compilation will fail when called in a cpp file, like `branch_and_bound.cpp`. We need to change to a `.cu` file to reuse the code from `raft`

nguidotti: `std::random_device` is non-deterministic because it uses the entropy from the hardware (according to cppreference). We need to use a pseudorandom generator here so we can set the seed and have a reproducible result.

nguidotti: Also PCG is a higher quality and faster random generator than the ones offer by the `STL`.

cpp/src/utilities/pcg.hpp
nguidotti: Even for integer, we need to generate a floating point number between 0 and 1 for this algorithm to work. See this article: [https://www.pcg-random.org/posts/bounded-rands.html](https://www.pcg-random.org/posts/bounded-rands.html)

nguidotti: It is slightly bias and do not cover the full 64-bit range, but should be enough for our purposes. In the article, its called FP Multiply.

cpp/src/dual_simplex/branch_and_bound.cpp
```diff
@@ -552,36 +574,54 @@ branch_and_bound_t<i_t, f_t>::child_selection(mip_node_t<i_t, f_t>* node_ptr)
...
   lp_settings.set_log(false);
   lp_settings.cut_off    = upper_bound + settings_.dual_tol;
   lp_settings.inside_mip = 2;
   lp_settings.time_limit = settings_.time_limit - toc(stats_.start_time);
 
-  // in B&B we only have equality constraints, leave it empty for default
-  std::vector<char> row_sense;
+  // Limit the number of simplex iterations when diving.
+  if (thread_type == thread_type_t::DIVING) {
+    lp_settings.iteration_limit = 0.05 * stats_.total_lp_iters - stats.total_lp_iters;
```
chris-maes, line 606: I'm having trouble understanding what iteration limit is being set here. There is a `stats_` and a `stats`. Maybe a comment would clarify this.

cpp/src/dual_simplex/branch_and_bound.cpp
```diff
@@ -590,40 +630,52 @@ node_status_t branch_and_bound_t<i_t, f_t>::solve_node(search_tree_t<i_t, f_t>&
...
+                                                                           leaf_vstatus,
+                                                                           leaf_edge_norms);
+
       lp_status = convert_lp_status_to_dual_status(second_status);
     }
 
-    stats_.total_lp_solve_time += toc(lp_start_time);
-    stats_.total_lp_iters += node_iter;
+    stats.total_lp_solve_time += toc(lp_start_time);
+    stats.total_lp_iters += node_iter;
```
chris-maes, line 663: What does the local stats represent?

cpp/src/dual_simplex/branch_and_bound.cpp
chris-maes: If you pass stats_ in for stats could this set a negative iteration limit?

cpp/src/dual_simplex/branch_and_bound.cpp
```diff
@@ -744,12 +801,29 @@ void branch_and_bound_t<i_t, f_t>::exploration_ramp_up(search_tree_t<i_t, f_t>*
...
+                                         leaf_problem,
+                                         basis_update,
+                                         basic_list,
+                                         nonbasic_list,
+                                         presolver,
+                                         thread_type_t::EXPLORATION,
+                                         true,
+                                         original_lp_.lower,
+                                         original_lp_.upper,
+                                         stats_,
```
chris-maes, line 825: Passing stats_ in for stats?

cpp/src/dual_simplex/branch_and_bound.cpp
```diff
@@ -944,73 +1052,92 @@ void branch_and_bound_t<i_t, f_t>::best_first_thread(i_t id,
...
 
     if (start_node.has_value()) {
       if (get_upper_bound() < start_node->node.lower_bound) { continue; }
 
+      bool recompute_bounds_and_basis = true;
       search_tree_t<i_t, f_t> subtree(std::move(start_node->node));
       std::deque<mip_node_t<i_t, f_t>*> stack;
       stack.push_front(&subtree.root);
 
+      stats_t lp_stats;
```
chris-maes, line 1091: I see so this tracking how much work we have done on the diving thread? Maybe we could call this `dive_stats` for clarity?

cpp/src/dual_simplex/branch_and_bound.cpp
```diff
@@ -1174,40 +1300,35 @@ mip_status_t branch_and_bound_t<i_t, f_t>::solve(mip_solution_t<i_t, f_t>& solut
   min_diving_queue_size_      = 4 * settings_.num_diving_threads;
   status_                     = mip_exploration_status_t::RUNNING;
   lower_bound_ceiling_        = inf;
+  diving_queue_.set_rng_seed(settings_.random_seed);
```
chris-maes, line 1303: Do you see an improvement by using randomness? If so, by how much?

cpp/src/dual_simplex/pseudo_costs.cpp
```diff
@@ -208,19 +208,18 @@ template <typename i_t, typename f_t>
 void pseudo_costs_t<i_t, f_t>::update_pseudo_costs(mip_node_t<i_t, f_t>* node_ptr,
                                                    f_t leaf_objective)
 {
-  mutex.lock();
+  std::lock_guard<omp_mutex_t> lock(mutex);
```
chris-maes, line 211: Why did we need to change the mutex here?

cpp/src/dual_simplex/pseudo_costs.cpp
chris-maes, line 259: There are a lot of unnecessary changes to this function. Can we revert back to the original function here? 

The logs have been removed which are useful for debugging. The variables have been renamed unnecessarily.

cpp/src/dual_simplex/pseudo_costs.cpp
```diff
@@ -261,65 +260,76 @@ i_t pseudo_costs_t<i_t, f_t>::variable_selection(const std::vector<i_t>& fractio
...
+  constexpr f_t eps = 1e-6;
+  f_t estimate      = lower_bound;
+
+  i_t num_initialized_down;
+  i_t num_initialized_up;
+  f_t pc_down_avg;
+  f_t pc_up_avg;
+  initialized(num_initialized_down, num_initialized_up, pc_down_avg, pc_up_avg);
+
+  for (auto j : fractional) {
```
chris-maes, line 318: It would be helpful to include a reference or even add the math and the intuition behind this estimate in the comments.

cpp/src/dual_simplex/pseudo_costs.cpp
```diff
@@ -261,65 +260,76 @@ i_t pseudo_costs_t<i_t, f_t>::variable_selection(const std::vector<i_t>& fractio
...
+    f_t f_down = solution[j] - std::floor(solution[j]);
+    f_t f_up   = std::ceil(solution[j]) - solution[j];
+
+    f_t pc_down = pseudo_cost_num_down[j] != 0 ? pseudo_cost_sum_down[j] / pseudo_cost_num_down[j]
+                                               : pc_down_avg;
+
+    f_t pc_up =
+      pseudo_cost_num_up[j] != 0 ? pseudo_cost_sum_up[j] / pseudo_cost_num_up[j] : pc_up_avg;
+    estimate += std::min(std::max(pc_down * f_down, eps), std::max(pc_up * f_up, eps));
```
chris-maes, line 327: This might be easier to read if you introduced some named variables 

```
const f_t change_in_obj_down = pc_down * f_down;
const f_t change_in_obj_up = pc_up * f_up;
```

chris-maes (commented):
Thanks for implementing this @nguidotti 

Please revert the changes to the pseudocost code before merging. 

Let's also get the other PRs merged so the diff of this one is smaller and easier to understand.

chris-maes (changes requested):
Thanks for implementing this @nguidotti 

Please revert the changes to the pseudocost code before merging. 

Let's also get the other PRs merged so the diff of this one is smaller and easier to understand.

cpp/src/dual_simplex/pseudo_costs.cpp
aliceb-nv: I made this suggestion to align with much of the rest of the codebase, which uses scoped locks to ensure mutexes are properly released at scope exit. Threading/deadlock errors are a nightmare to debug, so might as well ease our task a little by taking advantage of RAII :)

aliceb-nv: To be clear: there is no functional change. This is a code style/best practices suggestion

PR #570, opened 2025-11-06
Fix explored nodes counter

nguidotti:
This PR changes how the explored node counter is updated, such that it is only incremented after solving a node in the B&B.

## Checklist

- Testing
   - [ ] Added tests
   - [ ] NA
- Documentation
   - [ ] Added new documentation
   - [ ] NA

## Summary by CodeRabbit

* **Refactor**
  * Improved progress reporting cadence and coordination in the branch-and-bound solver.
  * More accurate and consistent node-counting in progress displays.
  * Reduced noisy/duplicate log output and ensured coordinated reporting from solver start.

### DIFF
--- a/cpp/src/dual_simplex/branch_and_bound.cpp
+++ b/cpp/src/dual_simplex/branch_and_bound.cpp
@@ -788,9 +788,7 @@ void branch_and_bound_t<i_t, f_t>::exploration_ramp_up(mip_node_t<i_t, f_t>* nod
   if (lower_bound > upper_bound || rel_gap < settings_.relative_mip_gap_tol) {
     search_tree->graphviz_node(settings_.log, node, "cutoff", node->lower_bound);
     search_tree->update(node, node_status_t::FATHOMED);
-    ++exploration_stats_.nodes_explored;
     --exploration_stats_.nodes_unexplored;
-    ++exploration_stats_.nodes_since_last_log;
     return;
   }
 
@@ -856,6 +854,10 @@ void branch_and_bound_t<i_t, f_t>::exploration_ramp_up(mip_node_t<i_t, f_t>* nod
                                         original_lp_.upper,
                                         settings_.log);
 
+  ++exploration_stats_.nodes_since_last_log;
+  ++exploration_stats_.nodes_explored;
+  --exploration_stats_.nodes_unexplored;
+
   if (status == node_solve_info_t::TIME_LIMIT) {
     solver_status_ = mip_exploration_status_t::TIME_LIMIT;
     return;
@@ -879,10 +881,6 @@ void branch_and_bound_t<i_t, f_t>::exploration_ramp_up(mip_node_t<i_t, f_t>* nod
       mutex_heap_.unlock();
     }
   }
-
-  ++exploration_stats_.nodes_explored;
-  --exploration_stats_.nodes_unexplored;
-  ++exploration_stats_.nodes_since_last_log;
 }
 
 template <typename i_t, typename f_t>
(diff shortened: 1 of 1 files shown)


cpp/src/dual_simplex/branch_and_bound.cpp
```diff
@@ -736,6 +735,10 @@ void branch_and_bound_t<i_t, f_t>::exploration_ramp_up(search_tree_t<i_t, f_t>*
                            nodes_explored > 0 ? stats_.total_lp_iters / nodes_explored : 0,
                            gap_user.c_str(),
                            now);
+
+      stats_.nodes_since_last_log = 0;
+      stats_.last_log             = tic();
+      should_report_              = true;
     }
```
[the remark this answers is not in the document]
nguidotti: While this is technically true, there are an if condition surrounding this piece of code that prevents more than one thread to access it.

cpp/src/dual_simplex/branch_and_bound.cpp
```diff
@@ -697,17 +697,17 @@ void branch_and_bound_t<i_t, f_t>::exploration_ramp_up(search_tree_t<i_t, f_t>*
...
-  stats_.nodes_since_last_log++;
+  f_t lower_bound = node->lower_bound;
+  f_t upper_bound = get_upper_bound();
+  f_t rel_gap     = user_relative_gap(original_lp_, upper_bound, lower_bound);
+  f_t abs_gap     = upper_bound - lower_bound;
 
   if (lower_bound > upper_bound || rel_gap < settings_.relative_mip_gap_tol) {
     search_tree->graphviz_node(settings_.log, node, "cutoff", node->lower_bound);
     search_tree->update_tree(node, node_status_t::FATHOMED);
+    ++stats_.nodes_explored;
```
chris-maes: This is a bit strange. I think of nodes explored as the number of nodes we have solved, or perhaps proved infeasible through node presolve. If we are fathoming the node here I would not count it as explored.

cpp/src/dual_simplex/branch_and_bound.cpp
```diff
@@ -697,17 +697,17 @@ void branch_and_bound_t<i_t, f_t>::exploration_ramp_up(search_tree_t<i_t, f_t>*
...
+    --stats_.nodes_unexplored;
+    ++stats_.nodes_since_last_log;
```
chris-maes: Same here. The purpose of nodes since last log is for us to print every so often in a deterministic way. Here we did no work, so I don't think we want to increase this counter.

cpp/src/dual_simplex/branch_and_bound.cpp
```diff
@@ -717,15 +717,14 @@ void branch_and_bound_t<i_t, f_t>::exploration_ramp_up(search_tree_t<i_t, f_t>*
   if (((stats_.nodes_since_last_log >= 10 || abs_gap < 10 * settings_.absolute_mip_gap_tol) &&
        (time_since_last_log >= 1)) ||
       (time_since_last_log > 30) || now > settings_.time_limit) {
-    // Check if no new node was explored until now. If this is the case,
-    // only the last thread should report the progress
-    if (stats_.nodes_explored.load() == nodes_explored) {
-      stats_.nodes_since_last_log = 0;
-      stats_.last_log             = tic();
+    bool should_report = should_report_.exchange(false);
```
chris-maes, line 803: Do we need to have member variable to decide if we should log? I think this is increasing the complexity of the code for little return.

cpp/src/dual_simplex/branch_and_bound.cpp
```diff
@@ -775,6 +778,10 @@ void branch_and_bound_t<i_t, f_t>::exploration_ramp_up(search_tree_t<i_t, f_t>*
       mutex_heap_.unlock();
     }
   }
+
+  ++stats_.nodes_explored;
```
chris-maes: Nit: is there a reason why all of these are ++x instead x++?

chris-maes: Why are we incrementing the nodes explored here at all?

cpp/src/dual_simplex/branch_and_bound.cpp
nguidotti: Since this piece of code is running by multiple threads, the atomic here prevent multiple reports from happening (potentially out-of-order). But I agree that it is an imperfect solution. I will probably change when the ramp-up phase is reworked or the report is moved to a separated thread.

PR #573, opened 2025-11-06
Use latest pre-commit-hooks, clean up copyright notices

KyleFromNVIDIA:
## Issue
https://github.com/rapidsai/build-infra/issues/297

## Checklist

- Testing
   - [ ] Added tests
   - [ ] NA
- Documentation
   - [ ] Added new documentation
   - [ ] NA

### DIFF
--- a/python/cuopt_server/utils/microservice_data_gen.py
+++ b/python/cuopt_server/utils/microservice_data_gen.py
@@ -1,4 +1,3 @@
-# noqa
 # SPDX-FileCopyrightText: Copyright (c) 2023-2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 
--- a/benchmarks/linear_programming/utils/get_datasets.py
+++ b/benchmarks/linear_programming/utils/get_datasets.py
@@ -1,4 +1,3 @@
-# noqa
 # SPDX-FileCopyrightText: Copyright (c) 2021-2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 
--- a/ci/utils/dli_nb_strip.py
+++ b/ci/utils/dli_nb_strip.py
@@ -1,4 +1,3 @@
-# noqa
 # SPDX-FileCopyrightText: Copyright (c) 2022-2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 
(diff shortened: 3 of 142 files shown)


python/cuopt_server/utils/microservice_data_gen.py
```diff
@@ -1,19 +1,6 @@
-# SPDX-FileCopyrightText: Copyright (c) 2023-2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.  # noqa
+# noqa
```
bdice: Does this `# noqa` do anything? I think the rule is that `# noqa` works on the current line, and `# ruff: noqa` ignores linter violations on the whole file. I think a line containing _only_ `# noqa` would do nothing. What happens if we delete this line?

KyleFromNVIDIA: A `# noqa` at the top causes linting to be disabled for the entire file. I don't know why it was here, but I kept it in to maintain the status quo.

KyleFromNVIDIA: Hmmm. I wonder if it was there to avoid triggering the line length check. Let me try removing it from all files.

KyleFromNVIDIA: Removed.

ramakrishnap-nv: Yeah, it was there to skip length check.

cpp/tests/utilities/cxxopts.hpp
```diff
@@ -1,3 +1,8 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2014-2022 Jarryd Beck
```
ramakrishnap-nv, line 2: I am not sure whether we need this or not.

rgsl888prabhu (approved):
Minor question, rest looks good.

cpp/tests/utilities/cxxopts.hpp
KyleFromNVIDIA: We do need it. Across RAPIDS, every file that has third-party copyright has been getting appropriate `SPDX-FileCopyrightText` and `SPDX-License-Identifier` headers so that anyone can build a full software BOM from the repository.

python/cuopt_server/utils/microservice_data_gen.py
KyleFromNVIDIA: I also see where I got the precedent to leave the `# noqa` in place rather than taking it out:

https://github.com/rapidsai/rapidsmpf/commit/7304b5f8e8e0050ef2db13f583d198d8e4b72396#diff-5ff2f7259cc0e4c4d0d99e9e1c7d11f6c188bd201927554e2b271b182848caebR1

But that one was a specific directive for an undocumented public module. Completely different purpose from this `# noqa` that suppresses the line length check.

PR #580, opened 2025-11-11
Add Probing Implications and Improve Variable Order

akifcorduk:
We were computing the probing cache but actually were not using the probing implications to change the problem. We were just using the cache during solve. In this PR, we add implications to the problem during presolve. When a probing a range is infeasible, we fix the variable bounds to other range. If both probing ranges imply some global bounds on other variables, we update those global bounds. Also if both probings of a binary variable imply some singletons on some other variable, we can substitute that variable with the current variable(with some coefficients).

This PR also changes the variable ordering to prioritize max possible implications: number of other integers that the variable shares constraint with. 

With this PR, we can add more variable substitutions since the function is ready now.

Overall there is a consistent increase in number of feasible on H100 system. Compared to baseline main over two batches of miplib runs: 226(average) compared to before: 223(average). Objective gap: 13.4% compared to 14.4%. Overall 3 feasible and 1% gap improvement.

## Summary by CodeRabbit

* **New Features**
  * Added variable substitution, variable-bound updates, and row-sorting capabilities for problem instances.

* **Bug Fixes**
  * Relaxed strict dual-residual checks that could trigger crashes.

* **Performance**
  * Enhanced presolve probing behavior and cache configuration for better presolve effectiveness.
  * Improved constraint-statistics gathering and faster handling of bound updates.

### DIFF
--- a/benchmarks/linear_programming/cuopt/run_mip.cpp
+++ b/benchmarks/linear_programming/cuopt/run_mip.cpp
@@ -1,6 +1,6 @@
 /* clang-format off */
 /*
- * SPDX-FileCopyrightText: Copyright (c) 2022-2025, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ * SPDX-FileCopyrightText: Copyright (c) 2022-2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
  * SPDX-License-Identifier: Apache-2.0
  */
 /* clang-format on */
@@ -231,10 +231,16 @@ int run_single_file(std::string file_path,
   }
   std::stringstream ss;
   int decimal_places = 2;
+  double mip_gap     = solution.get_mip_gap();
+  int is_optimal     = solution.get_termination_status() ==
+                       cuopt::linear_programming::mip_termination_status_t::Optimal
+                         ? 1
+                         : 0;
   ss << std::fixed << std::setprecision(decimal_places) << base_filename << "," << sol_found << ","
      << obj_val << "," << benchmark_info.objective_of_initial_population << ","
      << benchmark_info.last_improvement_of_best_feasible << ","
-     << benchmark_info.last_improvement_after_recombination << "\n";
+     << benchmark_info.last_improvement_after_recombination << "," << mip_gap << "," << is_optimal
+     << "\n";
   write_to_output_file(out_dir, base_filename, device, n_gpus, batch_id, ss.str());
   CUOPT_LOG_INFO("Results written to the file %s", base_filename.c_str());
   return sol_found;


cpp/src/mip/problem/problem.cu
```diff
@@ -1085,6 +1095,168 @@ void problem_t<i_t, f_t>::set_implied_integers(const std::vector<i_t>& implied_i
...
+  presolve_data.objective_offset += objective_offset_value;
+  const i_t num_segments = n_constraints;
+  f_t initial_value{0.};
+
+  auto input_transform_it = thrust::make_transform_iterator(
+    thrust::make_counting_iterator(0),
+    [coefficients           = make_span(coefficients),
+     variables              = make_span(variables),
+     variable_fix_mask      = make_span(fixing_helpers.variable_fix_mask),
+     substitute_coefficient = make_span(d_coefficient_values),
```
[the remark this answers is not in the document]
akifcorduk: I handled your review, can you check again?

chris-maes:
@akifcorduk should this PR go into the 25.12 release?

cpp/src/mip/problem/problem.cu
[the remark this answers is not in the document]
akifcorduk: Can you check again if this issue is fully resolved?

cpp/src/mip/problem/problem.cu
```diff
@@ -1220,6 +1236,185 @@ void problem_t<i_t, f_t>::set_implied_integers(const std::vector<i_t>& implied_i
...
+  return (end_iter == intersection.begin());  // true if no overlap
+}
+
+// note that this only substitutes the variables, for problem modification trivial_presolve needs to
+// be called.
+// note that, this function assumes var_indices and var_to_substitude_indices don't contain any
+// common indices
+template <typename i_t, typename f_t>
+void problem_t<i_t, f_t>::substitute_variables(const std::vector<i_t>& var_indices,
+                                               const std::vector<i_t>& var_to_substitude_indices,
```
aliceb-nv: harmless typo - "var_to_substituTe_indices"

cpp/CMakeLists.txt
aliceb-nv, line 28: You can actually just pass `-DCMAKE_CUDA_ARCHITECTURES='90-real'` for example to the `--cmake-args` arg of ./build.sh; cmake takes the last specified value on the cmdline for its define so it will override the default

cpp/src/mip/presolve/probing_cache.cu
```diff
@@ -360,17 +363,25 @@ void compute_cache_for_var(i_t var_idx,
...
+  std::vector<f_t> h_improved_lower_bounds_0(h_var_bounds.size());
+  std::vector<f_t> h_improved_upper_bounds_0(h_var_bounds.size());
+  std::vector<f_t> h_improved_lower_bounds_1(h_var_bounds.size());
+  std::vector<f_t> h_improved_upper_bounds_1(h_var_bounds.size());
   std::pair<val_interval_t<i_t, f_t>, val_interval_t<i_t, f_t>> probe_vals;
   auto bounds = h_var_bounds[var_idx];
   f_t lb      = get_lower(bounds);
   f_t ub      = get_upper(bounds);
+  // note that is_binary does not always mean the bound difference is one
+  bool is_binary = ub == 1 && lb == 0;
```
aliceb-nv, line 384: We might want to add a canonization pass to the trivial_presolve to turn any integer with [lb, lb+1] bounds into [0, 1]; FJ benefits greatly from recognizing binary variables
Papilo might already do that!

cpp/src/mip/presolve/probing_cache.cu
```diff
@@ -423,12 +434,46 @@ void compute_cache_for_var(i_t var_idx,
...
+      problem_is_infeasible.store(true);
+      return;
+    }
+    i_t infeas_constraints_count  = i == 0 ? multi_probe_presolve.infeas_constraints_count_0
+                                           : multi_probe_presolve.infeas_constraints_count_1;
+    const auto& probe_val         = i == 0 ? probe_vals.first : probe_vals.second;
+    auto& h_improved_lower_bounds = i == 0 ? h_improved_lower_bounds_0 : h_improved_lower_bounds_1;
+    auto& h_improved_upper_bounds = i == 0 ? h_improved_upper_bounds_0 : h_improved_upper_bounds_1;
+    if (infeas_constraints_count > 0) {
+      CUOPT_LOG_DEBUG("Var %d is infeasible for probe %d on value %f. Fixing other interval",
```
aliceb-nv: should these be _LOG_TRACE() maybe? I fear this is going to clutter the logging output

cpp/src/mip/presolve/probing_cache.cu
```diff
@@ -449,21 +494,364 @@ void compute_cache_for_var(i_t var_idx,
...
+  }
+
+  if (!var_indices.empty()) {
+    problem.substitute_variables(
+      var_indices, substituting_var_indices, offset_values, coefficient_values);
+  }
+}
+
+template <typename i_t, typename f_t>
+std::vector<i_t> compute_priority_indices_by_implied_integers(problem_t<i_t, f_t>& problem)
```
aliceb-nv, line 739: What does the "implied integers" refer to here?

cpp/src/mip/presolve/probing_cache.cu
```diff
@@ -449,21 +494,364 @@ void compute_cache_for_var(i_t var_idx,
...
+
+template <typename i_t, typename f_t>
+bool compute_probing_cache(bound_presolve_t<i_t, f_t>& bound_presolve,
                            problem_t<i_t, f_t>& problem,
                            timer_t timer)
 {
   raft::common::nvtx::range fun_scope("compute_probing_cache");
   // we dont want to compute the probing cache for all variables for time and computation resources
-  auto priority_indices = compute_prioritized_integer_indices(bound_presolve, problem);
+  // auto priority_indices = compute_prioritized_integer_indices(bound_presolve, problem);
```
aliceb-nv: dead code

cpp/src/mip/presolve/probing_cache.cu
```diff
@@ -475,51 +863,83 @@ void compute_probing_cache(bound_presolve_t<i_t, f_t>& bound_presolve,
...
   std::vector<multi_probe_t<i_t, f_t>> multi_probe_presolve_pool;
+  std::vector<std::vector<std::tuple<f_t, i_t, f_t, f_t>>> modification_vector_pool(max_threads);
+  std::vector<std::vector<substitution_t<i_t, f_t>>> substitution_vector_pool(max_threads);
 
   // Initialize multi_probe_presolve_pool
   for (size_t i = 0; i < max_threads; i++) {
     multi_probe_presolve_pool.emplace_back(bound_presolve.context);
     multi_probe_presolve_pool[i].resize(problem);
-    multi_probe_presolve_pool[i].compute_stats = false;
+    multi_probe_presolve_pool[i].compute_stats = true;
```
aliceb-nv, line 872: Maybe we should gate this behind a define or something, as to not compute stats on release runs

cpp/src/mip/problem/problem.cu
```diff
@@ -823,6 +823,22 @@ void problem_t<i_t, f_t>::post_process_assignment(rmm::device_uvector<f_t>& curr
...
   }
+
+  // Apply variable substitutions from probing: x_substituted = offset + coefficient *
+  // x_substituting
+  for (const auto& sub : presolve_data.variable_substitutions) {
+    cuopt_assert(sub.substituted_var < (i_t)h_assignment.size(), "substituted_var out of bounds");
+    cuopt_assert(sub.substituting_var < (i_t)h_assignment.size(), "substituting_var out of bounds");
+    h_assignment[sub.substituted_var] =
+      sub.offset + sub.coefficient * h_assignment[sub.substituting_var];
+    CUOPT_LOG_DEBUG("Post-process substitution: x[%d] = %f + %f * x[%d] = %f",
```
aliceb-nv, line 834: _TRACE here as well, I think

cpp/src/mip/problem/problem.cu
```diff
@@ -1220,6 +1236,185 @@ void problem_t<i_t, f_t>::set_implied_integers(const std::vector<i_t>& implied_i
...
+                    objective_offset          = objective_offset.data()] __device__(i_t idx) {
+                     i_t var_idx                = var_indices[idx];
+                     i_t substituting_var_idx   = var_to_substitude_indices[idx];
+                     variable_fix_mask[var_idx] = idx;
+                     f_t objective_offset_difference =
+                       objective_coefficients[var_idx] * substitute_offset[idx];
+                     atomicAdd(objective_offset, objective_offset_difference);
+                     atomicAdd(&objective_coefficients[substituting_var_idx],
+                               objective_coefficients[var_idx] * substitute_coefficient[idx]);
+                   });
```
aliceb-nv: Maybe this could be a for_each followed by a reduce on a vector of the objective offsets - the atomicAdd on the offset scalar may cause a lot of L2 pressure

aliceb-nv (approved):
Awesome work Akif :) Very good idea to add some determinism future-proofing
My comments are minor, but I think it could be beneficial as well to add nvtx ranges to these functions for ease of profiling, as probing can take a substantial amount of runtime
LGTM!

cpp/CMakeLists.txt
akifcorduk: oh that's great pointer. i was going to change the whole build script to do that. i was going to remove cmake changes anyway.

cpp/src/mip/presolve/probing_cache.cu
akifcorduk: Let me test on few instances to see if papilo actually does that. Thanks for the pointer.

cpp/src/mip/presolve/probing_cache.cu
akifcorduk: Sounds good!

cpp/src/mip/presolve/probing_cache.cu
akifcorduk: The implied integer means that a change in ones variable bounds might imply whole integer bound change on other variables. We prioritize variables that are neighbors to more integer variables. Probing on variables that are neighbors with mosly continuous variables is not super useful.

cpp/src/mip/presolve/probing_cache.cu
akifcorduk: The stats here is probably a wrong naming. It actually computes the activities and number of infeasible constraints after a propagation. So this is needed in release.

cpp/src/mip/problem/problem.cu
akifcorduk: Sounds good. Actually, keeping the determinism in mind, we shouldn't even do that for the objective_coefficients. But the amount of memory for that will probably explode. I could keep a sparse matrix that we later sort with the var indices. But for now i will just keep a note there for the objective_coefficients, if you are okay.

aliceb-nv: Sounds good!

cpp/src/mip/presolve/probing_cache.cu
aliceb-nv: Gotcha! I got confused at first because Papilo also produces a list of "implied integers" - in the sense of continuous variables that are proven to only take integer values in feasible solutions

PR #587, opened 2025-11-13
WIP: remote execution

tmckayus:
This change provides  remote execution from solve_lp and solve_mip, thereby available from all interfaces and transparent to the user.

The remote mode is triggered by setting CUOPT_REMOTE_HOST and CUOPT_REMOTE_PORT env vars currently. Later,  API calls will be added that allow the mode to be specified through the API.

The essential change here is to make all problem input and solution output use CPU memory exclusively outside of solve_lp and solve_mip. This alllows efficient serialization and deserialziation to and from the server, and also will allow cuopt to be run on a host without a GPU for remote solves.

Still to do: enhancing the server implementation to provide job queues, job abort, etc.

### DIFF
--- a/cpp/src/linear_programming/utilities/remote_solve.cu
+++ b/cpp/src/linear_programming/utilities/remote_solve.cu
@@ -478,6 +478,13 @@ bool is_remote_solve_enabled(const char** host, const char** port)
   return (*host != nullptr && *port != nullptr);
 }
 
+// Check if sync mode is enabled (default is async)
+static bool use_sync_mode()
+{
+  const char* sync_env = std::getenv("CUOPT_REMOTE_USE_SYNC");
+  return (sync_env != nullptr && std::string(sync_env) == "1");
+}
+
 // Helper: Create and connect socket
 static int connect_to_server(const char* host, const char* port)
 {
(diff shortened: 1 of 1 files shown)


cpp/CMakeLists.txt
```diff
@@ -208,6 +208,27 @@ create_logger_macros(CUOPT "cuopt::default_logger()" include/cuopt)
 
 find_package(CUDSS REQUIRED)
 
+# Find Protocol Buffers for remote solve support
+find_package(Protobuf REQUIRED)
```
ramakrishnap-nv, line 212: We may need to add protobuf requirement to both cli and library in dependecies.yaml, conda/receipe/libcuopt

cpp/cuopt_remote_server.cpp
```diff
@@ -0,0 +1,457 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2024-2025 NVIDIA CORPORATION & AFFILIATES. All rights
```
ramakrishnap-nv, line 2:
```suggestion
 * SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights
```

cpp/cuopt_remote_server.cpp
```diff
@@ -0,0 +1,457 @@
...
+ * reserved. SPDX-License-Identifier: Apache-2.0
```
ramakrishnap-nv, line 3: You may want to refer to this latest PR on adding new shortened licensed header. https://github.com/NVIDIA/cuopt/pull/573/files

rgsl888prabhu (commented):
Is there an approach on how to make a client backward compatible so new client will not break  with any old server and any other such scenarios ?

PR #601, opened 2025-11-18
Add Multi-GPU concurrent LP solve

hlinsen:
This PR enable solving PDLP and Barrier on multiple GPUs

## Summary by CodeRabbit

* **New Features**
  * Multi-GPU batch-mode support for LP (up to 2 GPUs); solvers can run in parallel on separate devices
  * Exposed num_gpus parameter across CLI, solver settings, Python API, benchmarks, and server config

* **Bug Fixes / Reliability**
  * Per-device initialization and memory setup, with multi-GPU runtime checks and clearer logging

* **Documentation**
  * Added "Multi-GPU Mode" guide

* **Chores / Tests**
  * Test CLI help assertion updated; build link dependencies adjusted (TBB)

### DIFF
--- a/conda/recipes/libcuopt/recipe.yaml
+++ b/conda/recipes/libcuopt/recipe.yaml
@@ -9,6 +9,7 @@ context:
   cuda_major: '${{ (env.get("RAPIDS_CUDA_VERSION") | split("."))[0] }}'
   date_string: '${{ env.get("RAPIDS_DATE_STRING") }}'
   head_rev: '${{ git.head_rev(".")[:8] }}'
+  build_type: '${{ env.get("RAPIDS_BUILD_TYPE") }}'
 
 recipe:
   name: libcuopt-split
@@ -28,12 +29,14 @@ cache:
         export CXXFLAGS=$(echo $CXXFLAGS | sed -E 's@\-fdebug\-prefix\-map[^ ]*@@g')
         set +x
 
-        ./build.sh -n -v -a libmps_parser libcuopt deb --allgpuarch --cmake-args=\"-DCMAKE_INSTALL_LIBDIR=lib\"
+        ./build.sh -n -v ${BUILD_EXTRA_FLAGS} libmps_parser libcuopt deb --allgpuarch --cmake-args=\"-DCMAKE_INSTALL_LIBDIR=lib\"
       secrets:
         - AWS_ACCESS_KEY_ID
         - AWS_SECRET_ACCESS_KEY
         - AWS_SESSION_TOKEN
       env:
+        # Enable assertions (-a flag) for PR builds, but not for nightly or release builds
+        BUILD_EXTRA_FLAGS: '${{ "-a" if build_type == "pull-request" else "" }}'
         CMAKE_C_COMPILER_LAUNCHER: ${{ env.get("CMAKE_C_COMPILER_LAUNCHER") }}
         CMAKE_CUDA_COMPILER_LAUNCHER: ${{ env.get("CMAKE_CUDA_COMPILER_LAUNCHER") }}
         CMAKE_CXX_COMPILER_LAUNCHER: ${{ env.get("CMAKE_CXX_COMPILER_LAUNCHER") }}
--- a/docs/cuopt/source/conf.py
+++ b/docs/cuopt/source/conf.py
@@ -245,6 +245,11 @@ texinfo_documents = [
 numpydoc_show_inherited_class_members = {
     # option_context inherits undocumented members from the parent class
     "cuopt.option_context": False,
+    # VType and CType inherit from str, sense inherits from int
+    # Hide inherited methods to avoid autodoc signature errors
+    "cuopt.linear_programming.problem.VType": False,
+    "cuopt.linear_programming.problem.CType": False,
+    "cuopt.linear_programming.problem.sense": False,
 }
 
 # Rely on toctrees generated from autosummary on each of the pages we define
@@ -269,6 +274,7 @@ nitpick_ignore = [
         "cuopt.distance_engine.waypoint_matrix_wrapper.WaypointMatrix",
     ),
     ("py:class", "enum.Enum"),
+    ("py:class", "object)"),  # malformed type annotation reference
     ("py:obj", "cuopt.routing.DataModel.add_order_precedence"),
     ("py:obj", "cuopt_sh_client.SolverMethod.denominator"),
     ("py:obj", "cuopt_sh_client.SolverMethod.imag"),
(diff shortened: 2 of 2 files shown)


cpp/include/cuopt/linear_programming/constants.h
```diff
@@ -49,6 +49,7 @@
 #define CUOPT_CUDSS_DETERMINISTIC         "cudss_deterministic"
 #define CUOPT_PRESOLVE                    "presolve"
 #define CUOPT_DUAL_POSTSOLVE              "dual_postsolve"
+#define CUOPT_MULTI_GPU                   "multi_gpu"
```
rg20:
```suggestion
#define CUOPT_NUM_GPUS                   "num_gpus"
```

This gives control for users.

cpp/src/linear_programming/solve.cu
```diff
@@ -679,25 +688,46 @@ optimization_problem_solution_t<i_t, f_t> run_concurrent(
...
-                             std::ref(barrier_problem),
-                             std::ref(settings_pdlp),
-                             std::ref(sol_barrier_ptr),
-                             std::ref(timer));
+  auto barrier_thread = std::thread([&]() {
+    if (settings.multi_gpu) {
+      raft::device_setter device_setter(1);  // Scoped variable
+      CUOPT_LOG_DEBUG("Barrier device: %d", device_setter.get_current_device());
+
+      rmm::cuda_stream_view barrier_stream = rmm::cuda_stream_per_thread;
```
rg20, line 698: avoid duplicating code. You can use lambda and call it here.

cpp/include/cuopt/linear_programming/constants.h
rg20: for using all gpus, we can ask users to provide` --num-gpus -1 `or something like that

cpp/src/dual_simplex/simplex_solver_settings.hpp
```diff
@@ -120,6 +121,7 @@ struct simplex_solver_settings_t {
   bool cudss_deterministic;   // true to use cuDSS deterministic mode, false for non-deterministic
   bool barrier;               // true to use barrier method, false to use dual simplex method
   bool eliminate_dense_columns;  // true to eliminate dense columns from A*D*A^T
+  int num_gpus;                  // Number of GPUs to use (1 for single-GPU, 2 for multi-GPU)
```
rg20:
```suggestion
  int num_gpus;                  // Number of GPUs to use ( maximum of 2 gpus are supported at the moment)
```

cpp/src/linear_programming/solve.cu
```diff
@@ -679,25 +689,41 @@ optimization_problem_solution_t<i_t, f_t> run_concurrent(
...
+      barrier_problem.handle_ptr           = &barrier_handle;
+
+      run_barrier_thread<i_t, f_t>(std::ref(barrier_problem),
+                                   std::ref(settings_pdlp),
+                                   std::ref(sol_barrier_ptr),
+                                   std::ref(timer));
+    };
 
+    if (settings.num_gpus > 1) {
+      raft::device_setter device_setter(1);  // Scoped variable
```
rg20, line 711: We need to formalize this better, but thats for later.  Right now we are hard-coding to use the second gpu for barrier.

cpp/cuopt_cli.cpp
```diff
@@ -333,7 +334,13 @@ int main(int argc, char* argv[])
   const auto initial_solution_file = program.get<std::string>("--initial-solution");
   const auto solve_relaxation      = program.get<bool>("--relaxation");
 
-  auto memory_resource = make_async();
-  rmm::mr::set_current_device_resource(memory_resource.get());
+  std::vector<std::shared_ptr<rmm::mr::device_memory_resource>> memory_resources;
+  for (int i = 0; i < raft::device_setter::get_device_count(); ++i) {
```
rg20: Is this looping over all the devices?   

Maybe we should --gpus flag to specify which gpus to use?

hlinsen: This is looping over all the devices yes. Shouldn't be harmful but let's loop up to 2 for now.
I think the user should specify the gpus with `CUDA_VISIBLE_DEVICES` and by default it's 0 and 1. The same way it defaults to GPU 0 for a single gpu program.

rg20 (approved):

docs/cuopt/source/lp-features.rst
```diff
@@ -132,3 +132,8 @@ Batch Mode
 ----------
 
 Users can submit a set of problems which will be solved in a batch. Problems will be solved at the same time in parallel to fully utilize the GPU. Checkout :ref:`self-hosted client <generic-example-with-normal-and-batch-mode>` example in thin client.
+
+Multi-GPU Mode
+--------------
+
+Users can use multiple GPUs to solve a problem by specifying the ``num_gpus`` parameter. The feature is restricted to LP problems that uses concurrent mode and supports up to 2 GPUs at the moment. Using this mode fill run PDLP and IPM in parallel on different GPUs to avoid sharing single GPU resources.
```
chris-maes: FILL -> WILL

benchmarks/linear_programming/run_mps_files.sh
```diff
@@ -74,6 +80,9 @@ Examples:
     # Run all MPS files in /data/lp using 2 GPUs with 1 hour time limit
     $(basename "$0") --path /data/lp --ngpus 2 --time-limit 3600
 
+    # Run with multi-GPU mode using 4 GPUs (2 GPUs per instance, 2 parallel instances)
+    $(basename "$0") --path /data/lp --ngpus 4 --gpus-per-instance 2 --time-limit 3600
```
akifcorduk, line 84: Does ngpus need to be multiple of gpus-per-instance?

benchmarks/linear_programming/run_mps_files.sh
```diff
@@ -198,7 +221,48 @@ else
...
+    # Single GPU per instance
+    echo "Using single GPU per instance..."
+    GPU_GROUPS=("${GPU_LIST[@]}")
+fi
+
+GPU_COUNT=${#GPU_GROUPS[@]}
+echo "Total GPU groups created: $GPU_COUNT"
+
+# Print configuration information
+echo "=========================================="
```
akifcorduk, line 258: I think we should check if GPU_COUNT is multiple of GPUS_PER_INSTANCE. And if there are enough GPU_COUNT gpus (from nvidia-smi).

cpp/src/dual_simplex/sparse_cholesky.cuh
```diff
@@ -348,7 +363,7 @@ class sparse_cholesky_cudss_t : public sparse_cholesky_base_t<i_t, f_t> {
     CUDSS_CALL_AND_CHECK_EXIT(cudssDestroy(handle), status, "cudssDestroy");
     CUDA_CALL_AND_CHECK_EXIT(cudaStreamSynchronize(stream), "cudaStreamSynchronize");
 #if CUDART_VERSION >= 13000
-    if (settings_.concurrent_halt != nullptr) {
+    if (settings_.concurrent_halt != nullptr && settings_.num_gpus == 1) {
       auto cuStreamDestroy_func = cuopt::detail::get_driver_entry_point("cuStreamDestroy");
```
akifcorduk, line 367: I know it is not part of this PR but can't we put this create/destroy logic into RAII struct?

cpp/src/linear_programming/solve.cu
akifcorduk, line 711: I think we can synchronize the stream before changing the device. Probably there are some sync in functions, but this way might be safer.

cpp/src/dual_simplex/sparse_cholesky.cuh
hlinsen: Yes I agree. Even getting the driver entry point can be wrapped. I think they do it in cccl.

benchmarks/linear_programming/run_mps_files.sh
hlinsen: I added a check for gpu count and for gpus_per_instance I am assuming a max of 2 gpus. I am now returning a hard error instead of a warning: https://github.com/NVIDIA/cuopt/pull/601/commits/b9d6916a2e92964e16c3041464834199ba49aa6a

PR #608, opened 2025-11-19
Fix nightly build to build for both main and release and one shouldn't cancel other

rgsl888prabhu:
## Description
nightly workflow triggers build.yaml for both main and release/25.12 branches, they both share the same concurrency group. because of which one would cancel another as it thinks the new one is latest rather than different.

Also it was passing git hub ref for main instead of release, which also causes other issue.

Both have been fixed.

## Checklist

- Testing
   - [ ] Added tests
   - [x] NA
- Documentation
   - [ ] Added new documentation
   - [x] NA

## Summary by CodeRabbit

* **Chores**
  * Updated internal build system concurrency handling and workflow orchestration configurations.
  * Enhanced dependency tracking to include test support components.
  * Updated test framework dependencies to compatible versions.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -38,7 +38,7 @@ on:
         default: false
 
 concurrency:
-  group: ${{ github.workflow }}-${{ github.ref }}-${{ inputs.branch || github.ref_name }}
+  group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
 
 jobs:


.github/workflows/build.yaml
```diff
@@ -38,7 +38,7 @@ on:
         default: false
 
 concurrency:
-  group: ${{ github.workflow }}-${{ github.ref }}
+  group: ${{ github.workflow }}-${{ github.ref }}-${{ inputs.branch || github.ref_name }}
```
bdice: Doesn't `github.ref` already make this unique?

bdice: We have been using `${{ github.workflow }}-${{ github.ref }}-${{ github.event_name }}` in other RAPIDS repositories. Is that sufficient?

ramakrishnap-nv: May be with nightly update, this change might not be required.

PR #623, opened 2025-11-24
Add message when optimal solution is found during presolve

adityasinghz:
## Description

Added an explicit log message "Optimal solution found during presolve" that is displayed when the problem is fully reduced during presolve (i.e., when the problem is reduced to 0 constraints, 0 variables, and 0 nonzeros). This provides clear feedback to users that the optimal solution was discovered during presolve rather than through the main solving process.

**Changes Made:**
- `cpp/src/mip/solve.cu`: Added log message when `problem.empty` is true before running the MIP solver
- `cpp/src/mip/solver.cu`: Added log message in two locations:
  - When `context.problem_ptr->empty` is true before running presolve
  - When `context.problem_ptr->empty` becomes true after running presolve

**Example Output:**
Before this change, when the optimal solution was found during presolve, users would see:
```
Presolve status: reduced the problem
Presolve removed: 100 constraints, 200 variables, 560 nonzeros
Presolved problem: 0 constraints, 0 variables, 0 nonzeros
Best feasible: 37.000000
```

After this change, users will now see:
```
Presolve status: reduced the problem
Presolve removed: 100 constraints, 200 variables, 560 nonzeros
Presolved problem: 0 constraints, 0 variables, 0 nonzeros
Optimal solution found during presolve
Problem fully reduced in presolve
Best feasible: 37.000000
```

## Checklist

- Testing
   - [ ] Added tests
   - [ ] NA

- Documentation
   - [ ] Added new documentation
   - [ ] NA

## Summary by CodeRabbit

* **Bug Fixes**
  * When preprocessing fully eliminates a problem (no rows or columns remain), the solver now clearly reports that an optimal solution was found during presolve.
  * Improved presolve logging for clearer interpretation of outcomes and solver status during initial processing.

### DIFF
--- a/.github/workflows/nightly.yaml
+++ b/.github/workflows/nightly.yaml
@@ -1,4 +1,4 @@
-# SPDX-FileCopyrightText: Copyright (c) 2025, NVIDIA CORPORATION. All rights reserved.
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026, NVIDIA CORPORATION. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 
 name: Trigger Nightly cuOpt Pipeline
@@ -17,7 +17,7 @@ jobs:
       matrix:
         cuopt_branch:
           - "main"
-          - "release/25.12"
+          - "release/26.02"
     steps:
       - uses: actions/checkout@v4
       - name: Trigger Pipeline
--- a/build.sh
+++ b/build.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-# SPDX-FileCopyrightText: Copyright (c) 2021-2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-FileCopyrightText: Copyright (c) 2021-2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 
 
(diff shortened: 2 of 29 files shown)


cpp/src/mip/solve.cu
```diff
@@ -70,6 +70,7 @@ mip_solution_t<i_t, f_t> run_mip(detail::problem_t<i_t, f_t>& problem,
   pdlp_hyper_params::update_step_size_on_initial_solution     = true;
   // if the input problem is empty: early exit
   if (problem.empty) {
+    CUOPT_LOG_INFO("Optimal solution found during presolve");
```
rg20: This is the wrong place for this message. Problem could be empty for different reason as well.   

It makes sense to move this to presolve logic.

cpp/src/mip/solver.cu
```diff
@@ -100,6 +100,7 @@ solution_t<i_t, f_t> mip_solver_t<i_t, f_t>::run_solver()
                 "preprocess_problem should be called before running the solver");
 
   if (context.problem_ptr->empty) {
+    CUOPT_LOG_INFO("Optimal solution found during presolve");
```
rg20: same comment as above

cpp/src/mip/solver.cu
```diff
@@ -118,6 +119,7 @@ solution_t<i_t, f_t> mip_solver_t<i_t, f_t>::run_solver()
     return sol;
   }
   if (context.problem_ptr->empty) {
+    CUOPT_LOG_INFO("Optimal solution found during presolve");
```
rg20: same comment as above

cpp/src/mip/solve.cu
adityasinghz: Thanks for the feedback! I've moved the "Optimal solution found during presolve" message to the presolve function itself (`third_party_presolve.cpp`). 

The message now appears in `third_party_presolve_t::apply()` right after checking the presolved problem dimensions, where we can accurately detect when presolve finds the optimal solution (i.e., when the problem is reduced to 0 constraints and 0 variables).

I've removed the message from the generic empty problem checks in:
- `cpp/src/mip/solve.cu` (run_mip function)
- `cpp/src/mip/solver.cu` (run_solver function - both locations)

This ensures the message only appears when presolve actually found the solution, not when the problem is empty for other reasons.

cpp/src/mip/solver.cu
adityasinghz: Moved

cpp/src/mip/solver.cu
adityasinghz: Moved

adityasinghz:
Hi @rgsl888prabhu,

Thank you for the question. This is a user experience improvement (adding a log message when optimal solution is found during presolve), so I'm flexible on which release it targets.

**Option 1: If this should go into 25.12:**
I can rebase this onto `release/25.12` and create a new PR targeting that branch. Should I proceed with this?

**Option 2: If this can wait for 26.02:**
I can keep the current PR targeting `main` (which will be part of 26.02).
Could you please let me know which release you'd prefer? I'm happy to adjust accordingly.

Thanks!

PR #625, opened 2025-11-24
Fix lower bound calculation during ramp-up phase in parallel B&B

adityasinghz:
## Description

Fixed the lower bound calculation during the ramp-up phase in parallel Branch and Bound. Previously, cuOpt was using the root objective as the global lower bound when the heap was empty, which is a pessimistic estimation that doesn't reflect the actual lower bounds calculated by active threads during ramp-up.

**Changes Made:**
- **Line 1229**: Changed from `heap_.size() > 0 ? heap_.top()->lower_bound : search_tree.root.lower_bound` to `get_lower_bound()`
- **Line 736**: Changed from `root_objective_` to `get_lower_bound()` in ramp-up logging

**Why This Fix Works:**
The `get_lower_bound()` function properly considers all sources of lower bound information:
- `lower_bound_ceiling_` (updated during ramp-up when threads encounter numerical issues)
- Heap top node's lower bound (if heap is not empty)
- All `local_lower_bounds_` from active threads (updated as threads solve nodes during ramp-up)

This ensures the lower bound accurately reflects the work done by active threads during ramp-up, not just the pessimistic root objective.

**Example Impact:**
During ramp-up, if threads have solved nodes with lower bounds better than the root objective, the global lower bound will now reflect these improved bounds instead of falling back to the root objective.

## Checklist

- Testing

   - [ ] Added tests
   - [ ] NA

- Documentation
   - [ ] Added new documentation
   - [ ] NA

### DIFF
--- a/cpp/src/dual_simplex/branch_and_bound.cpp
+++ b/cpp/src/dual_simplex/branch_and_bound.cpp
@@ -733,7 +733,7 @@ void branch_and_bound_t<i_t, f_t>::exploration_ramp_up(search_tree_t<i_t, f_t>* 
       stats_.last_log             = tic();
 
       f_t obj              = compute_user_objective(original_lp_, upper_bound);
-      f_t user_lower       = compute_user_objective(original_lp_, root_objective_);
+      f_t user_lower       = compute_user_objective(original_lp_, get_lower_bound());
       std::string gap_user = user_mip_gap<f_t>(obj, user_lower);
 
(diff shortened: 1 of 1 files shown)


nguidotti:
I am sorry that I assume that you used a LLM (the way that you write seems like a LLM response). In any case, we should check the correctness of you changes by running the MIPLIB dataset and attach the relevant logs.

PR #626, opened 2025-11-25
Add support for --split-compile and --jobserver in build.sh

aliceb-nv:
## Description

NVCC has support for intra-TU and target arch parallelization opportunities, which are enabled with --split-compile and --threads. This PR adds a build flag to enable them.
Additionally, support for Make >=4.4 jobserver has been added, in order to prevent oversubscription. NVCC can act as a jobserver client with the --jobserver flag, support for which has been added in this PR.

On a 256 logical cores machine, this cut down build times by about 1.5x.

## Checklist

- Testing
   - [ ] Added tests
   - [ ] NA
- Documentation
   - [ ] Added new documentation
   - [ ] NA

## Summary by CodeRabbit

* **Chores**
  * Added automatic parallel build support (defaults to available CPU cores) and improved handling so builds can run in parallel even when using the no-install option.
  * Enhanced CUDA compilation to better utilize jobserver-style parallelism when parallelism is enabled.
  * Declared make as a build requirement and exposed a make-based build target to invoke parallel builds.


cpp/Makefile
```diff
@@ -0,0 +1,15 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025, NVIDIA CORPORATION. All rights reserved.
```
ramakrishnap-nv: Should we move this to cpp ?

build.sh
```diff
@@ -172,6 +174,26 @@ function cmakeArgs {
     read -ra EXTRA_CMAKE_ARGS <<< "$EXTRA_CMAKE_ARGS"
 }
 
+function parallelArgs {
+    # Check for -j option
+    if [[ -n $(echo "$ARGS" | { grep -E "\-j" || true; } ) ]]; then
```
jameslamb: There is already a `hasArg` function doing this:

https://github.com/NVIDIA/cuopt/blob/7543358c9caca5e557d6e48fefa8d94baadcbab9/build.sh#L97-L99

Using that might make this a bit easier to read, would you consider it?

build.sh
```diff
@@ -349,13 +372,16 @@ if buildAll || hasArg libcuopt; then
...
           -DHOST_LINEINFO=${HOST_LINEINFO} \
+          -DPARALLEL_LEVEL="${PARALLEL_LEVEL}" \
           "${CACHE_ARGS[@]}" \
           "${EXTRA_CMAKE_ARGS[@]}" \
           "${REPODIR}"/cpp
+    JFLAG="${PARALLEL_LEVEL:+-j${PARALLEL_LEVEL}}"
     if hasArg -n; then
-        cmake --build "${LIBCUOPT_BUILD_DIR}" ${VERBOSE_FLAG}
+        # Manual make invocation to start its jobserver
+        make ${JFLAG} -C "${REPODIR}/cpp" LIBCUOPT_BUILD_DIR="${LIBCUOPT_BUILD_DIR}" VERBOSE_FLAG="${VERBOSE_FLAG}" PARALLEL_LEVEL="${PARALLEL_LEVEL}" ninja-build
```
jameslamb, line 391: Instead of needing to directly invoke `make`, you could explicitly select that in the `cmake` configure step above if it's important, like this:

```shell
cmake -G 'Unix Makefiles' \
    -DDEFINE_BENCHMARK="${DEFINE_BENCHMARK}" \
    # ... etc., etc.
```

That'd allow you to continue to just use `cmake --build` instead of invoking the preferred build tool yourself.

Would you consider that?

build.sh
jameslamb: And actually instead of doing all this command-line parsing and needing to care about spellings like `-j2`, `-j 2`, `-j=2` etc. I think you might find it easier to just allow the shell variable `PARALLEL_LEVEL` to be overridden by an environment variable, like `cudf` does:

```shell
PARALLEL_LEVEL=${PARALLEL_LEVEL:=$(nproc)}
```

([cudf build.sh](https://github.com/rapidsai/cudf/blob/2063035f4742a6ac8f7fd54d67a56618a611194c/build.sh#L82))

Would you consider that?

And in CI if you explicitly want to use GNU Make, you could override it here:

https://github.com/NVIDIA/cuopt/blob/7543358c9caca5e557d6e48fefa8d94baadcbab9/ci/build_cpp.sh#L13

With

```shell
CMAKE_GENERATOR="Unix Makefiles"
```

docs: https://cmake.org/cmake/help/latest/manual/cmake-generators.7.html#makefile-generators

build.sh
aliceb-nv: We use Ninja as the default generator; this step is only necessary in order to create a make jobserver that ninja can hook into (as a jobserver client). This workaround is (to the best of my knowledge) necessary to get nvcc's jobserver support to play nice with Ninja, which is a jobsever client itself

build.sh
aliceb-nv: This shell variable override approach sounds good, thanks for the suggestion! Will implement it.

rgsl888prabhu:
@jameslamb May I get your review on this PR

build.sh
jameslamb: Ah ok I see. I don't know of a better way to do that, that seems fine.

PR #628, opened 2025-11-25
Objective cut in recombiners 

akifcorduk:
This PR extends the objective cut of the problem to the recombiners as well. There is also some cleanup of unused functions.

Pending: benchmark results/comparison.
Not critical for 25.12

## Release Notes

* **Refactor**
  * Optimized CPU thread initialization for efficient parallel processing in benchmarks
  * Reorganized diversity management and population-handling workflows
  * Improved local search algorithm structure and problem context handling
  * Enhanced feasibility validation in solution processing

* **Chores**
  * Updated copyright year in headers
  * Removed redundant code and blank lines

### DIFF
--- a/cpp/CMakeLists.txt
+++ b/cpp/CMakeLists.txt
@@ -75,7 +75,6 @@ if(CMAKE_COMPILER_IS_GNUCXX)
   list(APPEND CUOPT_CXX_FLAGS -Werror -Wno-error=deprecated-declarations)
 endif(CMAKE_COMPILER_IS_GNUCXX)
 
-
 # To use sanitizer with cuda runtime, one must follow a few steps:
 # 1. Run the binary with env var set: LD_PRELOAD="$(gcc -print-file-name=libasan.so)" ASAN_OPTIONS='protect_shadow_gap=0:replace_intrin=0'
 # 2. (Optional) To run with a debugger (gdb or cuda-gdb) use the additional ASAN option alloc_dealloc_mismatch=0
--- a/cpp/src/mip/diversity/diversity_manager.cu
+++ b/cpp/src/mip/diversity/diversity_manager.cu
@@ -648,15 +648,7 @@ diversity_manager_t<i_t, f_t>::recombine_and_local_search(solution_t<i_t, f_t>& 
   ls_config_t<i_t, f_t> ls_config;
   ls_config.best_objective_of_parents    = best_objective_of_parents;
   ls_config.at_least_one_parent_feasible = at_least_one_parent_feasible;
-  // offspring.swap_problem_pointers();
-  // population.weights_with_cuts.cstr_weights.resize(offspring.problem_ptr->n_constraints,
-  //                                                  offspring.handle_ptr->get_stream());
-  // raft::copy(population.weights_with_cuts.cstr_weights.data(),
-  //            population.weights.cstr_weights.data(),
-  //            population.weights.cstr_weights.size(),
-  //            offspring.handle_ptr->get_stream());
-  success = this->run_local_search(offspring, population.weights_with_cuts, timer, ls_config);
-  // offspring.swap_problem_pointers();
+  success = this->run_local_search(offspring, population.weights, timer, ls_config);
   if (!success) {
     // add the attempt
     mab_recombiner.add_mab_reward(mab_recombiner.last_chosen_option,
--- a/cpp/src/mip/diversity/population.cu
+++ b/cpp/src/mip/diversity/population.cu
@@ -40,7 +40,6 @@ population_t<i_t, f_t>::population_t(std::string const& name_,
     max_solutions(max_solutions_),
     infeasibility_importance(infeasibility_weight_),
     weights(0, context.problem_ptr->handle_ptr),
-    weights_with_cuts(0, context.problem_ptr->handle_ptr),
     rng(cuopt::seed_generator::get_seed()),
     population_hash_map(*problem_ptr),
     timer(0)
@@ -64,34 +63,12 @@ i_t get_max_var_threshold(i_t n_vars)
 }
 
 template <typename i_t, typename f_t>
-void population_t<i_t, f_t>::apply_problem_ptr_to_all_solutions()
-{
-  for (size_t i = 0; i < indices.size(); i++) {
-    solutions[indices[i].first].second.problem_with_cuts_ptr = problem_ptr_with_cuts;
-  }
-}
-
-template <typename i_t, typename f_t>
 void population_t<i_t, f_t>::allocate_solutions()
 {
   for (size_t i = 0; i < max_solutions; ++i) {
     bool occupied = false;
     solutions.emplace_back(occupied, solution_t<i_t, f_t>(*problem_ptr));
   }
-}
-
-template <typename i_t, typename f_t>
-void population_t<i_t, f_t>::set_problem_ptr_with_cuts(problem_t<i_t, f_t>* problem_ptr_with_cuts)
-{
-  constexpr f_t ten           = 10.;
-  this->problem_ptr_with_cuts = problem_ptr_with_cuts;
-  weights_with_cuts.cstr_weights.resize(problem_ptr_with_cuts->n_constraints,
-                                        problem_ptr_with_cuts->handle_ptr->get_stream());
-  // fill last element with default
-  thrust::uninitialized_fill(problem_ptr_with_cuts->handle_ptr->get_thrust_policy(),
-                             weights_with_cuts.cstr_weights.begin() + problem_ptr->n_constraints,
-                             weights_with_cuts.cstr_weights.end(),
-                             ten);
 }
 
 template <typename i_t, typename f_t>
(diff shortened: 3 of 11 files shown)


cpp/src/mip/diversity/recombiners/fp_recombiner.cuh
```diff
@@ -126,6 +137,7 @@ class fp_recombiner_t : public recombiner_t<i_t, f_t> {
         fp_recombiner_config_t::decrease_max_n_of_vars_from_other();
       }
     }
+    offspring.swap_problem_pointers();
```
aliceb-nv: Could we guard this kind of problem pointer swapping with RAII, say, a utility class like what scoped_guard does with mutexes?
I think it will be very easy to get confused on which problem is currently the active one

cpp/src/mip/diversity/diversity_manager.cu
```diff
@@ -584,7 +591,15 @@ diversity_manager_t<i_t, f_t>::recombine_and_local_search(solution_t<i_t, f_t>&
...
   ls_config.best_objective_of_parents    = best_objective_of_parents;
   ls_config.at_least_one_parent_feasible = at_least_one_parent_feasible;
-  success = this->run_local_search(offspring, population.weights, timer, ls_config);
+  offspring.swap_problem_pointers();
+  population.weights_with_cuts.cstr_weights.resize(offspring.problem_ptr->n_constraints,
+                                                   offspring.handle_ptr->get_stream());
+  raft::copy(population.weights_with_cuts.cstr_weights.data(),
+             population.weights.cstr_weights.data(),
+             population.weights.cstr_weights.size(),
+             offspring.handle_ptr->get_stream());
```
aliceb-nv: I am a little worried about maintaining the two weights arrays as well. I fear it would be easy to end up with broken invariants
Could we maybe imagine a double_buffer_t or flip_buffer_t or ping_pong_t class that would take, say,
thrust::pair<problem_t*, rmm:device_vector<T>> as its type, and allow for swapping between the two sets with the guarantee of no mismatched state?

cpp/src/mip/diversity/population.cu
```diff
@@ -64,6 +73,14 @@ i_t get_max_var_threshold(i_t n_vars)
   return n_vars - 10;
 }
 
+template <typename i_t, typename f_t>
+void population_t<i_t, f_t>::apply_problem_ptr_to_all_solutions()
+{
+  for (size_t i = 0; i < indices.size(); i++) {
+    solutions[indices[i].first].second.problem_with_cuts_ptr = problem_ptr_with_cuts;
```
aliceb-nv: I think we could benefit from adding a `flags` field to problem_t, one of which could be `HasCuts` or `HasObjectiveCuts`. Otherwise I think it will be difficult to check for invariants and ensure we didn't mess up pointer assignments

cpp/src/mip/local_search/local_search.cu
```diff
@@ -558,68 +636,88 @@ void local_search_t<i_t, f_t>::resize_to_old_problem(problem_t<i_t, f_t>* old_pr
...
   }
   population_ptr->update_weights();
-  save_solution_and_add_cutting_plane(
-    population_ptr->best_feasible(), best_solution, best_objective);
-  raft::copy(solution.assignment.data(),
-             best_solution.data(),
-             solution.assignment.size(),
-             solution.handle_ptr->get_stream());
+  save_solution_and_add_cutting_plane(population_ptr->best_feasible(), solution, best_objective);
+  population_ptr->apply_problem_ptr_to_all_solutions();
```
aliceb-nv: I think updating all problem_with_cuts pointers should be the responsibility of the function adding the objective cut (unless I misundertand)
I am worried about the decoupling between the problem modifications and the updating of pointers; it may be all too easy to forget once, and have a lot of trouble tracking down the issue

cpp/src/mip/local_search/local_search.cu
```diff
@@ -558,68 +636,88 @@ void local_search_t<i_t, f_t>::resize_to_old_problem(problem_t<i_t, f_t>* old_pr
...
+                                                             population_t<i_t, f_t>* population_ptr,
+                                                             i_t i,
+                                                             i_t last_improved_iteration,
+                                                             f_t& best_objective)
+{
+  raft::common::nvtx::range fun_scope("reset_alpha_and_save_solution");
+  fp.reset();
+  solution_t<i_t, f_t> solution_copy(solution);
+  solution_copy.problem_ptr = old_problem_ptr;
+  solution_copy.resize_to_problem();
```
aliceb-nv: We should probably have a `update_problem_pointer` method or similar for `solution_t`, to avoid having to call resize_to_problem and/or ensure any current or future invariant is maintained
With the addition of cuts soon I think we need to be very robust there

cpp/CMakeLists.txt
```diff
@@ -452,7 +454,7 @@ install(TARGETS cuopt_cli
 endif()
 
 
-option(BUILD_MIP_BENCHMARKS "Build MIP benchmarks" OFF)
+option(BUILD_MIP_BENCHMARKS "Build MIP benchmarks" ON)
```
aliceb-nv: We should probably leave this CMake option as OFF by default, and set it to ON in our own workflows
e.g. typically I call `build.sh` with `--cmake-args=\\\"-DBUILD_MIP_BENCHMARKS=ON -DBUILD_LP_BENCHMARKS=ON ...` as the build task in vscode/cursor

aliceb-nv (commented):
Thanks for the PR Akif!
I have some concerns regarding code structure/refactoring, I just want to make sure we have robust code since we're going to be manipulating problem_t instances much more with the involvement of cuts soon
Plus I am not sure I fully understand and track the swapping of problem pointers as it stands 😅

rg20:
@akifcorduk is this going for 25.12?

cpp/src/mip/diversity/recombiners/fp_recombiner.cuh
akifcorduk: I think it will just prevent the forgetting to swap back. It will probably be a better coding practice to do that. But it won't help with which problem is active, because we don't always revert.

cpp/src/mip/diversity/diversity_manager.cu
akifcorduk: I think it makes sense. We should make it robust as more cuts are coming.

cpp/src/mip/diversity/population.cu
akifcorduk: There is already a flag for the objective cut. We can use that.

cpp/src/mip/local_search/local_search.cu
akifcorduk: The issue here is that a cut might be added/modified multiple times. This is a lazy initialization which applies the problem and recomputes the qualities. If we recompute qualities everytime a cut is applied, we would lose perf.

cpp/src/mip/local_search/local_search.cu
akifcorduk: Makes sense.

cpp/src/mip/local_search/local_search.cu
aliceb-nv: That makes sense :) 
I could imagine a 'dirty' flag that would be set everytime a cut is added, and reset once apply_problem_ptr() is called and invariants are restored. That way we could assert(!dirty) when we need qualities to be up to date 
Sorry for all the nitpicks 😅 (I just know I would, sooner or later, mess this up, so thinking defensively helps me in my code)

cpp/src/mip/diversity/recombiners/fp_recombiner.cuh
akifcorduk: I tried RAII swaps but I think it complicates the code. I think we should do more robust arch as we will add more cuts soon. Probably whole heuristics also needs to see the same problem, we should probably think about population/infeasibility logic then.

cpp/src/mip/diversity/diversity_manager.cu
akifcorduk: The issue here is that problem is member of the solution and weights is the member of population. I think this might cause more invariants to break. I think the solution to all the problems you pointed out is to use uniform problem for all population and diversity management. Do you think we should tackle this now?

cpp/src/mip/local_search/local_search.cu
akifcorduk: What apply_problem_ptr_to_all_solutions really does is that it adds problem_ptr_with_cuts to the newly added solutions from add_external_solution. Otherwise, we could do it once in the beginning with the first cut.

cpp/CMakeLists.txt
```diff
@@ -169,15 +176,10 @@ if(CMAKE_BUILD_TYPE MATCHES Debug)
 elseif(CMAKE_CUDA_LINEINFO)
   message(STATUS "Enabling line info")
   list(APPEND CUOPT_CUDA_FLAGS -lineinfo)
+  list(APPEND CUOPT_CXX_FLAGS  -lineinfo)
```
aliceb-nv: Would that work with the host compiler instead of nvcc? iirc we already have a switch for turning lineinfo on for host code - `--host-lineinfo` in build.sh

akifcorduk: I couldn't find --host-lineinfo. I think xcompile should propagate lineinfo option, but I am not sure if it is working.

akifcorduk: Oh it is a cmake arg, found it. Let me try some configs to see what works.

aliceb-nv: Here in build.sh :)
https://github.com/NVIDIA/cuopt/blob/ce3c759634ad2cb42a3a7c1cce19d587a3dfb7e3/build.sh#L255

PR #629, opened 2025-11-26
Restart helper function fix

Kh4ster:
Calrify naming on a restart helper function and avoid potentially picking the wrong restart strategy

## Summary by CodeRabbit

* **Refactor**
  * Revised restart-mode logic in the linear programming solver: restart checks were inverted to base initialization on the trust-region mode rather than the previous KKT-based condition. This changes which internal buffers and views are allocated/initialized for different restart strategies. No changes to public interfaces or exported signatures.

### DIFF
--- a/cpp/src/dual_simplex/branch_and_bound.cpp
+++ b/cpp/src/dual_simplex/branch_and_bound.cpp
@@ -780,17 +780,15 @@ void branch_and_bound_t<i_t, f_t>::exploration_ramp_up(mip_node_t<i_t, f_t>* nod
   // to repair the heuristic solution.
   repair_heuristic_solutions();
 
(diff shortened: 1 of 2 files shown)


cpp/src/linear_programming/restart_strategy/pdlp_restart_strategy.cu
```diff
@@ -154,34 +156,42 @@ pdlp_restart_strategy_t<i_t, f_t>::pdlp_restart_strategy_t(
     gap_reduction_ratio_last_trial_{stream_view_},
     last_restart_length_{0},
     // If KKT restart, don't need to init all of those
```
aliceb-nv: Could update this comment to reflect the new logic (not-trust-regsion), but rly minor nit

PR #670, opened 2025-12-08
Rounding and crash fixes

aliceb-nv:
This PR fixes errouneous logic in the simple rounding and edge cases causing supposedly integer solutions to contain fractional variables. RINS fixes are also included (wrong problem pointer being used for some checks). 

Also, it seems that the cuSparse CSR-to-CSC routine occasionally produces an incorrect output transpose. As a workaround, a manual transpose implementation is used instead. Further work will be required to identify a more easily reproducible example and identify the underlying issue (and/or file a cuSparse bug)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved numerical robustness with tighter clamping, added bounds checks, and reduced spurious runtime assertions.
* **Performance Improvements**
  * Faster matrix transpose and improved GPU rounding kernels for better throughput.
* **Stability**
  * Consistent use of copied problem state during heuristics and safer handling of bound/rounding edge cases.
* **Behavioral**
  * Rounding and feasibility flows now enforce post-clamp validation and per-variable locking/decision logic.

### DIFF
--- a/cpp/src/mip/problem/problem_helpers.cuh
+++ b/cpp/src/mip/problem/problem_helpers.cuh
@@ -222,7 +222,6 @@ static bool check_transpose_validity(const rmm::device_uvector<f_t>& coefficient
       raft::device_span<const i_t>(reverse_offsets.data(), reverse_offsets.size()),
       raft::device_span<const i_t>(reverse_variables.data(), reverse_variables.size()),
       failed.data());
-  RAFT_CUDA_TRY(cudaStreamSynchronize(handle_ptr->get_stream()));
   RAFT_CUDA_TRY(cudaPeekAtLastError());
   return !failed.value(handle_ptr->get_stream());
 }


cpp/src/mip/problem/problem.cu
```diff
@@ -299,6 +299,112 @@ problem_t<i_t, f_t>::problem_t(const problem_t<i_t, f_t>& problem_, bool no_deep
...
+
+    i_t p          = atomicAdd(&next_pos[col], 1);
+    row_ind_out[p] = row;
+    val_out[p]     = csr_val[k];
+  }
+}
+
+// CSR to CSC transpose with sorted row indices within each column
+template <typename i_t, typename f_t>
+void csr_to_csc_transpose(const i_t* csr_offsets,
```
rg20, line 330: Lets move this to utility folders.  We are using a lot of these for Barrier solver as well.   You can do this after the release though.

cpp/src/mip/problem/problem_helpers.cuh
```diff
@@ -220,9 +224,7 @@ static bool check_transpose_validity(const rmm::device_uvector<f_t>& coefficient
       failed.data());
   RAFT_CUDA_TRY(cudaStreamSynchronize(handle_ptr->get_stream()));
   RAFT_CUDA_TRY(cudaPeekAtLastError());
-  cuopt_assert(!failed.value(handle_ptr->get_stream()),
-               "Difference between the matrix and its transpose");
-  return true;
+  return !failed.value(handle_ptr->get_stream());
```
rg20, line 226: do you need to sync stream?

cpp/src/mip/problem/problem.cu
aliceb-nv: Will do👍

PR #691, opened 2025-12-11
CI crash fixes

aliceb-nv:
## Release Notes

* **Performance**
  * Optimized branch-and-bound algorithm with improved search termination conditions
  
* **Improvements**
  * Enhanced concurrency control mechanisms across solver components
  * Improved logger initialization and lifecycle management for better resource handling

### DIFF
--- a/cpp/src/utilities/logger.cpp
+++ b/cpp/src/utilities/logger.cpp
@@ -182,7 +182,7 @@ init_logger_t::init_logger_t(std::string log_file, bool log_to_console)
     cuopt::default_logger().log(entry.level, entry.msg.c_str());
   }
 
-  // 1Create guard and store weak reference for future instances to find
+  // Create guard and store weak reference for future instances to find
   auto guard     = std::make_shared<logger_config_guard>();
   g_active_guard = guard;
   guard_         = guard;


cpp/src/dual_simplex/simplex_solver_settings.hpp
```diff
@@ -145,8 +145,8 @@ struct simplex_solver_settings_t {
   std::function<void()> heuristic_preemption_callback;
   std::function<void(std::vector<f_t>&, std::vector<f_t>&, f_t)> set_simplex_solution_callback;
   mutable logger_t log;
-  volatile int* concurrent_halt;  // if nullptr ignored, if !nullptr, 0 if solver should
-                                  // continue, 1 if solver should halt
+  std::atomic<int>* concurrent_halt;  // if nullptr ignored, if !nullptr, 0 if solver should
```
Kh4ster, line 148: Shouldn't we set to nullptr by default to be extra safe?

aliceb-nv: It's initialized in the constructor right above :)
`concurrent_halt(nullptr)`

aliceb-nv: but yea, we should definitely move to c++11-style initialization where we aren't doing it already

Kh4ster: Yup, if possible i'd rather have it here to make sure we never forget about it

Kh4ster: The risk of dereferencing a non-nullptr is too big

aliceb-nv: (Assuming the rest of the PR looks okay to you) 
Would you be okay with addressing this in a follow-up PR (on main brain) right after 25.12 is merged?

Kh4ster: Sure

Kh4ster (approved):
Thanks!

PR #714, opened 2025-12-15
Adjusted column spacing in B&B logs

nguidotti:
This PR adjusts the column spacing the B&B logs to allow 2 letter symbols when finding a feasible solution. This PR allows specifying the opening mode for the logger (used for debugging).

## Checklist

- Testing
   - [ ] Added tests
   - [ ] NA
- Documentation
   - [ ] Added new documentation
   - [ ] NA

## Summary by CodeRabbit

* **New Features**
  * Unified per-node and heuristic reporting for consolidated node-level logs.

* **Bug Fixes**
  * Improved log formatting and alignment; node headers now show depth, branch direction and status; optional LP-status blocks added.
  * Feasible-solution symbols padded for consistent alignment.
  * Initial node count shown differently in final/root logs for clearer tables.

* **Chores**
  * Logging API accepts explicit file-open modes and more robust file handling.

### DIFF
--- a/.github/workflows/nightly.yaml
+++ b/.github/workflows/nightly.yaml
@@ -1,4 +1,4 @@
-# SPDX-FileCopyrightText: Copyright (c) 2025, NVIDIA CORPORATION. All rights reserved.
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026, NVIDIA CORPORATION. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 
 name: Trigger Nightly cuOpt Pipeline
@@ -17,7 +17,7 @@ jobs:
       matrix:
         cuopt_branch:
           - "main"
-          - "release/25.12"
+          - "release/26.02"
     steps:
       - uses: actions/checkout@v4
       - name: Trigger Pipeline
--- a/build.sh
+++ b/build.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-# SPDX-FileCopyrightText: Copyright (c) 2021-2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-FileCopyrightText: Copyright (c) 2021-2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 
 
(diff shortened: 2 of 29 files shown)


cpp/src/dual_simplex/branch_and_bound.cpp
```diff
@@ -534,17 +534,19 @@ void branch_and_bound_t<i_t, f_t>::add_feasible_solution(f_t leaf_objective,
...
-      nodes_unexplored,
-      obj,
-      lower,
-      leaf_depth,
-      nodes_explored > 0 ? exploration_stats_.total_lp_iters / nodes_explored : 0,
-      user_mip_gap<f_t>(obj, lower).c_str(),
-      toc(exploration_stats_.start_time));
+    f_t iter_node   = nodes_explored > 0 ? static_cast<f_t>(exploration_stats_.total_lp_iters) /
+                                           static_cast<f_t>(nodes_explored)
+                                         : 0;
```
aliceb-nv: Should we display "-" instead of "0" in this case maybe? It seems this is what SCIP does

nguidotti: It is a good suggestion. Actually, I just realize that this part of the code is always called within the B&B, so it is unlikely for the `nodes_explored` to be equal to `0`. Heuristics solutions are treated separately.

nguidotti: I think we can just count the root node as an explored node, and this check is not needed anymore.

chris-maes (changes requested):
Can we discuss before merging?

cpp/src/dual_simplex/branch_and_bound.cpp
```diff
@@ -183,21 +183,21 @@ std::string user_mip_gap(f_t obj_value, f_t lower_bound)
...
   }
 }
 
 inline const char* feasible_solution_symbol(thread_type_t type)
 {
   switch (type) {
-    case thread_type_t::EXPLORATION: return "B";
-    case thread_type_t::DIVING: return "D";
-    default: return "U";
+    case thread_type_t::EXPLORATION: return "B ";
```
chris-maes, line 198: Nit: Why not return a single char here and do the spacing in the format string itself?

chris-maes:
LGTM

PR #721, opened 2025-12-16
C api additions

### LINKED ISSUE #719: [FEA] Add way to provide initial primal/dual solution for C API
Hi, I've got a workflow that involves solving a sequence of LPs, where problem `n+1` is a slight modification of problem `n`. As a result, the solution from problem `n` is often a very good starting guess for problem `n+1`. However, it seems that while there is a way to prescribe an initial primal/dual solution with the C++ API, I didn't see a way to do that from the (stable) C API.

I'd like to extend the C API by adding new functions that allow the user to specify an initial primal/dual solution. It appears that the implementation is already there, it's just a matter of exposing it.

I've tried using the C++ API directly, but this is undesirable for two main reasons:
- it requires building cuOpt from source, which is challenging due to dependency management
- it is risky in the sense that the C++ API does not make any stability guarantees

### LINKED ISSUE #720: [FEA] Add a way to export MPS files from C API
I have an application that procedurally generates LPs and solves them with cuOpt. In some cases, I have observed unusual behavior or crashes when trying to solve these LPs. I would like a way to serialize the problems to disk (e.g., MPS files) so that I could share them to help the troubleshooting process.

-----------

I would like to add a new function to the C API that allows a user to export a problem they've constructed to MPS format. For example:

```cpp
cuopt_int_t cuOptWriteProblem(
    const char *filename,
    const cuOptOptimizationProblem *problem_ptr
)
```

> Note: It appears as though the ability to write MPS files is already implemented in the C++ API, so it would just be a matter of exposing this capability to the C API.

------------

I've considered using cuOpt's C++ API to export the files, but it is more challenging due to missing headers in the binary releases and the lack of stability in the API.

aliceb-nv:
This PR extends the C API to add support for saving the problem to a file, adding initial primal and dual solutions for LP, and initial solutions for MIP. This functionality exists in the underlying C++ layer, and is a feature of most LP/MIP solver APIs available.
Kept as a draft for now, as it is intended as a proposal to iterate on.

closes #719
closes #720 

## Summary by CodeRabbit

* **New Features**
  * Write optimization problems to MPS files.
  * Set initial primal and dual solutions for LP solves and specify initial MIP starts.
  * Check semantic equivalence between optimization problems.

* **Enhancements**
  * Improved MPS writer: explicit handling of orphan variables and refined bounds/objective emissions.

* **Tests**
  * Added MPS write/read roundtrip tests to validate I/O and equivalence.

### DIFF
--- a/cpp/include/cuopt/linear_programming/cuopt_c.h
+++ b/cpp/include/cuopt/linear_programming/cuopt_c.h
@@ -705,6 +705,7 @@ cuopt_int_t cuOptGetFloatParameter(cuOptSolverSettings settings,
  *            of size num_variables containing the initial primal values.
  * @param[in] num_variables - The number of variables (size of the primal_solution array).
  *
+ * @note All pointer arguments (primal_solution) refer to host memory.
  * @return A status code indicating success or failure.
  */
 cuopt_int_t cuOptSetInitialPrimalSolution(cuOptSolverSettings settings,
@@ -721,6 +722,7 @@ cuopt_int_t cuOptSetInitialPrimalSolution(cuOptSolverSettings settings,
  *            of size num_constraints containing the initial dual values.
  * @param[in] num_constraints - The number of constraints (size of the dual_solution array).
  *
+ * @note All pointer arguments (dual_solution) refer to host memory.
  * @return A status code indicating success or failure.
  */
 cuopt_int_t cuOptSetInitialDualSolution(cuOptSolverSettings settings,
@@ -740,6 +742,7 @@ cuopt_int_t cuOptSetInitialDualSolution(cuOptSolverSettings settings,
  *
  * @attention Currently unsupported with presolve on.
  *
+ * @note All pointer arguments (solution) refer to host memory.
  * @return A status code indicating success or failure.
  */
 cuopt_int_t cuOptAddMIPStart(cuOptSolverSettings settings,
--- a/cpp/src/linear_programming/optimization_problem.cu
+++ b/cpp/src/linear_programming/optimization_problem.cu
@@ -126,14 +126,12 @@ static bool csr_matrices_equivalent_with_permutation(const rmm::device_uvector<i
   rmm::device_uvector<i_t> this_cols(nnz, stream);
   rmm::device_uvector<f_t> this_vals(nnz, stream);
 
-  rmm::device_uvector<i_t> entry_indices(nnz, stream);
-  thrust::sequence(policy, entry_indices.begin(), entry_indices.end());
-
+  // upper_bound returns 1-based indices; convert to 0-based
   thrust::upper_bound(policy,
                       this_offsets.begin(),
                       this_offsets.end(),
-                      entry_indices.begin(),
-                      entry_indices.end(),
+                      thrust::make_counting_iterator<i_t>(0),
+                      thrust::make_counting_iterator<i_t>(nnz),
                       this_rows.begin());
   thrust::transform(
     policy, this_rows.begin(), this_rows.end(), this_rows.begin(), [] __device__(i_t r) {
@@ -151,8 +149,8 @@ static bool csr_matrices_equivalent_with_permutation(const rmm::device_uvector<i
   thrust::upper_bound(policy,
                       other_offsets.begin(),
                       other_offsets.end(),
-                      entry_indices.begin(),
-                      entry_indices.end(),
+                      thrust::make_counting_iterator<i_t>(0),
+                      thrust::make_counting_iterator<i_t>(nnz),
                       other_rows.begin());
   thrust::transform(
     policy, other_rows.begin(), other_rows.end(), other_rows.begin(), [] __device__(i_t r) {
@@ -259,19 +257,23 @@ bool optimization_problem_t<i_t, f_t>::is_equivalent(
   };
 
   // Compare variable-indexed arrays
+  if (c_.size() != other.c_.size()) { return false; }
   if (!permuted_eq(c_.begin(), c_.end(), other.c_.begin(), d_var_perm.begin())) { return false; }
+  if (variable_lower_bounds_.size() != other.variable_lower_bounds_.size()) { return false; }
   if (!permuted_eq(variable_lower_bounds_.begin(),
                    variable_lower_bounds_.end(),
                    other.variable_lower_bounds_.begin(),
                    d_var_perm.begin())) {
     return false;
   }
+  if (variable_upper_bounds_.size() != other.variable_upper_bounds_.size()) { return false; }
   if (!permuted_eq(variable_upper_bounds_.begin(),
                    variable_upper_bounds_.end(),
                    other.variable_upper_bounds_.begin(),
                    d_var_perm.begin())) {
     return false;
   }
+  if (variable_types_.size() != other.variable_types_.size()) { return false; }
   if (!permuted_eq(variable_types_.begin(),
                    variable_types_.end(),
                    other.variable_types_.begin(),
@@ -280,19 +282,23 @@ bool optimization_problem_t<i_t, f_t>::is_equivalent(
   }
 
   // Compare constraint-indexed arrays
+  if (b_.size() != other.b_.size()) { return false; }
   if (!permuted_eq(b_.begin(), b_.end(), other.b_.begin(), d_row_perm.begin())) { return false; }
+  if (constraint_lower_bounds_.size() != other.constraint_lower_bounds_.size()) { return false; }
   if (!permuted_eq(constraint_lower_bounds_.begin(),
                    constraint_lower_bounds_.end(),
                    other.constraint_lower_bounds_.begin(),
                    d_row_perm.begin())) {
     return false;
   }
+  if (constraint_upper_bounds_.size() != other.constraint_upper_bounds_.size()) { return false; }
   if (!permuted_eq(constraint_upper_bounds_.begin(),
                    constraint_upper_bounds_.end(),
                    other.constraint_upper_bounds_.begin(),
                    d_row_perm.begin())) {
     return false;
   }
+  if (row_types_.size() != other.row_types_.size()) { return false; }
   if (!permuted_eq(
         row_types_.begin(), row_types_.end(), other.row_types_.begin(), d_row_perm.begin())) {
     return false;


cpp/include/cuopt/linear_programming/cuopt_c.h
```diff
@@ -681,6 +695,51 @@ cuopt_int_t cuOptGetFloatParameter(cuOptSolverSettings settings,
                                    const char* parameter_name,
                                    cuopt_float_t* parameter_value);
 
+/**
+ * @brief Set the initial primal solution for an LP solve.
+ *
+ * @param[in] settings - The solver settings object.
+ * @param[in] primal_solution - A pointer to an array of type cuopt_float_t
+ *            of size num_variables containing the initial primal values.
```
Kh4ster, line 705: Do we want to specify if data should be on the CPU and/or GPU?

cpp/include/cuopt/linear_programming/cuopt_c.h
```diff
@@ -681,6 +695,51 @@ cuopt_int_t cuOptGetFloatParameter(cuOptSolverSettings settings,
...
+cuopt_int_t cuOptSetInitialPrimalSolution(cuOptSolverSettings settings,
+                                          const cuopt_float_t* primal_solution,
+                                          cuopt_int_t num_variables);
+
+/**
+ * @brief Set the initial dual solution for an LP solve.
+ *
+ * @param[in] settings - The solver settings object.
+ * @param[in] dual_solution - A pointer to an array of type cuopt_float_t
+ *            of size num_constraints containing the initial dual values.
```
Kh4ster, line 722: Same here

cpp/include/cuopt/linear_programming/cuopt_c.h
```diff
@@ -681,6 +695,51 @@ cuopt_int_t cuOptGetFloatParameter(cuOptSolverSettings settings,
...
+
+/**
+ * @brief Add an initial solution (MIP start) for MIP solving.
+ *
+ * This function can be called multiple times to add multiple MIP starts.
+ * The solver will use these as starting points for the MIP search.
+ *
+ * @param[in] settings - The solver settings object.
+ * @param[in] solution - A pointer to an array of type cuopt_float_t
+ *            of size num_variables containing the solution values.
```
Kh4ster, line 740: Same here

cpp/src/linear_programming/cuopt_c.cpp
```diff
@@ -92,6 +92,25 @@ cuopt_int_t cuOptReadProblem(const char* filename, cuOptOptimizationProblem* pro
   return CUOPT_SUCCESS;
 }
 
+cuopt_int_t cuOptWriteProblem(cuOptOptimizationProblem problem,
+                              const char* filename,
+                              cuopt_int_t format)
+{
+  if (problem == nullptr) { return CUOPT_INVALID_ARGUMENT; }
```
Kh4ster, line 75: Are we also checking somewhere that the filename is not empty?

cpp/tests/linear_programming/c_api_tests/c_api_test.c
```diff
@@ -1200,3 +1200,84 @@ cuOptDestroySolution(&solution);
...
+
+  printf("Write problem test passed\n");
+
+DONE:
+  cuOptDestroyProblem(&problem);
+  cuOptDestroyProblem(&problem_read);
+  cuOptDestroySolverSettings(&settings);
+  cuOptDestroySolution(&solution);
+  return status;
+}
```
Kh4ster, line 1284: Shouldn't we add a test where we check equality between an original crafted problem and a problem after read/write?

cpp/include/cuopt/linear_programming/cuopt_c.h
aliceb-nv: My understanding is that the C API abstracts away the reality of a GPU implementation - therefore all pointers involved are host pointers (as a result, we do not expose the stream(s) involved)
I think that if the need arises, we could create alternative versions of these functions to upload GPU data, or use something similar to cuSparse's pointer mode flag

cpp/src/linear_programming/cuopt_c.cpp
aliceb-nv: Nice catch :) Likely later down the line, but we might as well check as early as in the C API wrapper

cpp/tests/linear_programming/c_api_tests/c_api_test.c
aliceb-nv: That sounds good, this will just require a bit of trickery since we would need to access the C++ API in the test. Let me check how it can be done

cpp/include/cuopt/linear_programming/cuopt_c.h
smish-nvidia: I'll note that as a new cuOpt user, I had the exact same question about host/device pointer for inputs. 

I think it would be helpful to explicitly say they expect host pointers (either in the individual docstrings or in some high level place in the user guide "all C API pointers are expected to be pointers to host memory").

cpp/src/linear_programming/cuopt_c.cpp
```diff
@@ -706,6 +702,60 @@ cuopt_int_t cuOptGetFloatParameter(cuOptSolverSettings settings,
...
+                                          cuopt_int_t num_variables)
+{
+  if (settings == nullptr) { return CUOPT_INVALID_ARGUMENT; }
+  if (primal_solution == nullptr) { return CUOPT_INVALID_ARGUMENT; }
+  if (num_variables <= 0) { return CUOPT_INVALID_ARGUMENT; }
+
+  solver_settings_t<cuopt_int_t, cuopt_float_t>* solver_settings =
+    static_cast<solver_settings_t<cuopt_int_t, cuopt_float_t>*>(settings);
+  try {
+    solver_settings->set_initial_pdlp_primal_solution(primal_solution, num_variables);
```
rg20, line 716: This is only valid for PDLP right?    How about for MIP or other LP solvers?   

If this is only for PDLP, then C API should also be explicit.

cpp/src/linear_programming/cuopt_c.cpp
```diff
@@ -706,6 +702,60 @@ cuopt_int_t cuOptGetFloatParameter(cuOptSolverSettings settings,
...
+                                        cuopt_int_t num_constraints)
+{
+  if (settings == nullptr) { return CUOPT_INVALID_ARGUMENT; }
+  if (dual_solution == nullptr) { return CUOPT_INVALID_ARGUMENT; }
+  if (num_constraints <= 0) { return CUOPT_INVALID_ARGUMENT; }
+
+  solver_settings_t<cuopt_int_t, cuopt_float_t>* solver_settings =
+    static_cast<solver_settings_t<cuopt_int_t, cuopt_float_t>*>(settings);
+  try {
+    solver_settings->set_initial_pdlp_dual_solution(dual_solution, num_constraints);
```
rg20, line 734: same here

cpp/src/linear_programming/cuopt_c.cpp
```diff
@@ -706,6 +702,60 @@ cuopt_int_t cuOptGetFloatParameter(cuOptSolverSettings settings,
...
+  } catch (const std::exception& e) {
+    return CUOPT_INVALID_ARGUMENT;
+  }
+  return CUOPT_SUCCESS;
+}
+
+cuopt_int_t cuOptAddMIPStart(cuOptSolverSettings settings,
```
rg20, line 741: Combine MIP and LP API?

cpp/src/linear_programming/cuopt_c.cpp
aliceb-nv: I've added a note clarifying this is only supported for PDLP yet. That way we can add support later on without having to change the API itself

cpp/src/linear_programming/cuopt_c.cpp
aliceb-nv: I've added a note clarifying this is only supported for PDLP yet. That way we can add support later on without having to change the API itself

cpp/src/linear_programming/cuopt_c.cpp
aliceb-nv: I've made it a separate function because initial primal solutions are commonly referred to as MIP starts in the context of MIP, and the behavior is somewhat different compared to the LP API (e.g. there can be multiple MIP starts)

cpp/src/linear_programming/optimization_problem.cu
```diff
@@ -78,6 +88,251 @@ optimization_problem_t<i_t, f_t>::optimization_problem_t(
...
+bool optimization_problem_t<i_t, f_t>::is_equivalent(
+  const optimization_problem_t<i_t, f_t>& other) const
+{
+  if (maximize_ != other.maximize_) { return false; }
+  if (n_vars_ != other.n_vars_) { return false; }
+  if (n_constraints_ != other.n_constraints_) { return false; }
+  if (objective_scaling_factor_ != other.objective_scaling_factor_) { return false; }
+  if (objective_offset_ != other.objective_offset_) { return false; }
+  if (problem_category_ != other.problem_category_) { return false; }
+  if (A_.size() != other.A_.size()) { return false; }
```
Kh4ster, line 215: Shouldn't we also compare the sizes of var bounds / constraints bounds or am I missing something?

Kh4ster: Same for b_ c_

cpp/src/linear_programming/optimization_problem.cu
```diff
@@ -78,6 +88,251 @@ optimization_problem_t<i_t, f_t>::optimization_problem_t(
...
+
+  auto policy = rmm::exec_policy(stream);
+
+  // Expand CSR row offsets to row indices for 'this'
+  rmm::device_uvector<i_t> this_rows(nnz, stream);
+  rmm::device_uvector<i_t> this_cols(nnz, stream);
+  rmm::device_uvector<f_t> this_vals(nnz, stream);
+
+  rmm::device_uvector<i_t> entry_indices(nnz, stream);
+  thrust::sequence(policy, entry_indices.begin(), entry_indices.end());
```
Kh4ster: Minor and we don't care about perf here anyways but couldn't we have used a counting_iterator here to avoid the sequence materialization?

cpp/src/linear_programming/optimization_problem.cu
```diff
@@ -78,6 +88,251 @@ optimization_problem_t<i_t, f_t>::optimization_problem_t(
...
+  if (!thrust::equal(policy, this_keys.begin(), this_keys.end(), other_keys.begin())) {
+    return false;
+  }
+
+  if (!thrust::equal(policy, this_vals.begin(), this_vals.end(), other_vals.begin())) {
+    return false;
+  }
+
+  return true;
+}
```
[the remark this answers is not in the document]
aliceb-nv: This function is only used in CI testing on known inputs

cpp/src/linear_programming/optimization_problem.cu
aliceb-nv: We check for n_vars and n_constraints beforehand, and we assume the problems in input are well-formed; but I'll add some extra checks

cpp/src/linear_programming/optimization_problem.cu
aliceb-nv: Good catch :) That's cleaner, implementing

PR #726, opened 2025-12-17
Populate dual solution when dual post solve is enabled

hlinsen:
This PR fixes a bug in populating dual post solve information.

## Summary by CodeRabbit

* **New Features**
  * Added configurable "dual postsolve" to optionally restore dual variables and reduced costs during solution recovery.
  * Presolve adapts options when dual postsolve is enabled for improved compatibility.

* **Bug Fixes**
  * When dual postsolve is used, duals and reduced costs are preserved instead of being reset to NaN.

* **Tests**
  * Added tests validating dual postsolve behavior and solution sizing when enabled vs. disabled.

### DIFF
--- a/python/cuopt/cuopt/linear_programming/problem.py
+++ b/python/cuopt/cuopt/linear_programming/problem.py
@@ -935,7 +935,7 @@ class LinearExpression:
                 return other * self
 
     def __rmul__(self, other):
-        return other * self
+        return self * other
 
     def __itruediv__(self, other):
         # Compute expr /= constant


cpp/src/linear_programming/solve.cu
```diff
@@ -924,16 +924,19 @@ optimization_problem_solution_t<i_t, f_t> solve_lp(
...
 
-      thrust::fill(rmm::exec_policy(op_problem.get_handle_ptr()->get_stream()),
-                   dual_solution.data(),
-                   dual_solution.data() + dual_solution.size(),
-                   std::numeric_limits<f_t>::signaling_NaN());
-      thrust::fill(rmm::exec_policy(op_problem.get_handle_ptr()->get_stream()),
-                   reduced_costs.data(),
-                   reduced_costs.data() + reduced_costs.size(),
-                   std::numeric_limits<f_t>::signaling_NaN());
+      if (!settings.dual_postsolve) {
```
Kh4ster: I'm not sure I understand, are we returnning an empty dual or a dual filled with NaNs?

hlinsen: It will be of size 0 so I should actually delete this block entirely you're right

PR #732, opened 2025-12-18
Fix missing headers for thrust::pair, add trailing return types where needed

bdice:
## Description
This fixes some issues with missing headers and missing return types on device lambdas that are needed for the compiler to correctly infer the return types.

This makes cuOpt compatible with CCCL 3.2.

## Checklist

- Testing
   - [ ] Added tests
   - [ ] NA
- Documentation
   - [ ] Added new documentation
   - [ ] NA

## Summary by CodeRabbit

* **Refactor**
  * Device lambdas across solver modules now have explicit return types and several inline lambdas were replaced with named functors to improve clarity, reliability, and maintainability.
* **Chores**
  * Added necessary includes to support device-side utilities and types used by internal algorithms.

### DIFF
--- a/cpp/src/routing/utilities/md_utils.hpp
+++ b/cpp/src/routing/utilities/md_utils.hpp
@@ -12,7 +12,6 @@
 
 #include <thrust/functional.h>
 #include <thrust/logical.h>
-#include <cuda/functional>
 #include <rmm/device_uvector.hpp>
 #include <utilities/copy_helpers.hpp>
 #include <utilities/macros.cuh>


cpp/src/mip/utils.cuh
```diff
@@ -188,12 +188,29 @@ inline i_t compute_number_of_integer_var_diff(const rmm::device_uvector<i_t>& in
...
                      indices.end(),
-                     [first_ptr, second_ptr, int_tol] __host__ __device__(i_t idx) {
+                     [first_ptr, second_ptr, int_tol] __host__ __device__(i_t idx) -> bool {
                        return integer_equal<f_t>(first_ptr[idx], second_ptr[idx], int_tol);
                      });
   return same_vars;
 }
 
+template <typename i_t, typename f_t>
+struct integer_equal_on_indices_functor {
```
bdice, line 198: I'm seeing if I can make these functors back into device lambdas. I think the issue before was that it was a `__host__ __device__` lambda and we're using it in a device-only context.

cpp/src/routing/utilities/md_utils.hpp
```diff
@@ -12,6 +12,7 @@
 
 #include <thrust/functional.h>
 #include <thrust/logical.h>
+#include <cuda/functional>
```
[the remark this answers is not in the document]
bdice: This is correct.
```suggestion
```

cpp/src/mip/utils.cuh
bdice: Doesn't look like that's working for me. I recommend we merge this as-is. Converting lambdas to functors fixes a lot of compiler type inference issues, we've seen this many times in cudf.

PR #746, opened 2026-01-05
Move augmented sytem computations in barrier to GPU

### LINKED ISSUE #705: QP/QCP/SOCP
Tasks:
• QP performance improvements
    ◦ Move augmented system to GPU
        ▪︎ May also help for LPs that use augmented system
    ◦ Reduce latency for small (portfolio optimization) problems
• QCP/SOCP solver

rg20:
## Description
This PR improves the performance of augmented system computation within barrier method by moving iterative refinement, forming augmented system, and few other computations to the GPU.  

## Issue
Closes #705


rg20:
@coderabbit review

cpp/src/linear_programming/optimization_problem.cu
```diff
@@ -158,46 +158,84 @@ void optimization_problem_t<i_t, f_t>::set_quadratic_objective_matrix(
...
+    i_t row_start = Q_offsets[i];
+    i_t row_end   = Q_offsets[i + 1];
+    for (i_t p = row_start; p < row_end; ++p) {
+      i_t j = Q_indices[p];
+      f_t x = Q_values[p];
+      // Add H(i,j)
+      H_i.push_back(i);
+      H_j.push_back(j);
+      if (i == j) { H_x.push_back(2 * x); }
+      if (i != j) {
```
chris-maes, line 180: I made a mistake here. We always want to execute this if (even when i == j)

cpp/src/dual_simplex/barrier.cu
```diff
@@ -442,14 +429,35 @@ class iteration_data_t {
...
       }
       augmented.col_start[n + m] = q;
       cuopt_assert(q == 2 * nnzA + n + m + off_diag_Qnz, "augmented nnz != predicted");
       cuopt_assert(A.col_start[n] == AT.col_start[m], "A nz != AT nz");
 
+      csr_matrix_t<i_t, f_t> augmented_CSR(n + m, n + m, augmented.col_start[n + m]);
+      augmented.to_compressed_row(augmented_CSR);
+
+      std::vector<i_t> augmented_diagonal_indices(augmented_CSR.n, -1);
+      // Extract the diagonal indices from augmented_CSR
```
chris-maes: This is less efficient than forming `augemented_diagonal_indices` in the loop. Since you need to do a search.

chris-maes: I think the issue is that you are forming augmented in CSC. And then you want to have it in CSR format. 

I would just form augmented in CSR from the start. And generate `augmented_diagonal_indicies` as you do that.

cpp/src/dual_simplex/barrier.cu
```diff
@@ -464,16 +472,38 @@ class iteration_data_t {
       cuopt_assert(error.norm1() <= 1e-2, "|| Aug - Aug^T ||_1 > 1e-2");
 #endif
     } else {
-      for (i_t j = 0; j < n; ++j) {
-        f_t q_diag = nnzQ > 0 ? Qdiag[j] : 0.0;
+      /*
```
chris-maes: Remove dead code?

cpp/src/dual_simplex/barrier.cu
```diff
@@ -1902,6 +1891,7 @@ int barrier_solver_t<i_t, f_t>::initial_point(iteration_data_t<i_t, f_t>& data)
     for (i_t k = 0; k < lp.num_rows; k++) {
       data.y[k] = py[lp.num_cols + k];
     }
+#if 0
```
chris-maes: Just remove this debug code

cpp/src/dual_simplex/barrier.cu
```diff
@@ -2450,25 +2362,30 @@ i_t barrier_solver_t<i_t, f_t>::gpu_compute_search_direction(iteration_data_t<i_
...
-      void solve(const dense_vector_t<i_t, f_t>& b, dense_vector_t<i_t, f_t>& x) const
+
+      void solve(rmm::device_uvector<f_t>& b, rmm::device_uvector<f_t>& x) const
       {
         data_.chol->solve(b, x);
       }
     } op(data);
     iterative_refinement(op, augmented_rhs, augmented_soln);
+
+#if 0
```
chris-maes: It's probably a good idea to reenable this code. It's good to know when the residual on the augmented system is large

chris-maes: Alternatively, you could return some info from the call to `iterative_refinement`

cpp/src/dual_simplex/barrier.cu
```diff
@@ -2483,6 +2400,7 @@ i_t barrier_solver_t<i_t, f_t>::gpu_compute_search_direction(iteration_data_t<i_
     // TMP should only be init once
     data.cusparse_dy_ = data.cusparse_view_.create_vector(data.d_dy_);
 
+#if 0
```
chris-maes: I guess we should delete this code as well. It's too bad because it does provide some information about where the residual is large.

cpp/src/dual_simplex/barrier.cu
```diff
@@ -2996,101 +2901,68 @@ void barrier_solver_t<i_t, f_t>::compute_target_mu(
...
+  raft::copy(data.d_dz_aff_.data(), data.dz_aff.data(), data.dz_aff.size(), stream_view_);
+
+  f_t step_primal_aff = std::min(gpu_max_step_to_boundary(data, data.d_w_, data.d_dw_aff_),
+                                 gpu_max_step_to_boundary(data, data.d_x_, data.d_dx_aff_));
+  f_t step_dual_aff   = std::min(gpu_max_step_to_boundary(data, data.d_v_, data.d_dv_aff_),
+                               gpu_max_step_to_boundary(data, data.d_z_, data.d_dz_aff_));
+
+  if (data.Q.n > 0) { step_primal_aff = step_dual_aff = std::min(step_primal_aff, step_dual_aff); }
+
+  f_t complementarity_xz_aff_sum = data.transform_reduce_helper_.transform_reduce(
```
chris-maes, line 2894: Can we add a comment here about what is being computed mathematically? 

Is this computing both 
complementatrity_xz_aff
and the sum?

chris-maes: Ah I guess we don't store the product x_aff * z_aff here, we just want the sum? Ah comment would be helpful.

cpp/src/linear_programming/optimization_problem.cu
rg20: Yes.  If i == j, I am just doubling the coefficient here. That is equivalent.

cpp/src/dual_simplex/barrier.cu
rg20: This is done only once upfront.  Should not impact the performance.

cpp/src/dual_simplex/barrier.cu
```diff
@@ -3138,79 +2932,62 @@ template <typename i_t, typename f_t>
...
-    // TODO Nicolas: Redundant copies
-    data.d_y_.resize(data.y.size(), stream_view_);
-    data.d_dy_aff_.resize(data.dy_aff.size(), stream_view_);
-    raft::copy(data.d_y_.data(), data.y.data(), data.y.size(), stream_view_);
-    raft::copy(data.d_dy_aff_.data(), data.dy_aff.data(), data.dy_aff.size(), stream_view_);
+  // TODO Nicolas: Redundant copies
+  data.d_y_.resize(data.y.size(), stream_view_);
+  data.d_dy_aff_.resize(data.dy_aff.size(), stream_view_);
+  raft::copy(data.d_y_.data(), data.y.data(), data.y.size(), stream_view_);
+  raft::copy(data.d_dy_aff_.data(), data.dy_aff.data(), data.dy_aff.size(), stream_view_);
```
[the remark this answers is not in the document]
rg20: Changing this would require a lot of testing again. I will do this as a followup PR.

cpp/src/dual_simplex/barrier.cu
```diff
@@ -1437,6 +1412,20 @@ class iteration_data_t {
...
+                          dense_vector_t<i_t, f_t>& y)
+  {
+    rmm::device_uvector<f_t> d_x(x.size(), handle_ptr->get_stream());
+    raft::copy(d_x.data(), x.data(), x.size(), handle_ptr->get_stream());
+    rmm::device_uvector<f_t> d_y(y.size(), handle_ptr->get_stream());
+    raft::copy(d_y.data(), y.data(), y.size(), handle_ptr->get_stream());
+    augmented_multiply(alpha, d_x, beta, d_y);
+    raft::copy(y.data(), d_y.data(), y.size(), handle_ptr->get_stream());
+    handle_ptr->sync_stream();
+  }
```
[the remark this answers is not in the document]
rg20: This is used only in debug code.

cpp/src/dual_simplex/barrier.cu
chris-maes: I think it would be best to fix this before merging. In addition to avoiding searching and a transpose, the code will also become simpler to understand.

cpp/src/dual_simplex/barrier.cu
```diff
@@ -484,99 +501,66 @@ class iteration_data_t {
...
-          stream_view_);
-      } else {
-        d_inv_diag_prime.resize(inv_diag.size(), stream_view_);
-        raft::copy(d_inv_diag_prime.data(), d_inv_diag.data(), inv_diag.size(), stream_view_);
-      }
+    // TODO do we really need this copy? (it's ok since gpu to gpu)
+    raft::copy(device_AD.x.data(),
+               d_original_A_values.data(),
+               d_original_A_values.size(),
+               handle_ptr->get_stream());
```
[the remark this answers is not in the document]
chris-maes: The TODO comment is bogus and should be removed. The copy is necessary because the `device_AD.x` values depend on the value of D which changes from iteration. We need to copy the original A values over, before scaling by the matrix D.

cpp/src/dual_simplex/barrier.cu
chris-maes: can you add a comment as request?

chris-maes:
LGTM but I would form the augmented matrix in CSR before merging. Do all these error checks have an impact on performance? It would probably be good to rerun the benchmarks again before merging to ensure that there are no performance regressions.

chris-maes (approved):
LGTM. Thanks for making these changes. It's always a great sign when a PR improves performance and reduces the amount of code :)

Iroy30 (approved):
Thanks!

ramakrishnap-nv (approved):

PR #766, opened 2026-01-13
Scheduler-Worker Model + Reliability Branching

### LINKED ISSUE #445: [FEA] Improve the lower bound calculation in the ramp-up phase
During the ramp-up phase in the parallel B&B, `cuopt` set the root objective as the global lower bound. This is a pessimistic estimation and does not reflect the actual lower bound calculated by the threads. Note that not all threads are active in this phase.

### LINKED ISSUE #526: [FEA] Move report/repair to a separated task during the ramp up phase
Currently, the report and repair of the heuristic solutions can be done by any task during the ramp-up phase, which requires additional synchronisation to avoid data races or incorrect reporting. Moving the report/repair to a separate task should lead to cleaner code.

### LINKED ISSUE #700: Parallel Branch and Bound
- Master-worker parallel model
- Reliability branching in parallel using dual simplex
- Parallel subtree

nguidotti:
This PR implements the schedule-worker model for B&B, where one thread (the scheduler) is responsible for scheduling tasks to the other threads (the workers). With this model, the CPU resources can be shared efficiently among the different tasks within the MIP solver. It also allows the scheduling policy to be changed in runtime (e.g., in the future, the solver can dynamically set the number of workers allocated to each diving heuristics at runtime). 

This also implements a parallel reliability branching (Section 5.7 from [1]). This also fixes the incorrect pseudocost update in each node in the B&B tree.

Closes #526.
Closes #445.
Closes #700.

Performance over the MIPLIB2017 dataset (GH200):

```
main - 046eafd (1) vs reliability-branching-cuts (2)
Feasible solutions = 225 vs 224 (-1)
Optimal solutions = 44 vs 68 (+24)
Solutions with <0.1% primal gap = 109 vs 125 (+16)
Average nodes explored = 7298973 vs 4300451 (-2998522, -41.081%)
Shifted geomean for MIP gap = 0.2979 vs 0.2374 (-0.0605, -20.302%)
Average primal gap = 11.9935 vs 11.0833 (-0.9102, -7.589%)
Average primal integral per time = 19.3326 vs 30.7054 (+11.3728, +37.038%)

```

## Reference 
[1] T. Achterberg, “Constraint Integer Programming,” PhD, Technischen Universität Berlin, Berlin, 2007. doi: [10.14279/depositonce-1634](https://doi.org/10.14279/depositonce-1634).

## Checklist

- Testing
   - [ ] Added tests
   - [ ] NA
- Documentation
   - [ ] Added new documentation
   - [ ] NA

## Summary by CodeRabbit

* **New Features**
  * New CLI option --reliability-branching and reliability-based trial branching in variable selection.
  * New per-worker branch-and-bound framework and a CPU PCG random generator.

* **Improvements**
  * Master–worker scheduling for branch-and-bound, improved multithreaded efficiency and diagnostics.
  * Safer concurrency (protected queues, reduced coarse locking) and refined pseudocost scoring for more robust variable choices.
  * Adjusted threading and solver defaults.

* **Chores**
  * Simplified sub-MIP / RINS configuration for leaner execution.

### DIFF
--- a/cpp/src/dual_simplex/branch_and_bound.cpp
+++ b/cpp/src/dual_simplex/branch_and_bound.cpp
@@ -273,7 +273,8 @@ branch_and_bound_t<i_t, f_t>::branch_and_bound_t(
   }
 #endif
 
-  upper_bound_ = inf;
+  upper_bound_    = inf;
+  root_objective_ = std::numeric_limits<f_t>::quiet_NaN();
 }
 
 template <typename i_t, typename f_t>
@@ -283,7 +284,14 @@ f_t branch_and_bound_t<i_t, f_t>::get_lower_bound()
   f_t heap_lower_bound = node_queue_.get_lower_bound();
   lower_bound          = std::min(heap_lower_bound, lower_bound);
   lower_bound          = std::min(worker_pool_.get_lower_bound(), lower_bound);
-  return std::isfinite(lower_bound) ? lower_bound : -inf;
+
+  if (std::isfinite(lower_bound)) {
+    return lower_bound;
+  } else if (std::isfinite(root_objective_)) {
+    return root_objective_;
+  } else {
+    return -inf;
+  }
 }
 
 template <typename i_t, typename f_t>


cpp/src/dual_simplex/branch_and_bound.cpp
```diff
@@ -1567,56 +1450,36 @@ mip_status_t branch_and_bound_t<i_t, f_t>::solve(mip_solution_t<i_t, f_t>& solut
...
   exploration_stats_.last_log             = tic();
-  active_subtrees_                        = 0;
   is_running                              = true;
   lower_bound_ceiling_                    = inf;
-  should_report_                          = true;
+  min_node_queue_size_                    = 2 * settings_.num_threads;
+
+  if (settings_.diving_settings.coefficient_diving != 0) {
+    calculate_variable_locks(original_lp_, var_up_locks_, var_down_locks_);
+  }
```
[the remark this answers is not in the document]
nguidotti: @coderabbit Check again. This issue has already been fixed.

benchmarks/linear_programming/cuopt/run_mip.cpp
```diff
@@ -147,6 +147,7 @@ int run_single_file(std::string file_path,
                     int num_cpu_threads,
                     bool write_log_file,
                     bool log_to_console,
+                    bool reliability_branching,
```
chris-maes: Remove? This should be normal parameter right?

benchmarks/linear_programming/cuopt/run_mip.cpp
```diff
@@ -197,13 +198,16 @@ int run_single_file(std::string file_path,
     }
   }
 
+  CUOPT_LOG_INFO("Reliability branching: %d\n", reliability_branching);
```
chris-maes: Remove debug code?

benchmarks/linear_programming/cuopt/run_mip.cpp
```diff
@@ -354,6 +360,10 @@ int main(int argc, char* argv[])
     .help("track allocations (t/f)")
     .default_value(std::string("f"));
 
+  program.add_argument("--reliability-branching")
```
chris-maes, line 360: This should be handled as a regular parameter. No need to change run_mip at all.

I would suggest making the parameter work as follows:
reliability_branching = -1 (automatic)
reliability_branching = 0 ( off)
reliability_branching = k > 0, a variable is considered reliable if it has been branched on k times.

chris-maes: See `math_optimization/solver_settings.cu` for an example of how to add a parameter.

benchmarks/linear_programming/cuopt/run_mip.cpp
```diff
@@ -376,12 +386,13 @@ int main(int argc, char* argv[])
   std::string result_file;
   int batch_num = -1;
 
-  bool heuristics_only   = program.get<std::string>("--heuristics-only")[0] == 't';
-  int num_cpu_threads    = program.get<int>("--num-cpu-threads");
-  bool write_log_file    = program.get<std::string>("--write-log-file")[0] == 't';
-  bool log_to_console    = program.get<std::string>("--log-to-console")[0] == 't';
-  double memory_limit    = program.get<double>("--memory-limit");
-  bool track_allocations = program.get<std::string>("--track-allocations")[0] == 't';
+  bool heuristics_only       = program.get<std::string>("--heuristics-only")[0] == 't';
```
chris-maes: Remove all changes

cpp/include/cuopt/linear_programming/mip/solver_settings.hpp
```diff
@@ -83,8 +82,10 @@ class mip_solver_settings_t {
   bool heuristics_only = false;
   i_t num_cpu_threads  = -1;  // -1 means use default number of threads in branch and bound
   i_t num_gpus         = 1;
-  i_t mip_batch_pdlp_strong_branching = 0;
-  bool log_to_console                 = true;
+  bool log_to_console  = true;
+
+  bool reliability_branching = true;
```
chris-maes: Use int as suggested above.

cpp/src/dual_simplex/bnb_worker.hpp
```diff
@@ -0,0 +1,274 @@
...
+namespace cuopt::linear_programming::dual_simplex {
+
+constexpr int bnb_num_worker_types = 5;
+
+// Indicate the search and variable selection algorithms used by each thread
+// in B&B (See [1]).
+//
+// [1] T. Achterberg, “Constraint Integer Programming,” PhD, Technischen Universität Berlin,
+// Berlin, 2007. doi: 10.14279/depositonce-1634.
+enum bnb_worker_type_t : int {
```
chris-maes: This seems more like the search strategy than the worker type

cpp/src/utilities/pcg.hpp
```diff
@@ -0,0 +1,158 @@
...
+#pragma once
+
+#include <stdint.h>
+#include <vector>
+
+// Copied from raft/PCGenerator (rng_device.cuh). We cannot use raft implementation
+// on the CPU (.cpp file) since the raft header includes CUDA code.
+// The original code is from https://www.pcg-random.org/.
+namespace cuopt {
+class PCG {
```
chris-maes: PCG isn't a very descriptive name. Can we call this `random_generator_pcg_t`?

chris-maes: Why do we need this specific random number generator?

cpp/src/dual_simplex/bnb_worker.hpp
```diff
@@ -0,0 +1,274 @@
...
+        leaf_problem.lower, leaf_problem.upper, bounds_changed);
+    }
+
+    return node_presolver.bounds_strengthening(
+      leaf_problem.lower, leaf_problem.upper, bounds_changed, settings);
+  }
+
+ private:
+  // For diving, we need to store the full node instead of
+  // of just a pointer, since it is not store in the tree anymore.
```
chris-maes: Nit: since it is not *stored* in the tree anymore.

cpp/src/dual_simplex/branch_and_bound_worker.hpp
```diff
@@ -0,0 +1,274 @@
...
+      node_ptr->update_branched_variable_bounds(
+        leaf_problem.lower, leaf_problem.upper, bounds_changed);
+    }
+
+    return node_presolver.bounds_strengthening(
+      leaf_problem.lower, leaf_problem.upper, bounds_changed, settings);
+  }
+
+ private:
+  // For diving, we need to store the full node instead of
```
chris-maes, line 155: Not for this PR, but we should discuss this strategy of separating a node off. This limits our ability to do reduced cost strengthening and propagate bounds changes at the root to the child nodes.

cpp/src/dual_simplex/branch_and_bound_worker.hpp
```diff
@@ -0,0 +1,274 @@
...
+class bnb_worker_pool_t {
+ public:
+  void init(i_t num_workers,
+            const lp_problem_t<i_t, f_t>& original_lp,
+            const csr_matrix_t<i_t, f_t>& Arow,
+            const std::vector<variable_type_t>& var_type,
+            const simplex_solver_settings_t<i_t, f_t>& settings)
+  {
+    workers_.resize(num_workers);
+    num_idle_workers_ = num_workers;
```
chris-maes, line 174: Nit: is `num_idle_workers_` just equal to `idle_workers.size()`? If so, can the variable be removed?

cpp/src/dual_simplex/branch_and_bound_worker.hpp
```diff
@@ -0,0 +1,274 @@
...
+    for (i_t i = 0; i < workers_.size(); ++i) {
+      if (workers_[i]->worker_type == BEST_FIRST && workers_[i]->is_active) {
+        lower_bound = std::min(workers_[i]->lower_bound.load(), lower_bound);
+      }
+    }
+
+    return lower_bound;
+  }
+
+  i_t num_idle_workers() { return num_idle_workers_; }
```
chris-maes, line 231: `return idle_workers_.size()`?

cpp/src/dual_simplex/bnb_worker.hpp
```diff
@@ -0,0 +1,274 @@
...
+    workers_.resize(num_workers);
+    num_idle_workers_ = num_workers;
+    for (i_t i = 0; i < num_workers; ++i) {
+      workers_[i] =
+        std::make_unique<bnb_worker_data_t<i_t, f_t>>(i, original_lp, Arow, var_type, settings);
+      idle_workers_.push_front(i);
+    }
+  }
+
+  // Here, we are assuming that the master is the only
```
chris-maes: We should avoid the terminology "master" throughout. I would suggest coming up with a different name.

cpp/src/dual_simplex/bnb_worker.hpp
```diff
@@ -0,0 +1,274 @@
...
+    std::lock_guard<omp_mutex_t> lock(mutex_);
+    idle_workers_.push_back(worker->worker_id);
+    num_idle_workers_++;
+  }
+
+  f_t get_lower_bounds()
+  {
+    f_t lower_bound = std::numeric_limits<f_t>::infinity();
+
+    for (i_t i = 0; i < workers_.size(); ++i) {
```
chris-maes: Do you need to check the idle_workers as well? Or just the active workers?

cpp/src/dual_simplex/bnb_worker.hpp
```diff
@@ -0,0 +1,274 @@
...
+
+  void return_worker_to_pool(bnb_worker_data_t<i_t, f_t>* worker)
+  {
+    worker->is_active = false;
+    std::lock_guard<omp_mutex_t> lock(mutex_);
+    idle_workers_.push_back(worker->worker_id);
+    num_idle_workers_++;
+  }
+
+  f_t get_lower_bounds()
```
chris-maes: Nit: Shouldn't this be `get_lower_bound`? As it is returning a single lower bound?

cpp/src/dual_simplex/branch_and_bound.cpp
```diff
@@ -286,7 +282,7 @@ void branch_and_bound_t<i_t, f_t>::report(char symbol, f_t obj, f_t lower_bound,
   i_t nodes_unexplored = exploration_stats_.nodes_unexplored;
   f_t user_obj         = compute_user_objective(original_lp_, obj);
   f_t user_lower       = compute_user_objective(original_lp_, lower_bound);
-  f_t iter_node        = exploration_stats_.total_lp_iters / nodes_explored;
+  f_t iter_node = nodes_explored > 0 ? (f_t)exploration_stats_.total_lp_iters / nodes_explored : 0;
```
chris-maes: If the nodes explored is zero, we need to return the total_lp_iters. This is important at the root node, where we have explored zero nodes, but still want to correctly report the number of iterations performed.

cpp/src/dual_simplex/simplex_solver_settings.hpp
```diff
@@ -22,18 +22,69 @@ namespace cuopt::linear_programming::dual_simplex {
...
+  // iterations in the best-first threads.
   f_t iteration_limit_factor = 0.05;
-  i_t backtrack_limit        = 5;
+
+  // The maximum backtracking allowed.
+  i_t backtrack_limit = 5;
+};
+
+template <typename i_t, typename f_t>
+struct reliability_branching_settings_t {
```
chris-maes: Having a separate setting struct for reliability branching is cubersome. Because of how we use the solver_settings.cu/solver_settings.hpp to change settings.

I think you should decide which settings should be configurable by the user and include those directly in simplex_solver_settings struct. The settings that aren't configurable by the user should probably go directly in pseudo_costs.hpp.

cpp/src/dual_simplex/simplex_solver_settings.hpp
```diff
@@ -22,18 +22,69 @@ namespace cuopt::linear_programming::dual_simplex {
...
+  // - <0: automatic
+  // - 0: disable (use pseudocost branching instead)
+  // - >0: will use the value for the threshold.
+  i_t reliable_threshold = -1;
+
+  // Maximum and minimum points of the curve to determine the value
+  // of the `reliable_threshold` based on the current number of LP
+  // iterations in strong branching and B&B.
+  // Only used when `reliable_threshold` is negative
+  i_t max_reliable_threshold = 5;
```
chris-maes: 5 is low. I've seen 8 discussed in the literature.

cpp/src/dual_simplex/branch_and_bound.cpp
```diff
@@ -569,24 +565,37 @@ template <typename i_t, typename f_t>
...
-    type = bnb_worker_type_t::PSEUDOCOST_DIVING;
-  }
-
-  switch (type) {
+  switch (worker_data->worker_type) {
     case bnb_worker_type_t::BEST_FIRST:
-      branch_var = pc_.variable_selection(fractional, solution, log);
-      round_dir  = martin_criteria(solution[branch_var], root_relax_soln_.x[branch_var]);
+
+      if (settings_.reliability_branching_settings.enable) {
```
chris-maes: I think this should be
`if (settings_.reliability_branching != 0)`

cpp/src/dual_simplex/branch_and_bound_worker.hpp
```diff
@@ -0,0 +1,274 @@
+/* clang-format off */
```
chris-maes, line 1: Nit: I think we should avoid the abbreviation `bnb`. I would call this file `branch_and_bound_worker.hpp`. So that it is grouped with the `branch_and_bound.hpp/cpp` file.

cpp/src/dual_simplex/branch_and_bound.cpp
```diff
@@ -667,56 +666,40 @@ dual::status_t branch_and_bound_t<i_t, f_t>::solve_node_lp(
...
-    i_t node_iter                    = 0;
-    f_t lp_start_time                = tic();
-    std::vector<f_t> leaf_edge_norms = edge_norms_;  // = node.steepest_edge_norms;
+    i_t node_iter     = 0;
+    f_t lp_start_time = tic();
 
     lp_status = dual_phase2_with_advanced_basis(2,
                                                 0,
-                                                recompute_bounds_and_basis,
+                                                worker_data->recompute_basis,
```
chris-maes: Nit: Keep the name `recompute_bounds_and_basis` if we are recomputing the bounds as well.

cpp/src/dual_simplex/branch_and_bound.cpp
```diff
@@ -667,56 +666,40 @@ dual::status_t branch_and_bound_t<i_t, f_t>::solve_node_lp(
...
-      leaf_problem.lower, leaf_problem.upper, node_presolver.bounds_changed);
-  }
-
-  bool feasible =
-    node_presolver.bounds_strengthening(leaf_problem.lower, leaf_problem.upper, lp_settings);
-
-  dual::status_t lp_status = dual::status_t::DUAL_UNBOUNDED;
+  bool feasible                = worker_data->set_lp_variable_bounds_for(node_ptr, settings_);
+  dual::status_t lp_status     = dual::status_t::DUAL_UNBOUNDED;
+  worker_data->leaf_edge_norms = edge_norms_;  // = node.steepest_edge_norms;
```
chris-maes: Nit: remove the leftover `// node.steepest_edge_norm`

cpp/src/dual_simplex/bnb_worker.hpp
```diff
@@ -0,0 +1,274 @@
...
+#include <utilities/pcg.hpp>
+
+#include <array>
+#include <deque>
+#include <mutex>
+#include <vector>
+
+namespace cuopt::linear_programming::dual_simplex {
+
+constexpr int bnb_num_worker_types = 5;
```
chris-maes: Nit: bnb_num_worker_types -> branch_and_bound_worker_types

cpp/src/dual_simplex/bnb_worker.hpp
chris-maes: Let's not use the abbreviation `bnb`

cpp/src/dual_simplex/bnb_worker.hpp
```diff
@@ -0,0 +1,274 @@
...
+enum bnb_worker_type_t : int {
+  BEST_FIRST         = 0,  // Best-First + Plunging.
+  PSEUDOCOST_DIVING  = 1,  // Pseudocost diving (9.2.5)
+  LINE_SEARCH_DIVING = 2,  // Line search diving (9.2.4)
+  GUIDED_DIVING      = 3,  // Guided diving (9.2.3).
+  COEFFICIENT_DIVING = 4   // Coefficient diving (9.2.1)
+};
+
+template <typename i_t, typename f_t>
+struct bnb_stats_t {
```
chris-maes: Nit: Let's not use the abbreviation bnb

cpp/src/dual_simplex/bnb_worker.hpp
```diff
@@ -0,0 +1,274 @@
...
+  omp_atomic_t<f_t> total_lp_solve_time  = 0.0;
+  omp_atomic_t<int64_t> nodes_explored   = 0;
+  omp_atomic_t<int64_t> nodes_unexplored = 0;
+  omp_atomic_t<int64_t> total_lp_iters   = 0;
+  omp_atomic_t<i_t> nodes_since_last_log = 0;
+  omp_atomic_t<f_t> last_log             = 0.0;
+};
+
+template <typename i_t, typename f_t>
+class bnb_worker_data_t {
```
chris-maes: Nit: Let's not use the abbreviation bnb

cpp/src/dual_simplex/branch_and_bound.cpp
```diff
@@ -606,29 +615,19 @@ branch_variable_t<i_t> branch_and_bound_t<i_t, f_t>::variable_selection(
...
-  std::vector<i_t>& basic_list,
-  std::vector<i_t>& nonbasic_list,
-  bounds_strengthening_t<i_t, f_t>& node_presolver,
-  bnb_worker_type_t thread_type,
-  bool recompute_bounds_and_basis,
-  const std::vector<f_t>& root_lower,
-  const std::vector<f_t>& root_upper,
-  bnb_stats_t<i_t, f_t>& stats,
-  logger_t& log)
+dual::status_t branch_and_bound_t<i_t, f_t>::solve_node_lp(mip_node_t<i_t, f_t>* node_ptr,
```
chris-maes: This ties `solve_node_lp` to the `bnb_worker_data_t` struct. Is this what we want to do?

cpp/src/dual_simplex/branch_and_bound.cpp
```diff
@@ -667,56 +666,40 @@ dual::status_t branch_and_bound_t<i_t, f_t>::solve_node_lp(
...
-  } else {
-    node_ptr->update_branched_variable_bounds(
-      leaf_problem.lower, leaf_problem.upper, node_presolver.bounds_changed);
-  }
-
-  bool feasible =
-    node_presolver.bounds_strengthening(leaf_problem.lower, leaf_problem.upper, lp_settings);
-
-  dual::status_t lp_status = dual::status_t::DUAL_UNBOUNDED;
+  bool feasible                = worker_data->set_lp_variable_bounds_for(node_ptr, settings_);
```
chris-maes: Nit: why does the method name `set_lp_variable_bounds_for` end with `for`?

cpp/src/dual_simplex/branch_and_bound.cpp
```diff
@@ -834,133 +816,20 @@ std::pair<node_status_t, rounding_direction_t> branch_and_bound_t<i_t, f_t>::upd
...
-
-template <typename i_t, typename f_t>
-void branch_and_bound_t<i_t, f_t>::plunge_from(i_t task_id,
-                                               mip_node_t<i_t, f_t>* start_node,
-                                               lp_problem_t<i_t, f_t>& leaf_problem,
-                                               bounds_strengthening_t<i_t, f_t>& node_presolver,
-                                               basis_update_mpf_t<i_t, f_t>& basis_factors,
-                                               std::vector<i_t>& basic_list,
-                                               std::vector<i_t>& nonbasic_list)
+void branch_and_bound_t<i_t, f_t>::plunge_with(bnb_worker_data_t<i_t, f_t>* worker_data,
```
chris-maes: Why did the name change from `plunge_from` to `plunge_with`?

cpp/src/dual_simplex/branch_and_bound.cpp
```diff
@@ -834,133 +816,20 @@ std::pair<node_status_t, rounding_direction_t> branch_and_bound_t<i_t, f_t>::upd
...
 {
-  bool recompute_bounds_and_basis = true;
   std::deque<mip_node_t<i_t, f_t>*> stack;
-  stack.push_front(start_node);
-
-  while (stack.size() > 0 && solver_status_ == mip_status_t::UNSET && is_running) {
-    if (task_id == 0) { repair_heuristic_solutions(); }
+  stack.push_front(worker_data->start_node);
+  worker_data->recompute_basis  = true;
+  worker_data->recompute_bounds = true;
```
chris-maes: Ah you have separate flags for recomputing the basis and the bounds? What is the case where these would be recomputed separately?

cpp/src/dual_simplex/branch_and_bound.cpp
```diff
@@ -1055,102 +893,45 @@ void branch_and_bound_t<i_t, f_t>::plunge_from(i_t task_id,
...
-                                             basis_update_mpf_t<i_t, f_t>& basis_factors,
-                                             std::vector<i_t>& basic_list,
-                                             std::vector<i_t>& nonbasic_list,
-                                             bnb_worker_type_t diving_type)
+void branch_and_bound_t<i_t, f_t>::dive_with(bnb_worker_data_t<i_t, f_t>* worker_data)
 {
   logger_t log;
   log.log = false;
 
+  bnb_worker_type_t diving_type    = worker_data->worker_type;
```
chris-maes: Diving type is a better name than `bnb_worker_type`. Maybe `bnb_worker_type_t` -> `diving_type_t`

cpp/src/dual_simplex/branch_and_bound.cpp
```diff
@@ -1220,68 +991,219 @@ void branch_and_bound_t<i_t, f_t>::dive_from(mip_node_t<i_t, f_t>& start_node,
...
+    if (!launched_any_task) { std::this_thread::sleep_for(std::chrono::milliseconds(1)); }
   }
+
+  is_running_ = false;
+}
+
+template <typename i_t, typename f_t>
+void branch_and_bound_t<i_t, f_t>::single_threaded_solve()
+{
+  bnb_worker_data_t<i_t, f_t> worker(0, original_lp_, Arow_, var_types_, settings_);
```
chris-maes: Keep in mind that `Arow_`, `var_types_` and `original_lp_` all change during the cut pass. It probably isn't a good idea to create the worker until after the cut passes are completed.

cpp/src/dual_simplex/branch_and_bound.cpp
```diff
@@ -1220,68 +991,219 @@ void branch_and_bound_t<i_t, f_t>::dive_from(mip_node_t<i_t, f_t>& start_node,
...
+    // execution of the master. As of 8/Jan/2026, GCC does not
+    // implement taskyield, but LLVM does.
+    if (!launched_any_task) { std::this_thread::sleep_for(std::chrono::milliseconds(1)); }
   }
+
+  is_running_ = false;
+}
+
+template <typename i_t, typename f_t>
+void branch_and_bound_t<i_t, f_t>::single_threaded_solve()
```
chris-maes, line 1355: This is a nice function to have. It is good for sub MIP.

cpp/src/dual_simplex/branch_and_bound.cpp
```diff
@@ -1388,41 +1310,18 @@ lp_status_t branch_and_bound_t<i_t, f_t>::solve_root_relaxation(
 }
 
 template <typename i_t, typename f_t>
-mip_status_t branch_and_bound_t<i_t, f_t>::solve(mip_solution_t<i_t, f_t>& solution)
+mip_status_t branch_and_bound_t<i_t, f_t>::solve(mip_solution_t<i_t, f_t>& solution,
+                                                 mip_solve_mode_t solve_mode)
```
chris-maes: Nit: I don't think solve_mode should be a special argument to `solve`. The solve mode can be a parameter in settings. Just like any other parameter.

cpp/src/dual_simplex/branch_and_bound.hpp
```diff
@@ -35,38 +36,14 @@ enum class mip_status_t {
...
-//
-// [1] T. Achterberg, “Constraint Integer Programming,” PhD, Technischen Universität Berlin,
-// Berlin, 2007. doi: 10.14279/depositonce-1634.
-enum class bnb_worker_type_t {
-  BEST_FIRST         = 0,  // Best-First + Plunging.
-  PSEUDOCOST_DIVING  = 1,  // Pseudocost diving (9.2.5)
-  LINE_SEARCH_DIVING = 2,  // Line search diving (9.2.4)
-  GUIDED_DIVING = 3,  // Guided diving (9.2.3). If no incumbent is found yet, use pseudocost diving.
-  COEFFICIENT_DIVING = 4  // Coefficient diving (9.2.1)
+enum class mip_solve_mode_t {
```
chris-maes: Put this in simplex_solver_settings. I would also keep it an int. Perhaps `num_mip_threads`.

cpp/src/dual_simplex/branch_and_bound.hpp
```diff
@@ -114,7 +91,7 @@ class branch_and_bound_t {
   lp_status_t solve_root_relaxation(simplex_solver_settings_t<i_t, f_t> const& lp_settings);
 
   // The main entry routine. Returns the solver status and populates solution with the incumbent.
-  mip_status_t solve(mip_solution_t<i_t, f_t>& solution);
+  mip_status_t solve(mip_solution_t<i_t, f_t>& solution, mip_solve_mode_t solve_mode);
```
chris-maes: Please revert the special argument for solver_mode

cpp/src/dual_simplex/branch_and_bound.hpp
```diff
@@ -204,72 +185,43 @@ class branch_and_bound_t {
...
-                 std::vector<i_t>& basic_list,
-                 std::vector<i_t>& nonbasic_list,
-                 bnb_worker_type_t diving_type);
-
-  // Each diving thread pops the first node from the dive queue and then performs
-  // a deep dive into the subtree determined by the node.
-  void diving_thread(bnb_worker_type_t diving_type);
+  void dive_with(bnb_worker_data_t<i_t, f_t>* worker_data);
+
+  // Run the scheduler (aka the master) whose will schedule and manage
```
chris-maes: Ah scheduler is a better name. Let's remove the word `master` throughout. And refer to this a the scheduler-worker model.

cpp/src/dual_simplex/simplex_solver_settings.hpp
```diff
@@ -22,18 +22,69 @@ namespace cuopt::linear_programming::dual_simplex {
...
+template <typename i_t, typename f_t>
+struct reliability_branching_settings_t {
+  // Enable or disable reliability branching
+  bool enable = false;
+
+  // Lower bound for the maximum number of LP iterations for a single trial branching
+  i_t lower_max_lp_iter = 10;
+
+  // Upper bound for the maximum number of LP iterations for a single trial branching
+  i_t upper_max_lp_iter = 500;
```
chris-maes: 500 seems pretty high here. How sensitive are the timings to this upper bound on the iterations?

cpp/src/dual_simplex/simplex_solver_settings.hpp
```diff
@@ -22,18 +22,69 @@ namespace cuopt::linear_programming::dual_simplex {
...
+
+  // Priority of the tasks created when running the trial branching in parallel.
+  // Set to 1 to have the same priority as the other tasks.
+  i_t task_priority = 5;
```
chris-maes: Does 5 mean they have higher or lower priority than the best-first/diving tasks? I would think we would want reliability branching to have higher priority since this stalls us from exploring nodes in the tree.

cpp/src/dual_simplex/pseudo_costs.cpp
```diff
@@ -135,92 +133,81 @@ void strong_branch_helper(i_t start,
...
+  child_settings.inside_mip      = 2;
+  child_settings.scale_columns   = false;
+
+  lp_solution_t<i_t, f_t> solution(original_lp.num_rows, original_lp.num_cols);
+  i_t iter                                         = 0;
+  std::vector<variable_status_t> child_vstatus     = vstatus;
+  std::vector<f_t> child_edge_norms                = edge_norms;
+  std::vector<i_t> child_basic_list                = basic_list;
+  std::vector<i_t> child_nonbasic_list             = nonbasic_list;
+  basis_update_mpf_t<i_t, f_t> child_basis_factors = basis_factors;
```
chris-maes, line 177: This is expensive. Both in terms of time and memory. It will copy the basis factorization. 

We should think about how we want to do this.

cpp/src/dual_simplex/pseudo_costs.cpp
```diff
@@ -379,7 +287,7 @@ template <typename i_t, typename f_t>
 void pseudo_costs_t<i_t, f_t>::initialized(i_t& num_initialized_down,
                                            i_t& num_initialized_up,
                                            f_t& pseudo_cost_down_avg,
-                                           f_t& pseudo_cost_up_avg) const
+                                           f_t& pseudo_cost_up_avg)
```
chris-maes: Why drop the const?

cpp/src/dual_simplex/pseudo_costs.cpp
```diff
@@ -470,8 +379,193 @@ i_t pseudo_costs_t<i_t, f_t>::variable_selection(const std::vector<i_t>& fractio
...
+i_t pseudo_costs_t<i_t, f_t>::reliable_variable_selection(
+  mip_node_t<i_t, f_t>* node_ptr,
+  const std::vector<i_t>& fractional,
+  const std::vector<f_t>& solution,
+  const simplex_solver_settings_t<i_t, f_t>& settings,
+  const std::vector<variable_type_t>& var_types,
+  bnb_worker_data_t<i_t, f_t>* worker_data,
+  const bnb_stats_t<i_t, f_t>& bnb_stats,
+  f_t upper_bound,
+  logger_t& log)
```
chris-maes, line 551: Nit: why do you need a separate `logger_t` here? Can't you get the logger_t from `settings`? If you want it to be a different logger, just switch it in the caller.

cpp/src/dual_simplex/pseudo_costs.cpp
```diff
@@ -470,8 +379,193 @@ i_t pseudo_costs_t<i_t, f_t>::variable_selection(const std::vector<i_t>& fractio
...
+
+  log.printf("PC: num initialized down %d up %d avg down %e up %e\n",
+             num_initialized_down,
+             num_initialized_up,
+             pseudo_cost_down_avg,
+             pseudo_cost_up_avg);
+
+  const int64_t bnb_total_lp_iter  = bnb_stats.total_lp_iters;
+  const int64_t bnb_nodes_explored = bnb_stats.nodes_explored;
+  const i_t bnb_lp_iter_per_node   = bnb_total_lp_iter / bnb_stats.nodes_explored;
```
chris-maes: Are you not using reliability branching at the root node? You use the previous strong branching code there?

cpp/src/dual_simplex/pseudo_costs.cpp
```diff
@@ -470,8 +379,193 @@ i_t pseudo_costs_t<i_t, f_t>::variable_selection(const std::vector<i_t>& fractio
...
+
+  i_t reliable_threshold = settings.reliability_branching_settings.reliable_threshold;
+  if (reliable_threshold < 0) {
+    const i_t max_threshold = settings.reliability_branching_settings.max_reliable_threshold;
+    const i_t min_threshold = settings.reliability_branching_settings.min_reliable_threshold;
+    const f_t iter_factor   = settings.reliability_branching_settings.bnb_lp_factor;
+    const i_t iter_offset   = settings.reliability_branching_settings.bnb_lp_offset;
+    const int64_t alpha     = iter_factor * bnb_total_lp_iter;
+    const int64_t max_iter  = alpha + settings.reliability_branching_settings.bnb_lp_offset;
```
chris-maes: Nit: Is `max_iter` the total number of simplex iterations you want to spend in reliability branching? If so maybe it should be called `max_total_reliability_iterations`

cpp/src/dual_simplex/pseudo_costs.cpp
```diff
@@ -470,8 +379,193 @@ i_t pseudo_costs_t<i_t, f_t>::variable_selection(const std::vector<i_t>& fractio
...
+
+    f_t gamma          = (max_iter - sb_total_lp_iter) / (sb_total_lp_iter + 1.0);
```
chris-maes: What is gamma?  It would be helpful to give this a descriptive name.

Is gamma the fraction of the strong branching iterations that you want to devote to reliability branching?  If so, maybe you can use the name `fraction_of_strong_branching_iterations`

cpp/src/dual_simplex/pseudo_costs.cpp
```diff
@@ -470,8 +379,193 @@ i_t pseudo_costs_t<i_t, f_t>::variable_selection(const std::vector<i_t>& fractio
...
+  const i_t max_num_candidates = settings.reliability_branching_settings.max_num_candidates;
+  const i_t num_candidates     = std::min<size_t>(unreliable_list.size(), max_num_candidates);
+
+  assert(task_priority > 0);
+  assert(max_num_candidates > 0);
+  assert(num_candidates > 0);
+
+  log.printf(
+    "RB iters = %d, B&B iters = %d, unreliable = %d, num_tasks = %d, reliable_threshold = %d\n",
+    sb_total_lp_iter.load(),
```
chris-maes: Let's not use the abbreviation sb. Call it `strong_branching_total_iters`

cpp/src/dual_simplex/pseudo_costs.cpp
```diff
@@ -470,8 +379,193 @@ i_t pseudo_costs_t<i_t, f_t>::variable_selection(const std::vector<i_t>& fractio
...
+
+  initialized(num_initialized_down, num_initialized_up, pseudo_cost_down_avg, pseudo_cost_up_avg);
+
+  log.printf("PC: num initialized down %d up %d avg down %e up %e\n",
+             num_initialized_down,
+             num_initialized_up,
+             pseudo_cost_down_avg,
+             pseudo_cost_up_avg);
+
+  const int64_t bnb_total_lp_iter  = bnb_stats.total_lp_iters;
```
chris-maes: Let's not use the abbreviation bnb. Call it `branch_and_bound_total_lp_iters`  or just `branch_and_bound_lp_iters` or `branch_and_bound_simplex_iters`

cpp/src/dual_simplex/pseudo_costs.cpp
```diff
@@ -470,8 +379,193 @@ i_t pseudo_costs_t<i_t, f_t>::variable_selection(const std::vector<i_t>& fractio
...
+    num_tasks,
+    reliable_threshold);
+
+  // Shuffle the unreliable list so every variable has the same chance to be selected.
+  if (unreliable_list.size() > max_num_candidates) { worker_data->rng.shuffle(unreliable_list); }
+
+#pragma omp taskloop if (num_tasks > 1) priority(task_priority) num_tasks(num_tasks)
+  for (i_t i = 0; i < num_candidates; ++i) {
+    const i_t j = unreliable_list[i];
+    std::lock_guard<omp_mutex_t> lock(pseudo_cost_mutex[j]);
```
chris-maes: Do you want to lock the mutex for this entire section of code? 

Wouldn't it be better to only lock when you read or write `pseudo_cost_num_{down,up}` or `pseudo_cost_sum_{down,up}`. The LP solves can take a long time. Ideally, you wouldn't have a thread blocking on an LP solve.

cpp/src/dual_simplex/pseudo_costs.cpp
```diff
@@ -470,8 +379,193 @@ i_t pseudo_costs_t<i_t, f_t>::variable_selection(const std::vector<i_t>& fractio
...
+
+      if (!std::isnan(obj)) {
+        f_t change_in_obj = obj - node_ptr->lower_bound;
+        f_t change_in_x   = solution[j] - std::floor(solution[j]);
+        pseudo_cost_sum_down[j] += change_in_obj / change_in_x;
+        pseudo_cost_num_down[j]++;
+      }
+    }
+
+    if (toc(start_time) > settings.time_limit) { continue; }
```
chris-maes, line 681: Why continue if you've hit the time limit? Shouldn't you break out of the loop?

cpp/src/dual_simplex/pseudo_costs.cpp
```diff
@@ -470,8 +379,193 @@ i_t pseudo_costs_t<i_t, f_t>::variable_selection(const std::vector<i_t>& fractio
...
+    unreliable_list.size(),
+    num_tasks,
+    reliable_threshold);
+
+  // Shuffle the unreliable list so every variable has the same chance to be selected.
+  if (unreliable_list.size() > max_num_candidates) { worker_data->rng.shuffle(unreliable_list); }
+
+#pragma omp taskloop if (num_tasks > 1) priority(task_priority) num_tasks(num_tasks)
+  for (i_t i = 0; i < num_candidates; ++i) {
+    const i_t j = unreliable_list[i];
```
chris-maes, line 647: If you need to continue when you hit the time limit. Probably best to check the time limit at the start of the loop.

cpp/src/dual_simplex/pseudo_costs.cpp
```diff
@@ -470,8 +379,193 @@ i_t pseudo_costs_t<i_t, f_t>::variable_selection(const std::vector<i_t>& fractio
...
+    if (toc(start_time) > settings.time_limit) { continue; }
+    f_t pc_up   = pseudo_cost_num_up[j] > 0 ? pseudo_cost_sum_up[j] / pseudo_cost_num_up[j]
+                                            : pseudo_cost_up_avg;
+    f_t pc_down = pseudo_cost_sum_down[j] > 0 ? pseudo_cost_sum_down[j] / pseudo_cost_num_down[j]
+                                              : pseudo_cost_down_avg;
+
+    constexpr f_t eps = 1e-6;
+    const f_t f_down  = solution[j] - std::floor(solution[j]);
+    const f_t f_up    = std::ceil(solution[j]) - solution[j];
+    f_t score         = std::max(f_down * pc_down, eps) * std::max(f_up * pc_up, eps);
```
chris-maes: Nit: Maybe pull the score out into a separate function, since it is repeated in several places now.

cpp/src/dual_simplex/pseudo_costs.cpp
```diff
@@ -470,8 +379,193 @@ i_t pseudo_costs_t<i_t, f_t>::variable_selection(const std::vector<i_t>& fractio
...
+
+    std::lock_guard<omp_mutex_t> score_lock(score_mutex);
```
chris-maes: I find a bare lock guard harder to read. I would suggest either
```
{
  std::lock_gaurd<omp_mutex_t> score_lock
  if (score > max_score)
  {
  }
}
```

Or put an explict `lock` and `unlock` before and after the if

cpp/src/dual_simplex/pseudo_costs.cpp
```diff
@@ -135,92 +133,81 @@ void strong_branch_helper(i_t start,
...
+  i_t iter                                         = 0;
+  std::vector<variable_status_t> child_vstatus     = vstatus;
+  std::vector<f_t> child_edge_norms                = edge_norms;
+  std::vector<i_t> child_basic_list                = basic_list;
+  std::vector<i_t> child_nonbasic_list             = nonbasic_list;
+  basis_update_mpf_t<i_t, f_t> child_basis_factors = basis_factors;
+
+  dual::status_t status = dual_phase2_with_advanced_basis(2,
+                                                          0,
+                                                          false,
```
chris-maes: Nit: let's not have a bare `false` as an argument. It's easier to understand if you do a
`const bool initialize_basis = false` and give `initialize_basis` as an argument.

cpp/src/dual_simplex/pseudo_costs.cpp
```diff
@@ -135,92 +133,81 @@ void strong_branch_helper(i_t start,
...
+                     branch_var,
+                     child_problem.lower[branch_var],
+                     child_problem.upper[branch_var],
+                     iter,
+                     dual::status_to_string(status).c_str(),
+                     compute_objective(child_problem, solution.x));
+
+  if (status == dual::status_t::DUAL_UNBOUNDED) {
+    // LP was infeasible
+    return std::numeric_limits<f_t>::infinity();
```
chris-maes, line 206: We didn't take advantage of discovering infeasiblity when initializing the psuedocosts at the root node. (We should have, but infeasibilities at the root are pretty rare, so it did not matter as much). 

But inside the tree, branching on a variable leading to an infeasible node should be much more common. If we discover, that both the down branch and the up branch lead to infeasible nodes, than that tells us the parent node is integer infeasible and can fathom it.  So we should probably have two infeasibilities give us an infinite score, so that ensure that we select that variable.

Similarly, if trial branching finds a variable where both the up branch and the down branch of a variable lead to cut off objectives, then branching on this variable, would cause the parent node to be fathomed. So we should probably ensure that we select that variable.

I'm not sure, what to do if only the down branch or the up branch is infeasible/cutoff. For binary variables, branching on these variables results in a fixing. Probably these should be preferred over variables where both the down branch and the up branch is feasible. But we probably still need to pick the best one sided objective among those variables with one infeasible branch. 

I guess this is not for this PR. But we should probably run some experiments soon to determine if there are some problems where trial branching detects infeasibility/cut off many times.

cpp/src/dual_simplex/pseudo_costs.hpp
```diff
@@ -52,22 +57,31 @@ class pseudo_costs_t {
...
+  std::vector<omp_atomic_t<f_t>> pseudo_cost_sum_down;
+  std::vector<omp_atomic_t<i_t>> pseudo_cost_num_up;
+  std::vector<omp_atomic_t<i_t>> pseudo_cost_num_down;
   std::vector<f_t> strong_branch_down;
   std::vector<f_t> strong_branch_up;
-
-  omp_mutex_t mutex;
+  std::vector<omp_mutex_t> pseudo_cost_mutex;
   omp_atomic_t<i_t> num_strong_branches_completed = 0;
+  omp_atomic_t<int64_t> sb_total_lp_iter          = 0;
```
chris-maes: I would call this `strong_branching_lp_iter` or `trial_branching_simplex_iter` or just `trial_branching_iter`

cpp/src/dual_simplex/solve.cpp
```diff
@@ -599,7 +599,7 @@ i_t solve(const user_problem_t<i_t, f_t>& problem,
   if (is_mip(problem) && !settings.relaxation) {
     branch_and_bound_t branch_and_bound(problem, settings);
     mip_solution_t<i_t, f_t> mip_solution(problem.num_cols);
-    mip_status_t mip_status = branch_and_bound.solve(mip_solution);
+    mip_status_t mip_status = branch_and_bound.solve(mip_solution, mip_solve_mode_t::BNB_PARALLEL);
```
chris-maes: Remove special mode argument

cpp/src/dual_simplex/solve.cpp
```diff
@@ -638,7 +638,7 @@ i_t solve_mip_with_guess(const user_problem_t<i_t, f_t>& problem,
   if (is_mip(problem)) {
     branch_and_bound_t branch_and_bound(problem, settings);
     branch_and_bound.set_initial_guess(guess);
-    mip_status_t mip_status = branch_and_bound.solve(solution);
+    mip_status_t mip_status = branch_and_bound.solve(solution, mip_solve_mode_t::BNB_PARALLEL);
```
chris-maes: Remove special mode argument

cpp/src/mip/solver.cu
```diff
@@ -167,20 +167,19 @@ solution_t<i_t, f_t> mip_solver_t<i_t, f_t>::run_solver()
...
     }
 
-    i_t num_threads                           = branch_and_bound_settings.num_threads;
-    i_t num_bfs_workers                       = std::max(1, num_threads / 4);
-    i_t num_diving_workers                    = std::max(1, num_threads - num_bfs_workers);
-    branch_and_bound_settings.num_bfs_workers = num_bfs_workers;
-    branch_and_bound_settings.diving_settings.num_diving_workers = num_diving_workers;
-    branch_and_bound_settings.mip_batch_pdlp_strong_branching =
-      context.settings.mip_batch_pdlp_strong_branching;
+    dual_simplex::mip_solve_mode_t solve_mode =
```
chris-maes: This is a case where the code could be simplified by just having a `mip_num_threads` and not having a special mode enum.

benchmarks/linear_programming/cuopt/run_mip.cpp
nguidotti: The `run_mip` is used for benchmarking. I will keep it since allow us to experiment with different reliability threshold. 
I will add the option to `cuopt_cli` as well.

benchmarks/linear_programming/cuopt/run_mip.cpp
nguidotti: The code is automatically formatted for every commit. The only changes here is the spacing.

cpp/src/utilities/pcg.hpp
nguidotti: It is the same generator used by the heuristics from `raft`. However, we cannot include the same header  since it includes CUDA code. PCG (Permuted Congruential Generator) produces a higher quality random number compared to the ones in the STL. I can follow `raft` naming and rename to `random_pcgenerator_t`.

cpp/src/dual_simplex/branch_and_bound_worker.hpp
nguidotti: This is only used for diving, where we are creating a new node that are detached from the main tree. The bounds at that node are stored in the `start_lower/upper`. We need to do this so diving do not change the main tree (which leads to incorrect lower bounds).

cpp/src/dual_simplex/branch_and_bound_worker.hpp
nguidotti: `num_idle_workers_` is atomic, while `idle_workers.size()` is not. So, we can read `num_idle_workers_` without needing to lock the entire object.

cpp/src/dual_simplex/bnb_worker.hpp
nguidotti: Ideally, only the active workers. However, it is very difficult to keep track which workers are active efficiently (an active worker can become idle at any time, when it happens we need to remove it from the "active" list and this will happen in parallel). It is significant easier to just ignore inactive workers, like I did here.

cpp/src/dual_simplex/simplex_solver_settings.hpp
nguidotti: I took the value from SCIP. Note that `max_reliable_threshold` and `min_reliable_threshold` defines the curve that will be used for determining the value of the `reliable_threshold` based on the number of LP iterations in strong branching and B&B. If it strong branching is cheap, then `reliable_threshold` may be greater than `max_reliable_threshold`. 

Achterberg's thesis describes a similar strategy and this is how is implemented in SCIP.

nguidotti: Nevertheless, we may spent some time tuning these parameters.

cpp/src/dual_simplex/simplex_solver_settings.hpp
nguidotti: These bounds are used in both Achterberg's thesis and SCIP. We can experiment with different values later.

cpp/src/dual_simplex/simplex_solver_settings.hpp
nguidotti: Higher number = Higher priority. By default, all task have priority `0`. Note this is just a hint to the runtime.

cpp/src/dual_simplex/pseudo_costs.cpp
nguidotti: Yes. For the root node, we are still using the old codepath.

cpp/src/dual_simplex/pseudo_costs.cpp
nguidotti: The variable is classified as unreliable for more than one thread at a time. In this way, we prevent that we call the trial branching multiple times for the same unreliable variable (imagine that `reliable_threshold = 1` and we have 4 threads trying to initialize the same variable with strong branching`). 

The blocking can be avoided via `try_lock`, but in practice this make very little difference in performance as we progress down the tree, the number of variable that needs to be initialized (and the probability of being the same across threads) drops significantly. In the beginning of the tree, many threads are idling (so little contention there) .

cpp/src/dual_simplex/pseudo_costs.cpp
nguidotti: There are a few restrictions for parallelizing a loop with OpenMP:

- You cannot break out of the loop
- The start and end iterations must be defined in terms of a range 

So, when we reach the time limit, the only thing that we can do to end the loop is to skip the iterations until the end (which only takes a few milliseconds).

cpp/src/dual_simplex/pseudo_costs.cpp
nguidotti: See the comment above

cpp/src/dual_simplex/branch_and_bound.cpp
nguidotti: I think it is better this way, otherwise there are too many arguments.

cpp/src/dual_simplex/bnb_worker.hpp
nguidotti: The name start to become too long... Maybe we could move all branch_and_bound routines and types into a separated namespace.

benchmarks/linear_programming/cuopt/run_mip.cpp
chris-maes: I see. Fine to change run_mip. (Although, run_mip should just use regular settings). The important thing is to make reliability branching a parameter.

cpp/src/dual_simplex/bnb_worker.hpp
chris-maes: Typing should not be the bottleneck. If you want to make the name shorter, just remove bnb and call it `worker_type`. Again, though worker type does not seem to be the best name. This is the strategy for the worker.

cpp/src/utilities/pcg.hpp
chris-maes: If you are using existing code, maybe it is better to create a CPU only header. But call the existing code in a .cu or .cpp file. That way you avoid repeating code.

cpp/src/dual_simplex/branch_and_bound_worker.hpp
chris-maes: Diving can still benefit from fixing bounds at the root.

cpp/src/dual_simplex/simplex_solver_settings.hpp
chris-maes: When you run this, what values are you seeing? How sensitive is the run time to the reliable threshold?

cpp/src/dual_simplex/branch_and_bound.cpp
chris-maes: I think the question is more: are we always calling solve_node_lp with a worker? Or do we anticipate that we would want to call this without a worker?

cpp/src/dual_simplex/simplex_solver_settings.hpp
chris-maes: The average number of iterations per node across MIPLIB is 100. I'm not sure 5X times the average makes sense.

cpp/src/dual_simplex/pseudo_costs.cpp
chris-maes: If I understand correctly, you are locking while computing  both the up and down branch of the variable. If you only locked for the down branch, could another thread compute the up branch?

cpp/src/dual_simplex/pseudo_costs.cpp
chris-maes: Thanks for the clarification. But then you should check the time limit at the start of the loop.

cpp/src/dual_simplex/pseudo_costs.cpp
chris-maes: See reply above. If you don't check the time limit at the start of the loop, won't you do extra work until the loop has finished.

cpp/src/dual_simplex/pseudo_costs.cpp
```diff
@@ -195,15 +270,32 @@ void strong_branching(const lp_problem_t<i_t, f_t>& original_lp,
...
+{
+  constexpr f_t eps = 1e-6;
+  i_t num_up        = pseudo_cost_num_up[j];
+  i_t num_down      = pseudo_cost_num_down[j];
+  f_t pc_up         = num_up > 0 ? pseudo_cost_sum_up[j] / num_up : pseudo_cost_up_avg;
+  f_t pc_down       = num_down > 0 ? pseudo_cost_sum_down[j] / num_down : pseudo_cost_down_avg;
+  f_t f_down        = solution[j] - std::floor(solution[j]);
+  f_t f_up          = std::ceil(solution[j]) - solution[j];
+  return std::max(f_down * pc_down, eps) * std::max(f_up * pc_up, eps);
+}
```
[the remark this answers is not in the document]
nguidotti: `pseudo_cost_sum_* / pseudo_cost_num_*` are atomic, so there is no race condition

cpp/src/dual_simplex/pseudo_costs.cpp
nguidotti: Yes, this is a good idea. I will modify the code accordingly.

Ahh I forgot to mention `pseudo_cost_num_{down,up}` or `pseudo_cost_sum_{down,up}` are atomic, so multiple threads can read/write at the same time without locks. The code only locks when calling the trial branches.

cpp/src/dual_simplex/branch_and_bound.cpp
nguidotti: The enum is used for both best-first and diving. I renamed to `search_strategy_t`

cpp/src/dual_simplex/branch_and_bound.cpp
nguidotti: I do not think so

cpp/src/dual_simplex/branch_and_bound_worker.hpp
nguidotti: We are detaching the node from the main tree after it is popped from the queue. In the queue, it is just a pointer to a node in the main tree. So all updates to the nodes in the tree will be reflected on the pending diving nodes.

cpp/src/dual_simplex/pseudo_costs.cpp
nguidotti: This is just a temporary variable

cpp/src/dual_simplex/branch_and_bound.cpp
nguidotti: I do not think we will call this function without a worker

cpp/src/dual_simplex/pseudo_costs.cpp
nguidotti: It should be cheaper than computing from scratch, no?

nguidotti: If we could backtrack to the previous state, then we could avoid the copy here. This will also be useful for backtracking in the B&B code.

cpp/src/dual_simplex/branch_and_bound.cpp
```diff
@@ -1470,68 +1195,207 @@ void branch_and_bound_t<i_t, f_t>::dive_from(mip_node_t<i_t, f_t>& start_node,
...
+
+template <typename i_t, typename f_t>
+void branch_and_bound_t<i_t, f_t>::single_threaded_solve()
+{
+  branch_and_bound_worker_t<i_t, f_t> worker(0, original_lp_, Arow_, var_types_, settings_);
+
+  f_t lower_bound = get_lower_bound();
+  f_t abs_gap     = upper_bound_ - lower_bound;
+  f_t rel_gap     = user_relative_gap(original_lp_, upper_bound_.load(), lower_bound);
+
```
chris-maes, line 1362: Not for this PR, but we should add the cuts in the single-threaded solve.

cpp/src/dual_simplex/branch_and_bound_worker.hpp
```diff
@@ -0,0 +1,279 @@
...
+  omp_atomic_t<int64_t> nodes_unexplored = 0;
+  omp_atomic_t<int64_t> total_lp_iters   = 0;
+  omp_atomic_t<i_t> nodes_since_last_log = 0;
+  omp_atomic_t<f_t> last_log             = 0.0;
+};
+
+template <typename i_t, typename f_t>
+class branch_and_bound_worker_t {
+ public:
+  const i_t worker_id;
```
chris-maes, line 53: Nit: There are a lot of member variables here. It would be good to add the `_` post-fix so it is clear to the reader in the functions that they are accessing the member variables. For example, when reading the code it would have helped if I saw `recompute_basis_` to know that you were accessing the member variable.

cpp/src/dual_simplex/pseudo_costs.cpp
```diff
@@ -88,7 +89,7 @@ void strong_branch_helper(i_t start,
       }
 
       if (branch == 0) {
-        pc.strong_branch_down[k] = obj - root_obj;
+        pc.strong_branch_down[k] = std::max(obj - root_obj, eps);
```
chris-maes: It's possible for us to branch on an integer variable, and the resulting LPs have an objective that is equal to its parent. I think we should take that into account when initializing the psuedocosts. So I would do std::max(obj - root_obj, 0.0).

cpp/src/dual_simplex/pseudo_costs.cpp
```diff
@@ -135,6 +136,80 @@ void strong_branch_helper(i_t start,
...
+  child_settings.scale_columns   = false;
+
+  lp_solution_t<i_t, f_t> solution(original_lp.num_rows, original_lp.num_cols);
+  i_t iter                                         = 0;
+  std::vector<variable_status_t> child_vstatus     = vstatus;
+  std::vector<f_t> child_edge_norms                = edge_norms;
+  std::vector<i_t> child_basic_list                = basic_list;
+  std::vector<i_t> child_nonbasic_list             = nonbasic_list;
+  basis_update_mpf_t<i_t, f_t> child_basis_factors = basis_factors;
+
```
chris-maes, line 178: Another idea. Set the refactorization frequency to be greater than the iteration limit. This will ensure that we only refactor the basis if we run into numerical trouble. That should speed up reliability branching for iteration limits >= refactor_frequency = 100.

cpp/src/dual_simplex/pseudo_costs.cpp
```diff
@@ -350,15 +425,32 @@ void strong_branching(const user_problem_t<i_t, f_t>& original_problem,
...
+}
+
 template <typename i_t, typename f_t>
 void pseudo_costs_t<i_t, f_t>::update_pseudo_costs(mip_node_t<i_t, f_t>* node_ptr,
                                                    f_t leaf_objective)
 {
-  std::lock_guard<omp_mutex_t> lock(mutex);
-  const f_t change_in_obj = leaf_objective - node_ptr->lower_bound;
+  constexpr f_t eps       = 1e-6;
+  const f_t change_in_obj = std::max(leaf_objective - node_ptr->lower_bound, eps);
```
chris-maes: Same here: I would make change_in_ob = std::max(... , 0.0)

cpp/src/dual_simplex/pseudo_costs.cpp
```diff
@@ -412,13 +504,55 @@ i_t pseudo_costs_t<i_t, f_t>::variable_selection(const std::vector<i_t>& fractio
...
-  std::vector<f_t> score(num_fractional);
+  initialized(num_initialized_down, num_initialized_up, pseudo_cost_down_avg, pseudo_cost_up_avg);
 
+  log.printf("PC: num initialized down %d up %d avg down %e up %e\n",
+             num_initialized_down,
+             num_initialized_up,
+             pseudo_cost_down_avg,
+             pseudo_cost_up_avg);
+
+  for (auto j : fractional) {
```
chris-maes: j is an i_t. I would use i_t instead of auto to make it clear to the reader.

cpp/src/dual_simplex/pseudo_costs.cpp
```diff
@@ -432,39 +566,159 @@ i_t pseudo_costs_t<i_t, f_t>::variable_selection(const std::vector<i_t>& fractio
...
+  i_t reliable_threshold = settings.reliability_branching;
+  if (reliable_threshold < 0) {
+    const i_t max_threshold            = reliability_branching_settings.max_reliable_threshold;
+    const i_t min_threshold            = reliability_branching_settings.min_reliable_threshold;
+    const f_t iter_factor              = reliability_branching_settings.bnb_lp_factor;
+    const i_t iter_offset              = reliability_branching_settings.bnb_lp_offset;
+    const int64_t alpha                = iter_factor * branch_and_bound_lp_iters;
+    const int64_t max_reliability_iter = alpha + reliability_branching_settings.bnb_lp_offset;
+
+    f_t gamma =
```
chris-maes: Can we have a more descriptive name than gamma?

cpp/src/dual_simplex/pseudo_costs.cpp
```diff
@@ -432,39 +566,159 @@ i_t pseudo_costs_t<i_t, f_t>::variable_selection(const std::vector<i_t>& fractio
...
+    gamma = std::min(1.0, gamma);
+    gamma = std::max((alpha - strong_branching_lp_iter) / (strong_branching_lp_iter + 1.0), gamma);
+    reliable_threshold = (1 - gamma) * min_threshold + gamma * max_threshold;
+    reliable_threshold = strong_branching_lp_iter < max_reliability_iter ? reliable_threshold : 0;
+  }
+
+  std::vector<i_t> unreliable_list;
+  omp_mutex_t score_mutex;
+
+  for (auto j : fractional) {
```
chris-maes: Same here: `i_t` instead of `auto`

cpp/src/dual_simplex/pseudo_costs.cpp
```diff
@@ -432,39 +566,159 @@ i_t pseudo_costs_t<i_t, f_t>::variable_selection(const std::vector<i_t>& fractio
...
+  log.printf(
+    "RB iters = %d, B&B iters = %d, unreliable = %d, num_tasks = %d, reliable_threshold = %d\n",
+    strong_branching_lp_iter.load(),
+    branch_and_bound_lp_iters,
+    unreliable_list.size(),
+    num_tasks,
+    reliable_threshold);
+
+  // Shuffle the unreliable list so every variable has the same chance to be selected.
+  if (unreliable_list.size() > max_num_candidates) { worker->rng.shuffle(unreliable_list); }
```
chris-maes, line 637: Where are you setting the seed from the random number generator? Could you set it to the seed inside simplex_solver_settings.hpp so that all generators use the same seed. That makes it possible for us to expose the seed easily later on

cpp/src/dual_simplex/pseudo_costs.cpp
```diff
@@ -432,39 +566,159 @@ i_t pseudo_costs_t<i_t, f_t>::variable_selection(const std::vector<i_t>& fractio
...
+                                std::floor(solution[j]),
+                                upper_bound,
+                                branch_and_bound_lp_iter_per_node,
+                                start_time,
+                                reliability_branching_settings.upper_max_lp_iter,
+                                reliability_branching_settings.lower_max_lp_iter,
+                                strong_branching_lp_iter);
+
+      if (!std::isnan(obj)) {
+        f_t change_in_obj = std::max(obj - node_ptr->lower_bound, eps);
```
chris-maes, line 673: Same here: change_in_obj = std::max(..., 0.0)

cpp/src/dual_simplex/pseudo_costs.cpp
```diff
@@ -432,39 +566,159 @@ i_t pseudo_costs_t<i_t, f_t>::variable_selection(const std::vector<i_t>& fractio
...
+  }
+
+#pragma omp taskloop if (num_tasks > 1) priority(task_priority) num_tasks(num_tasks) \
+  shared(score_mutex)
+  for (i_t i = 0; i < num_candidates; ++i) {
+    const i_t j = unreliable_list[i];
+
+    if (toc(start_time) > settings.time_limit) { continue; }
+
+    pseudo_cost_mutex_down[j].lock();
```
chris-maes, line 651: Maybe not for this PR: but if two threads want deem the same variable unreliable, one thread will be locked out while the other solves the LP for the down branch. Could we have a thread try to grab the down lock and if it fails try to grab the up lock? That way we could have multiple threads working on the same variable.

cpp/src/dual_simplex/pseudo_costs.cpp
```diff
@@ -432,39 +566,159 @@ i_t pseudo_costs_t<i_t, f_t>::variable_selection(const std::vector<i_t>& fractio
...
+                                worker->leaf_problem.upper[j],
+                                upper_bound,
+                                branch_and_bound_lp_iter_per_node,
+                                start_time,
+                                reliability_branching_settings.upper_max_lp_iter,
+                                reliability_branching_settings.lower_max_lp_iter,
+                                strong_branching_lp_iter);
+
+      if (!std::isnan(obj)) {
+        f_t change_in_obj = std::max(obj - node_ptr->lower_bound, eps);
```
chris-maes, line 704: std::max(..., 0.0)

cpp/src/dual_simplex/pseudo_costs.cpp
```diff
@@ -432,39 +566,159 @@ i_t pseudo_costs_t<i_t, f_t>::variable_selection(const std::vector<i_t>& fractio
...
+        pseudo_cost_num_up[j]++;
+      }
+    }
+    pseudo_cost_mutex_up[j].unlock();
+
+    if (toc(start_time) > settings.time_limit) { continue; }
+
+    f_t score = calculate_pseudocost_score(j, solution, pseudo_cost_up_avg, pseudo_cost_down_avg);
+
+    if (std::lock_guard<omp_mutex_t> score_lock(score_mutex); score > max_score) {
```
chris-maes: Putting the lock guard inside the if predicate is even harder to read than before. You want these locks to be easy to spot for the reader. Could you please move the lock guard outside the if predicate.

cpp/src/dual_simplex/pseudo_costs.cpp
```diff
@@ -487,26 +739,15 @@ f_t pseudo_costs_t<i_t, f_t>::obj_estimate(const std::vector<i_t>& fractional,
...
-    } else {
-      pseudo_cost_down = pseudo_cost_down_avg;
-    }
-
-    if (pseudo_cost_num_up[j] != 0) {
-      pseudo_cost_up = pseudo_cost_sum_up[j] / pseudo_cost_num_up[j];
-    } else {
-      pseudo_cost_up = pseudo_cost_up_avg;
-    }
+  for (auto j : fractional) {
```
chris-maes: `auto` -> `i_t`

cpp/src/dual_simplex/pseudo_costs.hpp
```diff
@@ -52,17 +92,37 @@ class pseudo_costs_t {
                          const std::vector<f_t>& solution,
                          logger_t& log);
 
+  i_t reliable_variable_selection(mip_node_t<i_t, f_t>* node_ptr,
```
chris-maes, line 95: Nit: We try to follow the Google style guide where input arguments (those with const) appear first, followed by input output arguments, and then pure output arguments. 

https://google.github.io/styleguide/cppguide.html#Inputs_and_Outputs

"When ordering function parameters, put all input-only parameters before any output parameters. In particular, do not add new parameters to the end of the function just because they are new; place new input-only parameters before the output parameters. This is not a hard-and-fast rule. Parameters that are both input and output muddy the waters, and, as always, consistency with related functions may require you to bend the rule. Variadic functions may also require unusual parameter ordering."

chris-maes (approved):
Thanks for merging with cuts and congratulations on the number of problems solved to optimality with this PR Nicolas. 

Some minor feedback. Would appreciate if max(..,0.0) was used for the objective and lock guard moved out of if predicate before merging. Thank you! 

LGTM.

cpp/src/dual_simplex/pseudo_costs.cpp
nguidotti: In the `branch_and_bound_worker`. Each worker has it own `rng` with its own seed and stream (note that the `rng` cannot be shared between workers due to the internal state). They need to have separated seeds and stream in order to ensure proper randomness.

cpp/src/dual_simplex/pseudo_costs.cpp
nguidotti: Yes, but it do not think the additional complexity is worth it at the moment. According to the profiler results, there is little contention here as two threads are unlikely to try to initialize the same variable at the same time.

PR #779, opened 2026-01-15
Add C api callbacks for getting and setting solutions

akifcorduk:
This PR adds callbacks to C API. Users can now use C API to implement get_solution and set_solution callbacks.

This PR is a breaking PR that changes callback input/output pointers from device memory to host memory. We are also adding a user_data pointer for user to be more flexible inside the callbacks as well as dual bound for the B&B tree.

Previously callbacks were not enabled when third party presolved was enabled. Now we are able to handle uncrushing of solutions when third party presolve is enabled on get_solution callbacks. Registering a set_solution callback disables presolve as there is no way of crushing a solution in original space to reduced space when using third party presolve.

The examples and tests are modified to reflect the change.

## Summary by CodeRabbit

* **New Features**
  * MIP incumbent get/set callbacks with optional user_data and registration; server/client flag to enable set-callback behavior.

* **Behavior**
  * Incumbent records now include bound and user_data; solution-bound handling made thread-safe. Registering a set-solution callback may disable presolve and enables host-side incumbent injection.

* **Tests**
  * Added tests for get-only and get-and-set flows, user_data propagation, bound reporting, and presolve interactions.

* **Documentation**
  * Examples and docs updated to show callbacks, user_data, bound reporting, and set-callback usage.

### DIFF
--- a/cpp/src/mip/problem/presolve_data.cu
+++ b/cpp/src/mip/problem/presolve_data.cu
@@ -183,10 +183,6 @@ void presolve_data_t<i_t, f_t>::set_papilo_presolve_data(
                     original_num_variables);
     return;
   }
-  if (reduced_to_original.empty()) {
-    CUOPT_LOG_DEBUG("Papilo presolve data invalid: reduced_to_original is empty");
-    return;
-  }
   if (original_to_reduced.empty()) {
     CUOPT_LOG_DEBUG("Papilo presolve data invalid: original_to_reduced is empty");
     return;
@@ -232,8 +228,10 @@ template <typename i_t, typename f_t>
 void presolve_data_t<i_t, f_t>::papilo_uncrush_assignment(
   problem_t<i_t, f_t>& problem, rmm::device_uvector<f_t>& assignment) const
 {
-  if (papilo_presolve_ptr == nullptr) { return; }
-  cuopt_assert(!papilo_reduced_to_original_map.empty(), "Papilo reduced_to_original map is empty");
+  if (papilo_presolve_ptr == nullptr) {
+    CUOPT_LOG_INFO("Papilo presolve data not set, skipping uncrushing assignment");
+    return;
+  }
   cuopt_assert(assignment.size() == papilo_reduced_to_original_map.size(),
                "Papilo uncrush assignment size mismatch");
   auto h_assignment = cuopt::host_copy(assignment, problem.handle_ptr->get_stream());


cpp/include/cuopt/linear_programming/utilities/internals.hpp
```diff
@@ -31,16 +31,20 @@ class base_solution_callback_t : public Callback {
     this->n_variables = n_variables_;
   }
 
+  void set_user_data(void* user_data_) { user_data = user_data_; }
```
chris-maes: Nit: Google C++ style guide and other common conventions reserves the underscore postfix to refer to the protected member variable. So it's a bit confusing to have `user_data_` be the input argument, as this is usually the name of the member variable. Since you aren't using underscores for member variable names here, maybe use a different name for the argument (e.g. `input_user_data`)

cpp/include/cuopt/linear_programming/cuopt_c.h
```diff
@@ -681,6 +681,56 @@ cuopt_int_t cuOptGetFloatParameter(cuOptSolverSettings settings,
...
                                    cuopt_float_t* parameter_value);
 
+/**
+ * @brief Callback for receiving incumbent MIP solutions with user context.
+ *
+ * @param[in] solution - Device pointer to incumbent solution values.
+ * @param[in] objective_value - Device pointer to incumbent objective value.
+ * @param[in] user_data - User context pointer.
+ */
+typedef void (*cuOptMipGetSolutionCallback)(const cuopt_float_t* solution,
```
chris-maes: In other parts of the C API we capitalize MIP. So I would call this `cuOptMIPGetSolutionCallback`

cpp/include/cuopt/linear_programming/cuopt_c.h
```diff
@@ -681,6 +681,56 @@ cuopt_int_t cuOptGetFloatParameter(cuOptSolverSettings settings,
                                    const char* parameter_name,
                                    cuopt_float_t* parameter_value);
 
+/**
+ * @brief Callback for receiving incumbent MIP solutions with user context.
```
chris-maes: Nit: I would say `Signature of callback for receiving ...` or `Type of callback for receiving ...` so it is clear to the reader that this is declaring the type.

cpp/include/cuopt/linear_programming/cuopt_c.h
```diff
@@ -681,6 +681,56 @@ cuopt_int_t cuOptGetFloatParameter(cuOptSolverSettings settings,
...
+                                            void* user_data);
+
+/**
+ * @brief Callback for injecting MIP solutions with user context.
+ *
+ * @param[out] solution - Device pointer to solution values to set.
+ * @param[out] objective_value - Device pointer to objective value to set.
+ * @param[in] user_data - User context pointer.
+ */
+typedef void (*cuOptMipSetSolutionCallback)(cuopt_float_t* solution,
```
chris-maes: Same here: I would call this `cuOptMIPSetSolutionCallback` for consistency with the rest of the API

cpp/include/cuopt/linear_programming/cuopt_c.h
```diff
@@ -681,6 +681,56 @@ cuopt_int_t cuOptGetFloatParameter(cuOptSolverSettings settings,
                                    const char* parameter_name,
                                    cuopt_float_t* parameter_value);
 
+/**
+ * @brief Callback for receiving incumbent MIP solutions with user context.
+ *
+ * @param[in] solution - Device pointer to incumbent solution values.
+ * @param[in] objective_value - Device pointer to incumbent objective value.
+ * @param[in] user_data - User context pointer.
```
chris-maes: Maybe `User context pointer` -> `Pointer to user data`

cpp/include/cuopt/linear_programming/cuopt_c.h
```diff
@@ -681,6 +681,56 @@ cuopt_int_t cuOptGetFloatParameter(cuOptSolverSettings settings,
...
+ * @brief Register a callback to receive incumbent MIP solutions.
+ *
+ * @param[in] settings - The solver settings object.
+ * @param[in] callback - Callback function to receive incumbent solutions.
+ * @param[in] user_data - User-defined pointer passed through to the callback.
+ *  It will be forwarded to ``cuOptMipGetSolutionCallback`` when invoked.
+ *
+ * @return A status code indicating success or failure.
+ */
+cuopt_int_t cuOptSetMipGetSolutionCallback(cuOptSolverSettings settings,
```
chris-maes: Change to `cuOptSetMIPSolutionCallback`

cpp/include/cuopt/linear_programming/cuopt_c.h
```diff
@@ -681,6 +681,56 @@ cuopt_int_t cuOptGetFloatParameter(cuOptSolverSettings settings,
...
+ * @param[in] user_data - User context pointer.
+ */
+typedef void (*cuOptMipGetSolutionCallback)(const cuopt_float_t* solution,
+                                            const cuopt_float_t* objective_value,
+                                            void* user_data);
+
+/**
+ * @brief Callback for injecting MIP solutions with user context.
+ *
+ * @param[out] solution - Device pointer to solution values to set.
```
chris-maes: We probably should mention the dimension of the `solution_values` (i.e. that it needs to be equal to the number of variables in the original problem).

Also, do we crush the solution from the original variables into the presolve variables when MIP presolve is turned on?

cpp/include/cuopt/linear_programming/cuopt_c.h
```diff
@@ -681,6 +681,56 @@ cuopt_int_t cuOptGetFloatParameter(cuOptSolverSettings settings,
...
+ * @brief Register a callback to inject MIP solutions.
+ *
+ * @param[in] settings - The solver settings object.
+ * @param[in] callback - Callback function to inject solutions.
+ * @param[in] user_data - User-defined pointer passed through to the callback.
+ *  It will be forwarded to ``cuOptMipSetSolutionCallback`` when invoked.
+ *
+ * @return A status code indicating success or failure.
+ */
+cuopt_int_t cuOptSetMipSetSolutionCallback(cuOptSolverSettings settings,
```
chris-maes: Change to `cuOptSetMIPSolutionCallback`

cpp/include/cuopt/linear_programming/cuopt_c.h
```diff
@@ -681,6 +681,56 @@ cuopt_int_t cuOptGetFloatParameter(cuOptSolverSettings settings,
                                    const char* parameter_name,
                                    cuopt_float_t* parameter_value);
 
+/**
+ * @brief Callback for receiving incumbent MIP solutions with user context.
+ *
+ * @param[in] solution - Device pointer to incumbent solution values.
+ * @param[in] objective_value - Device pointer to incumbent objective value.
```
aliceb-nv: I wonder if these should be host pointers instead, in terms of API design. So far the API has been abstracting away the reality of the CUDA implementation (such as the corresponding context, the streams used...) away from the user, so this may represent a break in the philosophy. Plus we don't expose the GPU problem structure either, so it is likely the user will have to copy it back to the host either way to do any useful work
Maybe device-space callbacks could be implemented later on if need be? We might have to discuss this

chris-maes: I agree. The rest of the API uses abstracts out the device. Probably we should provide a host pointer here.

cpp/include/cuopt/linear_programming/cuopt_c.h
chris-maes: And this should probably be a host pointer rather than a device pointer

akifcorduk: @chris-maes Good catch! All callbacks are disabled if 3rd party presolve(Papilo) is enabled. Our own presolve changes are handled though. I believe this is a blocker for primal integral experiments. I will give it a try.

cpp/include/cuopt/linear_programming/cuopt_c.h
akifcorduk: We are agnostic to device/host memory in all APIs AFAIK. I can modify it to "Host/Device" pointer? The only caveat of supporting device pointers officialy is that the copies will be done by internal stream.

akifcorduk: I made all callbacks to input/output host pointers.

cpp/include/cuopt/linear_programming/cuopt_c.h
akifcorduk: I have enabled third party presolve and handled crushing/uncrushing of solutions.

cpp/src/mip/presolve/third_party_presolve.cpp
```diff
@@ -500,6 +510,22 @@ void third_party_presolve_t<i_t, f_t>::undo(rmm::device_uvector<f_t>& primal_sol
...
+  papilo::Solution<f_t> full_sol;
+  papilo::Message Msg{};
+  Msg.setVerbosityLevel(papilo::VerbosityLevel::kQuiet);
+  papilo::Postsolve<f_t> post_solver{Msg, post_solve_storage_.getNum()};
+
+  bool is_optimal = false;
+  auto status     = post_solver.undo(reduced_sol, full_sol, post_solve_storage_, is_optimal);
+  check_postsolve_status(status);
+  full_primal = std::move(full_sol.primal);
+}
```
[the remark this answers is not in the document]
akifcorduk: uncrush_will only be called from a single thread. so there is no need to worry about thread-safety here.

cpp/include/cuopt/linear_programming/cuopt_c.h
```diff
@@ -681,6 +681,77 @@ cuopt_int_t cuOptGetFloatParameter(cuOptSolverSettings settings,
...
+ * @note Registering a set-solution callback disables presolve.
+ * @note The callback arguments refer to host memory and are only valid during the callback
+ * invocation. Do not pass device/GPU pointers. Copy any data you need to keep after the callback
+ * returns.
+ *
+ * @return A status code indicating success or failure.
+ */
+cuopt_int_t cuOptSetMIPSetSolutionCallback(cuOptSolverSettings settings,
+                                           cuOptMIPSetSolutionCallback callback,
+                                           void* user_data);
```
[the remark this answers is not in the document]
akifcorduk: That's not correct. presolve is disabled when set-solution callback is enabled.

cpp/include/cuopt/linear_programming/mip/solver_stats.hpp
```diff
@@ -1,17 +1,41 @@
...
+
+  f_t get_solution_bound() const { return solution_bound.load(std::memory_order_relaxed); }
+
+  void set_solution_bound(f_t value) { solution_bound.store(value, std::memory_order_relaxed); }
+
+  f_t total_solve_time = 0.;
+  f_t presolve_time    = 0.;
+  std::atomic<f_t> solution_bound;
   i_t num_nodes              = 0;
   i_t num_simplex_iterations = 0;
```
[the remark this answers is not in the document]
akifcorduk: We don't officially support c++ api, so it is an intermediate class which is converted to public apis which are C and python.

cpp/src/linear_programming/cuopt_c.cpp
```diff
@@ -574,16 +628,15 @@ cuopt_int_t cuOptGetVariableTypes(cuOptOptimizationProblem problem, char* variab
 cuopt_int_t cuOptCreateSolverSettings(cuOptSolverSettings* settings_ptr)
 {
   if (settings_ptr == nullptr) { return CUOPT_INVALID_ARGUMENT; }
-  solver_settings_t<cuopt_int_t, cuopt_float_t>* settings =
-    new solver_settings_t<cuopt_int_t, cuopt_float_t>();
-  *settings_ptr = static_cast<cuOptSolverSettings>(settings);
+  solver_settings_handle_t* settings_handle = new solver_settings_handle_t();
```
nguidotti, line 627: It is not better to use an `unique_ptr` with `make_unique` and `reset` here, then the `new/delete`? Just to avoid memory leaks if the user forget to call `cuOptDestroySolverSettings`.

akifcorduk: `new` is used everywhere in C API implementation, i wanted to stay consistent. Maybe we can consider it for anotehr PR.

PR #786, opened 2026-01-20
Several bug fixes

nguidotti:
- Fixed early termination in `CMS750_4` due to an incorrect root objective from the PDLP. It provides an objective in user space and B&B expects the objective in the solver space.
- Fixed setting the number of threads to a fixed value when running `cuopt_cli`. Now the number of threads in the probing cache can be controlled by the `bounds_update::settings_t`.
- Inverted loop order in probing cache to avoid creating and deleting the OpenMP thread pool each iteration.
- PDLP/Barrier no longer set the root objective and solution when the B&B is already running.
- Improved the logs when solving the root relaxation using the concurrent mode. 

## Checklist

- Testing
   - [ ] Added tests
   - [ ] NA
- Documentation
   - [ ] Added new documentation
   - [ ] NA

## Summary by CodeRabbit

* **New Features**
  * Per-instance threading for presolve with sensible defaults and clamping.
  * Bidirectional objective conversion so callbacks receive both solver-space and user-space objective values.

* **Bug Fixes**
  * Root relaxation solution handling made idempotent to avoid overwrites.
  * Reduced verbose logging inside MIP solves and refined root/crossover status messages.

* **Style**
  * Updated header/license years and minor logging formatting improvements.

### DIFF
--- a/cpp/src/dual_simplex/branch_and_bound.cpp
+++ b/cpp/src/dual_simplex/branch_and_bound.cpp
@@ -1370,14 +1370,20 @@ lp_status_t branch_and_bound_t<i_t, f_t>::solve_root_relaxation(
   }
 
   settings_.log.printf("\n");
-  settings_.log.printf("Root relaxation solution found in %d iterations and %.2fs by %s\n",
-                       iter,
-                       toc(start_time),
-                       solver_name.c_str());
-  settings_.log.printf("Root relaxation objective %+.8e\n", user_objective);
-  settings_.log.printf("\n");
+  if (root_status == lp_status_t::OPTIMAL) {
+    settings_.log.printf("Root relaxation solution found in %d iterations and %.2fs by %s\n",
+                         iter,
+                         toc(start_time),
+                         solver_name.c_str());
+    settings_.log.printf("Root relaxation objective %+.8e\n", user_objective);
+  } else {
+    settings_.log.printf("Root relaxation returned status: %s\n",
+                         lp_status_to_string(root_status).c_str());
+  }
 
+  settings_.log.printf("\n");
   is_root_solution_set = true;
+
   return root_status;
 }
 
--- a/cpp/src/dual_simplex/solve.hpp
+++ b/cpp/src/dual_simplex/solve.hpp
@@ -1,6 +1,6 @@
 /* clang-format off */
 /*
- * SPDX-FileCopyrightText: Copyright (c) 2025, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ * SPDX-FileCopyrightText: Copyright (c) 2025-2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
  * SPDX-License-Identifier: Apache-2.0
  */
 /* clang-format on */
@@ -29,6 +29,22 @@ enum class lp_status_t {
   CONCURRENT_LIMIT = 7,
   UNSET            = 8
 };
+
+static std::string lp_status_to_string(lp_status_t status)
+{
+  switch (status) {
+    case lp_status_t::OPTIMAL: return "OPTIMAL";
+    case lp_status_t::INFEASIBLE: return "INFEASIBLE";
+    case lp_status_t::UNBOUNDED: return "UNBOUNDED";
+    case lp_status_t::ITERATION_LIMIT: return "ITERATION_LIMIT";
+    case lp_status_t::TIME_LIMIT: return "TIME_LIMIT";
+    case lp_status_t::NUMERICAL_ISSUES: return "NUMERICAL_ISSUES";
+    case lp_status_t::CUTOFF: return "CUTOFF";
+    case lp_status_t::CONCURRENT_LIMIT: return "CONCURRENT_LIMIT";
+    case lp_status_t::UNSET: return "UNSET";
+  }
+  return "UNKNOWN";
+}
 
 template <typename i_t, typename f_t>
 f_t compute_objective(const lp_problem_t<i_t, f_t>& problem, const std::vector<f_t>& x);


cpp/src/mip/presolve/probing_cache.cu
```diff
@@ -879,12 +877,14 @@ bool compute_probing_cache(bound_presolve_t<i_t, f_t>& bound_presolve,
...
-    if (timer.check_time_limit() || early_exit || problem_is_infeasible.load()) { break; }
-    size_t step_end = std::min(step_start + step_size, priority_indices.size());
+
 // Main parallel loop
-#pragma omp parallel
-    {
+#pragma omp parallel num_threads(num_threads)
+  {
+    for (size_t step_start = 0; step_start < priority_indices.size(); step_start += step_size) {
+      if (timer.check_time_limit() || early_exit || problem_is_infeasible.load()) { break; }
```
akifcorduk, line 888: I think this is not the same behavior as before. The logic depends on the fact that threads synchonize at the end of parallel block some problem modifications and checks are done and then new parallel block is launched. You can see the modification on omp single part. This change will cause race conditions.

cpp/src/mip/diversity/diversity_manager.cu
```diff
@@ -434,15 +434,13 @@ solution_t<i_t, f_t> diversity_manager_t<i_t, f_t>::run_solver()
                  problem_ptr->handle_ptr->get_stream());
       problem_ptr->handle_ptr->sync_stream();
 
-      auto user_obj   = problem_ptr->get_user_obj_from_solver_obj(lp_result.get_objective_value());
+      // PDLP returns user-space objective (it applies objective_scaling_factor internally)
+      auto user_obj   = lp_result.get_objective_value();
```
rg20, line 438: Is this uniform across the solvers? Or just PDLP behaves differently than others? We need to standardize this.

cpp/src/mip/problem/problem.cu
```diff
@@ -1988,6 +1988,13 @@ void problem_t<i_t, f_t>::get_host_user_problem(
         : cuopt::linear_programming::dual_simplex::variable_type_t::INTEGER;
   }
 }
+
+template <typename i_t, typename f_t>
+f_t problem_t<i_t, f_t>::get_solver_obj_from_user_obj(f_t user_obj) const
+{
+  return (user_obj / presolve_data.objective_scaling_factor) - presolve_data.objective_offset;
```
rg20, line 1995: Should n't this be: 

`(user_obj + presolve_data.objective_offset) / presolve_data.objective_scaling_factor` ?

cpp/src/mip/presolve/probing_cache.cu
nguidotti: There is an implicit barrier at the end of the `for` and `single` directives. So the behavior should be exactly the same: for loop --> synchronize --> single --> synchronize

cpp/src/mip/presolve/probing_cache.cu
```diff
@@ -856,17 +856,15 @@ bool compute_probing_cache(bound_presolve_t<i_t, f_t>& bound_presolve,
   bound_presolve.settings.iteration_limit = 50;
   bound_presolve.settings.time_limit      = timer.remaining_time();
 
-  // Set the number of threads
-  const size_t max_threads = 8;
-  omp_set_num_threads(max_threads);
+  const int num_threads = 8;
```
chris-maes: Do we want to hard code this to 8 threads? Shouldn't we be looking at a user setting here?

cpp/src/mip/problem/problem.cu
nguidotti: @aliceb-nv Is this correct?

cpp/src/mip/presolve/probing_cache.cu
nguidotti: Sure, I will move to the user settings.

cpp/src/mip/presolve/probing_cache.cu
akifcorduk: Didn't realize that actual parallel loop is later. Thanks

cpp/src/linear_programming/solve.cu
```diff
@@ -42,6 +42,9 @@
 
 #include <thread>  // For std::thread
 
+#define CUOPT_LP_SOLVER_LOG_INFO(...) \
```
akifcorduk: Can't we undef the previous macro and still use CUOPT_LOG_INFO here?

nguidotti: I think it is better to add another macro, since there are some methods where we do not pass the `setting` as argument, hence they need to use the old version.

nguidotti: Ideally, we should add  `logger_t` to the settings rather than `CUOPT_LOG_INFO` or `rapids::logger_t`.

cpp/src/dual_simplex/branch_and_bound.hpp
```diff
@@ -82,16 +85,24 @@ class branch_and_bound_t {
...
   {
-    root_crossover_soln_.x              = primal;
-    root_crossover_soln_.y              = dual;
-    root_crossover_soln_.z              = reduced_costs;
-    root_objective_                     = objective;
-    root_crossover_soln_.objective      = objective;
-    root_crossover_soln_.user_objective = user_objective;
-    root_crossover_soln_.iterations     = iterations;
-    root_crossover_solution_set_.store(true, std::memory_order_release);
+    if (root_solver_type_ == root_solver_type_t::NONE) {
```
chris-maes: Let's move this printing to inside `solve_root_relaxation` and only print when crossover is successful. 

Otherwise we could tell the user we have found a solution with PDLP/Barrier. And then the user could wait a long time only to have us use the dual simplex solution because we either: a) failed in crossover, or b) crossover took a very long time.

cpp/src/dual_simplex/branch_and_bound.hpp
```diff
@@ -162,6 +173,7 @@ class branch_and_bound_t {
   std::atomic<bool> root_crossover_solution_set_{false};
   bool enable_concurrent_lp_root_solve_{false};
   std::atomic<int> root_concurrent_halt_{0};
+  std::atomic<root_solver_type_t> root_solver_type_{root_solver_type_t::NONE};
```
hlinsen: Does this need to be an atomic? I think we can never have a race condition because we write to this var only when one of the solves is optimal which means the other solver returned with halt status.

cpp/src/dual_simplex/branch_and_bound.hpp
```diff
@@ -70,6 +70,9 @@ struct bnb_stats_t {
 template <typename i_t, typename f_t>
 class branch_and_bound_t {
  public:
+  // Specify which solver was used for solving the root LP relaxation
+  enum class root_solver_type_t { NONE = 0, CROSSOVER = 1, DUAL_SIMPLEX = 2 };
```
chris-maes: Do we need this? Can we just determine who won in `solve_root_relaxation`?

cpp/src/dual_simplex/branch_and_bound.cpp
```diff
@@ -1338,23 +1340,24 @@ lp_status_t branch_and_bound_t<i_t, f_t>::solve_root_relaxation(
                                                     root_crossover_soln_,
                                                     crossover_vstatus_);
 
-    if (crossover_status == crossover_status_t::OPTIMAL) {
-      settings_.log.printf("Crossover status: %d\n", crossover_status);
-    }
-
     // Check if crossover was stopped by dual simplex
     if (crossover_status == crossover_status_t::OPTIMAL) {
+      settings_.log.printf("\nCrossover found an optimal solution for the root relaxation\n\n");
```
chris-maes: This is where we should indicate that PDLP/Barrier won. Can you use the is_pdlp_solution to determine which of them won and print it here?

cpp/src/dual_simplex/branch_and_bound.cpp
```diff
@@ -1308,6 +1308,8 @@ lp_status_t branch_and_bound_t<i_t, f_t>::solve_root_relaxation(
   }
 
   if (root_crossover_solution_set_.load(std::memory_order_acquire)) {
+    settings_.log.printf("\nRunning crossover\n\n");
```
chris-maes: I would remove this print. I don't think we want to interrupt the dual simplex logs for the root solve, unless we know that crossover was successful.

cpp/src/dual_simplex/branch_and_bound.cpp
nguidotti: I did not find a way to get the winner solver from the `optimization_problem_solution_t` unless we change the API (#787). This object is user-facing, so it needs to be reflect on the Python side as well.

PR #804, opened 2026-01-27
tighten wheel size limits, expand CI-skipping logic, other small build changes

jameslamb:
## Description

Proposes a batch of miscellaneous build / packaging / CI changes.

### Tightens wheel size limits

Contributes to https://github.com/rapidsai/build-planning/issues/219

CUDA 13 wheels can be as much as 50% smaller than the equivalent CUDA 12 wheels, because of some new compression features in `nvcc`.

To ensure surprising package-size growth is caught in CI, this PR tightens the limits in the following ways:

* setting different limits based on CUDA major version where package size differs significantly
* setting `libcuopt` limits to `{compressed_size} + 50Mi`, rounded to the nearest 5Mi
* setting all other limits to `{compressed_size} + 10Mi`, rounded to the nearest 5Mi

### Expands CI-skipping logic

Contributes to https://github.com/rapidsai/build-planning/issues/243

Tries to avoid unnecessary CI runs by making the CI-skipping rules finer-grained. For example, PRs that only touch `.pre-commit-config.yaml` should now not require any runners with GPUs 😁 

### Removes some reliance on `pypi.nvidia.com`

Contributes to https://github.com/rapidsai/build-planning/issues/241

```shell
git grep -i -E 'pypi\.nvidia\.com'
git grep -i -E 'rapids\-init\-pip'
```

And removed/updated all relevant references. This project does not need any wheels from `pypi.nvidia.com` at build-time or runtime, it can safely avoid searching that index.

Its own packages are not available on `pypi.org` though, so uses of `pypi.nvidia.com` in docs are preserved here.

### Enforces PEP 639 license metadata in `pyproject.toml`

Contributes to https://github.com/rapidsai/pre-commit-hooks/issues/95

## Checklist

- Testing
- Documentation

## Summary by CodeRabbit

* **New Features**
  * Added a public CLI entrypoint (cuopt_cli).

* **Chores**
  * More granular CI gating so jobs run based on changed files.
  * Updated pre-commit hooks to rev 1.3.3 and added license verification.
  * Removed NVIDIA PyPI extra-index usage and introduced an opt-out env flag for CI pip init.
  * Tightened and standardized compressed package size limits and improved packaging validation.

### DIFF
--- a/python/cuopt/pyproject.toml
+++ b/python/cuopt/pyproject.toml
@@ -75,7 +75,7 @@ select = [
 ]
 
 # PyPI hard limit is 1GiB, but try to keep this as small as possible
-max_allowed_size_compressed = '15Mi'
+max_allowed_size_compressed = '55Mi'
 
 [tool.pytest.ini_options]
 testpaths = ["cuopt/tests"]


.github/workflows/pr.yaml
```diff
@@ -210,7 +291,7 @@ jobs:
   conda-python-tests:
     needs: [conda-python-build, changed-files, compute-matrix-filters]
     uses: rapidsai/shared-workflows/.github/workflows/conda-python-tests.yaml@main
-    #if: fromJSON(needs.changed-files.outputs.changed_file_groups).test_python
+    if: fromJSON(needs.changed-files.outputs.changed_file_groups).test_python_conda
```
jameslamb, line 292: These skips have all been disabled for the last year: https://github.com/rapidsai/cuopt/pull/2244

From that PR, I'm really not sure why or what the issue was. But either way, that's a year of running more CI than necessary, which is a waste of money and time. Hopefully we can re-enable this now... it'd help get changes integrated faster and reduce `cuopt`'s demands on our shared CI runners.

ci/thirdparty-testing/run_cvxpy_tests.sh
```diff
@@ -25,7 +25,6 @@ pip wheel \
 echo "installing 'cvxpy' with cuopt"
 python -m pip install \
     --constraint "${PIP_CONSTRAINT}" \
-    --extra-index-url=https://pypi.nvidia.com \
```
jameslamb, line 28: This shouldn't ever be necessary.

The only packages installed in this command that aren't on `pypi.org` are cuOpt packages themselves, which should all have been downloaded from the same CI run before this runs.

.github/workflows/pr.yaml
```diff
@@ -111,66 +111,147 @@ jobs:
     uses: rapidsai/shared-workflows/.github/workflows/changed-files.yaml@main
     with:
       files_yaml: |
-        test_cpp:
+        build_docs:
```
jameslamb, line 114: docs builds use GPU runners. Those runners are scarce, so it's worth skipping those jobs when they're not necessary (like when a PR only changes the Helm chart).

.github/workflows/pr.yaml
```diff
@@ -111,66 +111,147 @@ jobs:
...
+          - '!ci/run_ctests.sh'
+          - '!ci/run_*.pytests.sh'
+          - '!ci/test_cpp*.sh'
+          - '!ci/test_notebooks.sh'
+          - '!ci/test_python.sh'
+          - '!ci/test_self_hosted_service.sh'
+          - '!ci/test_wheel*.sh'
           - '!container-builder/**'
-          - '!helm-chart/**'
+          - '!helmchart/**'
```
jameslamb, line 141: This directory is called `helmchart/` (no `-`). This was wrong before.

.github/workflows/pr.yaml
```diff
@@ -111,66 +111,147 @@ jobs:
...
           - '!ngc/**'
-          - '!omniverse/**'
-          - '!regression/**'
-          - '!resources/**'
```
jameslamb, line 130: These directories no longer exist in the repo.

.github/workflows/pr.yaml
```diff
@@ -111,66 +111,147 @@ jobs:
...
+          - '!container-builder/**'
           - '!docs/**'
+          - '!helmchart/**'
           - '!img/**'
+          - '!ngc/**'
           - '!notebooks/**'
-          - '!python/cuopt_self_hosted/**'
-          - '!python/cuopt_server/**'
-          - '!python/nvcf_client/**'
-        test_python_cuopt_server:
```
jameslamb, line 163: I'm hoping the `test_python_wheels:` group is granular enough that we don't need `test_python_cuopt_server` separately.

But please do break it down into more-specific groups if you feel that's doable! The more unnecessary CI runs we can save, the better 😁

python/libcuopt/pyproject.toml
```diff
@@ -49,15 +49,14 @@ Source = "https://github.com/nvidia/cuopt"
 [project.entry-points."cmake.prefix"]
 libcuopt = "libcuopt"
 
+[project.scripts]
+cuopt_cli = "libcuopt._cli_wrapper:main"
```
jameslamb, line 52: This needed to be moved up to work around this: https://github.com/rapidsai/pre-commit-hooks/issues/105

dependencies.yaml
```diff
@@ -351,11 +351,6 @@ dependencies:
           - &pandas pandas>=2.0
           - &pyyaml pyyaml>=6.0.0
           - scipy>=1.14.1
-      - output_types: requirements
-        packages:
-          # pip recognizes the index as a global option for the requirements.txt file
-          - --extra-index-url=https://pypi.nvidia.com
-          - --extra-index-url=https://pypi.anaconda.org/rapidsai-wheels-nightly/simple
```
jameslamb, line 358: every package in this group is available from `pypi.org`

ci/validate_wheel.sh
```diff
@@ -1,19 +1,44 @@
...
+# SPDX-FileCopyrightText: Copyright (c) 2024-2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 
 set -euo pipefail
 
 package_dir=$1
 wheel_dir_relative_path=$2
 
+RAPIDS_CUDA_MAJOR="${RAPIDS_CUDA_VERSION%%.*}"
+
```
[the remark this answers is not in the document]
jameslamb: No. We want this failure if that variable's unset. "unbound variable RAPIDS_CUDA_VERSION" is clear enough, we don't need and explicit guard and custom error message.

.github/workflows/pr.yaml
```diff
@@ -111,66 +111,147 @@ jobs:
...
           - '!img/**'
+          - '!ngc/**'
           - '!notebooks/**'
-          - '!python/cuopt_self_hosted/**'
-          - '!python/nvcf_client/**'
+          - '!python/**'
+          - '!sonar-project.properties'
+          - '!sonarqube/**'
+          - '!ucf/**'
+          - '!utilities/**'
```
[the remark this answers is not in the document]
jameslamb: Yes that's a pretty big omission! Thank you.

.github/workflows/pr.yaml
```diff
@@ -111,66 +111,147 @@ jobs:
...
+          - '!conda/**'
+          - '!container-builder/**'
           - '!docs/**'
+          - '!helmchart/**'
           - '!img/**'
+          - '!ngc/**'
           - '!notebooks/**'
-          - '!python/cuopt_self_hosted/**'
-          - '!python/nvcf_client/**'
+          - '!python/**'
```
```suggestion
```

.github/workflows/pr.yaml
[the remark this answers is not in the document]
jameslamb: fixed in https://github.com/NVIDIA/cuopt/pull/804/changes/9a7902d3bab4a611a6515755546f03ba7d4d146a

.github/workflows/pr.yaml
```diff
@@ -111,66 +111,146 @@ jobs:
...
+          - '!.ai/**'
+          - '!.coderabbit.yaml'
+          - '!.github/AGENTS.md'
+          - '!.github/CODE_OF_CONDUCT.md'
+          - '!.github/CODEOWNERS'
+          - '!.github/ISSUE_TEMPLATE/**'
+          - '!.github/PULL_REQUEST_TEMPLATE.md'
+          - '!.github/SECURITY.md'
+          - '!.github/.ai/**'
+          - '!.github/.coderabbit_review_guid.md'
```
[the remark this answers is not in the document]
jameslamb: you are right, thank you. Fixed in https://github.com/NVIDIA/cuopt/pull/804/commits/0726dec6e53611603ad094267e2aed5f834b506c

rgsl888prabhu:
> It seems that some recent changes (maybe #805) added 10s of MB to the package sizes for `libcuopt` and `cuopt` wheels... I've had to adjust the thresholds a few times since the one I set when I first opened this PR, which were based on a recent nightly build.
> 
> Not a complaint, just letting you know @rgsl888prabhu
> 
> Merging this PR will make such binary-size growth more obvious in the future so it's not a surprise.

Thank you @jameslamb, this would definitely help. There might be additional growth in size due the grpc issue I was talking to you. In case of conda I think I have to build it as static and in case of wheel, I might need to bundle the .so with wheel.

PR #799, opened 2026-01-26
WIP: Splitting PR (Add gRPC server and remote execution integrated in cuOpt solver)

tmckayus:
This change adds a new remote execution system which will ultimately replace the cuopt Python server. In this new implementation, remote solves are identical to local solves except that the env vars CUOPT_REMOTE_HOST and CUOPT_REMOTE_PORT are set to indicate a remote server.

The remote server code is included and can be built with "build.sh cuopt_grpc_server". It builds as part of the default build.

Internal code paths have been modified to use data_model_view_t with non-owning pointers to store problems and solutions as GPU memory or CPU memory, the former for local solves and the latter for remote solves.

The server is built with gRPC and has support for TLS and mTLS.

## Summary by CodeRabbit

* **New Features**
  * Remote solving for LP and MIP over gRPC (submit, poll, stream/retrieve results, cancel) and a new cuopt_grpc_server executable
  * Protobuf-backed problem/solution serialization and pluggable serializers
  * CLI: added --print-grpc-max diagnostic option

* **Documentation**
  * Added gRPC architecture, serialization plugin guide, and solution memory architecture docs

* **Quality of Life**
  * Python package and utilities now use lazy-loading to reduce import overhead

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

PR #807, opened 2026-01-28
Memory model changes preliminary to CPU-host remote solve

tmckayus:
This change modifies the memory model for problems and solutions so that CPU-only hosts will be able to do remote solves in the future (remote solve is stubbed out here).

If remote solving is enabled, then problems and solutions will be in host memory. For local solves, problems will be moved to GPU memory before the solve (if not already in GPU memory) and solutions will be in GPU memory.

The data_model_view_t from the mps parser is used to store non-owning pointers to CPU or GPU data for problems. A similar setup is used for solutions.

## Summary by CodeRabbit

* **New Features**
  * Remote LP/MIP solve via env-configured host/port; lazy-loaded submodules to avoid unnecessary init.

* **Enhancements**
  * Unified memory model with host/device residency, host accessors and on-demand transfers; view-based solve APIs and lazy CUDA initialization; clearer logging and remote/local dispatch.

* **Documentation**
  * Added memory-model overview describing local vs remote execution and data residency.

* **Tests**
  * Expanded C++ and Python tests for memory model, remote stubs, conversions, and import/edge cases.

### DIFF
--- a/cpp/src/linear_programming/utilities/cython_solve.cu
+++ b/cpp/src/linear_programming/utilities/cython_solve.cu
@@ -17,6 +17,7 @@
 #include <mps_parser/mps_data_model.hpp>
 #include <mps_parser/writer.hpp>
 #include <utilities/copy_helpers.hpp>
+#include <utilities/macros.cuh>
 
 #include <raft/common/nvtx.hpp>
 #include <raft/core/handle.hpp>


tmckayus:
This PR is a portion of #799, which will be split into multiple PRs.  This change covers the memory model changes in the engine, and local solves should behave the same.

cpp/include/cuopt/linear_programming/data_model_view.hpp
```diff
@@ -0,0 +1,58 @@
+/* clang-format off */
```
tmckayus, line 1: Instead of this alias, the data_model_view_t could be moved to a small common library shared by mps parser and libcuopt.  Currently mps parser is also statically linked into libcuopt to avoid runtime path issues. A small common library would eliminate that too, but that could be done as a second pass.

cpp/libmps_parser/CMakeLists.txt
```diff
@@ -77,6 +78,12 @@ add_library(mps_parser SHARED
   src/utilities/cython_mps_parser.cpp
 )
 
+# Shared library for standalone use
+add_library(mps_parser SHARED ${MPS_PARSER_SOURCES})
+
+# Static library for linking into libcuopt
+add_library(mps_parser_static STATIC ${MPS_PARSER_SOURCES})
```
tmckayus, line 85: This is for static linking into libcuopt, since the data_model_view_t is shared now. If we move that type to a common .so, we could eliminate this

cpp/src/linear_programming/utilities/cython_solve.cu
```diff
@@ -327,7 +459,13 @@ std::pair<std::vector<std::unique_ptr<solver_ret_t>>, double> call_batch_solve(
   auto start_solver = std::chrono::high_resolution_clock::now();
 
   // Limit parallelism as too much stream overlap gets too slow
-  const int max_thread = compute_max_thread(data_models);
+  int max_thread = 1;
+  if (linear_programming::is_remote_solve_enabled()) {
```
tmckayus: batch should be handled by sending the whole batch to the remote server and letting it run batch mode. A check has been added in solve_lp to throw an error currently if batch is used with remote solve.

This change on max_thread will be removed in an update

cpp/include/cuopt/linear_programming/mip/solver_solution.hpp
```diff
@@ -75,8 +135,105 @@ class mip_solution_t : public base_solution_t {
...
  private:
-  rmm::device_uvector<f_t> solution_;
+  // GPU (device) storage - populated for local GPU solves
+  std::unique_ptr<rmm::device_uvector<f_t>> solution_;
+
+  // CPU (host) storage - populated for remote solves
+  std::unique_ptr<std::vector<f_t>> solution_host_;
+
+  // Flag indicating where solution data is stored
+  bool is_device_memory_ = true;
```
rg20, line 179: This can be concluded from the unique_ptrs right?

tmckayus: probably, yes, based on a null test. Let me take a look at that

cpp/include/cuopt/linear_programming/mip/solver_solution.hpp
```diff
@@ -75,8 +135,105 @@ class mip_solution_t : public base_solution_t {
...
 
+  //============================================================================
+  // Setters for remote solve deserialization
+  //============================================================================
+
+  /**
+   * @brief Set the solution in host memory
+   * @param solution The solution vector
+   */
+  void set_solution_host(std::vector<f_t> solution);
```
rg20: The setters blows up the API. Can we avoid?

cpp/include/cuopt/linear_programming/pdlp/solver_solution.hpp
```diff
@@ -260,6 +331,128 @@ class optimization_problem_solution_t : public base_solution_t {
...
 
+  //============================================================================
+  // Setters for host solution data (used by remote solve deserialization)
+  //============================================================================
+
+  /**
+   * @brief Set the primal solution in host memory
+   * @param solution The primal solution vector
+   */
+  void set_primal_solution_host(std::vector<f_t> solution);
```
rg20: same here

cpp/include/cuopt/linear_programming/utilities/cython_solve.hpp
```diff
@@ -25,9 +25,19 @@ namespace cython {
...
   std::unique_ptr<rmm::device_buffer> dual_solution_;
   std::unique_ptr<rmm::device_buffer> reduced_cost_;
+
+  // CPU (host) storage - populated for remote solves
+  std::vector<double> primal_solution_host_;
+  std::vector<double> dual_solution_host_;
+  std::vector<double> reduced_cost_host_;
+
+  // Flag indicating where solution data is stored
+  bool is_device_memory_ = true;
```
rg20, line 39: This can be derived from the vectors right?

cpp/include/cuopt/linear_programming/utilities/cython_solve.hpp
```diff
@@ -64,8 +74,15 @@ struct linear_programming_ret_t {
...
 
 struct mip_ret_t {
+  // GPU (device) storage - populated for local GPU solves
   std::unique_ptr<rmm::device_buffer> solution_;
 
+  // CPU (host) storage - populated for remote solves
+  std::vector<double> solution_host_;
+
+  // Flag indicating where solution data is stored
+  bool is_device_memory_ = true;
```
rg20, line 84: same here

cpp/include/cuopt/linear_programming/utilities/remote_solve.hpp
```diff
@@ -0,0 +1,142 @@
...
+
+/**
+ * @brief Solve an LP problem on a remote server.
+ *
+ * Stub implementation for the memory-model-only branch.
+ */
+template <typename i_t, typename f_t>
+optimization_problem_solution_t<i_t, f_t> solve_lp_remote(
+  const remote_solve_config_t&,
+  const cuopt::mps_parser::data_model_view_t<i_t, f_t>& view,
```
rg20, line 73:
```suggestion
  const cuopt::mps_parser::data_model_view_t<i_t, f_t>& model_view,
```

cpp/include/cuopt/linear_programming/utilities/remote_solve.hpp
```diff
@@ -0,0 +1,142 @@
...
+
+/**
+ * @brief Solve a MIP problem on a remote server.
+ *
+ * Stub implementation for the memory-model-only branch.
+ */
+template <typename i_t, typename f_t>
+mip_solution_t<i_t, f_t> solve_mip_remote(
+  const remote_solve_config_t&,
+  const cuopt::mps_parser::data_model_view_t<i_t, f_t>& view,
```
rg20, line 120:
```suggestion
  const cuopt::mps_parser::data_model_view_t<i_t, f_t>& model_view,
```

cpp/include/cuopt/linear_programming/mip/solver_solution.hpp
tmckayus: yes we could do that with null ptr checks I believe.

cpp/libmps_parser/include/mps_parser/data_model_view.hpp
```diff
@@ -406,8 +406,24 @@ class data_model_view_t {
    */
   bool has_quadratic_objective() const noexcept;
 
+  /**
+   * @brief Set whether the data pointed to by this view is in device (GPU) memory.
+   * @note Default is false (CPU memory). Set to true when view points to GPU buffers.
+   *
+   * @param is_device true if data is in GPU memory, false if in CPU memory
+   */
+  void set_is_device_memory(bool is_device) noexcept { is_device_memory_ = is_device; }
```
rg20, line 422: Do we need this?  

If we are using raft copies everywhere, it takes care right?  Also, how do we ensure correctness? What if a user provides host memory but forgets to set this?

cpp/libmps_parser/src/data_model_view.cpp
```diff
@@ -348,7 +350,8 @@ bool data_model_view_t<i_t, f_t>::has_quadratic_objective() const noexcept
 
 // NOTE: Explicitly instantiate all types here in order to avoid linker error
 template class data_model_view_t<int, float>;
-
 template class data_model_view_t<int, double>;
+template class data_model_view_t<int64_t, float>;
```
rg20: are extending to 64 bit integers?   the solver does not support it though right?

cpp/src/linear_programming/cuopt_c.cpp
```diff
@@ -1,52 +1,281 @@
...
 using namespace cuopt::linear_programming;
 
+/**
+ * @brief CPU-side storage for problem data.
+ *
+ * This struct stores all problem data in CPU memory. At solve time, a data_model_view_t
+ * is created pointing to this data, and the solve_lp/solve_mip routines handle
+ * local vs remote solve automatically.
+ */
+struct problem_cpu_data_t {
```
rg20: You can use optimization_problem_t class directly right?  

This seems like a duplicate

rg20: ah, this is a CPU version of optimization_problem_t,  Can you move this to the same file?  You can rename this as optimization_problem_host_t. 

I think its cleaner to template optimization_problem_t with the memory type.

cpp/src/linear_programming/cuopt_c.cpp
```diff
@@ -1,52 +1,281 @@
...
+  /**
+   * @brief Create a view pointing to GPU data from the gpu_problem.
+   * Call this after gpu_problem is fully populated.
+   */
+  void create_view_from_gpu_problem()
+  {
+    if (!gpu_problem) return;
+    auto& gpu = *gpu_problem;
+
+    view.set_maximize(gpu.get_sense());
```
rg20, line 98: I am worried about lot of boilerplate code here. We do similar things in optimization_problem_t::view_t

cpp/src/mip/diversity/population.cu
```diff
@@ -275,6 +280,7 @@ void population_t<i_t, f_t>::run_solution_callbacks(solution_t<i_t, f_t>& sol)
 
     for (auto callback : user_callbacks) {
       if (callback->get_type() == internals::base_solution_callback_type::GET_SOLUTION) {
+        callback->set_memory_location(internals::callback_memory_location::DEVICE);
```
rg20, line 283: explicitly having to set these is error prone.

cpp/src/mip/solve.cu
```diff
@@ -285,13 +288,280 @@ mip_solution_t<i_t, f_t> solve_mip(optimization_problem_t<i_t, f_t>& op_problem,
...
+                                   mps_data_model.get_initial_dual_solution().size());
+  }
+
+  return view;
+}
+
+// Helper struct to hold CPU copies of GPU data for remote solve
+template <typename i_t, typename f_t>
+struct cpu_problem_data_t {
+  std::vector<f_t> A_values;
```
rg20: This looks similar to problem_cpu_data.

cpp/src/mip/solve.cu
```diff
@@ -285,13 +288,280 @@ mip_solution_t<i_t, f_t> solve_mip(optimization_problem_t<i_t, f_t>& op_problem,
...
+  std::vector<f_t> variable_upper_bounds;
+  std::vector<char> variable_types;
+  std::vector<f_t> quadratic_objective_values;
+  std::vector<i_t> quadratic_objective_indices;
+  std::vector<i_t> quadratic_objective_offsets;
+  bool maximize;
+  f_t objective_scaling_factor;
+  f_t objective_offset;
+
+  data_model_view_t<i_t, f_t> create_view() const
```
rg20: This seems redundant

cpp/src/mip/solve.cu
```diff
@@ -285,13 +288,280 @@ mip_solution_t<i_t, f_t> solve_mip(optimization_problem_t<i_t, f_t>& op_problem,
...
+  cpu_data.objective_offset         = gpu_view.get_objective_offset();
+
+  auto copy_to_host = [stream](auto& dst_vec, auto src_span) {
+    if (src_span.size() > 0) {
+      dst_vec.resize(src_span.size());
+      raft::copy(dst_vec.data(), src_span.data(), src_span.size(), stream);
+    }
+  };
+
+  copy_to_host(cpu_data.A_values, gpu_view.get_constraint_matrix_values());
```
rg20: this is error prone as well

rg20: I understand this is unavoidable, but its better to keep this contained in a single source file

cpp/src/linear_programming/utilities/cython_solve.cu
```diff
@@ -229,38 +190,63 @@ std::unique_ptr<solver_ret_t> call_solve(
   bool is_batch_mode)
 {
   raft::common::nvtx::range fun_scope("Call Solve");
+
+  // Data from Python is always in CPU memory
+  data_model->set_is_device_memory(false);
```
tmckayus: this should be an assert, based on the fact that DataModel.get_data_ptr() in Python only allows numpy arrays, and so currently all data from Python coming from DataModel will be CPU. But it shouldn't be setting this flag explicitly at this point.  Will be fixed in an update.

tmckayus:
Closing, superseded by #819

Issue #205, opened 2025-07-14
[FEA] MIP cuts


afender:
Implement cutting plane generation. This is a fundamental algorithmic improvement that directly impacts time to optimality in MIP.  

- Upgrade the code base to support the cut framework 
- Add core cut types after considering GPU optimization perspective

PR #814, opened 2026-01-29
Add cuts to the MIP solver

### LINKED ISSUE #698: Cuts
Tasks:
• Cut Infrastructure
• MIR/Mixed Integer Gomory
• Clique, Knapsack
• Additional Cuts

chris-maes:
This PR adds cuts to the MIP solver. This includes the following:

1.  Add constraints in the form C*x <= d to an LP that has already been solved to optimality (and has basis information).
    -  The constraints must be violated at the current relaxation solution x^star. That is, C*x^star > d. 
    - The constraint matrix, rhs, basis, and basis factorization, are all updated to include the additional constraints.
    - Dual simplex is started in phase 2 from a dual feasible solution. 
2.  Remove constraints from an LP that has already been solved to optimality.
    - The constraints must have slacks in the basis
    - The basis is refactored from scratch
3. Add cut pass loop after solving the root relaxation 
4. Add a cut pool to store cuts and select cuts
     - We currently score cuts  based on distance and orthogonality.  
6. Add Mixed Integer Gomory Cuts
     - These are computed via a MIR cut on a row of the simplex tableau
7. Add Mixed Integer Rounding (MIR) Cuts
    - These are constructed by aggregating rows of the constraint matrix.
8. Add Strong Chvatal-Gomory Cuts
   - These are constructed from a row of the tableau matrix and from rows of the constraint matrix.
9. Fixes to Handling of Steepest Edge Norms in Dual Simplex
    - Ensure that all basic variables have a positive steepest edge norms
10. Reduced Costs Fixing at Root Node 
   -  These are applied after each cut pass and after strong branching, if a heuristic solution is available.
12. Fix issues in Crossover when solving the dual problem
   - We were not correctly populating slack variables when solving the dual. This issue appeared on graph20-80-1rand
14. Fix issue in Crossover when basis became rank-deficient in dual push
15. Fix issues across the code with handling and propagating concurrent halt. 
16. New solver options: mip-cut-passes, mip-mixed-integer-gomory-cuts, mip-mir-cuts, mip-strong-chvatal-gomory-cuts, mip-knapsack-cuts, mip-cut-change-threshold, mip-cut-min-orthogonality.

Closes #698, #205 

Results from a GH200 A/B test with 64 threads and a 300 second time limit. Further runs needed with larger time limit.  Further work is needed to get the full benefit of cuts.

A: cuts PR with --mip-cut-passes=10
B: cuts PR with --mip-cut-passes=0

Geomean MIP GAP A / (B = baseline): 0.97
Geomean Time to Optimal A/B: 0.96
A optimal 45
B optimal 37
A problems with feasible solutions 225
B problems with feasible solutions 224

A wins
drayage-100-23 1.14 8.61
gfd-schedulen180f7d50m30k18 87.85 300.0
neos-827175 12.28 25.86
neos-1171448 28.85 98.2
n5-3 180.62 300.0
neos859080 0.78 1.38
seymour1 38.78 300.0
neos-860300 71.98 90.91
neos-3083819-nubu 30.48 300.0
neos-933966 248.26 300.0
neos-957323 46.36 300.0
neos-960392 50.12 115.68
netdiversion 121.59 300.0
ns1208400 162.15 300.0
nw04 39.89 45.98
piperout-27 8.22 14.09
supportcase7 30.48 300.0
supportcase6 70.22 300.0
uccase12 26.95 41.87
A wins  19

A losses
app1-1 2.3 0.74
cbs-cta 2.21 1.83
irp 17.59 12.94
istanbul-no-cutoff 18.26 13.56
mas76 300.0 20.98
neos-1122047 8.3 6.46
neos-1445765 1.86 1.52
neos-1582420 134.3 93.44
neos-3004026-krka 6.56 1.31
neos8 0.88 0.68
ns1952667 99.84 64.92
piperout-08 10.75 8.23
pk1 224.32 91.84
qap10 134.76 62.82
swath1 60.04 50.3
trento1 300.0 172.28
triptim1 246.65 130.92
A losses 17

### DIFF
--- a/regression/benchmark_scripts/benchmark.py
+++ b/regression/benchmark_scripts/benchmark.py
@@ -1,0 +1,382 @@
+# SPDX-FileCopyrightText: Copyright (c) 2024-2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+
+import os
+from multiprocessing import Process
+import rmm
+import time
+import pandas as pd
+import glob
+import logging as log
+from datetime import datetime
+import argparse
+
+log.getLogger().setLevel(log.INFO)
+
+
+def create_regression_markdown(data, regression_path, test_type_string):
+    regression_md_file = (
+        regression_path + "/" + test_type_string + "_regressions.md"
+    )
+
+    md_data = "*No regressions*"
+    # This to reduce size of slack message
+    limit_no_of_regression_list = 5
+
+    if len(data) > 0:
+        status = "*!! Regressions found !!*"
+        end_msg = (
+            "\n*Continues ...*"
+            if len(data) > limit_no_of_regression_list
+            else ""
+        )
+        table = data[:limit_no_of_regression_list].to_string(index=False)
+        md_data = status + f"\n```\n{table}\n```" + end_msg
+
+    with open(regression_md_file, "w") as fp:
+        fp.write(md_data)
+
+
+def record_regressions(
+    test_name, data, req_metrics, regression_path, test_type_string
+):
+    regression_file = (
+        regression_path + "/" + test_type_string + "_regressions.csv"
+    )
+
+    regression_df = pd.DataFrame(
+        {
+            "Test Name": [],
+            "Metric Name": [],
+            "Value": [],
+            "Avg Value": [],
+            "Regression(%)": [],
+        }
+    )
+    for name in req_metrics:
+        if name.startswith("bks_change_"):
+            pchange = data[name].iloc[-1].item()
+            metric_name = name.replace("bks_change_", "")
+            limit = req_metrics[metric_name]["bks"].get("threshold", 5)
+            prev_val_mean = pchange
+            latest_val = pchange
+        else:
+            limit = req_metrics[name].get("threshold", 5)
+            prev_val_mean = (
+                data[name][:-1][-30:].mean().item()
+                if len(data) > 1
+                else data[name].iloc[-1].item()
+            )
+            latest_val = data[name].iloc[-1].item()
+
+            if prev_val_mean == 0:
+                pchange = latest_val
+            else:
+                pchange = ((latest_val - prev_val_mean) / prev_val_mean) * 100
+
+        if abs(pchange) >= limit:
+            regression_df.loc[len(regression_df)] = [
+                test_name,
+                name,
+                latest_val,
+                prev_val_mean,
+                pchange,
+            ]
+
+    regression_df.to_csv(regression_file)
+    create_regression_markdown(
+        regression_df, regression_path, test_type_string
+    )
+
+
+def get_bks_change(metrics, required_metrics):
+    bks_metrics = {}
+    for metric, value in required_metrics.items():
+        if "bks" in value.keys():
+            bks = value["bks"]["value"]
+            if bks is None:
+                continue
+            current = metrics[metric]
+            if bks == 0:
+                bks_metrics["bks_change_" + metric] = abs(current) * 100
+            elif current == 0:
+                bks_metrics["bks_change_" + metric] = abs(bks) * 100
+            else:
+                bks_metrics["bks_change_" + metric] = abs(
+                    ((current - bks) / bks) * 100
+                )
+
+    return bks_metrics
+
+
+def record_result(
+    test_name, metrics, required_metrics, csv_path, test_type_string
+):
+    file_path = csv_path + "/"
+    if test_type_string == "lp" or test_type_string == "mip":
+        file_path += test_type_string + "_" + test_name + ".csv"
+    else:
+        file_path += test_name + ".csv"
+    bks_metrics = get_bks_change(metrics, required_metrics)
+
+    # Add default metrics to data
+    required_metrics.update(bks_metrics)
+    metrics.update(bks_metrics)
+    req_metrics = list(required_metrics.keys()) + ["date_time", "git_commit"]
+
+    current_data = pd.DataFrame(
+        {key: [metrics[key]] for key in sorted(req_metrics)}
+    )
+    if os.path.isfile(file_path):
+        previous_data = pd.read_csv(file_path, index_col=0)
+        updated_data = pd.concat(
+            [previous_data, current_data], ignore_index=True
+        )
+    else:
+        updated_data = current_data
+    record_regressions(
+        test_name, updated_data, required_metrics, csv_path, test_type_string
+    )
+    updated_data.to_csv(file_path)
+
+
+def run_benchmark(
+    test_name,
+    data_model,
+    solver_settings,
+    required_metrics,
+    csv_path,
+    git_commit,
+    test_status_file,
+    d_type,
+):
+    import rmm
+
+    mr = rmm.mr.get_current_device_resource()
+
+    from utils import LPMetrics, RoutingMetrics
+    from cuopt import linear_programming
+    from cuopt import routing
+
+    start_time = time.time()
+    if d_type == "lp" or d_type == "mip":
+        metrics = LPMetrics()._asdict()
+        solver_settings.set_parameter("infeasibility_detection", False)
+        solver_settings.set_parameter("time_limit", 60)
+        solution = linear_programming.Solve(data_model, solver_settings)
+    else:
+        metrics = RoutingMetrics()._asdict()
+        solution = routing.Solve(data_model)
+    end_time = time.time()
+
+    metrics["gpu_memory_usage"] = int(
+        mr.allocation_counts.peak_bytes / (1024 * 1024)
+    )
+    metrics["date_time"] = datetime.now().strftime("%m_%d_%Y_%H_%M_%S")
+    metrics["git_commit"] = git_commit
+
+    success_status = False
+
+    if d_type == "lp" or d_type == "mip":
+        ## Optimal solution
+        acceptable_termination = ["Optimal", "TimeLimit", "FeasibleFound"]
+        if solution.get_termination_reason() in acceptable_termination:
+            success_status = True
+            metrics["solver_time"] = solution.get_solve_time()
+            metrics["primal_objective_value"] = solution.get_primal_objective()
+            if d_type == "lp":
+                lp_stats = solution.get_lp_stats()
+                metrics["nb_iterations"] = lp_stats["nb_iterations"]
+            else:
+                milp_stats = solution.get_milp_stats()
+                metrics["mip_gap"] = milp_stats["mip_gap"]
+                metrics["max_constraint_violation"] = milp_stats[
+                    "max_constraint_violation"
+                ]
+                metrics["max_int_violation"] = milp_stats["max_int_violation"]
+                metrics["max_variable_bound_violation"] = milp_stats[
+                    "max_variable_bound_violation"
+                ]
+            record_result(
+                test_name, metrics, required_metrics, csv_path, d_type
+            )
+    else:
+        if solution.get_status() == 0:
+            success_status = True
+            metrics["solver_time"] = end_time - start_time
+            metrics["total_objective_value"] = solution.get_total_objective()
+            metrics["vehicle_count"] = solution.get_vehicle_count()
+
+            objectives = solution.get_objective_values()
+            if "prize" in required_metrics:
+                metrics["prize"] = objectives[routing.Objective.PRIZE]
+            if "cost" in required_metrics:
+                metrics["cost"] = objectives[routing.Objective.COST]
+            if "travel_time" in required_metrics:
+                metrics["travel_time"] = objectives[
+                    routing.Objective.TRAVEL_TIME
+                ]
+            record_result(
+                test_name, metrics, required_metrics, csv_path, d_type
+            )
+    return "SUCCESS" if success_status is True else "FAILED"
+
+
+def reinitialize_rmm():
+    pool_size = 2**30
+    rmm.reinitialize(pool_allocator=True, initial_pool_size=pool_size)
+
+    base_mr = rmm.mr.get_current_device_resource()
+    stats_mr = rmm.mr.StatisticsResourceAdaptor(base_mr)
+    rmm.mr.set_current_device_resource(stats_mr)
+
+    return base_mr, stats_mr
+
+
+def worker(
+    gpu_id,
+    dataset_file_path,
+    csv_path,
+    git_commit,
+    log_path,
+    test_status_file,
+    n_gpus,
+    d_type="routing",
+):
+    import os
+
+    os.environ["CUDA_VISIBLE_DEVICES"] = gpu_id
+
+    from utils import get_configuration
+
+    data_files = []
+    if d_type == "lp" or d_type == "mip":
+        data_files = glob.glob(dataset_file_path + "/*.mps")
+    else:
+        data_files = glob.glob(dataset_file_path + "/*_config.json")
+    idx = int(gpu_id)
+    n_files = 1  # len(data_files)
+
+    while idx < n_files:
+        mr, stats_mr = reinitialize_rmm()
+
+        data_file = data_files[idx]
+        test_name = str(data_file)
+        status = "FAILED"
+        try:
+            test_name, data_model, solver_settings, requested_metrics = (
+                get_configuration(data_file, dataset_file_path, d_type)
+            )
+            log.basicConfig(
+                level=log.INFO,
+                filename=log_path + "/" + test_name + "_log.txt",
+                filemode="a+",
+                format="%(asctime)-15s %(levelname)-8s %(message)s",
+            )
+            log.getLogger().setLevel(log.INFO)
+            log.info(
+                f"------------- Test Start : {test_name} gpu id : {gpu_id} -------------------"
+            )
+            status = run_benchmark(
+                test_name,
+                data_model,
+                solver_settings,
+                requested_metrics,
+                csv_path,
+                git_commit,
+                test_status_file,
+                d_type,
+            )
+        except Exception as e:
+            log.error(str(e))
+
+        with open(test_status_file, "a") as f:
+            f.write("\n")
+            f.write(test_name + ": " + status)
+
+        # Delete instance of rmm
+        del mr
+        del stats_mr
+
+        log.info(
+            f"------------- Test End : {test_name} gpu id : {gpu_id} -------------------"
+        )
+        idx = idx + n_gpus
+
+
+def run(
+    dataset_file_path,
+    csv_path,
+    git_commit,
+    log_path,
+    test_status_file,
+    n_gpus,
+    d_type,
+):
+    # Restricting n_gpus to one to avoid resource sharing
+    # n_gpus = 1
+    procs = []
+    for gpu_id in range(int(n_gpus)):
+        p = Process(
+            target=worker,
+            args=(
+                str(gpu_id),
+                dataset_file_path,
+                csv_path,
+                git_commit,
+                log_path,
+                test_status_file,
+                int(n_gpus),
+                d_type,
+            ),
+        )
+        p.start()
+        procs.append(p)
+
+    for p in procs:
+        p.join()
+    print("All processes finished.")
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser()
+
+    parser.add_argument(
+        "-c", "--config-path", type=str, help="Path to all configuration file"
+    )
+    parser.add_argument(
+        "-r",
+        "--csv-path",
+        type=str,
+        help="Path to store result files, this would be for github where results gets stored",
+    )
+    parser.add_argument(
+        "-g",
+        "--git-commit",
+        type=str,
+        help="git commit sha to keep track of runs",
+    )
+    parser.add_argument("-l", "--log-path", type=str, help="Path to log files")
+    parser.add_argument(
+        "-s",
+        "--test-status-file",
+        type=str,
+        help="All test status will be stored in this file",
+    )
+    parser.add_argument(
+        "-n", "--num-gpus", type=str, help="Number of GPUs available"
+    )
+    parser.add_argument(
+        "-t", "--type", type=str, default="", help="Type of benchmark"
+    )
+    args = parser.parse_args()
+    run(
+        args.config_path,
+        args.csv_path,
+        args.git_commit,
+        args.log_path,
+        args.test_status_file,
+        args.num_gpus,
+        args.type,
+    )
--- a/regression/benchmark_scripts/configs/README.md
+++ b/regression/benchmark_scripts/configs/README.md
@@ -1,0 +1,19 @@
+# Creating configuration and data file for routing
+
+- For each test, create a configuration file and a corresponding data file.
+- Refer `test_name_config.json` for the format of the configuration file.
+- Supported metrics can be found in `cuopt/regression/benchmark_scripts/utils.py`
+- File names should start with test names followed by `config` or data depending on type of it.
+- Data file should be as per openapi spec of cuopt server
+- These configuration and data files needs to be uploaded to `s3://cuopt-datasets/regression_datasets/`
+
+   ```
+   aws s3 cp /path/to/files s3://cuopt-datasets/regression_datasets/
+   ```
+
+# Creating configuration and data file for lp and milp
+
+- For each test, create a mps file
+- Refer `lp_config.json` and `mip_config.json` for the format of the configuration file.
+- Supported metrics can be found in `cuopt/regression/benchmark_scripts/utils.py`
+- These configuration and data files needs to be in the LP_DATASETS_PATH set in config.sh
--- a/regression/benchmark_scripts/configs/example_test_creation.py
+++ b/regression/benchmark_scripts/configs/example_test_creation.py
@@ -1,0 +1,76 @@
+# SPDX-FileCopyrightText: Copyright (c) 2024-2026, NVIDIA CORPORATION.
+# SPDX-License-Identifier: Apache-2.0
+# CONFIDENTIAL, provided under NDA.
+
+from cuopt.routing import utils
+import json
+
+"""
+This is an example of creating a modified test from Homberger dataset.
+In this test, the RC2_10_5 test is modified so that the vehicle count is reduced to 12 and the order prizes are set.
+The prizes are high enough so that prize always becomes the primary objective.
+One can easily use an existing json file and modify the data as well by loading the json as a dictionary
+"""
+test_name = "prize_collection_vrp"
+# test_name = "LC1_10_9"
+
+# base_file_name = "/home/nfs/rgandham/git-repos/reopt/datasets/pdptw/LC1_10_9.pdptw"
+base_file_name = (
+    "/home/nfs/rgandham/git-repos/reopt/datasets/cvrptw/RC2_10_5.TXT"
+)
+
+# model_dict = utils.create_model_dictionary_from_file(base_file_name, is_pdp=True)
+model_dict = utils.create_model_dictionary_from_file(base_file_name)
+
+
+# Reduce the fleet size to 12
+num_vehicles = 12
+fleet_data = model_dict["fleet_data"]
+
+vehicle_locations = fleet_data["vehicle_locations"]
+vehicle_tw = fleet_data["vehicle_time_windows"]
+capacities = fleet_data["capacities"]
+
+new_locs = [vehicle_locations[i] for i in range(num_vehicles)]
+new_tw = [vehicle_tw[i] for i in range(num_vehicles)]
+new_cap = [[capacities[0][i] for i in range(num_vehicles)]] * 1
+
+fleet_data["vehicle_locations"] = new_locs
+fleet_data["vehicle_time_windows"] = new_tw
+fleet_data["capacities"] = new_cap
+
+# Add prizes
+task_data = model_dict["task_data"]
+
+n_tasks = len(task_data["demand"][0])
+
+prizes = [10000.0] * n_tasks
+task_data["prizes"] = prizes
+
+
+# Set 10 min time limit
+solver_config = {}
+solver_config["time_limit"] = 600
+
+model_dict["solver_config"] = solver_config
+
+test_config_file_name = test_name + "_config.json"
+model_data_file_name = test_name + "_data.json"
+
+test_config = {}
+test_config["test_name"] = test_name
+test_config["file_name"] = model_data_file_name
+test_config["metrics"] = [
+    "vehicle_count",
+    "total_cost",
+    "prize",
+    "memory_usage",
+]
+
+with open(test_config_file_name, "w") as fp:
+    json.dump(test_config, fp)
+    fp.close()
+
+with open(model_data_file_name, "w") as fp:
+    json.dump(model_dict, fp)
+    fp.close()
--- a/regression/benchmark_scripts/configs/test_name_config.json
+++ b/regression/benchmark_scripts/configs/test_name_config.json
@@ -1,0 +1,23 @@
+{
+    "test_name": "test_name",
+    "file_name": "test_name_data.json",
+    "metrics": {
+        "total_objective_value": {
+            "threshold": 5,
+            "unit": "total_objective_value"
+        },
+	    "vehicle_count": {
+            "threshold": 5,
+            "unit": "vehicle_count"
+        },
+	    "solver_time": {
+            "threshold": 5,
+            "unit": "seconds"
+        },
+        "gpu_memory_usage": {
+            "threshold": 20,
+            "unit": "MB"
+        }
+    },
+    "details": "Add details about you test"
+}
--- a/regression/benchmark_scripts/configs/test_name_data.json
+++ b/regression/benchmark_scripts/configs/test_name_data.json
@@ -1,0 +1,117 @@
+{
+    "cost_waypoint_graph_data": {
+        "waypoint_graph": null
+    },
+    "travel_time_waypoint_graph_data": {
+        "waypoint_graph": null
+    },
+    "cost_matrix_data": {
+        "data": {
+            "0": [
+                [
+                    0,
+                    1,
+                    1
+                ],
+                [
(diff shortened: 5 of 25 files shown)


cpp/src/dual_simplex/branch_and_bound.hpp
```diff
@@ -138,6 +151,13 @@ class branch_and_bound_t {
   // Local lower bounds for each thread
   std::vector<omp_atomic_t<f_t>> local_lower_bounds_;
 
+  // Mutex for the original LP
+  // The heuristics threads look at the original LP. But the main thread modifies the
+  // size of the original LP by adding slacks for cuts. Heuristic threads should lock
+  // this mutex when accessing the original LP. The main thread should lock this mutex
+  // when modifying the original LP.
+  omp_mutex_t mutex_original_lp_;
```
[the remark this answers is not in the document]
chris-maes: The diving threads are started after the root cut pass. The original_lp_ is fixed after the cut passes. 
Besides the diving threads are there are unprotected reads?

cpp/src/dual_simplex/crossover.cpp
```diff
@@ -1378,18 +1378,22 @@ crossover_status_t crossover(const lp_problem_t<i_t, f_t>& lp,
...
+        if (toc(start_time) > settings.time_limit) {
+          settings.log.printf("Time limit exceeded\n");
+          return crossover_status_t::TIME_LIMIT;
+        }
+        if (settings.concurrent_halt != nullptr && *settings.concurrent_halt == 1) {
+          settings.log.printf("Concurrent halt\n");
+          return crossover_status_t::CONCURRENT_LIMIT;
+        }
+        solution.iterations += iter;
       }
```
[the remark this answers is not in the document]
chris-maes: I don't understand this. If we can't get dual feasible, the solution.z values are not returned to the user.

cpp/src/mip/presolve/third_party_presolve.cpp
```diff
@@ -409,7 +409,7 @@ std::optional<third_party_presolve_result_t<i_t, f_t>> third_party_presolve_t<i_
                  papilo_problem.getNCols(),
                  papilo_problem.getConstraintMatrix().getNnz());
 
-  CUOPT_LOG_INFO("Calling Papilo presolver");
+  CUOPT_LOG_INFO("Calling Papilo presolver (git hash %s)", PAPILO_GITHASH);
```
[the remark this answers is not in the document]
chris-maes: In what builds would PAPILO_GITHASH be missing?

cpp/src/dual_simplex/branch_and_bound.hpp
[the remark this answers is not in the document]
chris-maes: The mutex for original_lp is needed because `set_new_solution` can be called from another thread during the cut passes. And `set_new_solution` access original_lp. 

The other functions `best_first_thread`, `exploration_ramp_up` and `plunge_from` all occur after the cut passes have been completed, and thus `original_lp_` is fixed.

cpp/src/dual_simplex/crossover.cpp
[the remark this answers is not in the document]
chris-maes: Ah I see. I think the issue is there is a mismatch between `vstatus` (which was set from `compute_dual_solution_from_basis`) and `solution.y` and `solution.z` which come from `phase1_solution` in the call to `dual_infeasibility`. 

However, wouldn't a better fix be to remove the line `solution = phase1_solution`.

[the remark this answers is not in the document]
chris-maes: That doesn't make sense. The primal values will be overwritten by the phase2 call if we are dual feasible. And if we are dual infeasible, we return NUMERIC and the primal values should not be accessed by the user.

cpp/src/dual_simplex/cuts.cpp
```diff
@@ -0,0 +1,2860 @@
...
+            break;
+          }
+        }
+        inequality.x[k] = 0.0;
+      }
+    }
+
+    if (negate_inequality == -1) {
+      break;  // TODO: this stops us from generating further MIR cuts for other rows.
+    }
```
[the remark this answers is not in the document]
chris-maes: A `continue` won't work because of how we get `i` and `max_score`.

cpp/src/dual_simplex/branch_and_bound.cpp
```diff
@@ -280,47 +317,122 @@ void branch_and_bound_t<i_t, f_t>::report_heuristic(f_t obj)
...
+  for (i_t j = 0; j < num_cols_to_check; j++) {
+    if (!std::isfinite(reduced_costs[j])) {
+      printf("Variable %d reduced cost is %e\n", j, reduced_costs[j]);
+      continue;
+    }
+    if (std::abs(reduced_costs[j]) > threshold) {
+      const f_t lower_j            = original_lp_.lower[j];
+      const f_t upper_j            = original_lp_.upper[j];
+      const f_t abs_gap            = upper_bound - root_obj;
+      f_t reduced_cost_upper_bound = upper_j;
```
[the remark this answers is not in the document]
chris-maes: The root objective is always less than the upper bound. If it is equal to the upper bound we have solved the problem to optimality.

cpp/src/dual_simplex/crossover.cpp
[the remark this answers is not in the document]
chris-maes: I believe that simply removing the line `solution = phase1_solution` solves the problem and avoids the mismatch.

cpp/src/dual_simplex/cuts.cpp
```diff
@@ -0,0 +1,2859 @@
...
+template <typename i_t, typename f_t>
+void cut_pool_t<i_t, f_t>::score_cuts(std::vector<f_t>& x_relax)
+{
+  const f_t weight_distance      = 1.0;
+  const f_t weight_orthogonality = 1.0;
+  const f_t min_cut_distance     = 1e-4;
+  cut_distances_.resize(cut_storage_.m, 0.0);
+  cut_norms_.resize(cut_storage_.m, 0.0);
+  cut_orthogonality_.resize(cut_storage_.m, 1);
+  cut_scores_.resize(cut_storage_.m, 0.0);
```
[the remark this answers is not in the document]
chris-maes: Nice catch

cpp/src/dual_simplex/cuts.hpp
```diff
@@ -0,0 +1,470 @@
...
+        num_knapsack_cuts++;
+      } else if (cut_type == cut_type_t::CHVATAL_GOMORY) {
+        num_cg_cuts++;
+      }
+    }
+  }
+  i_t num_gomory_cuts   = 0;
+  i_t num_mir_cuts      = 0;
+  i_t num_knapsack_cuts = 0;
+  i_t num_cg_cuts       = 0;
```
aliceb-nv: This is really minor, but you could have a cut_type_t num_cuts[CUT_TYPE_MAX]; member instead of multiple members and branches

aliceb-nv: And/Or a second static array like this to fetch the cut name when printing:
```
static const char* cut_names[] = 
{"Gomory", "MIR", "Knapsack", "CG"};

File: cpp/src/dual_simplex/basis_solves.cpp [hunk 8]
```diff
@@ -363,7 +363,7 @@ i_t factorize_basis(const csc_matrix_t<i_t, f_t>& A,
                                  S_perm_inv);
         if (settings.concurrent_halt != nullptr && *settings.concurrent_halt == 1) {
           settings.log.printf("Concurrent halt\n");
-          return -1;
+          return -2;
```
nguidotti: Nitpick: Can we replace the `-1`/`-2` with enums? I think it will be easier to understand.

cpp/src/dual_simplex/branch_and_bound.cpp
```diff
@@ -280,47 +317,123 @@ void branch_and_bound_t<i_t, f_t>::report_heuristic(f_t obj)
...
+i_t branch_and_bound_t<i_t, f_t>::find_reduced_cost_fixings(f_t upper_bound,
+                                                            std::vector<f_t>& lower_bounds,
+                                                            std::vector<f_t>& upper_bounds)
+{
+  std::vector<f_t> reduced_costs = root_relax_soln_.z;
+  lower_bounds                   = original_lp_.lower;
+  upper_bounds                   = original_lp_.upper;
+  std::vector<bool> bounds_changed(original_lp_.num_cols, false);
+  const f_t root_obj    = compute_objective(original_lp_, root_relax_soln_.x);
+  const f_t threshold   = 1e-3;
```
nguidotti: Should this go into the settings instead of being a hard-coded value?

cpp/src/dual_simplex/branch_and_bound.cpp
```diff
@@ -1297,13 +1544,16 @@ lp_status_t branch_and_bound_t<i_t, f_t>::solve_root_relaxation(
   lp_status_t root_status;
   std::future<lp_status_t> root_status_future;
   root_status_future = std::async(std::launch::async,
```
nguidotti, line 1556: Since we are getting the basis and its factorization from the root relaxation, can they be passed to the strong branching for improve its performance?

cpp/src/dual_simplex/branch_and_bound.cpp
```diff
@@ -1510,33 +1793,292 @@ mip_status_t branch_and_bound_t<i_t, f_t>::solve(mip_solution_t<i_t, f_t>& solut
...
+                                   settings_,
+                                   Arow_,
+                                   new_slacks_,
+                                   var_types_,
+                                   basis_update,
+                                   root_relax_soln_.x,
+                                   basic_list,
+                                   nonbasic_list);
+      f_t cut_generation_time = toc(cut_start_time);
+      if (1 || cut_generation_time > 1.0) {
```
nguidotti: Debug leftover?

cpp/src/dual_simplex/branch_and_bound.cpp
```diff
@@ -1510,33 +1793,292 @@ mip_status_t branch_and_bound_t<i_t, f_t>::solve(mip_solution_t<i_t, f_t>& solut
...
+      settings_.log.printf("Before A check\n");
+      original_lp_.A.check_matrix();
+#endif
+      original_lp_.A.to_compressed_row(Arow_);
+
+      f_t node_presolve_start_time = tic();
+      bounds_strengthening_t<i_t, f_t> node_presolve(original_lp_, Arow_, row_sense, var_types_);
+      mutex_original_lp_.lock();
+      bool feasible =
+        node_presolve.bounds_strengthening(original_lp_.lower, original_lp_.upper, settings_);
```
nguidotti: The `node_presolve` has a public attribute with the `bound_changed`, which will be used within the `bound_strengthening`.  I changed this on #766 so the `bound_changed` is external.

cpp/src/dual_simplex/branch_and_bound.cpp
```diff
@@ -1510,33 +1793,292 @@ mip_status_t branch_and_bound_t<i_t, f_t>::solve(mip_solution_t<i_t, f_t>& solut
...
+        mutex_upper_.lock();
+        incumbent_.set_incumbent_solution(root_objective_, root_relax_soln_.x);
+        mutex_upper_.unlock();
+      }
+      f_t obj         = upper_bound_.load();
+      f_t user_obj    = compute_user_objective(original_lp_, obj);
+      f_t user_lower  = compute_user_objective(original_lp_, root_objective_);
+      std::string gap = num_fractional != 0 ? user_mip_gap<f_t>(user_obj, user_lower) : "0.0%";
+
+      settings_.log.printf("  %10d   %10lu    %+13.6e    %+10.6e   %6d %6d   %7.1e     %s %9.2f\n",
```
nguidotti: Nitpick: should we use the `report` routine here?

cpp/src/dual_simplex/branch_and_bound.cpp
chris-maes: That's a good idea. We could save repeating the basis factorization for each trial branch LP solve.  But each trial branch needs its own `basis_update_t` because the basis will change for each LP solve. 

There are a couple ways we could try to do this:
1) Have each thread copy the basis_update_t structure. Including the updates etc. Since the root relaxation basis_update_t may have updates.
2)  Force a refactorization of the basis at the end of the root. Then initialize each LP solve basis_update_t from the LU factors of the basis. 

Both of those involve copying the LU factors of the matrix. There's probably another optimization, where the LU factors of the root basis are fixed, each subproblem makes updates. And if the the subproblem did not force a refactorization, you just throw away the basis updates and start again.

cpp/src/dual_simplex/branch_and_bound.cpp
chris-maes: Fixed. Thanks for catching this.

cpp/src/dual_simplex/cuts.cpp
```diff
@@ -0,0 +1,2878 @@
...
+      }
+    }
+
+    // Clear the aggregated mark
+    for (i_t row : aggregated_rows) {
+      aggregated_mark[row] = 0;
+    }
+    // Clear the aggregated rows
+    aggregated_rows.clear();
+    work_estimate += 2 * aggregated_rows.size();
```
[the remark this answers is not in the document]
chris-maes: Thanks. Fixed.

cpp/src/dual_simplex/branch_and_bound.cpp
```diff
@@ -280,47 +317,123 @@ void branch_and_bound_t<i_t, f_t>::report_heuristic(f_t obj)
...
     f_t bound_err;
     i_t num_fractional;
+    mutex_original_lp_.lock();
     is_feasible = check_guess(
       original_lp_, settings_, var_types_, crushed_solution, primal_err, bound_err, num_fractional);
-    if (is_feasible) {
+    mutex_original_lp_.unlock();
+    mutex_upper_.lock();
+    if (is_feasible && obj < upper_bound_) {
       upper_bound_ = obj;
```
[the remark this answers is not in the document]
chris-maes: Thanks. Fixed.

cpp/src/dual_simplex/branch_and_bound.cpp
```diff
@@ -280,47 +317,123 @@ void branch_and_bound_t<i_t, f_t>::report_heuristic(f_t obj)
...
+  std::vector<bool> bounds_changed(original_lp_.num_cols, false);
+  const f_t root_obj    = compute_objective(original_lp_, root_relax_soln_.x);
+  const f_t threshold   = 1e-3;
+  const f_t weaken      = 1e-5;
+  const f_t fixed_tol   = settings_.fixed_tol;
+  i_t num_improved      = 0;
+  i_t num_fixed         = 0;
+  i_t num_cols_to_check = reduced_costs.size();  // Reduced costs will be smaller than the original
+                                                 // problem because we have added slacks for cuts
+  for (i_t j = 0; j < num_cols_to_check; j++) {
```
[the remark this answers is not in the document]
chris-maes: var_types_ should be resized when the reduced costs are resized in add_cuts. So they should always be the same size.

cpp/src/dual_simplex/cuts.cpp
```diff
@@ -0,0 +1,2840 @@
...
+              aggregated_rows.push_back(pivot_row);
+              aggregated_mark[pivot_row] = 1;
+              work_estimate += inequality.i.size() + pivot_row_inequality.i.size();
+            } else {
+              // No potential rows to aggregate
+              break;
+            }
+          } else {
+            assert(col_len > 1);
+          }
```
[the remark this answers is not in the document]
chris-maes: Thanks. Fixed.

cpp/src/dual_simplex/branch_and_bound.cpp
chris-maes: Sure. Done.

cpp/src/dual_simplex/cuts.cpp
```diff
@@ -0,0 +1,2839 @@
...
+  const f_t min_orthogonality = settings_.cut_min_orthogonality;
+  best_cuts_.reserve(std::min(max_cuts, cut_storage_.m));
+  best_cuts_.clear();
+
+  if (!sorted_indices.empty()) {
+    const i_t i = sorted_indices.back();
+    sorted_indices.pop_back();
+    best_cuts_.push_back(i);
+    scored_cuts_++;
+  }
```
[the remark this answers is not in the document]
chris-maes: Nice catch. Fixed.

cpp/src/dual_simplex/basis_solves.cpp
chris-maes: This can't be an enum, because the return use numbers >= 0 to signal correct behavior. Numbers < 0 are used to signal errors or concurrent limit. 

Anyway, to make it easier to read/understand I added a constant that we can use.

cpp/src/dual_simplex/cuts.cpp
```diff
@@ -0,0 +1,2860 @@
...
+  // Build permutation
+  std::vector<i_t> perm(n);
+  std::iota(perm.begin(), perm.end(), 0);
+
+  std::vector<f_t> ratios;
+  ratios.resize(n);
+  for (i_t i = 0; i < n; i++) {
+    ratios[i] = values[i] / weights[i];
+  }
+
```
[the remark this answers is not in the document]
chris-maes: If weights[i] = 0, and values[i] != 0 and ratios[i] = inf. And we will correctly select the item, because it provides value and the weight is zero.

cpp/src/dual_simplex/cuts.cpp
```diff
@@ -0,0 +1,2840 @@
...
+    is_slack_(num_vars_, 0),
+    slack_rows_(num_vars_, 0),
+    bound_info_(num_vars_, 0)
+{
+  for (i_t j : new_slacks) {
+    is_slack_[j]        = 1;
+    const i_t col_start = lp.A.col_start[j];
+    const i_t i         = lp.A.i[col_start];
+    slack_rows_[j]      = i;
+    assert(std::abs(lp.A.x[col_start]) == 1.0);
```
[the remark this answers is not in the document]
chris-maes: Slack variables are always exactly 1.0 or -1.0.

cpp/src/dual_simplex/cuts.cpp
```diff
@@ -0,0 +1,2878 @@
...
+
+  // Construct C_B = C(:, basic_list)
+  std::vector<i_t> C_col_degree(lp.num_cols, 0);
+  i_t cuts_nz = cuts.row_start[p];
+  for (i_t q = 0; q < cuts_nz; q++) {
+    const i_t j = cuts.j[q];
+    if (j >= lp.num_cols) {
+      settings.log.printf("j %d is greater than p %d\n", j, p);
+      return -1;
+    }
```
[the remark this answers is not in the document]
chris-maes: Fixed.

cpp/src/dual_simplex/branch_and_bound.cpp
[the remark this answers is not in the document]
chris-maes: Nice catch. Fixed.

cpp/src/dual_simplex/branch_and_bound.cpp
chris-maes: Done

cpp/src/dual_simplex/branch_and_bound.cpp
```diff
@@ -1556,6 +2091,47 @@ mip_status_t branch_and_bound_t<i_t, f_t>::solve(mip_solution_t<i_t, f_t>& solut
...
+        std::vector<i_t> new_fractional;
+        new_fractional.reserve(fractional.size() - num_to_remove);
+        for (i_t k = 0; k < fractional.size(); k++) {
+          if (!to_remove[k]) { new_fractional.push_back(fractional[k]); }
+        }
+        fractional     = new_fractional;
+        num_fractional = fractional.size();
+      }
+    }
+  }
```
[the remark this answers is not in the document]
chris-maes: Fixed.

cpp/src/dual_simplex/branch_and_bound.cpp
```diff
@@ -280,47 +317,130 @@ void branch_and_bound_t<i_t, f_t>::report_heuristic(f_t obj)
...
+
+  if (num_fixed > 0 || num_improved > 0) {
+    settings_.log.printf(
+      "Reduced costs: Found %d improved bounds and %d fixed variables (%.1f%%)\n",
+      num_improved,
+      num_fixed,
+      100.0 * static_cast<f_t>(num_fixed) / static_cast<f_t>(num_integer_variables_));
+  }
+  return num_fixed;
+}
```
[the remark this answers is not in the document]
chris-maes: Nice catch. Fixed.

cpp/src/dual_simplex/branch_and_bound.cpp
chris-maes: But I think it's best to do that in a follow up PR.

cpp/src/dual_simplex/branch_and_bound.cpp
chris-maes: Yes, this will be removed before merging. Still using it for testing.

cpp/tests/mip/cuts_test.cu
```diff
@@ -0,0 +1,167 @@
...
+#include <raft/core/handle.hpp>
+#include <raft/util/cudart_utils.hpp>
+
+#include <gtest/gtest.h>
+
+#include <cstdint>
+#include <filesystem>
+#include <sstream>
+#include <string>
+#include <vector>
```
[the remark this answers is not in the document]
chris-maes: Fixed.

cpp/tests/mip/cuts_test.cu
```diff
@@ -0,0 +1,167 @@
...
+  auto problem = create_cuts_problem_1();
+
+  settings.time_limit                  = test_time_limit;
+  settings.max_cut_passes              = 1;
+  mip_solution_t<int, double> solution = solve_mip(&handle_, problem, settings);
+  EXPECT_EQ(solution.get_termination_status(), mip_termination_status_t::Optimal);
+
+  double obj_val = solution.get_objective_value();
+  // Expected objective value from documentation example is approximately 303.5
+  EXPECT_NEAR(-28, obj_val, 1e-3);
```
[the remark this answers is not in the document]
chris-maes: Fixed.

cpp/src/dual_simplex/branch_and_bound.cpp
```diff
@@ -122,10 +124,14 @@ bool check_guess(const lp_problem_t<i_t, f_t>& original_lp,
...
+                                           const std::vector<i_t>& basic_list,
+                                           std::vector<f_t>& edge_norms)
 {
-  for (i_t j = 0; j < edge_norms.size(); ++j) {
-    if (edge_norms[j] <= 0.0) { edge_norms[j] = 1e-4; }
+  if (edge_norms.size() != lp.num_cols) { edge_norms.resize(lp.num_cols, -1.0); }
+  for (i_t k = 0; k < lp.num_rows; k++) {
+    const i_t j = basic_list[k];
+    if (edge_norms[j] < 0.0) { edge_norms[j] = 1e-4; }
   }
```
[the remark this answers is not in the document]
chris-maes: NaN isn't an issue. But I will catch the == 0.0 case.

cpp/src/dual_simplex/branch_and_bound.cpp
```diff
@@ -1510,33 +1792,284 @@ mip_status_t branch_and_bound_t<i_t, f_t>::solve(mip_solution_t<i_t, f_t>& solut
...
+      f_t obj = upper_bound_.load();
+      report(' ', obj, root_objective_, 0, num_fractional);
+
+      f_t rel_gap = user_relative_gap(original_lp_, upper_bound_.load(), root_objective_);
+      f_t abs_gap = upper_bound_.load() - root_objective_;
+      if (rel_gap < settings_.relative_mip_gap_tol || abs_gap < settings_.absolute_mip_gap_tol) {
+        if (settings_.solution_callback != nullptr) {
+          settings_.solution_callback(root_relax_soln_.x, root_objective_);
+        }
+        set_final_solution(solution, root_objective_);
```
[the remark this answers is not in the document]
chris-maes: Fixed.

cpp/src/dual_simplex/bounds_strengthening.cpp
```diff
@@ -91,9 +90,10 @@ bounds_strengthening_t<i_t, f_t>::bounds_strengthening_t(
 
 template <typename i_t, typename f_t>
 bool bounds_strengthening_t<i_t, f_t>::bounds_strengthening(
+  const simplex_solver_settings_t<i_t, f_t>& settings,
+  const std::vector<bool>& bounds_changed,
   std::vector<f_t>& lower_bounds,
-  std::vector<f_t>& upper_bounds,
-  const simplex_solver_settings_t<i_t, f_t>& settings)
+  std::vector<f_t>& upper_bounds)
```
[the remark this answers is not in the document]
chris-maes: It should be plural. Fixed.

cpp/src/dual_simplex/basis_updates.cpp
```diff
@@ -1108,6 +1108,220 @@ i_t basis_update_t<i_t, f_t>::lower_triangular_multiply(const csc_matrix_t<i_t,
...
+    }
+  }
+  for (i_t j = m; j < m + cuts_basic.m; ++j) {
+    new_L.col_start[j] = L_nz;
+    new_L.i[L_nz]      = j;
+    new_L.x[L_nz]      = 1.0;
+    L_nz++;
+  }
+  new_L.col_start[m + cuts_basic.m] = L_nz;
+  if (L_nz != predicted_nz) {
```
akifcorduk: Stale debug code.

cpp/src/dual_simplex/bound_flipping_ratio_test.cpp
```diff
@@ -255,11 +255,11 @@ void bound_flipping_ratio_test_t<i_t, f_t>::heap_passes(const std::vector<i_t>&
     }
 
     if (toc(start_time_) > settings_.time_limit) {
-      entering_index = -2;
+      entering_index = -3;
       return;
     }
     if (settings_.concurrent_halt != nullptr && *settings_.concurrent_halt == 1) {
-      entering_index = -3;
+      entering_index = CONCURRENT_HALT_RETURN;
```
akifcorduk, line 259: Just ti make sure, in other places the error code for CONCURRENT_HALT_RETURN is -2. For bound flipping, did we swap the error codes for time limit and concurrent_halt ? If so, maybe it would be clearer to enum other error codes too.

cpp/src/dual_simplex/branch_and_bound.cpp
```diff
@@ -1510,33 +1792,280 @@ mip_status_t branch_and_bound_t<i_t, f_t>::solve(mip_solution_t<i_t, f_t>& solut
...
+                                     edge_norms_);
+      var_types_.resize(original_lp_.num_cols, variable_type_t::CONTINUOUS);
+      mutex_original_lp_.unlock();
+      f_t add_cuts_time = toc(add_cuts_start_time);
+      if (add_cuts_time > 1.0) {
+        settings_.log.debug("Add cuts time %.2f seconds\n", add_cuts_time);
+      }
+      if (add_cuts_status != 0) {
+        settings_.log.printf("Failed to add cuts\n");
+        return mip_status_t::NUMERICAL;
```
akifcorduk, line 1933: Why are we returning numerical if we can't add cuts? Can't we just continue with original problem?

cpp/src/dual_simplex/folding.cpp
```diff
@@ -577,7 +577,7 @@ i_t color_graph(const csc_matrix_t<i_t, f_t>& A,
     for (i_t k = 0; k < max_vertices; k++) {
       if (vertex_to_sum[k] != 0.0) {
         settings.log.printf("Folding: Vertex %d has sum %e\n", k, vertex_to_sum[k]);
-        return -2;
+        return -1;
```
akifcorduk: I think error code enums here would be more clear.

cpp/src/mip/solver.cu
```diff
@@ -108,16 +108,17 @@ solution_t<i_t, f_t> mip_solver_t<i_t, f_t>::run_solver()
   }
 
   diversity_manager_t<i_t, f_t> dm(context);
-  dm.timer              = timer_;
-  bool presolve_success = dm.run_presolve(timer_.remaining_time());
+  dm.timer                = timer_;
+  const bool run_presolve = context.settings.presolve;
+  bool presolve_success   = run_presolve ? dm.run_presolve(timer_.remaining_time()) : true;
```
akifcorduk, line 111: Don't you think it makes sense to have a separation between third party and internal presolve? Internal presolve has other important components like probing cache that impacts heuristic quality.

cpp/src/dual_simplex/cuts.cpp
```diff
@@ -0,0 +1,2822 @@
...
+  cut.i.reserve(inequality.i.size());
+  cut.x.reserve(inequality.i.size());
+  cut.i.clear();
+  cut.x.clear();
+
+  f_t a_0   = inequality_rhs;
+  f_t f_a_0 = fractional_part(a_0);
+
+  if (f_a_0 == 0.0) {
+    // f(a_0) == 0.0 so we do a weak CG cut
```
aliceb-nv, line 2132: If `fractional_part(inequality_rhs)` is in (0, 1e-2), neither a weak CG cut nor a strong CG cut are generated (since it will go through the fallthrough line 2169). Is this intended?

cpp/src/dual_simplex/branch_and_bound.cpp
chris-maes: Ideally, we would drop the cuts and continue. I have some code that attempted this, but it is not trivial to remove cuts. I will need to add that in a follow up PR.

akifcorduk: Can't we use a copy of the problem before cuts?

cpp/src/dual_simplex/cuts.cpp
chris-maes: Yes, that is intended, for now. The numerics of the cut are often bad for small fractional parts.

cpp/src/mip/solver.cu
chris-maes: I think when you want presolve off, you want all presolve off. But we can follow up and make presolve an int. So if presolve = 0 it's off, presolve = 1 cuopt internal, presolve = 2 cuopt internal + paplio, presolve = -1 automatic

cpp/src/dual_simplex/bound_flipping_ratio_test.cpp
chris-maes: Before bound flipping was returning -3 for concurrent halt. I made it so it now returns CONCURRENT_HALT_RETURN=-2 to be consistent with other functions.

cpp/src/dual_simplex/branch_and_bound.cpp
chris-maes: It's not just the problem, there is quite a bit of additional state.

cpp/include/cuopt/linear_programming/mip/solver_settings.hpp
```diff
@@ -83,12 +83,22 @@ class mip_solver_settings_t {
...
+  i_t max_cut_passes            = 10;  // number of cut passes to make
+  i_t mir_cuts                  = -1;
+  i_t mixed_integer_gomory_cuts = -1;
+  i_t knapsack_cuts             = -1;
+  i_t strong_chvatal_gomory_cuts      = -1;
+  i_t reduced_cost_strengthening      = -1;
+  f_t cut_change_threshold            = 1e-3;
+  f_t cut_min_orthogonality           = 0.5;
   i_t mip_batch_pdlp_strong_branching = 0;
+  i_t num_gpus                        = 1;
```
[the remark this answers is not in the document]
chris-maes: This is in the C++ side. So it should not change C API ABI stability.

cpp/src/dual_simplex/basis_updates.cpp
chris-maes: Fixed. Thanks

cpp/src/dual_simplex/bound_flipping_ratio_test.cpp
chris-maes: I added the constants for the other return codes. Thank you.

cpp/src/dual_simplex/cuts.hpp
chris-maes: This is nice. Done. Thanks

cpp/src/dual_simplex/folding.cpp
chris-maes: Done.

chris-maes: DOne

cpp/src/dual_simplex/basis_updates.cpp
```diff
@@ -1108,6 +1108,217 @@ i_t basis_update_t<i_t, f_t>::lower_triangular_multiply(const csc_matrix_t<i_t,
...
+        const i_t u_col = 2 * k;
+        const i_t v_col = 2 * k + 1;
+        const f_t mu    = mu_values_[k];
+
+        // dot = u^T * b
+        f_t dot         = dot_product(u_col, xi_workspace_, x_workspace_);
+        const f_t theta = dot / mu;
+        if (std::abs(theta) > zero_tol) {
+          add_sparse_column(S_, v_col, -theta, xi_workspace_, nz, x_workspace_);
+        }
```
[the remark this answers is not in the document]
chris-maes: We don't want to change the basis when we are appending cuts. mu_values_  is already guarded inside of `update()`

PR #820, opened 2026-02-02
[WIP, POC] Unified threading model in MIP solver

nguidotti:
This PR unifies the threading model across the MIP solver, such that the code now uses the tasking model from OpenMP. The only exception is the Papilo presolver that uses Intel TBB. 

More specifically, this PR
- Solves the CPU oversubscription problem. The solver now respect the number of threads set to the user, with the exception of Papilo or threads created by the CUDA runtime.
- Removes overheads from creating and destroying `std::thread`. This also 
- Migrates RINS from `std::thread` to `omp task`. Similar to previous logic, one instance of RINS can run at a time. 
- Migrates `CPU FJ` from `std::thread` to `omp task`. There are a few limitations 
   - `scratch_cpu_fj_on_lp_opt` and `scratch_cpu_fj` are running for the entire program. This essentially allocate two dedicated threads to these functions, while other routines needs to share the remaining CPU resources. This may hurt the performance for low core count CPUs.
   - Since there is a small delay between the task creation and its start (since the threads may be busy), the GPU FJ may finish before the CPU FJ even start when racing. 
-  Replaces locks in the `pseudocost` with atomics to reduce contention. 

This is a POC, hence it requires some polish before it can be merged.

## Checklist

- Testing
   - [ ] Added tests
   - [ ] NA
- Documentation
   - [ ] Added new documentation
   - [ ] NA

## Summary by CodeRabbit

* **New Features**
  * Added `--reliability-branching` CLI option to control advanced branching strategy in MIP solver (enabled by default)
  * Introduced parallel and single-threaded solve modes for branch-and-bound algorithm

* **Improvements**
  * Enhanced parallel branch-and-bound performance through improved work distribution and task scheduling
  * Improved solver synchronization and multi-threaded execution efficiency
  * Added reliability-based variable selection strategy for optimal branching decisions

* **Refactoring**
  * Internal solver architecture updated for better scalability with worker-pool model


nguidotti:
The base of this PR is too old. I will create a new PR with an updated base

PR #821, opened 2026-02-02
remove pip.conf migration code in CI scripts, fix pr-builder configuration

jameslamb:
## Description

Follow-up to changes for https://github.com/rapidsai/build-planning/issues/241

* removes `rapids-init-pip` feature flags now that those have become the default behavior (https://github.com/rapidsai/gha-tools/pull/242)

## Summary by CodeRabbit

* **Chores**
  * Simplified CI build and test initialization by removing several explicit environment overrides so dependency setup now relies on default behavior.
  * CI pull-request builder step now runs unconditionally and forwards full job context to invoked workflows, improving consistency of PR builds.

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -35,6 +35,9 @@ jobs:
       - test-self-hosted-server
     secrets: inherit
     uses: rapidsai/shared-workflows/.github/workflows/pr-builder.yaml@main
+    if: always()
+    with:
+      needs: ${{ toJSON(needs) }}
   check-lean-ci:
     runs-on: ubuntu-latest
     outputs:


ci/build_wheel.sh
```diff
@@ -10,8 +10,6 @@ package_dir=$2
 
 source rapids-configure-sccache
 source rapids-date-string
-RAPIDS_INIT_PIP_REMOVE_NVIDIA_INDEX="true"
-export RAPIDS_INIT_PIP_REMOVE_NVIDIA_INDEX
```
jameslamb, line 14: Just picking an arbitrary place on the diff to re-post this comment, so follow-ups can be grouped together in a thread.
---

hmmmm I'm not really sure why this isn't mergeable. `pr-builder / run` is getting skipped:

image

https://github.com/NVIDIA/cuopt/actions/runs/21615281140/job/62369023603?pr=821

I tried manually re-running it and it was skipped again.

Maybe the changes from #805 and/or this PR have left it in a weird state? I'll investigate. It shouldn't be the case that things like `docs-build` being skipped cause the entire `pr-builder` run to be skipped.

jameslamb: Ah, there's a GitHub outage affecting status checks: https://www.githubstatus.com/incidents/f314nlctbfs5

jameslamb: GitHub say they've pushed a fix but re-running the `pr-builder` job, it still showed as `skipped`. I've just fully re-run all CI on this PR... it's possible that other status checks from earlier jobs were lost during the incident or something.

If that still fails, I'll investigate whether there's something else going on here, maybe unrelated to that GitHub incident.

jameslamb: Jobs getting skipped should be no problem for `pr-builder`:

https://github.com/rapidsai/shared-workflows/blob/ce125aca3e92a02adb84a78f7361424bc08ee62c/.github/workflows/pr-builder.yaml#L20

I haven't seen this problem in any other repo using that. My theories right now are either that this PR was in a weird state because of the GitHub incident related to statuses today, or that some of this custom stuff done in this repo is confusing the check:

https://github.com/NVIDIA/cuopt/blob/ac400cdca1ae3e5f88247f50825d752cd180e531/.github/workflows/pr.yaml#L38

https://github.com/NVIDIA/cuopt/blob/ac400cdca1ae3e5f88247f50825d752cd180e531/.github/workflows/pr.yaml#L61

https://github.com/NVIDIA/cuopt/blob/ac400cdca1ae3e5f88247f50825d752cd180e531/.github/workflows/pr.yaml#L76

jameslamb: Aha! I see the problem!

 `pr-builder` is misconfigured here:

* needs an `if: always()` to ensure it isn't skipped when other jobs are skipped
* needs to actually pass the list of required jobs through to the workflow

Like this:

https://github.com/rapidsai/cudf/blob/b978750d0af8ddb20f4112fdbd6936e49297eafc/.github/workflows/pr.yaml#L47-L49

The fact that those were missing means this workflow has probably never worked in this repo and it's been doing a full run of CI on every commit 🙃 

Just pushed https://github.com/NVIDIA/cuopt/pull/821/commits/20c564e8309f47459d55cf7889f6691d480de05b fixing that.

ramakrishnap-nv: Thank you @jameslamb, I always assumed it was designed by default

PR #822, opened 2026-02-03
Update solution object to include the LP solver

### LINKED ISSUE #787: [FEA] Add a way to retrieve the solver type from the solution object
In the current API, there is no way to retrieve which solver was used for solving the LP from the solution object (`optimization_problem_solution_t` in C++) when using concurrent mode. To enable this, we need to change the API from C++, C and Python. 

We also should update the `solve_by_pdlp` flag since a LP can be solved with barrier, dual simplex and pdlp when running in concurrent mode.

nguidotti:
This PR replaced `solved_by_pdlp` with `solved_by` in `optimization_problem_solution_t` and all associated objects, such that now it is possible to retrieve which method was used for solving the LP when running in concurrent mode. This also fix a typo in the `CUOPT_TERMINATION_STATUS` and updates the B&B logs to display the method used for solving the root relaxation.

## Issue
Closes #787 

## Checklist

- Testing
   - [ ] Added tests
   - [ ] NA
- Documentation
   - [ ] Added new documentation
   - [ ] NA


nguidotti:
@coderabbit review

python/cuopt/cuopt/linear_programming/solution/solution.py
```diff
@@ -116,8 +117,8 @@ class Solution:
         Time used for pre-solve
     solve_time: Float64
         Solve time in seconds
-    solved_by_pdlp: bool
-        Whether the problem was solved by PDLP or Dual Simplex
+    solved_by: enum
+        Whether the problem was solved by Dual Simplex(1), PDLP(2) or Barrier(3)
```
ramakrishnap-nv: @nguidotti Should we be exposing these enums in python or we expect users to use direct integers as input ?

python/cuopt/cuopt/linear_programming/solution/solution.py
```diff
@@ -116,8 +117,8 @@ class Solution:
         Time used for pre-solve
     solve_time: Float64
         Solve time in seconds
-    solved_by_pdlp: bool
```
ramakrishnap-nv, line 119: This would be a breaking change since we are replacing this with something else. May for this time, we can cleanly remove it, but for next time lets plan a support for 2 releases with deprecation for major elements before completely removing it.

python/cuopt/cuopt/linear_programming/solution/solution.py
ramakrishnap-nv: May be provide details on SolverMethod so users would know how to use it.

nguidotti: I think all parameters in this object are set by the solver since this contains the solution of the problem. We are using `SolverMethod` for the enum here (which already existed).

nguidotti: Correct me if I am wrong @Iroy30

python/cuopt/cuopt/linear_programming/solution/solution.py
nguidotti: If you prefer, we can keep the `solved_by_pdlp` for now, set as `solved_by_pdlp = solved_by == SolverMethod.PDLP` and add a deprecation warning.

nguidotti: Done. See [cc5844b](https://github.com/NVIDIA/cuopt/pull/822/commits/cc5844b5f486f149804cbd91ce85a16bc7f481bd)

PR #846, opened 2026-02-10
Rename and split up cpp/src directories

chris-maes:
This PR makes the following changes:

- The `dual_simplex` directory is split into the following directories: `barrier`, `branch_and_bound`, `cuts`
- The `linear_programming` directory is renamed to `pdlp`
- The `mip` directory is renamed to `mip_heurisitics`

This better reflects the structure of the code. 

This PR has no other code changes. It only alters the `#include`  paths

This PR also includes a fix for C++20 in the cuOpt test builds

### DIFF
--- a/cpp/tests/CMakeLists.txt
+++ b/cpp/tests/CMakeLists.txt
@@ -1,4 +1,4 @@
-# cmake-format: off
+﻿# cmake-format: off
 # SPDX-FileCopyrightText: Copyright (c) 2021-2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 # cmake-format: on
@@ -36,7 +36,13 @@ set(CUOPT_TEST_DIR ${CMAKE_CURRENT_SOURCE_DIR})
 # ################################################################ ------------------------------------------------------------------
 function(ConfigureTest CMAKE_TEST_NAME)
     add_executable(${CMAKE_TEST_NAME} ${ARGN})
-
+    set_target_properties(${CMAKE_TEST_NAME}
+      PROPERTIES
+      CXX_STANDARD 20
+      CXX_STANDARD_REQUIRED ON
+      CUDA_STANDARD 20
+      CUDA_STANDARD_REQUIRED ON
+    )
     target_include_directories(${CMAKE_TEST_NAME}
         PRIVATE
         "${CUOPT_TEST_DIR}/../src"


cpp/src/branch_and_bound/CMakeLists.txt
```diff
@@ -0,0 +1,17 @@
...
+
+set(BRANCH_AND_BOUND_SRC_FILES
+  ${CMAKE_CURRENT_SOURCE_DIR}/branch_and_bound.cpp
+  ${CMAKE_CURRENT_SOURCE_DIR}/mip_node.cpp
+  ${CMAKE_CURRENT_SOURCE_DIR}/pseudo_costs.cpp
+  ${CMAKE_CURRENT_SOURCE_DIR}/diving_heuristics.cpp
+  )
+
+# Uncomment to enable debug info
+#set_source_files_properties(${BRANCH_AND_BOUND_SRC_FILES} DIRECTORY ${CMAKE_SOURCE_DIR} PROPERTIES COMPILE_OPTIONS "-g1")
```
akifcorduk: I think we can remove this, build configs should come from the upstream cmake files

cpp/src/cuts/CMakeLists.txt
```diff
@@ -0,0 +1,14 @@
...
+# SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+# cmake-format: on
+
+set(CUTS_SRC_FILES
+  ${CMAKE_CURRENT_SOURCE_DIR}/cuts.cpp
+  )
+
+# Uncomment to enable debug info
+#set_source_files_properties(${CUTS_SRC_FILES} DIRECTORY ${CMAKE_SOURCE_DIR} PROPERTIES COMPILE_OPTIONS "-g1")
```
akifcorduk: Here too, we can remove this

cpp/src/branch_and_bound/CMakeLists.txt
chris-maes: The use case here is to allow debugging just for files in the current directory. Without this I can't seem to get debugging symbols in the host code.

cpp/src/barrier/CMakeLists.txt
```diff
@@ -0,0 +1,17 @@
+# cmake-format: off
+# SPDX-FileCopyrightText: Copyright (c) 2024-2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
ramakrishnap-nv:
```suggestion
# SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```

cpp/src/branch_and_bound/CMakeLists.txt
ramakrishnap-nv: Agree, this should be passed from upper main Cmake which is being passed by build.sh. If it is not happening, then we are missing some configuration on top.

rgsl888prabhu (changes requested):

cpp/src/branch_and_bound/CMakeLists.txt
ramakrishnap-nv: But we can tackle this in a follow-up PR and fix it so we can merge this PR if all things pass. 

@chris-maes Can you please create an issue to follow-up for this ?

akifcorduk: you can pass a flag to `./build.sh --host-lineinfo ` which sets "HOST_LINEINFO". it does exactly the commented out code and should print out debug symbols.

chris-maes: I removed the commented code. But none of those options are simpler or better than simply adding the commented line of code.

PR #851, opened 2026-02-12
Consider objective function integrality when pruning + bug fixes

aliceb-nv:
This PR adds support for rounding the lower bounds in the branch and bound tree when the problem is proven to have an integral objective function, which allows for tighter pruning.
If objective coefficients are not integer, but are rational numbers with small enough denominators, the objective is scaled by the smallest possible integer that makes the objective function integral. (i.e. in the case of obj = 0.5x + 0.5y; it is scaled by two to obj = 1x + 1y)

Also included are fixes for bugs recently encountered:
 - Solutions sent to the repair queue would not be recrushed after the cuts passes
 - Column deduplication in `problem_t::substitute_variables` would not handle the case of more than two duplicates at once
 - Removed the cached data pointer in memory_instrumentation as it was bug-prone and is no longer necessary.

Number of optimals in 10min on Eos:
 - baseline: 44
 - PR: 57 

## Checklist

- Testing
   - [ ] Added tests
   - [ ] NA
- Documentation
   - [ ] Added new documentation
   - [ ] NA

## Summary by CodeRabbit

* **Bug Fixes**
  * Repair and queuing now consistently use the correct solution form across deterministic and nondeterministic flows.
  * MIP-gap logic updated to treat solutions at the bound as optimal.

* **Improvements**
  * Better handling for integral objectives: detection, bounds rounding, and cut-off adjustments to respect integrality.
  * New utilities to assess and (re)scale objective integrality for more robust presolve and solving.

* **Chores**
  * Increased CI package size thresholds.

### DIFF
--- a/ci/validate_wheel.sh
+++ b/ci/validate_wheel.sh
@@ -22,11 +22,11 @@ PYDISTCHECK_ARGS=(
 if [[ "${package_dir}" == "python/libcuopt" ]]; then
     if [[ "${RAPIDS_CUDA_MAJOR}" == "12" ]]; then
         PYDISTCHECK_ARGS+=(
-            --max-allowed-size-compressed '635Mi'
+            --max-allowed-size-compressed '645Mi'
         )
     else
         PYDISTCHECK_ARGS+=(
-            --max-allowed-size-compressed '475Mi'
+            --max-allowed-size-compressed '485Mi'
         )
     fi
 elif [[ "${package_dir}" != "python/cuopt" ]] && \


cpp/src/mip_heuristics/problem/problem.cu
```diff
@@ -1296,6 +1449,9 @@ void problem_t<i_t, f_t>::substitute_variables(const std::vector<i_t>& var_indic
       //  atomicAdd(objective_offset, objective_offset_difference);
       atomicAdd(&objective_coefficients[substituting_var_idx],
                 objective_coefficients[var_idx] * substitute_coefficient[idx]);
+      // Substitution changes the constraint coefficients on x_B, invalidating
+      // any implied-integrality proof that relied on the original structure.
+      var_flags[substituting_var_idx] &= ~(i_t)VAR_IMPLIED_INTEGER;
```
akifcorduk, line 1459: What are var flags?

aliceb-nv: I created this structure back when we added Papilo, in order to store additional information relating to columns. For now it is used only to store if the variable was proven to be implied integral :)

ci/validate_wheel.sh
```diff
@@ -22,11 +22,11 @@ PYDISTCHECK_ARGS=(
 if [[ "${package_dir}" == "python/libcuopt" ]]; then
     if [[ "${RAPIDS_CUDA_MAJOR}" == "12" ]]; then
         PYDISTCHECK_ARGS+=(
-            --max-allowed-size-compressed '635Mi'
+            --max-allowed-size-compressed '645Mi'
         )
     else
         PYDISTCHECK_ARGS+=(
-            --max-allowed-size-compressed '475Mi'
+            --max-allowed-size-compressed '485Mi'
```
ramakrishnap-nv, line 29: Was there any issue with cuda 13 wheel size, or it was just precautionary ?

aliceb-nv: I updated both to err on the safe side

Issue #855, opened 2026-02-13
[BUG] Negative gap on air05 instance


mtanneau:
[Edit]: issue updated following discussion in #858 to focus on negative difference between upper and lower bound.

**Describe the bug**

Solving instance `air05` (see below for MPS file) on a DGX Spark desktop results in a final dual bound that exceeds the final primal bound.

Sample log (see logs from additional runs on v26.02 below)
```txt
Setting CUDA_MODULE_LOADING to EAGER
Reading file air05.mps
cuOpt version: 26.4.0, git hash: b3bb21e4, host arch: aarch64, device archs: 121a-real
CPU: Unknown, threads (physical/logical): 1/20, RAM: 96.08 GiB
CUDA 12.9, device: NVIDIA GB10 (ID 0), VRAM: 119.70 GiB
CUDA device UUID: 43bce240-d7d6-65e5-ca7a-e351bab5de7a

Solving a problem with 426 constraints, 7195 variables (7195 integers), and 52121 nonzeros
Problem scaling:
Objective coefficents range:          [4e+01, 3e+03]
Constraint matrix coefficients range: [1e+00, 1e+00]
Constraint rhs / bounds range:        [1e+00, 1e+00]
Variable bounds range:                [0e+00, 1e+00]

Original problem: 426 constraints, 7195 variables, 52121 nonzeros
Calling Papilo presolver (git hash 741a2b9c)
Presolve status: reduced the problem
Presolve removed: 90 constraints, 1116 variables, 16171 nonzeros
Presolved problem: 336 constraints, 6079 variables, 35950 nonzeros
Objective function is integral
Papilo presolve time: 0.42
Objective offset 9167.000000 scaling_factor 1.000000
Model fingerprint: 0x3257a27
Running presolve!
Unused variables detected, eliminating them! Unused var count 4
After cuOpt presolve: 335 constraints, 6075 variables, objective offset 9167.000000.
cuOpt presolve time: 3.85

Solving LP root relaxation in concurrent mode
Skipping column scaling
Dual Simplex Phase 1
Dual feasible solution found.
Dual Simplex Phase 2
 Iter     Objective           Num Inf.  Sum Inf.     Perturb  Time
    0 -8.1936000000000000e+04     277 2.31020000e+04 0.00e+00 4.51
    1 -8.0926000000000000e+04     266 1.48790000e+04 0.00e+00 4.51
[...]

PR #858, opened 2026-02-13
Fix printing of non-zero coeffs after cut passes

mtanneau:
## Description

I tracked the issue of non-sensical number of non-zeros to this line:
https://github.com/NVIDIA/cuopt/blob/f91f9cc7ff983b0edd6a9653cd23320fa83f7277/cpp/src/branch_and_bound/branch_and_bound.cpp#L2346

And replaced the manual access to `A.col_start[A.n]` with `A.nnz()`.

For completeness, I tried the following variants:
* ❌ same as main
* ✅ `A.nnz()`
* ✅ Grab number of non-zeros outside of the `printf` statement, then print it. Use templated integer type.
  ```cpp
  const i_t num_non_zeros = original_lp_.A.col_start[original_lp_.A.n];
  ... printf("%d", num_non_zeros);  // illustrative, not functional as-is
  ```
* ❌ Grab number of non-zeros outside of the `printf` statement, then print it. Use `auto` type.
   ```cpp
  const auto num_non_zeros = original_lp_.A.col_start[original_lp_.A.n];
  ... printf("%d", num_non_zeros);
  ```

I suspect that directly accessing `A.col_start[A.n]` causes the value to be cast to the wrong type, which then causes some kind of issue inside `printf`. To see this, change the printf statement above to
```cpp
settings_.log.printf("Size with cuts : %d constraints, %d variables, %d %d %d nonzeros\n",
                         original_lp_.num_rows,
                         original_lp_.num_cols,
                         original_lp_.A.col_start[original_lp_.A.n],
                         12345);
```
which should output something like 
```txt
Size with cuts : 336 constraints, 6411 variables, [-994767476, -1061948824, 12345] nonzeros
```
when solving the `air05.mps` instance.

I also checked that all 4 variants are displayed as expected when printed directly to `std::cout`.

## Issue

Addresses part of #855, namely the number of non-zeros after cuts have been added

## Checklist

- Testing
   - [ ] Added tests
   - [ ] NA
- Documentation
   - [ ] Added new documentation
   - [x] NA

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected nonzero count reporting in problem size diagnostics to accurately reflect actual values, improving the accuracy of size metrics displayed during problem solving.


mtanneau:
Thanks @chris-maes and @aliceb-nv for looking into this!
I am closing this PR, and will update the issue to focus on the negative gap

PR #890, opened 2026-02-17
Fix issue with incorrect lower bound on air05

chris-maes:
Fix an issue with incorrect lower bound on air05.

In 26.02, with the new scheduler worker model, it looks like the lower bound is not computed correctly if there are nodes still remaining in the `node_queue`. This PR fathoms all the nodes and pulls the lower bound from the search tree. 

In 26.02 we were also incorrectly fathoming nodes with objectives less than the optimal objective, due to the `rel_gap` check. This PR removes that check. 

In 26.04 we were incorrectly setting the cutoff to one less than it should be when the objective was integral. This PR fixes that bug.


aliceb-nv (approved):

cpp/src/branch_and_bound/branch_and_bound.cpp
```diff
@@ -2471,8 +2471,25 @@ mip_status_t branch_and_bound_t<i_t, f_t>::solve(mip_solution_t<i_t, f_t>& solut
     lower_bound    = deterministic_compute_lower_bound();
     solver_status_ = deterministic_global_termination_status_;
   } else {
-    lower_bound = node_queue_.best_first_queue_size() > 0 ? node_queue_.get_lower_bound()
-                                                          : search_tree_.root.lower_bound;
+    if (node_queue_.best_first_queue_size() > 0) {
```
chris-maes, line 2474: Does this work correctly if we hit a time limit but still have nodes in the queue?

chris-maes: I think time limit does not go through this part of the code.

mtanneau: What if user sets a (large enough) gap tolerance and there are still nodes in the queue?
E.g.:
* Gap tolerance is 5%
* Incumbent obj is 100
* Current best bound is 95
* There's still an open node with lower bound 98

This could be problematic if the user wanted to restart the optimization afterwards with a tighter gap tolerance, since that open node would no longer be in the queue.

cpp/src/branch_and_bound/branch_and_bound.cpp
```diff
@@ -2471,8 +2471,25 @@ mip_status_t branch_and_bound_t<i_t, f_t>::solve(mip_solution_t<i_t, f_t>& solut
...
+      // We need to clear the queue and use the info in the search tree for the lower bound
+      while (node_queue_.best_first_queue_size() > 0) {
+        std::optional<mip_node_t<i_t, f_t>*> start_node = node_queue_.pop_best_first();
+
+        if (!start_node.has_value()) { continue; }
+        if (upper_bound_ < start_node.value()->lower_bound) {
```
mtanneau, line 2480: Since nodes are processed in increasing lower bound value, I think you may terminate this loop once this test evaluates to `true`, since any node remaining in the queue would have a higher dual bound (and therefore should not affect the lower bound computation).

This could save time in case where there remains a large number of open nodes with `lower_bound > upper_bound`

cpp/src/branch_and_bound/branch_and_bound.cpp
```diff
@@ -2471,8 +2471,25 @@ mip_status_t branch_and_bound_t<i_t, f_t>::solve(mip_solution_t<i_t, f_t>& solut
     lower_bound    = deterministic_compute_lower_bound();
     solver_status_ = deterministic_global_termination_status_;
   } else {
-    lower_bound = node_queue_.best_first_queue_size() > 0 ? node_queue_.get_lower_bound()
-                                                          : search_tree_.root.lower_bound;
```
mtanneau, line 2475: Apologies if I'm being naive, but wouldn't it be sufficient to set `lower_bound = min(lower_bound, upper_bound)` after querying the best bound from the queue/tree?

cpp/src/branch_and_bound/branch_and_bound.cpp
chris-maes: The point of this loop is to correctly mark the nodes as fathomed in the search tree, so that the search tree can correctly compute the lower bound. This means we want to go through all the nodes left in the queue.

cpp/src/branch_and_bound/branch_and_bound.cpp
chris-maes: No. That would lead to incorrect results on some problems. 

The upper bound doesn't enter into the lower bound computation at this point. The upper bound only effects whether a node is fathomed or not. 

We want to use the search tree to correctly calculate the lower bound as the minimum over all  leaf nodes in the tree.

cpp/src/branch_and_bound/branch_and_bound.cpp
chris-maes: Yes, that is correct. We need to handle the case where there are some nodes that still belong in the queue. This is done in the latest commit.

cpp/src/branch_and_bound/branch_and_bound.cpp
chris-maes: Ah I misunderstood. Yes, we can break out of the loop when lower bound is greater than the upper bound. And those nodes that remain in queue provide a valid lower bound.

cpp/src/branch_and_bound/branch_and_bound.cpp
```diff
@@ -2471,8 +2471,28 @@ mip_status_t branch_and_bound_t<i_t, f_t>::solve(mip_solution_t<i_t, f_t>& solut
...
+          continue;
+        } else {
+          break;
+        }
+      }
+      lower_bound = node_queue_.best_first_queue_size() > 0 ? node_queue_.get_lower_bound()
+                                                            : search_tree_.root.lower_bound;
+    } else {
+      lower_bound = search_tree_.root.lower_bound;
+    }
```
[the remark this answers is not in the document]
chris-maes: What if we just pushed `start_node` back into the node_queue inside the `else` branch?

PR #893, opened 2026-02-18
Fix batch pdlp and python api support

Kh4ster:
This PR fixes a few issues making batch PDLP crash occasionally.

- Refactor cuda graph
    Create a separate cu for CUDA graph
    Add fail safe in CUDA graph to avoid destruction while capturing
    Avoid throwing in case of error to avoid falling triggering destruction while capturing
- Stop using different streams in adaptive step size to avoid potential issues with cublas handle sharing
- Changed all thrust::transform_reduce to CUB::reduce to avoid unintentionally creating new streams and triggering cuda mallocs
- Use .value() on streams for all CUB calls to avoid having stream issue
- Added more asserts on nan
- Correctly use set_stream before trigger a cublasDgeam
- Use a separate handler and stream when starting batch pdlp

Additionally this also fixes a few compilation issues linked to PDLP.

This PR also allows batch PDLP to be toggled on from the Python side.

## Summary by CodeRabbit

* **New Features**
  * New string-based presolver option for LP solvers (multiple modes).
  * Added Batch PDLP strong branching toggle for MIP root behavior.

* **Bug Fixes / Stability**
  * Improved runtime validation and NaN/Inf checks across solver paths.
  * Stream and reduction changes for more robust GPU execution; CUDA graph capture added for performance.

* **Documentation**
  * Settings docs updated to include the new options.

* **Tests**
  * Added tests for batch PDLP strong branching.

### DIFF
--- a/cpp/src/pdlp/restart_strategy/pdlp_restart_strategy.cu
+++ b/cpp/src/pdlp/restart_strategy/pdlp_restart_strategy.cu
@@ -2008,465 +2008,466 @@ void pdlp_restart_strategy_t<i_t, f_t>::solve_bound_constrained_trust_region(
                    "Incorrect primal reverse iterator");
       index_start_primal = thrust::raw_pointer_cast(&*highest_negInf_primal) - threshold_.data() +
                            1;  // + 1 to go after last negInf
-      if (lowest_inf != end) {
+      testing_range_low_.set_value_async(index_start_primal, stream_view_);
+    } else  // No negInf found, start is 0
+      testing_range_low_.set_value_async(index_start_primal, stream_view_);
+    if (lowest_inf != end) {
+      cuopt_assert(device_to_host_value(thrust::raw_pointer_cast(&*lowest_inf)) ==
                      std::numeric_limits<f_t>::infinity(),
                    "Incorrect primal iterator");
-                     index_end_primal =
-                       thrust::raw_pointer_cast(lowest_inf) -
-                       threshold_
-                         .data();  // no - 1 to go before the first inf because end is not included
-                     testing_range_high_.set_value_async(index_end_primal, stream_view_);
-      } else  // No inf found, end is primal_size_h_
-        testing_range_high_.set_value_async(index_end_primal, stream_view_);
-      cuopt_assert(index_start_primal <= index_end_primal,
-                   "Start should be strictly smaller than end");
+      index_end_primal =
+        thrust::raw_pointer_cast(lowest_inf) -
+        threshold_.data();  // no - 1 to go before the first inf because end is not included
+      testing_range_high_.set_value_async(index_end_primal, stream_view_);
+    } else  // No inf found, end is primal_size_h_
+      testing_range_high_.set_value_async(index_end_primal, stream_view_);
+    cuopt_assert(index_start_primal <= index_end_primal,
+                 "Start should be strictly smaller than end");
 
-      cuopt_assert(!thrust::any_of(handle_ptr_->get_thrust_policy(),
-                                   threshold_.data() + index_start_primal,
-                                   threshold_.data() + index_end_primal,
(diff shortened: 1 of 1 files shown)


cpp/src/pdlp/step_size_strategy/adaptive_step_size_strategy.cu
```diff
@@ -371,19 +365,14 @@ void adaptive_step_size_strategy_t<i_t, f_t>::compute_interaction_and_movement(
...
     ||(y' - y)||
     (y' - y)_t . A @ (x' - x)
 
     And finally merge the results
   */
 
-  // We need to make sure both dot products happens after previous operations (next_primal/dual)
-  // Thus, we add another node in the main stream before starting the SpMVs
-
-  if (!batch_mode_) deltas_are_done_.record(stream_view_);
```
akifcorduk, line 385: Is removing those deltas_are_done_ lines intentional?

cpp/src/pdlp/utilities/ping_pong_graph.cu
```diff
@@ -0,0 +1,123 @@
...
+}
+
+template <typename i_t>
+ping_pong_graph_t<i_t>::~ping_pong_graph_t()
+{
+#ifndef CUPDLP_DEBUG_MODE
+  if (!is_legacy_batch_mode_) {
+    // This should not happen, but in case a graph was capturing while destroying the object
+    if (capture_even_active_ || capture_odd_active_) {
+      cancel_active_capture();
```
akifcorduk: I think this might hide real errors. If this is destroyed while its own stream is capturing, it is a logic issue. If this is destroyed while another stream/thread is capturing, we will lose the graph on the other stream/thread and cudaGraphLaunch will launch an incomplete/empty graph.

cpp/src/pdlp/pdlp.cu
```diff
@@ -1186,24 +1188,42 @@ static void compute_stats(const rmm::device_uvector<f_t>& vec,
...
+
+  rmm::device_scalar<f_t> d_smallest(stream);
+  rmm::device_scalar<f_t> d_largest(stream);
+  rmm::device_scalar<f_t> d_sum(stream);
+
+  auto min_nz_iter = thrust::make_transform_iterator(vec.cbegin(), min_nonzero);
+  auto abs_iter    = thrust::make_transform_iterator(vec.cbegin(), abs_op);
+
+  void* d_temp   = nullptr;
+  size_t bytes_1 = 0, bytes_2 = 0, bytes_3 = 1;
```
akifcorduk: bytes_3 should also be 0 right?

akifcorduk (approved):

cpp/src/pdlp/step_size_strategy/adaptive_step_size_strategy.cu
Kh4ster: Yes. It's actually illegal to run on the same cublas handler while using different streams which I was doing here. The error would extremly rarely trigger but it can results in an incorrect results for the dot product.

cpp/src/pdlp/utilities/ping_pong_graph.cu
Kh4ster: Yes I agree. This was to create a failsafe in case the PDLP object is destroyed while a capture is happenning which is an unitended behavior. Hence the log error above. You would still kill the application?

cpp/src/pdlp/pdlp.cu
Kh4ster: Yes that's weird ^^

cpp/src/pdlp/utilities/ping_pong_graph.cu
Kh4ster: This behavior has been changed now

PR #899, opened 2026-02-23
Run FJ heuristics before and during presolve

aliceb-nv:
This presolve lets FJ (both CPU and GPU) run before presolving is completed, in order to provide incumbent callbacks as early as possible.
While we cannot yet crush these incumbents into the fully-presolved problem representation, their primal values can be used to prune the B&B and provide stronger reduced cost fixings.

With this PR, we now find 191 feasible in the first 15s (145 before), and the primal integral in 10min is now 18.7% (was 19.8% before).

## Summary by CodeRabbit

* **New Features**
  * Early feasibility-jump heuristics (CPU & GPU) during presolve that report incumbents via callbacks and seed the main solve.
  * New early-incumbent heuristic framework with lifecycle, timing, and callback support.

* **Bug Fixes**
  * Null-pointer guards for scaling and safer Python callback invocation with GIL handling.
  * Improvement callbacks now fire only for strictly better solutions.

* **Chores**
  * Centralized cutoff handling and propagation of presolve incumbents; support for providing an initial cutoff.

### DIFF
--- a/ci/validate_wheel.sh
+++ b/ci/validate_wheel.sh
@@ -26,7 +26,7 @@ if [[ "${package_dir}" == "python/libcuopt" ]]; then
         )
     else
         PYDISTCHECK_ARGS+=(
-            --max-allowed-size-compressed '485Mi'
+            --max-allowed-size-compressed '490Mi'
         )
     fi
 elif [[ "${package_dir}" != "python/cuopt" ]] && \


cpp/src/branch_and_bound/branch_and_bound.hpp
```diff
@@ -108,6 +108,14 @@ class branch_and_bound_t {
 
   bool stop_for_time_limit(mip_solution_t<i_t, f_t>& solution);
 
+  // Set a cutoff bound from an external source (e.g., early FJ during presolve).
+  // Used for node pruning and reduced cost strengthening but NOT for gap computation.
+  // Unlike upper_bound_, this does not imply a verified incumbent solution exists.
+  void set_initial_cutoff(f_t bound) { initial_cutoff_ = bound; }
+
+  // Effective cutoff for node pruning: min of verified incumbent and external cutoff.
+  f_t get_cutoff() const { return std::min(upper_bound_.load(), initial_cutoff_); }
```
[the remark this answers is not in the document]
aliceb-nv: no, initial_cutoff is only set before B&B starts

cpp/src/mip_heuristics/feasibility_jump/early_cpufj.cu
```diff
@@ -0,0 +1,80 @@
...
+#include "early_cpufj.cuh"
+
+#include <mip_heuristics/feasibility_jump/fj_cpu.cuh>
+#include <mip_heuristics/mip_constants.hpp>
+#include <utilities/logger.hpp>
+
+namespace cuopt::linear_programming::detail {
+
+template <typename i_t, typename f_t>
+early_cpufj_t<i_t, f_t>::early_cpufj_t(
```
rg20, line 17: why do you need a separate class for this one? Can't you repurpose the existing one?

aliceb-nv: This is just a wrapper around said existing CPUFJ class for convenience in that specific scenario. This is helpful because we run cpufj twice in the early-heuristics: before papilo presolve (on the original problem), and before cuopt presolve (on the papilo-presolved problem)

cpp/src/mip_heuristics/feasibility_jump/feasibility_jump.cu
```diff
@@ -933,6 +933,22 @@ i_t fj_t<i_t, f_t>::host_loop(solution_t<i_t, f_t>& solution, i_t climber_idx)
...
+          last_reported_objective_ = user_obj;
+          // Copy assignment to host for callback
+          std::vector<f_t> h_assignment(solution.assignment.size());
+          raft::copy(h_assignment.data(),
+                     solution.assignment.data(),
+                     solution.assignment.size(),
+                     climber_stream);
+          climber_stream.synchronize();
+          improvement_callback(user_obj, h_assignment);
+        }
```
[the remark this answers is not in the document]
aliceb-nv: Was updated to compare the solver-space objectives, which are always minimization.

cpp/src/mip_heuristics/solve.cu
```diff
@@ -153,9 +166,38 @@ mip_solution_t<i_t, f_t> run_mip(detail::problem_t<i_t, f_t>& problem,
...
+      presolver_ptr->uncrush_primal_solution(assignment, user_assignment);
+      invoke_solution_callbacks(mip_callbacks, user_obj, user_assignment, no_bound);
+    };
+    early_cpufj = std::make_unique<detail::early_cpufj_t<i_t, f_t>>(
+      *problem.original_problem_ptr, settings.get_tolerances(), incumbent_callback);
+    early_cpufj->set_best_objective(initial_cutoff);
+    early_cpufj->start();
+    solver.context.early_cpufj_ptr = early_cpufj.get();
+    CUOPT_LOG_DEBUG("Started early CPUFJ on papilo-presolved problem during cuOpt presolve");
+  }
```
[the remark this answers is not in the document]
aliceb-nv: what? no, objective scaling factor pretty clearly dictates the direction of the user problem since the solver space problem is always minimization

cpp/src/mip_heuristics/solve.cu
```diff
@@ -288,12 +375,32 @@ mip_solution_t<i_t, f_t> solve_mip(optimization_problem_t<i_t, f_t>& op_problem,
...
+      early_cpufj.reset();
+    }
+
     if (settings.user_problem_file != "") {
       CUOPT_LOG_INFO("Writing user problem to file: %s", settings.user_problem_file.c_str());
       op_problem.write_to_mps(settings.user_problem_file);
     }
 
-    auto sol = run_mip(problem, settings, timer);
+    auto sol = run_mip(problem, settings, timer, early_best_objective.load());
```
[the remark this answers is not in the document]
aliceb-nv: check if `mip_scaling` actually affects the objective function coefficients

cpp/src/branch_and_bound/branch_and_bound.hpp
```diff
@@ -108,6 +108,20 @@ class branch_and_bound_t {
...
+  //
+  // IMPORTANT: `bound` must be in B&B's internal objective space, i.e. the space of
+  // original_lp_ where:  user_obj = obj_scale * (internal_obj + obj_constant).
+  // The caller (solver.cu) converts from user-space via
+  //   problem_ptr->get_solver_obj_from_user_obj(user_cutoff)
+  // which accounts for both the presolve objective offset and maximization.
+  void set_initial_cutoff(f_t bound) { initial_cutoff_ = bound; }
+
+  // Effective cutoff for node pruning: min of verified incumbent and external cutoff.
+  f_t get_cutoff() const { return std::min(upper_bound_.load(), initial_cutoff_); }
```
rg20, line 130:
```suggestion
  f_t get_current_best() const { return std::min(upper_bound_.load(), initial_cutoff_); }
```

cutoff is a bit generic/vague. Wdyt?

cpp/src/mip_heuristics/diversity/population.cu
```diff
@@ -267,9 +267,9 @@ void population_t<i_t, f_t>::invoke_get_solution_callback(
   f_t user_bound     = context.stats.get_solution_bound();
   solution_t<i_t, f_t> temp_sol(sol);
   problem_ptr->post_process_assignment(temp_sol.assignment);
-  if (context.settings.mip_scaling) {
+  if (context.settings.mip_scaling && context.scaling != nullptr) {
```
rg20: will there be cases where `settings.mip_scaling` is `false` and `context.scaling` is not `nullptr`?

cpp/src/branch_and_bound/branch_and_bound.hpp
aliceb-nv: I chose "cutoff" because the early heuristics only provide a bound, not an actualy incumbent (because we lack crushing), and "cutoff" was already the terminology employed in the dual simplex code for such a bound 🙂 I wanted to make it clear that this is not necessarily the objective of the current best incumbent, only a proof that a solution with an objective <= to this one exists.

cpp/src/mip_heuristics/diversity/population.cu
aliceb-nv: Well in this part of the code, never 🙂 This is just because I had to mark 'scaling' as an optional pointer in order to integrate the early CPU/GPUFJ heuristics with the code. Let me turn these into assertions instead.

nguidotti (approved):
Looks good to me. Just a note, Papilo uses all cpu cores via TBB. So, if you are creating new threads for early FJ, you are oversubscribing the node. If possible, I think you should reduce the number of TBB threads in Papilo accordingly.

PR #906, opened 2026-02-25
MIP Row and Objective Scaling

akifcorduk:
mip_scaling_strategy.cu implements a GPU-accelerated row-scaling and objective-scaling strategy for MIP problems that improves numerical conditioning before solving.

Objective scaling (scale_objective): scales small objective coefficients up so the smallest nonzero magnitude reaches a minimum threshold (~1e-2), guarded by a max-coefficient cap to avoid blowing up large values.

Iterative row scaling (scale_problem): runs up to 8 iterations of per-row scaling to reduce the log2-spread of row infinity norms across the constraint matrix. Each iteration:

- Computes a median-based target norm (excluding big-M and zero-RHS rows for outlier robustness)
- Picks a per-row scaling factor clamped to [1/32, 32], with tighter [1/16, 1] limits for big-M rows and rows that are already numerically large
- Preserves integer coefficient integrality for integer variables by preferring power-of-two scaling factors (exact in IEEE 754) or integer-GCD-preserving multipliers
- Tracks cumulative scaling per row within a [1/256, 256] band
- Stops early when the spread improvement stalls below a relative tolerance

The current optimal setup is with only row scaling, that's why I have left row scaling only as default.

This also fixes a bug of setting lower bound of an approximate PDLP and Barrier solution. Now we set the bound only if the relaxation solution is fully feasible on absolute per row tolerances. With main+bug fix, number of infinite mip_gaps are 5-6 instances, with scaling it is only 1. So, scaling is able find 4-5 instances in with the absolute feasible root relaxation (does not necessarily mean basic solution in dual simplex).

Overall we are able to find 2 more optimal solutions compared to main + lower bound bug fix. And ~1% mip gap improvement.

## Summary by CodeRabbit

* **New Features**
  * MIP row-scaling enabled by default for improved numeric robustness.
  * Solver returns partial results early when time limits are reached; time-limit enforcement is more proactive.
  * New selectable MIP scaling configurations with logging of the selected scaling setup.

* **Bug Fixes**
  * Fixed inconsistent unscaling and feasibility handling to prevent mismatched solution states.

* **Chores**
  * Removed a deprecated MIP configuration option.
  * Consolidated and replaced the scaling strategy; tests and build updated accordingly.

### DIFF
--- a/cpp/src/dual_simplex/phase2.cpp
+++ b/cpp/src/dual_simplex/phase2.cpp
@@ -3571,7 +3571,7 @@ dual::status_t dual_phase2_with_advanced_basis(i_t phase,
                           sum_perturb,
                           now);
       if (phase == 2 && settings.inside_mip == 1 && settings.dual_simplex_objective_callback) {
-        settings.dual_simplex_objective_callback(user_obj);
+        settings.dual_simplex_objective_callback(obj);
       }
     }
 


cpp/src/branch_and_bound/branch_and_bound.cpp
```diff
@@ -710,9 +695,9 @@ void branch_and_bound_t<i_t, f_t>::set_final_solution(mip_solution_t<i_t, f_t>&
     settings_.heuristic_preemption_callback();
   }
 
-  f_t gap              = upper_bound_ - lower_bound;
   f_t obj              = compute_user_objective(original_lp_, upper_bound_.load());
   f_t user_bound       = compute_user_objective(original_lp_, lower_bound);
+  f_t gap              = std::abs(obj - user_bound);
```
chris-maes, line 722: I don't think we should use the user objective and the user bound when computing the gap. We should also not use abs.

Calling compute_user_objective simply adds a constant to each term. In exact arithmetic the constant is removed when subtracting. But if the constant is large, as it often is, in MIP, in finite precision we can get  disastrous cancellation. 

 I also think the abs just hides errors. Suppose that we have a numerical issue and the lower bound is slightly larger than the upper bound. I think we should correctly exit because the gap is negative.

cpp/src/branch_and_bound/branch_and_bound.cpp
```diff
@@ -724,12 +709,13 @@ void branch_and_bound_t<i_t, f_t>::set_final_solution(mip_solution_t<i_t, f_t>&
                        obj,
                        is_maximization ? "Upper" : "Lower",
                        user_bound);
-  {
-    const f_t root_lp_obj = root_lp_current_lower_bound_.load();
-    if (std::isfinite(root_lp_obj)) {
-      settings_.log.printf("Root LP dual objective (last): %.16e\n", root_lp_obj);
-    }
-  }
+  settings_.log.printf(
```
chris-maes: Remove debug code?

cpp/src/dual_simplex/user_problem.hpp
```diff
@@ -46,7 +46,7 @@ struct user_problem_t {
   std::vector<std::string> row_names;
   std::vector<std::string> col_names;
   f_t obj_constant;
-  f_t obj_scale;  // 1.0 for min, -1.0 for max
+  f_t obj_scale;  // positive for min, netagive for max
```
chris-maes, line 49: Can you explain why these would ever not be 1.0 or -1.0?

chris-maes (changes requested):
Thanks for working on scaling Akif. 

I took a brief look at the changes to branch and bound. I don't think we should change how we are computing gaps in branch and bound.  Could we discuss before you merge?

cpp/src/branch_and_bound/branch_and_bound.cpp
chris-maes: And if the gap is large and negative we will print that in the log and can catch and fix it.

cpp/include/cuopt/linear_programming/mip/solver_settings.hpp
```diff
@@ -108,7 +107,7 @@ class mip_solver_settings_t {
 
   /** Initial primal solutions */
   std::vector<std::shared_ptr<rmm::device_uvector<f_t>>> initial_solutions;
-  bool mip_scaling = false;
+  bool mip_scaling = true;
```
chris-maes: Should this become an int with multiple options that the user can control?

cpp/src/branch_and_bound/branch_and_bound.cpp
akifcorduk: The reason for the abs was because of the maximization problems. When we are in the user space lower bound user_bound can be larger than the primal objective.

akifcorduk: It is fixed automatically via the scaling factor which determines maximization or minimization.

cpp/src/dual_simplex/user_problem.hpp
akifcorduk: This is because we have objective scaling factor now.

python/cuopt_server/cuopt_server/utils/linear_programming/data_definition.py
```diff
@@ -441,20 +441,21 @@ class SolverConfig(BaseModel):
...
+        "- 1: Full scaling (objective + row), default"
+        "<br>"
+        "- 2: Row scaling only (no objective scaling)",
     )
     mip_heuristics_only: Optional[bool] = Field(
         default=False,
         description="Set True to run heuristics only, False to run "
         "heuristics and branch and bound for MILP",
     )
-    mip_batch_pdlp_strong_branching: Optional[int] = Field(
```
Iroy30, line 453: Is this intentional ? I see we have this in constants.h

akifcorduk: This is not intentional. Claude removed it, it seems. Thanks for noticing it!

PR #908, opened 2026-02-26
Type and execution path cleanup for memory model changes

tmckayus:
This pull request eliminates ~1100 lines of code from the memory model changes by:

* merging the gpu_optimization_problem_t class and the original optimization_problem_t struct into a single optimization_problem_t class derived from the optimization_problem_interface_t base class. Also removes an extra copy.

* removing the GPU remote execution path. This converted from GPU to CPU for remote execution and then converted the solution back to CPU. There is no current use case for this, it was just future-proofing and can be removed.

It also splits the interface implementations out into separate files for cpu and gpu, and uses a standard naming convention for files and classes where "cpu_" prefix is used for cpu files and objects and no prefix is used for gpu files and objects.

## Summary by CodeRabbit

* **New Features**
  * Full CPU-backed optimization problem with host accessors, MPS I/O, and direct CPU solve paths for LP/MIP.

* **Improvements**
  * Unified problem interface and streamlined dispatch (CPU→GPU solve path) with move-based/zero-copy data transfer APIs.
  * Remote execution now consistently uses CPU memory by default.

* **Removals**
  * Legacy GPU-only classes, GPU-to-GPU conversion hooks, and GPU-specific remote solve overloads removed; GPU conversion entry points simplified.

### DIFF
--- a/cpp/include/cuopt/linear_programming/optimization_problem.hpp
+++ b/cpp/include/cuopt/linear_programming/optimization_problem.hpp
@@ -122,17 +122,72 @@ class optimization_problem_t : public optimization_problem_interface_t<i_t, f_t>
   // Setters
   // ============================================================================
 
+  /**
+   * @brief Set the sense of optimization to maximize.
+   * @note Setting before calling the solver is optional, default is false (minimize).
+   * @param[in] maximize true means to maximize the objective function, else minimize.
+   */
   void set_maximize(bool maximize) override;
+
+  /**
+   * @brief Set the constraint matrix (A) in CSR format.
+   * @note Setting before calling the solver is mandatory.
+   * Data is copied to GPU memory on the stream of the RAFT handle passed to the problem.
+   * @param[in] A_values Values of the CSR representation (device or host pointer)
+   * @param size_values Size of the A_values array
+   * @param[in] A_indices Indices of the CSR representation (device or host pointer)
+   * @param size_indices Size of the A_indices array
+   * @param[in] A_offsets Offsets of the CSR representation (device or host pointer)
+   * @param size_offsets Size of the A_offsets array
+   */
   void set_csr_constraint_matrix(const f_t* A_values,
                                  i_t size_values,
                                  const i_t* A_indices,
                                  i_t size_indices,
                                  const i_t* A_offsets,
                                  i_t size_offsets) override;
+
+  /**
+   * @brief Set the constraint bounds (b / right-hand side) array.
+   * @note Setting before calling the solver is mandatory.
+   * @param[in] b Device or host memory pointer to a floating point array of size size.
+   * @param size Size of the b array.
+   */
   void set_constraint_bounds(const f_t* b, i_t size) override;
+
+  /**
+   * @brief Set the objective coefficients (c) array.
+   * @note Setting before calling the solver is mandatory.
+   * @param[in] c Device or host memory pointer to a floating point array of size size.
+   * @param size Size of the c array.
+   */
   void set_objective_coefficients(const f_t* c, i_t size) override;
+
+  /**
+   * @brief Set the scaling factor of the objective function (scaling_factor * objective_value).
+   * @note Setting before calling the solver is optional, default value is 1.
+   * @param objective_scaling_factor Objective scaling factor value.
+   */
   void set_objective_scaling_factor(f_t objective_scaling_factor) override;
+
+  /**
+   * @brief Set the offset of the objective function (objective_offset + objective_value).
+   * @note Setting before calling the solver is optional, default value is 0.
+   * @param objective_offset Objective offset value.
+   */
   void set_objective_offset(f_t objective_offset) override;
+
+  /**
+   * @brief Set the quadratic objective matrix (Q) in CSR format.
+   * @note Used for quadratic programming: objective is x^T * Q * x + c^T * x
+   * @param[in] Q_values Values of the CSR representation
+   * @param size_values Size of the Q_values array
+   * @param[in] Q_indices Indices of the CSR representation
+   * @param size_indices Size of the Q_indices array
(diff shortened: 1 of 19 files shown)


cpp/include/cuopt/linear_programming/solve.hpp
```diff
@@ -148,6 +148,23 @@ optimization_problem_t<i_t, f_t> mps_data_model_to_optimization_problem(
   raft::handle_t const* handle_ptr,
   const cuopt::mps_parser::mps_data_model_t<i_t, f_t>& data_model);
 
+// ============================================================================
+// CPU problem overloads (convert to GPU, solve, convert solution back)
```
tmckayus, line 152: note this is for local testing, to be able to test the cpu execution path components without actually having to do a remote solve.  It won't be used in production and is only available through an env var (which will be undocumented)

cpp/src/pdlp/solve.cu
```diff
@@ -1201,21 +1203,25 @@ optimization_problem_solution_t<i_t, f_t> solve_lp(
       CUOPT_LOG_INFO("Third-party presolve is disabled, skipping");
     }
 
+    // Declare result in the same scope as problem so that result->reduced_problem
```
tmckayus: This particular change can be backed out, it was submitted as a separate PR

cpp/src/mip_heuristics/solve.cu
```diff
@@ -377,64 +379,86 @@ mip_solution_t<i_t, f_t> solve_mip(
...
 
-        // Temporarily set the handle on the CPU problem so it can create GPU resources
-        cpu_prob->set_handle(&handle);
+  // Convert CPU problem to GPU problem
+  auto gpu_problem = cpu_problem.to_optimization_problem();
 
-        // Convert CPU problem to GPU problem
-        auto op_problem = cpu_prob->to_optimization_problem();
+  // Synchronize before solving to ensure conversion is complete
+  stream.synchronize();
```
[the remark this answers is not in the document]
tmckayus: Local agent analysis: Invalid / disagree. CodeRabbit claims the conversion and solve share the same stream via the handle, but that's wrong. The handle is created locally in this function (raft::handle_t handle(stream)), and to_optimization_problem() runs on the CPU problem's temporarily-set handle. The resulting gpu_problem object then gets passed to solve_mip which uses gpu_problem's own handle/stream. These are not guaranteed to be the same stream. The synchronize ensures the H2D copies from conversion are visible before the solve starts. Keep it.

cpp/include/cuopt/linear_programming/optimization_problem.hpp
```diff
@@ -102,302 +111,229 @@ class optimization_problem_t {
...
-   passed to the problem.
-   * @param size_indices Size of the A_indices array.
-   * @param[in] A_offsets Offsets of the CSR representation of the constraint
-   matrix as a device or host memory pointer to a integer array of size
-   size_offsets.
-   * cuOpt copies this data. Copy happens on the stream of the raft:handler
-   passed to the problem.
-   * @param size_offsets Size of the A_offsets array.
-   */
+  void set_maximize(bool maximize) override;
```
rg20, line 130: Can you add the docs back in?

cpp/include/cuopt/linear_programming/optimization_problem.hpp
```diff
@@ -102,302 +111,229 @@ class optimization_problem_t {
...
-   * @brief Get the quadratic objective matrix offsets
-   * @return const reference to the Q_offsets vector
+   * @brief Move constraint matrix data without copying (transfers ownership).
+   * @note This is a zero-copy operation that just moves device pointers.
+   * @param[in] A_values rvalue reference to constraint matrix values
+   * @param[in] A_indices rvalue reference to constraint matrix column indices
+   * @param[in] A_offsets rvalue reference to constraint matrix row offsets
    */
-  const std::vector<i_t>& get_quadratic_objective_offsets() const;
+  void set_csr_constraint_matrix_move(rmm::device_uvector<f_t>&& A_values,
```
rg20: are these being used?  if they are not used, lets remove

if they are used, I think you can drop _move suffix

cpp/src/mip_heuristics/solve.cu
```diff
@@ -377,64 +379,86 @@ mip_solution_t<i_t, f_t> solve_mip(
...
-        CUOPT_LOG_INFO("Test mode: Converting CPU problem to GPU for local MIP solve");
+  // Create CUDA resources for the conversion
+  rmm::cuda_stream stream;
+  raft::handle_t handle(stream);
 
-        // Create CUDA resources for the conversion
-        rmm::cuda_stream stream;
-        raft::handle_t handle(stream);
+  // RAII guard: set handle for GPU resource creation, restore original on scope exit
+  auto* original_handle = cpu_problem.get_handle_ptr();
```
rg20: why would cpu_problem even hold a handle/stream? 

I think it makes sense to pass in a stream for cpu to gpu problem conversion function. This avoids any implicit assumptions on what stream being used.

cpp/include/cuopt/linear_programming/optimization_problem.hpp
tmckayus: done

cpp/include/cuopt/linear_programming/optimization_problem.hpp
tmckayus: removed

cpp/src/mip_heuristics/solve.cu
tmckayus: Removed from the constructor and passing as an argument.

rg20 (approved):
Thanks for the fixes!  

Getting better with each iteration!

PR #909, opened 2026-02-27
Fix race condition in add_external_solutions_to_population

akifcorduk:
B&B preemption will be set right after the solution is set but if we don't acquire the mutex before the early return check in population, the writes to solutions_in_external_queue_ is not visible to heuristic thread. Early return then finishes the heuristics and returns no solution. 

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed a race condition so external solutions are always processed safely, improving thread safety and stability when handling concurrent incoming solutions.

### DIFF
--- a/cpp/src/mip_heuristics/diversity/population.cu
+++ b/cpp/src/mip_heuristics/diversity/population.cu
@@ -179,17 +179,8 @@ void population_t<i_t, f_t>::add_external_solution(const std::vector<f_t>& solut
 template <typename i_t, typename f_t>
 void population_t<i_t, f_t>::add_external_solutions_to_population()
 {
-  std::vector<solution_t<i_t, f_t>> new_sol_vector;
-  {
-    // do the mutex here, otherwise it is a race condition.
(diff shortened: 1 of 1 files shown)


cpp/src/mip_heuristics/diversity/population.cu
```diff
@@ -198,7 +205,6 @@ void population_t<i_t, f_t>::preempt_heuristic_solver()
 template <typename i_t, typename f_t>
 std::vector<solution_t<i_t, f_t>> population_t<i_t, f_t>::get_external_solutions()
 {
-  std::lock_guard<std::mutex> lock(solution_mutex);
   std::vector<solution_t<i_t, f_t>> return_vector;
   i_t counter                     = 0;
   f_t new_best_feasible_objective = best_feasible_objective;
```
aliceb-nv, line 202: I think `get_external_solutions()` is called directly by diversity_manager, we may still need the lock in certain cases

akifcorduk: oh good catch. then let's keep it as it is before but skip the early return check.

PR #910, opened 2026-02-27
Add support for FP32 and mixed precision in PDLP

Kh4ster:
This PR adds support for FP32 and mixed precision in PDLP (not MIP, Dual Simplex or Barrier).

Those two new options are available through:
- FP32: creating the whole solver object as FP32 or passing "--pdlp-fp32" to the solve_LP binary
- Mixed precision: toggling the option "mixed_precision_spmv" in solver_settings or passing "--mixed-precision-spmv" to the solve_LP binary

Below, the detail of what each feature allows:

FP32 Precision Mode
-------------------

By default, PDLP operates in FP64 (double) precision. Users can switch to FP32 (float) precision for the entire solve. FP32 uses half the memory of FP64 and allows PDHG iterations to be on average twice as fast, but it may require more iterations to converge due to reduced numerical accuracy. FP32 mode is only supported with the PDLP method (not concurrent) and without crossover.

Note: The default precision is FP64 (double).

Mixed Precision SpMV
--------------------

When running PDLP in FP64 mode, users can enable mixed precision sparse matrix-vector products (SpMV) during PDHG iterations. In this mode, the constraint matrix and its transpose are stored in FP32 while vectors and the compute type remain in FP64. This allows SpMV operations to be faster thanks to reduced memory bandwidth requirements, while maintaining FP64 accuracy in the accumulation. This will make PDHG iterations faster while limiting the potential negative impact on convergence (compared to running in FP32 mode). Convergence checking and restart logic always use the full FP64 matrix, so this mode does not reduce memory usage since both the FP32 and FP64 copies of the matrix are kept in memory. Mixed precision SpMV only applies in FP64 mode and has no effect when running in FP32.

Note: The default value is false.

### DIFF
--- a/python/cuopt/cuopt/tests/linear_programming/test_lp_solver.py
+++ b/python/cuopt/cuopt/tests/linear_programming/test_lp_solver.py
@@ -745,25 +745,6 @@ def test_pdlp_precision_single():
     assert solution.get_solved_by_pdlp()
 
 
-def test_pdlp_precision_mixed():
-    file_path = (
-        RAPIDS_DATASET_ROOT_DIR + "/linear_programming/afiro_original.mps"
-    )
-    data_model_obj = cuopt_mps_parser.ParseMps(file_path)
-
-    settings = solver_settings.SolverSettings()
-    settings.set_parameter(CUOPT_METHOD, SolverMethod.PDLP)
-    settings.set_parameter(CUOPT_PDLP_PRECISION, 2)  # Mixed
-    settings.set_optimality_tolerance(1e-4)
-
-    solution = solver.Solve(data_model_obj, settings)
-
-    assert solution.get_termination_status() == LPTerminationStatus.Optimal
-    assert solution.get_primal_objective() == pytest.approx(
-        -464.7531, rel=1e-1
-    )
-
-
 def test_pdlp_precision_single_crossover():
     file_path = (
         RAPIDS_DATASET_ROOT_DIR + "/linear_programming/afiro_original.mps"


cpp/src/pdlp/solve.cu
```diff
@@ -1115,13 +1137,22 @@ optimization_problem_solution_t<i_t, f_t> solve_lp_with_method(
...
+    } else {
+      return run_pdlp(problem, settings, timer, is_batch_mode);
+    }
   } else {
+    // Float precision only supports PDLP without presolve/crossover
+    cuopt_expects(settings.method == method_t::PDLP,
+                  error_type_t::ValidationError,
+                  "Float precision only supports PDLP method. DualSimplex, Barrier, and Concurrent "
+                  "require double precision.");
     return run_pdlp(problem, settings, timer, is_batch_mode);
```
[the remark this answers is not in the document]
Kh4ster: No I don't want users thinking they are using concurrent while in FP32. They should realize that they should, and will, use PDLP only

cpp/tests/linear_programming/pdlp_test.cu
```diff
@@ -20,6 +20,7 @@
 #include <cuopt/linear_programming/pdlp/solver_settings.hpp>
 #include <cuopt/linear_programming/pdlp/solver_solution.hpp>
 #include <cuopt/linear_programming/solve.hpp>
+#include <mip_heuristics/mip_constants.hpp>
```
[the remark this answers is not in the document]
Kh4ster: In both cases the pdlp float32 tests should compile and run

cpp/src/mip_heuristics/presolve/third_party_presolve.cpp
```diff
@@ -734,26 +805,71 @@ void third_party_presolve_t<i_t, f_t>::undo_pslp(rmm::device_uvector<f_t>& prima
...
+    for (int i = 0; i < n_rows; ++i) {
+      h_dual_out[i] = static_cast<f_t>(uncrushed_sol->y[i]);
+    }
+
+    primal_solution.resize(n_cols, stream_view);
+    dual_solution.resize(n_rows, stream_view);
+    reduced_costs.resize(n_cols, stream_view);
+    raft::copy(primal_solution.data(), h_primal_out.data(), n_cols, stream_view);
+    raft::copy(dual_solution.data(), h_dual_out.data(), n_rows, stream_view);
+    raft::copy(reduced_costs.data(), h_reduced_costs_out.data(), n_cols, stream_view);
```
aliceb-nv: nit: it might be possible to unify that part of the function with thrust::transform calls that would automatically perform the implicit conversion through the iterators. It would materialize into a kernel rather than a memcpy but I don't think it matters much here

aliceb-nv: Come to think of it we can probably do the same for the input also. It might be possible to get rid of the constexpr branches for the most part that way

cpp/src/pdlp/cusparse_view.hpp
```diff
@@ -194,8 +195,56 @@ class cusparse_view_t {
...
+                                     cusparseOperation_t opA,
+                                     const double* alpha,
+                                     cusparseSpMatDescr_t matA,  // FP32 matrix
+                                     cusparseDnVecDescr_t vecX,  // FP64 vector
+                                     const double* beta,
+                                     cusparseDnVecDescr_t vecY,  // FP64 vector
+                                     cusparseSpMVAlg_t alg,
+                                     void* externalBuffer,
+                                     cudaStream_t stream);
+#endif
```
aliceb-nv, line 246: Do we even support CTK < 12.4 anymore? Might be worth checking

cpp/src/pdlp/pdhg.cu
```diff
@@ -319,17 +353,33 @@ void pdhg_solver_t<i_t, f_t>::compute_A_x()
...
+                                           CUSPARSE_OPERATION_NON_TRANSPOSE,
+                                           reusable_device_scalar_value_1_.data(),
+                                           cusparse_view_.A,
+                                           cusparse_view_.reflected_primal_solution,
+                                           reusable_device_scalar_value_0_.data(),
+                                           cusparse_view_.dual_gradient,
+                                           CUSPARSE_SPMV_CSR_ALG2,
+                                           (f_t*)cusparse_view_.buffer_non_transpose.data(),
+                                           stream_view_));
+    }
```
aliceb-nv, line 382: Another nit - it might be possible to unify this into a single A_spmv function, that takes an option to specify whether to use A or its transpose. Not fully confident if its worth the code changes though

cpp/src/pdlp/cusparse_view.hpp
Kh4ster: There are CUDA 12.2 tests in the CI

cpp/src/mip_heuristics/solution/solution.cu
```diff
@@ -660,7 +660,7 @@ mip_solution_t<i_t, f_t> solution_t<i_t, f_t>::get_solution(bool output_feasible
   }
 }
 
-#if MIP_INSTANTIATE_FLOAT
+#if MIP_INSTANTIATE_FLOAT || PDLP_INSTANTIATE_FLOAT
```
rg20, line 663: If this is done in all the files,  we can just have one flag, CUOPT_INSTANTIATE_FLOAT

cpp/src/pdlp/solution_conversion.cu
```diff
@@ -142,22 +155,22 @@ cpu_lp_solution_t<i_t, f_t>::to_cpu_linear_programming_ret_t()
   cuopt::cython::linear_programming_ret_t ret;
 
   cpu_solutions_t cpu;
-  cpu.primal_solution_ = std::move(primal_solution_);
-  cpu.dual_solution_   = std::move(dual_solution_);
-  cpu.reduced_cost_    = std::move(reduced_cost_);
+  cpu.primal_solution_ = to_cpu_buffer(primal_solution_);
```
rg20: I don't think we should bring up the mixed precision logic this far.  The conversion should have happened much before.

cpp/src/pdlp/solve.cu
```diff
@@ -606,82 +625,89 @@ optimization_problem_solution_t<i_t, f_t> run_pdlp(detail::problem_t<i_t, f_t>&
...
-    rmm::device_uvector<f_t> final_dual_solution =
-      cuopt::device_copy(vertex_solution.y, problem.handle_ptr->get_stream());
-    rmm::device_uvector<f_t> final_reduced_cost =
-      cuopt::device_copy(vertex_solution.z, problem.handle_ptr->get_stream());
-    problem.handle_ptr->sync_stream();
-    // Negate dual variables and reduced costs for maximization problems
-    if (problem.maximize) {
-      adjust_dual_solution_and_reduced_cost(
-        final_dual_solution, final_reduced_cost, problem.handle_ptr->get_stream());
+  if constexpr (std::is_same_v<f_t, double>) {
```
rg20, line 754: Does this mean crossover is disabled for single precision?   I don't think we want to make that decision.

cpp/src/pdlp/solve_remote.cu
```diff
@@ -108,6 +109,14 @@ std::unique_ptr<mip_solution_interface_t<i_t, f_t>> solve_mip_remote(
 }
 
 // Explicit template instantiations for remote execution stubs
+#if MIP_INSTANTIATE_FLOAT || PDLP_INSTANTIATE_FLOAT
+template std::unique_ptr<lp_solution_interface_t<int, float>> solve_lp_remote(
```
rg20: The APIs should be still in double precision and only the inner solves should be changed.

cpp/include/cuopt/linear_programming/pdlp/solver_settings.hpp
```diff
@@ -239,6 +239,15 @@ class pdlp_solver_settings_t {
...
   bool eliminate_dense_columns{true};
+  /**
+   * @brief Enable mixed precision SpMV during PDHG iterations (FP64 mode only).
+   *
+   * When true, the constraint matrix A and its transpose are stored in FP32 while
+   * vectors and compute type remain in FP64, reducing memory bandwidth during SpMV.
+   * Convergence checking and restarts always use the full FP64 matrix, so this does
+   * not reduce overall memory usage.  Has no effect in FP32 mode.
+   */
+  bool mixed_precision_spmv{false};
```
rg20: is there another flag for fp32? 

I think we should just have `--pdlp-precision`    flag with options: `default`, `single`, `double`, `mixed` as options. This allows us the change the definition of `mixed` in future.

docs/cuopt/source/lp-qp-features.rst
```diff
@@ -157,6 +157,20 @@ Batch Mode
 
 Users can submit a set of problems which will be solved in a batch. Problems will be solved at the same time in parallel to fully utilize the GPU. Checkout :ref:`self-hosted client <generic-example-with-normal-and-batch-mode>` example in thin client.
 
+FP32 Precision Mode
```
rg20: Lets just have one parameter for precision

cpp/src/mip_heuristics/solution/solution.cu
Kh4ster: This is not done in all files. I just tag the exact file that are needed to allow PDLP float compilation

cpp/src/pdlp/solution_conversion.cu
Kh4ster: Agree, changed

cpp/src/pdlp/solve.cu
Kh4ster: I agree, changed

cpp/src/pdlp/solve_remote.cu
Kh4ster: Agree, changed

cpp/include/cuopt/linear_programming/pdlp/solver_settings.hpp
Kh4ster: I agree, changed

docs/cuopt/source/lp-qp-features.rst
Kh4ster: Also updated in the doc

cpp/include/cuopt/linear_programming/constants.h
```diff
@@ -125,6 +126,12 @@
 #define CUOPT_METHOD_DUAL_SIMPLEX 2
 #define CUOPT_METHOD_BARRIER      3
 
+/* @brief PDLP precision mode constants */
+#define CUOPT_PDLP_DEFAULT_PRECISION 0
```
rg20: Let's use -1 for default.

cpp/src/pdlp/solve.cu
```diff
@@ -560,6 +560,188 @@ optimization_problem_solution_t<i_t, f_t> run_dual_simplex(
...
+      out_zip,
+      n_vars,
+      [] __device__(double2 b) {
+        return thrust::make_tuple(static_cast<float>(b.x), static_cast<float>(b.y));
+      },
+      stream_val));
+  }
+
+  // Build float optimization_problem_t from device pointers
+  optimization_problem_t<i_t, float> float_op(problem.handle_ptr);
```
rg20: I think its cleaner to move this to optimization_problem_t class itself. 

```
template<typename otherT> 
void convert_to_other_prec(optimization_problem_t<i_t, otherT>& other) const;
```

cpp/tests/linear_programming/c_api_tests/c_api_test.c
```diff
@@ -2122,3 +2122,69 @@ cuopt_int_t test_cpu_only_mip_execution(const char* filename)
   cuOptDestroySolution(&solution);
   return status;
 }
+
+cuopt_int_t test_pdlp_precision_single(const char* filename,
```
rg20, line 2192: how about a test for mixed precision?

cpp/src/mip_heuristics/presolve/third_party_presolve.cpp
```diff
@@ -30,6 +30,21 @@
 
 namespace cuopt::linear_programming::detail {
 
+// Helper to convert vector from one type to another (only when types differ)
+template <typename To, typename From>
```
rg20: do we need this now?  we are working with double precision at presolve level always right?

cpp/src/mip_heuristics/presolve/third_party_presolve.cpp
```diff
@@ -734,26 +805,71 @@ void third_party_presolve_t<i_t, f_t>::undo_pslp(rmm::device_uvector<f_t>& prima
...
+
+    primal_solution.resize(n_cols, stream_view);
+    dual_solution.resize(n_rows, stream_view);
+    reduced_costs.resize(n_cols, stream_view);
+    raft::copy(primal_solution.data(), uncrushed_sol->x, n_cols, stream_view);
+    raft::copy(dual_solution.data(), uncrushed_sol->y, n_rows, stream_view);
+    raft::copy(reduced_costs.data(), uncrushed_sol->z, n_cols, stream_view);
+  } else {
+    // Convert f_t to double for PSLP postsolve API
+    std::vector<f_t> h_primal_solution_ft(primal_solution.size());
```
rg20: you can remove this branch now

cpp/include/cuopt/linear_programming/constants.h
Kh4ster: You are right, changed

cpp/src/pdlp/solve.cu
Kh4ster: Agreed, moved

cpp/tests/linear_programming/c_api_tests/c_api_test.c
Kh4ster: Nice catch

cpp/src/mip_heuristics/presolve/third_party_presolve.cpp
Kh4ster: It has been moved but because we can presolve undo on the problem which is can be compiled as float we still need to compile presolve as float, just not handle if it's called in float as it's considered unsupported

cpp/src/mip_heuristics/presolve/third_party_presolve.cpp
Kh4ster: done

rg20 (approved):

PR #915, opened 2026-02-27
Deprecate LP batch solve across Python, server, and docs

anandhkb:
- Add DeprecationWarning to LP BatchSolve in solver.py
- Fix test_lp_solver: individual_solutions = [] (was [] * nb_solves)


tmckayus (dismissed):

python/cuopt_server/cuopt_server/utils/linear_programming/solver.py
```diff
@@ -341,20 +341,29 @@ def create_solution(sol):
...
             is_batch = True
-            data_model_list = []
-            warnings = []
+            warnings = [
+                "LP batch mode is deprecated. Multiple problems are now solved "
+                "sequentially. Implement your own parallelism if needed."
+            ]
+            sol = []
+            total_solve_time = 0.0
             for i_data in LP_data:
```
Iroy30, line 350: We are calling batch LP but have changed it to sequential solves under the hood which is misleading

tmckayus (commented):
we should discuss some more on just deprecation warning, or behavior change.

python/cuopt_server/cuopt_server/utils/linear_programming/solver.py
anandhkb: Understood. So what is your recommendation?

anandhkb:
Tagging @Iroy30 for advise

CC @rgsl888prabhu @tmckayus for viz

python/cuopt/cuopt/tests/linear_programming/test_lp_solver.py
```diff
@@ -492,7 +492,7 @@ def test_parser_and_batch_solver():
     batch_solution, solve_time = solver.BatchSolve(data_model_list, settings)
 
     # Call Solve on each individual data model object
-    individual_solutions = [] * nb_solves
```
ramakrishnap-nv, line 495: Why we updated this ?

ramakrishnap-nv (approved):
Rest looks good, just have a question.

python/cuopt/cuopt/tests/linear_programming/test_lp_solver.py
Iroy30: It was just a small bug in test. Unrelated to the deprecation. It doesn't change the test itself.

PR #916, opened 2026-02-28
Fix bugs causing primal simplex to cycle. Enable primal simplex clean up

chris-maes:
Fixed the following bugs that were causing primal simplex to cycle: 1) Swapped input/output arguments in b_solve()
2) Incorrectly setting variable status of leaving variable 3) Primal step length was not limited by bounds of entering variable.

Also fixed a bug/typo where the basis was reorderd twice after factorization.

Added code to switch to phase I if we loose primal feasibility, and switch back to phase II once feasibility is regained.

Tested on NETLIB LPs. Only 2 LPs pilot87 and pilot_ja need primal simplex to remove perturbations at the end of the dual simplex solve.

Tested on the 14 MIPLIB root relaxations that need primal simplex to remove perturbations at the end of the dual simplex solve.

### DIFF
--- a/cpp/src/dual_simplex/primal.cpp
+++ b/cpp/src/dual_simplex/primal.cpp
@@ -21,7 +21,6 @@ namespace {
 template <typename i_t, typename f_t>
 void set_primal_variables_on_bounds(const lp_problem_t<i_t, f_t>& lp,
                                     const simplex_solver_settings_t<i_t, f_t>& settings,
-                                    const std::vector<f_t>& z,
                                     std::vector<variable_status_t>& vstatus,
                                     std::vector<f_t>& x)
 {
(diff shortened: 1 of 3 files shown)


cpp/src/dual_simplex/primal.cpp
```diff
@@ -473,70 +554,104 @@ primal::status_t primal_phase2(i_t phase,
...
+    const bool basis_updated = (leaving_index != -1);
     for (i_t j = 0; j < n; ++j) {
       x[j] += step_length * delta_x[j];
     }
+    if (basis_updated) {
+      assert(step_length >= 0.0);
+      basic_list[basic_leaving]        = entering_index;
+      nonbasic_list[nonbasic_entering] = leaving_index;
+      vstatus[entering_index]          = variable_status_t::BASIC;
+      if (std::abs(lp.upper[leaving_index] - lp.lower[leaving_index]) < 1e-12) {
```
nguidotti, line 580: Nitpick: use `fixed_tol` instead of a hard coded value.

nguidotti (approved):
Thanks for the fix! Looks good to me.

Issue #861, opened 2026-02-13
[FEA] Routing Test Execution Optimization in CI/CD and Nightly runs


anandhkb:
Optimize routing CI/CD test execution in PRs.
Use selective triggers for PRs; run full tests nightly.

anandhkb:
> [@anandhkb](https://github.com/anandhkb) can i work on this ??

Hello @bala-bhargav 
We welcome your interest and contributions! Please do!

rgsl888prabhu:
@bala-bhargav Are you working on this problem ? Please create a draft PR so we can provide early feedback.

bala-bhargav:
Hi @rgsl888prabhu 
I've opened a PR to address this: #931
Happy to discuss the approach and iterate on feedback!

anandhkb:
Pending PR review of #931 from @mlubin - thanks @rgsl888prabhu and @mlubin and @bala-bhargav

mlubin:
We're stepping back from the plan to do selective triggers given flakiness in the tests has been fixed.

PR #931, opened 2026-03-04
[FEA] Add selective test execution for PR builds

bala-bhargav:
Implements **component-level selective test execution** for pull request CI builds. When a PR only touches routing code, only routing tests and datasets are downloaded/run. Same for LP and MIP. Nightly builds continue running everything unchanged.
#861


jameslamb (commented):
Why is this being done in shell scripts?

We already have a GitHub Action that runs on PR CI to do exactly this, and would prefer to not have multiple competing implementations of the same behavior.

https://github.com/NVIDIA/cuopt/blob/d61b19655f7b8535a5cafb960258934ff75f06da/.github/workflows/pr.yaml#L112-L114

ci/detect_test_components.sh
```diff
@@ -0,0 +1,154 @@
...
+
+    # In RAPIDS CI pull-request builds, the branch is a merge commit so
+    # HEAD~1..HEAD gives the full PR diff.  If the parent isn't reachable
+    # (e.g. shallow clone or non-merge workflow), fall back to running all.
+    local changed_files
+    if ! git rev-parse --verify HEAD~1 &>/dev/null; then
+        export CUOPT_TEST_COMPONENTS="all"
+        return
+    fi
+    if ! changed_files=$(git diff --name-only HEAD~1..HEAD 2>/dev/null); then
```
jameslamb: This is not true. `copy-pr-bot` copies commits from the PR branch to a branch on the repo called `pull-request/{pr-number}` and there HEAD is exactly HEAD of the PR branch.

It isn't like in other open-source projects on GitHub where HEAD is a merge commit created by GitHub that merges the PR branch into the target branch.

That is why in the RAPIDS `changed-files` action, we do some work to figure out what the target branch of the PR was:

https://github.com/rapidsai/shared-actions/blob/f722a09c99e883ccca580bbb7135945d8efb90e0/changed-files/action.yml#L40-L45

I think this check `HEAD~1..HEAD` is only going to compare between the difference between the latest and previous-to-that commit **on the PR branch**, which is not what you want.

You may want to look into something like `step-security/changed-files`, which the existing RAPIDS `changed-files` action uses:

* https://github.com/rapidsai/shared-actions/blob/f722a09c99e883ccca580bbb7135945d8efb90e0/changed-files/action.yml#L48
* https://github.com/step-security/changed-files

It produces an output listing all the files that changed: https://github.com/step-security/changed-files?tab=readme-ov-file#outputs-

jameslamb (changes requested):
> _Can you please suggest a better approach where we can use existing method to disable the tests? Should we separate these tests into different workflow or is there a way we can pass env variable to enable or disable these tests based on the files ?_

Yes, definitely!

Looking at this more closely and with the benefit of your description, I do understand how this is a bit different than the top-level "run / don't  run an entire workflow" `changed-files` gate.

I think you could re-use the output of the `changed-files` workflow for this purpose though. Each of the `changed-files` groups holds a boolean saying whether or not a group of file patterns was matched.

https://github.com/NVIDIA/cuopt/blob/d862480ff1a0f507912741aac2d094802a55635a/.github/workflows/pr.yaml#L296

I think you could re-use those to pass environment variables to the code (doing what `ci/detect_test_components.sh` is attempting to do).

Something like this:

```yaml
conda-python-tests:
  needs: [conda-python-build, changed-files, compute-matrix-filters]
  uses: rapidsai/shared-workflows/.github/workflows/conda-python-tests.yaml@main
  if: fromJSON(needs.changed-files.outputs.changed_file_groups).test_python_conda
  with:
    script: |
        CUOPT_ROUTING_FILES_CHANGED="${{ fromJSON(needs.changed-files.outputs.changed_file_groups).test_routing }}"
        export CUOPT_ROUTING_FILES_CHANGED
        ci/test_python.sh
```

And then handle those variables in test scripts however you want.

Untested, but in theory I think multi-line scripts like this should work: https://github.com/rapidsai/shared-workflows/blob/52b43724e7c9edba0acf675aaf6f526a22b3111f/.github/workflows/conda-python-tests.yaml#L304

If not, you can combine everything together with `;`.

Benefits of this approach:

* no need to re-implement the "figure out which files changed in the PR" logic (it is not trivial!)
* less shell code to maintain
* allows the groups to be re-used and combined for other jobs
* puts all of the "which files affect which workflows" logic in one place (`changed-files:` configuration in `pr.yaml`) instead 

And whether or not you do that, please do consider how these things will work when the tests scripts are run in workflow NOT triggered by a PR.

ci/run_ctests.sh
```diff
@@ -4,6 +4,44 @@
...
+    local components="${CUOPT_TEST_COMPONENTS:-all}"
+
+    # If all components should run, return true
+    if [[ "${components}" == "all" ]]; then
+        return 0
+    fi
+
+    # Map test binary names to components
+    case "${test_name}" in
+        ROUTING_*|WAYPOINT_*|VEHICLE_*|OBJECTIVE_*|RETAIL_*)
```
ramakrishnap-nv: This is just a suggestion, please explore if you can come across better approach. 

I am trying to see how to make it easier to manage, if we can have these tests installed per module, like routing, lp, milp, it would be easier to just add those directory instead of adding each new test to this list.

We may have to update cpp/test/CmakeLists.txt 
```
    install(
        TARGETS ${CMAKE_TEST_NAME}
        COMPONENT testing
        DESTINATION bin/gtests/libcuopt
        EXCLUDE_FROM_ALL
    )
```
to keep the lower folders intact so we can use them

Image

.github/workflows/pr.yaml
```diff
@@ -277,6 +277,67 @@ jobs:
...
+          - 'python/cuopt/cuopt/*.py'
+          - 'python/cuopt/cuopt/distance_engine/**'
+          - 'python/cuopt/cuopt/utils/**'
+          - 'python/libcuopt/**'
+          - 'conda/**'
+          - 'dependencies.yaml'
+        test_mip:
+          # Note: MIP has no separate Python tests (tested through LP tests).
+          # If MIP-specific Python tests are added, include their paths here.
+          - 'cpp/src/branch_and_bound/**'
```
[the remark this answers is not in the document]
ramakrishnap-nv: @bala-bhargav this is a good suggestion.

.github/workflows/pr.yaml
```diff
@@ -277,6 +277,67 @@ jobs:
...
+          - 'cpp/src/cuts/**'
+          - 'cpp/src/mip_heuristics/**'
+          - 'cpp/tests/mip/**'
+          - 'regression/mip*'
+          - 'cpp/include/**'
```
ramakrishnap-nv:
```suggestion
          - 'cpp/include/cuopt/error.hpp'
          - 'cpp/include/cuopt/linear_programming/**'
          - 'cpp/inlcude/cuopt/utilities/**'
          
          
```

.github/workflows/pr.yaml
```diff
@@ -277,6 +277,67 @@ jobs:
...
+          - 'cpp/src/barrier/**'
+          - 'cpp/src/pdlp/**'
+          - 'cpp/src/math_optimization/**'
+          - 'cpp/tests/linear_programming/**'
+          - 'cpp/tests/dual_simplex/**'
+          - 'cpp/tests/qp/**'
+          - 'python/cuopt/cuopt/tests/linear_programming/**'
+          - 'python/cuopt/cuopt/tests/quadratic_programming/**'
+          - 'regression/lp*'
+          - 'cpp/include/**'
```
ramakrishnap-nv:
```suggestion
          - 'cpp/include/cuopt/error.hpp'
          - 'cpp/include/cuopt/linear_programming/**'
          - 'cpp/inlcude/cuopt/utilities/**'
          
          
```

.github/workflows/pr.yaml
```diff
@@ -277,6 +277,67 @@ jobs:
...
+        test_routing:
+          - 'cpp/src/routing/**'
+          - 'cpp/src/distance/**'
+          - 'cpp/tests/routing/**'
+          - 'cpp/tests/distance_engine/**'
+          - 'cpp/tests/examples/routing/**'
+          - 'python/cuopt/cuopt/routing/**'
+          - 'python/cuopt/cuopt/tests/routing/**'
+          - 'regression/routing*'
+          - 'cpp/include/**'
```
ramakrishnap-nv:
```suggestion
          - 'cpp/include/cuopt/error.hpp'
          - 'cpp/include/cuopt/routing/**'
          - 'cpp/inlcude/cuopt/utilities/**'
          
          
```

rgsl888prabhu (changes requested):
@bala-bhargav Awesome work, I have few suggestions, please let me know if you have any questions

ci/run_ctests.sh
bala-bhargav: @rgsl888prabhu Great suggestion! I think we can update ConfigureTest() to install tests into per-module subdirectories (like bin/gtests/libcuopt/routing/, lp/, 

mip/
) so 

run_ctests.sh
 can just iterate the right folder instead of maintaining the case statement.

Since it touches the CMake build system and all the test CMakeLists, I'd prefer to keep it as a separate follow-up PR to avoid mixing it with the CI changes here. I'll open one once this is merged. Does that sound good?

.github/workflows/pr.yaml
```diff
@@ -277,6 +277,73 @@ jobs:
           - '!.cursor-plugin/**'
           - '!.claude-plugin/**'
           - '!gemini-extension.json'
+        test_routing:
```
jameslamb: Would you please alphabetize these lists? They're very long, I think they'll be easier to read if they're alphabetized as the file-system is.

That'll also resolve an inconsistency in grouping... e.g. not all `cpp/src/` paths are together right now.

ci/run_with_pr_changed_components.sh
```diff
@@ -0,0 +1,29 @@
...
+#     ci/test_example.sh
+#
+# Optional extra args are forwarded to the test script.
+
+set -euo pipefail
+
+export CUOPT_ROUTING_CHANGED="${1:-}"
+export CUOPT_LP_CHANGED="${2:-}"
+export CUOPT_MIP_CHANGED="${3:-}"
+shift 3
```
jameslamb:
```suggestion
export CUOPT_ROUTING_CHANGED="${1}"
export CUOPT_LP_CHANGED="${2}"
export CUOPT_MIP_CHANGED="${3}"
shift 3
```

It looks to me like this is always called with exactly 3 positional arguments... you should make it a lot stricter.

Removing the fallback to null would be a good start. That way, if anything is missing `set -u` will cause a loud error.

ci/run_ctests.sh
```diff
@@ -4,6 +4,44 @@
...
+            [[ "${components}" == *"routing"* ]] && return 0
+            ;;
+        LP_*|PDLP_*|C_API_*|DUAL_SIMPLEX_*|QP_*)
+            [[ "${components}" == *"lp"* ]] && return 0
+            ;;
+        MIP_*|PROBLEM_*|ELIM_*|STANDARDIZATION_*|MULTI_PROBE_*|INCUMBENT_*|DOC_EXAMPLE_*|CUTS_*|EMPTY_*|DETERMINISM_*)
```
jameslamb: Similar comment... I think it'd be worthwhile to go through the diff of this PR and alphabetize all hard-coded lists where ordering isn't important. That'll make it much easier for people to read, which migh help catch some omissions or duplicates.

jameslamb (approved):
I left some very minor comments, do what you want with them.

Otherwise, I'm approving the changes here. I'm very glad you were able to get it working re-using the `changed-files` mechanism.

I personally think this will be really hard to keep up to date and that the complexity of the interactions mean that some testing will accidentally be skipped... but I'll defer to @rgsl888prabhu and other cuOpt maintainers on how to weigh that risk against whatever testing time is saved by this. And I'll also defer to them on the correctness of the file mappings and testing of the implementation.

rgsl888prabhu:
@jameslamb If we want to redirect it to main (26.06), what would be the best way? Since this might have the 26.04 tag in it

ci/run_ctests.sh
mlubin: Have you considered cmake test labels? I'm pretty concerned about the maintainability of having hardcoded test names in this file

ramakrishnap-nv: Agree, harcoded test cases might create missing tests. I was thinking to tackle this in a follow-up PR, but since we have crossed 26.04 release, may be we can push updates as well in the same PR so there are no breakages in between.

ramakrishnap-nv: @mlubin Added label based gtests

ramakrishnap-nv: @mlubin May I get your review on this ?

ramakrishnap-nv:
Thank you @bala-bhargav for the contribution, cuOpt is currently planning to shift towards run all but make test set essential and also fix flaky issue rather than hiding them involuntarily. So I would be closing this PR.

PR #937, opened 2026-03-06
Add clique cuts

akifcorduk:
This PR adds clique cuts to the set of cuts we have. It uses the ideas from Preprocessing and Cutting Planes with Conflict Graphs, Brito et.al. 

In a previous PR, I have added the clique table generation as a preprocessing step (currently doesn't change the problem). This PR builds on that by using the extended cliques during cut passes. At each cut pass, a subgraph induced by the fractinoal variables on the current relaxation solution is used to generate clique cuts. We want to generate the set of violated clique inequalities that have at least min violation (1e-6). The maximal cliques in the fractional subgraph is found by Bron Kerbosh dynamic programming algorithm. The found maximal cliques are extended with the suitable integer valued variables on the original conflict graph. The extension is guided by reduced cost of the variables at the current relaxation optimal. We try to extend the cliques with variables that have lower reduced cost. The reason is that it disturbs dual simplex less (causes fewer refactors and iterations for resolve). Also the variables with lower reduced costs are likely to be active in the next resolve. This also results in more dominant cliques preventing multiple smaller sized cliques in the next iterations.

Benchmark results:
MIP GAP wins/losses defined as at least 10% difference:
main wins (6):
  Instance                                   dir1 gap   dir2 gap       diff
  app1-2                                       0.8077     1.1818    +0.3741
  bab2                                         0.4148     0.6148    +0.2001
  glass4                                       0.3333     0.4483    +0.1149
  map16715-04                                  0.9269     1.4085    +0.4816
  neos-4413714-turia                           0.7027     0.8454    +0.1427
  rail01                                       0.5476     0.7934    +0.2458

clique cuts wins (11):
  Instance                                   dir1 gap   dir2 gap       diff
  30n20b8                                      0.7688     0.6667    +0.1021
  bab6                                         0.9255     0.3611    +0.5643
  buildingenergy                               0.1706     0.0036    +0.1670
  co-100                                       0.8410     0.7089    +0.1321
  neos-3754480-nidda                          20.3512    19.6495    +0.7017
  neos-5188808-nattai                          0.2032     0.0000    +0.2032
  netdiversion                                 0.3710     0.1429    +0.2281
  physiciansched3-3                            0.2444     0.0269    +0.2175
  rocII-5-11                                   1.0893     0.7762    +0.3131
  satellites2-60-fs                           13.5000     1.4167   +12.0833
  sorrell3                                     9.5625     9.4375    +0.1250

Optimality wins/loses:
Wins

 graphdraw-domain  (443.68s)
 neos-5188808-nattai  (462.84s)
 supportcase26  (537.66s)

Losses

dano3_3  (121.30s)

Time to optimality geomean (shifted geomean by 1.0):
on 50 common optimal instances
Common optimal instances: 50
Geometric mean of ratios (main/clique): 1.1925

### DIFF
--- a/cpp/src/mip_heuristics/diversity/diversity_manager.cu
+++ b/cpp/src/mip_heuristics/diversity/diversity_manager.cu
@@ -209,32 +209,32 @@ bool diversity_manager_t<i_t, f_t>::run_presolve(f_t time_limit, timer_t global_
   const bool remap_cache_ids = true;
   if (!global_timer.check_time_limit()) { trivial_presolve(*problem_ptr, remap_cache_ids); }
   if (!problem_ptr->empty && !check_bounds_sanity(*problem_ptr)) { return false; }
-  if (!presolve_timer.check_time_limit() && !context.settings.heuristics_only &&
-      !problem_ptr->empty) {
-    f_t time_limit_for_clique_table = std::min(3., presolve_timer.remaining_time() / 5);
-    timer_t clique_timer(time_limit_for_clique_table);
-    dual_simplex::user_problem_t<i_t, f_t> host_problem(problem_ptr->handle_ptr);
-    problem_ptr->get_host_user_problem(host_problem);
-    std::shared_ptr<clique_table_t<i_t, f_t>> clique_table;
-    constexpr bool modify_problem_with_cliques = false;
-    find_initial_cliques(host_problem,
-                         context.settings.tolerances,
-                         &clique_table,
-                         clique_timer,
-                         modify_problem_with_cliques,
-                         nullptr);
-    if (modify_problem_with_cliques) {
-      problem_ptr->set_constraints_from_host_user_problem(host_problem);
-      cuopt_assert(host_problem.lower.size() == static_cast<size_t>(problem_ptr->n_variables),
-                   "host lower bound size mismatch");
-      cuopt_assert(host_problem.upper.size() == static_cast<size_t>(problem_ptr->n_variables),
-                   "host upper bound size mismatch");
-      std::vector<i_t> all_var_indices(problem_ptr->n_variables);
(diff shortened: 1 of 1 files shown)


chris-maes (approved):
The only concern is the increase in runtime. Ideally, we would see that this is performance neutral or speeds things  up. If it is a slow down, I think we should merge, but keep it disabled. 

Approving changes to branch_and_bound.{cpp,hpp} and cuts.{cpp, hpp} and simplex_solver_settings.hpp. Note I did not review the clique cut specific code, only how it interfaced with the existing code. 

One thought: cuts.cpp is getting pretty big. It might be good to put the clique cut specific functions in the files clique_cuts.{cpp, hpp} and I can separate out cuts into the additional files knapsack_cuts.{cpp, hpp}, mir_cuts.{cpp, hpp}, gomory_cuts.{cpp, hpp}, and chvatal_gomory_cuts.{cpp,hpp} in a follow up PR.  But fine to leave as is and I can do the separation later.

benchmarks/linear_programming/cuopt/run_mip.cpp
```diff
@@ -413,7 +418,16 @@ int main(int argc, char* argv[])
   int reliability_branching = program.get<int>("--reliability-branching");
   bool deterministic        = program.get<bool>("--determinism");
 
-  if (num_cpu_threads < 0) { num_cpu_threads = omp_get_max_threads() / n_gpus; }
+  if (num_cpu_threads < 0) {
+    num_cpu_threads = omp_get_max_threads() / n_gpus;
+    // std::ifstream smt_file("/sys/devices/system/cpu/smt/active");
```
chris-maes, line 423: Remove commented code?

akifcorduk: This is non-production code, I would rather merge it now as the time is a bit tight for this PR.

PR #939, opened 2026-03-06
Remote execution implementation

tmckayus:
This change replaces the solve_lp_remote and solve_mip_remote stubs with real routines that use an embedded grpc client to communicate with a remote cuopt server.

There are two documents included, GRPC_ARCHITECTURE.md and SERVER_ARCHITECTURE.md that act as developer guides.

The server is built by build.sh and is called cuopt_grpc_server.

Remote execution is still enabled via env vars CUOPT_REMOTE_HOST and CUOPT_REMOTE_PORT

To try this feature, checkout the branch/PR and do the following:

$ build.sh
$ cuopt_grpc_server // defaults are fine, run with -h for help
$ CUOPT_REMOTE_HOST=localhost CUOPT_REMOTE_PORT=8765 cuopt_cli myproblem.mps

All cuopt APIs will pick up remote execution if the env vars are set. So cuopt_cli, the C API, the Python API, all will solve problems on the server if the env vars are set and the server is running. Just use cuopt tools and APIs as you normally would for a local solve.

### DIFF
--- a/conda/recipes/libcuopt/recipe.yaml
+++ b/conda/recipes/libcuopt/recipe.yaml
@@ -29,14 +29,7 @@ cache:
         export CXXFLAGS=$(echo $CXXFLAGS | sed -E 's@\-fdebug\-prefix\-map[^ ]*@@g')
         set +x
 
-        # Workaround for abseil-cpp#1624: Mutex::Dtor() not exported from shared libabseil on Linux.
-        # Build gRPC/Protobuf/Abseil from source to ensure consistent ABI and avoid symbol issues.
-        GRPC_INSTALL_DIR="${SRC_DIR}/grpc_install"
-
-        bash ./ci/utils/install_protobuf_grpc.sh --prefix="${GRPC_INSTALL_DIR}" --build-dir="${SRC_DIR}" --skip-deps
-
-        export CMAKE_PREFIX_PATH="${GRPC_INSTALL_DIR}:${CMAKE_PREFIX_PATH:-}"
-        ./build.sh -n -v ${BUILD_EXTRA_FLAGS} libmps_parser libcuopt deb --allgpuarch --cmake-args=\"-DCMAKE_INSTALL_LIBDIR=lib -DCMAKE_PREFIX_PATH=${GRPC_INSTALL_DIR}\"
+        ./build.sh -n -v ${BUILD_EXTRA_FLAGS} libmps_parser libcuopt deb --allgpuarch --cmake-args=\"-DCMAKE_INSTALL_LIBDIR=lib\"
       secrets:
         - AWS_ACCESS_KEY_ID
         - AWS_SECRET_ACCESS_KEY
@@ -100,6 +93,10 @@ cache:
       - bzip2
       - openssl
       - c-ares
+      - libgrpc
+      - libprotobuf
+      - libabseil
+      - re2
 
 outputs:
   - package:
@@ -177,6 +174,10 @@ outputs:
         - openssl
         - c-ares
         - libuuid
+        - libgrpc
+        - libprotobuf
+        - libabseil
+        - re2
       run:
         - ${{ pin_compatible("cuda-version", upper_bound="x", lower_bound="x") }}
         - ${{ pin_subpackage("libmps-parser", exact=True) }}
@@ -187,6 +188,9 @@ outputs:
         - openssl
         - c-ares
         - libuuid
+        - libgrpc
+        - libprotobuf
+        - libabseil
       ignore_run_exports:
         by_name:
           - cuda-nvtx
@@ -231,6 +235,9 @@ outputs:
         - libcusparse-dev
         - openssl
         - c-ares
+        - libgrpc
+        - libprotobuf
+        - libabseil
       run:
         - ${{ pin_subpackage("libcuopt", exact=True) }}
         - ${{ pin_subpackage("libmps-parser", exact=True) }}
--- a/conda/environments/all_cuda-129_arch-aarch64.yaml
+++ b/conda/environments/all_cuda-129_arch-aarch64.yaml
@@ -28,11 +28,14 @@ dependencies:
 - gcc_linux-aarch64=14.*
 - ipython
 - jsonref==1.1.0
+- libabseil
 - libboost-devel
 - libcudss-dev >=0.7
 - libcurand-dev
 - libcusolver-dev
 - libcusparse-dev
+- libgrpc
+- libprotobuf
 - libraft-headers==26.4.*,>=0.0.0a0
 - librmm==26.4.*,>=0.0.0a0
 - make
@@ -60,6 +63,7 @@ dependencies:
 - pyyaml>=6.0.0
 - rapids-build-backend>=0.4.0,<0.5.0
 - rapids-logger==0.2.*,>=0.0.0a0
+- re2
 - requests
 - rmm==26.4.*,>=0.0.0a0
 - scikit-build-core>=0.11.0
--- a/conda/environments/all_cuda-129_arch-x86_64.yaml
+++ b/conda/environments/all_cuda-129_arch-x86_64.yaml
@@ -28,11 +28,14 @@ dependencies:
 - gcc_linux-64=14.*
 - ipython
 - jsonref==1.1.0
+- libabseil
 - libboost-devel
 - libcudss-dev >=0.7
 - libcurand-dev
 - libcusolver-dev
 - libcusparse-dev
+- libgrpc
+- libprotobuf
 - libraft-headers==26.4.*,>=0.0.0a0
 - librmm==26.4.*,>=0.0.0a0
 - make
@@ -60,6 +63,7 @@ dependencies:
 - pyyaml>=6.0.0
 - rapids-build-backend>=0.4.0,<0.5.0
 - rapids-logger==0.2.*,>=0.0.0a0
+- re2
 - requests
 - rmm==26.4.*,>=0.0.0a0
 - scikit-build-core>=0.11.0
--- a/conda/environments/all_cuda-131_arch-aarch64.yaml
+++ b/conda/environments/all_cuda-131_arch-aarch64.yaml
@@ -28,11 +28,14 @@ dependencies:
 - gcc_linux-aarch64=14.*
 - ipython
 - jsonref==1.1.0
+- libabseil
 - libboost-devel
 - libcudss-dev >=0.7
 - libcurand-dev
 - libcusolver-dev
 - libcusparse-dev
+- libgrpc
+- libprotobuf
 - libraft-headers==26.4.*,>=0.0.0a0
 - librmm==26.4.*,>=0.0.0a0
 - make
@@ -60,6 +63,7 @@ dependencies:
 - pyyaml>=6.0.0
 - rapids-build-backend>=0.4.0,<0.5.0
 - rapids-logger==0.2.*,>=0.0.0a0
+- re2
 - requests
 - rmm==26.4.*,>=0.0.0a0
 - scikit-build-core>=0.11.0
--- a/conda/environments/all_cuda-131_arch-x86_64.yaml
+++ b/conda/environments/all_cuda-131_arch-x86_64.yaml
@@ -28,11 +28,14 @@ dependencies:
 - gcc_linux-64=14.*
 - ipython
 - jsonref==1.1.0
+- libabseil
 - libboost-devel
 - libcudss-dev >=0.7
 - libcurand-dev
 - libcusolver-dev
 - libcusparse-dev
+- libgrpc
+- libprotobuf
 - libraft-headers==26.4.*,>=0.0.0a0
 - librmm==26.4.*,>=0.0.0a0
 - make
@@ -60,6 +63,7 @@ dependencies:
 - pyyaml>=6.0.0
 - rapids-build-backend>=0.4.0,<0.5.0
 - rapids-logger==0.2.*,>=0.0.0a0
+- re2
 - requests
 - rmm==26.4.*,>=0.0.0a0
 - scikit-build-core>=0.11.0
--- a/cpp/CMakeLists.txt
+++ b/cpp/CMakeLists.txt
@@ -111,9 +111,14 @@ endif(BUILD_MSAN)
 
 # Note: -UNDEBUG is applied via CUOPT_CXX_FLAGS / CUOPT_CUDA_FLAGS (not add_definitions)
 # to avoid leaking into dependencies that are built in-tree.
+#
+# For CXX: -UNDEBUG is applied per-source-file (not target-wide) to exclude gRPC
+# infrastructure files. Those files include abseil headers, and abseil's shared library
+# on conda-forge doesn't export Mutex::Dtor() in NDEBUG builds (abseil-cpp#1624).
+# Keeping NDEBUG defined for gRPC files makes the header inline an empty Dtor(),
+# avoiding the missing symbol at runtime.
 if(DEFINE_ASSERT)
   add_definitions(-DASSERT_MODE)
-  list(APPEND CUOPT_CXX_FLAGS -UNDEBUG)
   list(APPEND CUOPT_CUDA_FLAGS -UNDEBUG)
 endif(DEFINE_ASSERT)
 
@@ -346,6 +351,14 @@ set(CUOPT_SRC_FILES )
 add_subdirectory(src)
 if (HOST_LINEINFO)
   set_source_files_properties(${CUOPT_SRC_FILES} DIRECTORY ${CMAKE_SOURCE_DIR} PROPERTIES COMPILE_OPTIONS "-g1")
+endif()
+
+# Apply -UNDEBUG only to solver source files (not gRPC infrastructure).
+# Must happen before gRPC files are appended to CUOPT_SRC_FILES.
+# Uses APPEND to preserve any existing per-file options (e.g. -g1 from HOST_LINEINFO).
+if(DEFINE_ASSERT)
+  set_property(SOURCE ${CUOPT_SRC_FILES} DIRECTORY ${CMAKE_SOURCE_DIR}
+    APPEND PROPERTY COMPILE_OPTIONS "-UNDEBUG")
 endif()
 
 # Add gRPC mapper files and generated protobuf sources
--- a/dependencies.yaml
+++ b/dependencies.yaml
@@ -304,6 +304,10 @@ dependencies:
           - bzip2
           - openssl
           - c-ares
+          - libgrpc
+          - libprotobuf
+          - libabseil
+          - re2
   test_cpp:
     common:
       - output_types: [conda]


cpp/cuopt_grpc_server.cpp
```diff
@@ -0,0 +1,3116 @@
...
+ * - Pipe-based IPC for problem/result data
+ * - Result tracking and retrieval threads
+ * - Log streaming
+ *
+ * Only the client-facing network layer is different (gRPC vs TCP).
+ */
+
+#ifdef CUOPT_ENABLE_GRPC
+
+#include <grpcpp/grpcpp.h>
```
ramakrishnap-nv: Why do we have this file outside of src path? is there a better way to situate it ?

tmckayus: ah, I believe this should not be here. It was phase 1 before I broke up the server into multiple components.

tmckayus: dropped this file, it was the original POC monolith :)

cpp/src/grpc/client/solve_remote.cpp
```diff
@@ -0,0 +1,261 @@
...
+// ============================================================================
+// Remote execution via gRPC
+// ============================================================================
+
+template <typename i_t, typename f_t>
+std::unique_ptr<lp_solution_interface_t<i_t, f_t>> solve_lp_remote(
+  cpu_optimization_problem_t<i_t, f_t> const& cpu_problem,
+  pdlp_solver_settings_t<i_t, f_t> const& settings,
+  bool problem_checking,
+  bool use_pdlp_solver_mode)
```
rg20: looks like last two arguments are not used at all?

cpp/src/grpc/client/solve_remote.cpp
```diff
@@ -0,0 +1,261 @@
...
+
+// Apply env-var overrides for transfer, debug, and TLS configuration.
+static void apply_env_overrides(grpc_client_config_t& config)
+{
+  constexpr int64_t kMinChunkSize   = 4096;
+  constexpr int64_t kMaxChunkSize   = 2LL * 1024 * 1024 * 1024;  // 2 GiB
+  constexpr int64_t kMinMessageSize = 4096;
+  constexpr int64_t kMaxMessageSize = 2LL * 1024 * 1024 * 1024;
+
+  auto chunk = parse_env_int64("CUOPT_CHUNK_SIZE", config.chunk_size_bytes);
```
rg20, line 91: are you going to remove environment variables at some point?

cpp/src/grpc/client/solve_remote.cpp
```diff
@@ -0,0 +1,261 @@
...
+// ============================================================================
+
+template <typename i_t, typename f_t>
+std::unique_ptr<lp_solution_interface_t<i_t, f_t>> solve_lp_remote(
+  cpu_optimization_problem_t<i_t, f_t> const& cpu_problem,
+  pdlp_solver_settings_t<i_t, f_t> const& settings,
+  bool problem_checking,
+  bool use_pdlp_solver_mode)
+{
+  init_logger_t log(settings.log_file, settings.log_to_console);
```
rg20, line 129: the remote solve will have its own logger?   

also that would mean we won't be pulling in the logs from the remote.   How should we handle the logs?

cpp/src/grpc/client/solve_remote.cpp
```diff
@@ -0,0 +1,261 @@
...
+  // Build gRPC client configuration
+  grpc_client_config_t config;
+  config.server_address  = get_grpc_server_address();
+  config.timeout_seconds = solver_timeout_seconds(settings.time_limit);
+  apply_env_overrides(config);
+
+  // Configure log streaming based on settings
+  if (settings.log_to_console) {
+    config.stream_logs  = true;
+    config.log_callback = [](const std::string& line) { std::cout << line << std::endl; };
```
rg20: how does this work?

cpp/src/grpc/server/grpc_field_element_size.hpp
```diff
@@ -0,0 +1,32 @@
...
+  switch (field_id) {
+    case cuopt::remote::FIELD_A_INDICES:
+    case cuopt::remote::FIELD_A_OFFSETS:
+    case cuopt::remote::FIELD_Q_INDICES:
+    case cuopt::remote::FIELD_Q_OFFSETS: return 4;
+    case cuopt::remote::FIELD_ROW_TYPES:
+    case cuopt::remote::FIELD_VARIABLE_TYPES:
+    case cuopt::remote::FIELD_VARIABLE_NAMES:
+    case cuopt::remote::FIELD_ROW_NAMES: return 1;
+    default: return 8;
```
rg20: Usually I avoid implementing the default case.  This way, when we add a new field, we immediately get a compilation error and forces us to implement this correctly.

cpp/src/grpc/server/grpc_job_management.cpp
```diff
@@ -0,0 +1,336 @@
...
+
+// =============================================================================
+// Job management
+// =============================================================================
+
+// Reserve a shared-memory job queue slot, store the serialized request data,
+// and register the job in the tracker. Returns {true, job_id} on success.
+// Uses CAS on `claimed` for lock-free slot reservation and release semantics
+// on `ready` to publish all writes to the dispatch thread.
+std::pair<bool, std::string> submit_job_async(std::vector<uint8_t>&& request_data, bool is_mip)
```
rg20:
```suggestion
std::pair<bool, std::string> submit_job_async(std::vector<uint8_t>&& request_data)
```

is_mip is too specific to the problem structure, I think this function should be generic enough and easily extendable for many problem types

cpp/src/grpc/server/grpc_job_management.cpp
```diff
@@ -0,0 +1,336 @@
...
+      }
+      waiter->cv.notify_all();
+      waiting_threads.erase(wit);
+    }
+  }
+
+  return 0;
+}
+
+std::string generate_job_id()
```
rg20, line 332: are the job ids stored somewhere?  for tracking

cpp/src/grpc/server/grpc_server_main.cpp
```diff
@@ -0,0 +1,347 @@
...
+      std::cerr << "Error: " << flag << " requires a value\n";
+      print_usage(argv[0]);
+      return false;
+    }
+    return true;
+  };
+
+  for (int i = 1; i < argc; i++) {
+    std::string arg = argv[i];
+    if (arg == "-p" || arg == "--port") {
```
rg20: maybe use argparser for consistency? we use that in cuopt_cli

cpp/src/grpc/server/grpc_server_main.cpp
```diff
@@ -0,0 +1,347 @@
...
+  if (config.chunk_timeout_seconds < 0) {
+    std::cerr << "Error: --chunk-timeout must be >= 0\n";
+    print_usage(argv[0]);
+    return 1;
+  }
+
+  config.max_message_bytes =
+    std::clamp(config.max_message_bytes, kServerMinMessageBytes, kServerMaxMessageBytes);
+
+  std::cout << "cuOpt gRPC Remote Solve Server\n"
```
rg20: We want to avoid std::cout and rely instead on CUOPT_LOG_INFO mechanism.

cpp/src/grpc/server/grpc_server_main.cpp
```diff
@@ -0,0 +1,347 @@
...
+
+  ensure_log_dir_exists();
+
+  shm_unlink(SHM_JOB_QUEUE);
+  shm_unlink(SHM_RESULT_QUEUE);
+  shm_unlink(SHM_CONTROL);
+
+  int shm_fd = shm_open(SHM_JOB_QUEUE, O_CREAT | O_RDWR, 0600);
+  if (shm_fd < 0) {
+    std::cerr << "[Server] Failed to create shared memory for job queue: " << strerror(errno)
```
rg20: maybe cuopt_expects?

cpp/src/grpc/cuopt_remote.proto
```diff
@@ -0,0 +1,348 @@
...
+
+  // Quadratic objective matrix Q in CSR format
+  repeated double Q_values = 21;
+  repeated int32 Q_indices = 22;
+  repeated int32 Q_offsets = 23;
+}
+
+// PDLP solver mode enum (matches cuOpt pdlp_solver_mode_t)
+// Matches cuOpt pdlp_solver_mode_t enum values
+enum PDLPSolverMode {
```
rg20, line 63: these are solver parameters, I am not sure this is the right abstraction. May be we can just use the existing structs/enums?

cpp/src/grpc/cuopt_remote.proto
```diff
@@ -0,0 +1,348 @@
...
+  Stable1 = 0;
+  Stable2 = 1;
+  Methodical1 = 2;
+  Fast1 = 3;
+  Stable3 = 4;
+}
+
+// Matches cuOpt method_t enum values
+enum LPMethod {
```
rg20, line 72: Same here!   

Do we need this struct?

cpp/src/grpc/cuopt_remote.proto
```diff
@@ -0,0 +1,348 @@
...
+  int32 total_pdlp_iterations = 12;
+  int32 total_pdhg_iterations = 13;
+  double last_candidate_kkt_score = 14;
+  double last_restart_kkt_score = 15;
+  double sum_solution_weight = 16;
+  int32 iterations_since_last_restart = 17;
+}
+
+// MIP solver settings (field names match cuOpt Python/C++ API)
+message MIPSolverSettings {
```
rg20, line 147: I think these are error prone.

cpp/src/grpc/server/grpc_field_element_size.hpp
tmckayus: agreed, will make them explicit

cpp/src/grpc/cuopt_remote.proto
tmckayus: The .proto files can't use C++ structs or enums directly, they have to be their own types in the proto files so there is no way around the duplication/mirroring.  This is just par for the course with protobuf (the .proto files are compiled with protoc and are written in its own C-like IDL)

However, I have a follow-on PR ready to go with a Python code generator that creates all the proto files and the conversion code from a YAML file.  We have simple definitions of the structures in YAML matching cpu_optimization_problem_t for example, and the code generator takes care of the rest.  So only two places to update, C++ and the corresponding field in the YAML (which is pretty lightweight).

tmckayus: we could drop the abstraction and just make them int32.  Less self documenting but fine to do. What do you think?

cpp/src/grpc/cuopt_remote.proto
tmckayus: same as above

cpp/src/grpc/cuopt_remote.proto
tmckayus: similar, we have to mirror the structure in .proto files if we are going to use gRPC as a protocol, there's no way around it. 

If we are using a code generator though, and don't have to update them by hand, then the maintenance burden should be light. Just add a field to the YAML when we add a field to the corresponding C++ type that is going to go across the wire. The code generator handles all the mapping and conversion.

cpp/src/grpc/client/solve_remote.cpp
tmckayus: no, these should probably be removed. This was to mirror solve_lp(), but the truth is those two parameters are hardwired to True in Cython or defaulted (True) everywhere in the code solve_lp is called. Nobody ever sets them to False, and they are not exposed through the Python or C API.

And, the grpc_server wraps the same solver code, so they are already defaulted to True on the server side. I suggest removing them here, and maybe we have a larger discussion with the team to remove them everywhere.

Alternatively, I can add them to the proto definition and just send them across the wire.

cpp/src/grpc/client/solve_remote.cpp
tmckayus: These types of env vars about chunk size and similar are useful for grpc client/server modes so we can test array chunking and message size limits, etc, with small files that solve fast.  But, they won't be documented for the users.

The network config ones we'll keep, but I'll probably add a programmatic interface to set those as well.

cpp/src/grpc/client/solve_remote.cpp
tmckayus: this logger here is just to catch messages in the grpc client. The logs on the server are written to a file and a thread in the client uses an RPC to retrieve the logs and print them (via callback) to the user's console during the solve. So the effect is, they see the solver logs locally on the console.

cpp/src/grpc/client/solve_remote.cpp
tmckayus: This callback is invoked by a thread in the grpc client during a solve. It makes RPC calls to the server to retrieve the log lines, and then invokes the callback (which prints to the console). This is actually how the Python server and client handle logs, same exact design (except that it's http instead of gRPC, and Python instead of C++).

cpp/src/grpc/server/grpc_server_main.cpp
tmckayus: Hmm, so the server's internal logs, about message sizes and worker process activity and all of that, is handled this way. If we use CUOPT_LOG_INFO those messages will be thrown in with the server logs during a solve, and the client will see them interspersed with the solve. They are only useful when reviewing server activity or debugging.

We could maybe set up a second logging mechanism, like CUOPT_LOG_INFO but separated out? What do you think?  I can take a look at that too.

cpp/src/grpc/server/grpc_server_main.cpp
tmckayus: agreed, good idea

cpp/src/grpc/server/grpc_server_main.cpp
tmckayus: I think for a server startup error it's better to just have a clear message on the console.
We could use cuopt_expects and add a try/catch at the end of main to print out the message. 

This error is unlikely to happen, but it could if the server doesn't have permision to open a file.

cpp/src/grpc/client/solve_remote.cpp
tmckayus: perhaps instead of std::cout in this call back, the entries should be routed through the local grpc client logger. But we may get a double header on the log entry -- I'll take a look at this ....

cpp/src/grpc/server/grpc_server_main.cpp
tmckayus: We can use cuopt_expects with a try/catch in main. We can also use it in the worker.  Since the payload is JSON, if we want to simplify the message we have to parse the JSON string but that's not difficult.

cpp/src/grpc/server/grpc_field_element_size.hpp
tmckayus: done

cpp/src/grpc/server/grpc_server_main.cpp
tmckayus: added a second named standard logger (SERVER_LOG_xxxx) for use by the main process and worker so that serer meta logging is not interleaved with solver logs. By default both are logged to the server console, but they can be routed to files individually and the grpc client will only retrieve the solver logs.

cpp/src/grpc/server/grpc_server_main.cpp
tmckayus: implemented cuopt_expects with a try/catch in server main and worker that pulls out the error type and the message from the JSON string and logs to console or in the case of the worker sends as the result message

cpp/src/grpc/server/grpc_server_main.cpp
tmckayus: implemented argparser

cpp/src/grpc/server/grpc_job_management.cpp
tmckayus: job ids are used analogously to the existing Python server.  They are uuids, sent back to the client after a submission and used to query the job, or cancel, retrieve logs, incumbents, etc.  They are keys to job status and results stored in the main process. 

As with the existing Python service, there is no backing store on the server, all results are in memory.

cpp/src/grpc/server/grpc_job_management.cpp
tmckayus: replaced with a uint32 problem_type so it's general. These functions are internal to the server, not part of the API, and since the problem type is known from the data in the user-facing gRPC interface routines, it's convenient to pass it here as a second parameter rather than delay inferring it later in the worker from the pipe read.

cpp/src/grpc/cuopt_remote.proto
tmckayus: Another thing to consider here -- the .proto files are language independent, and are the "contract" with any user who is writing a gRPC client against the server (which they could do). Without the enums and field structure, they have to know from some other source what the valid values are.

tmckayus: on the plus side, we don't have to write documents for the interface. The .proto files *are* the interface. We  just link to them.

cpp/src/grpc/client/solve_remote.cpp
rg20: I would say lets remove this and cleanup the APIs.  That would also make it easy to unify solve_lp and solve_mip calls

tmckayus: okay will do

tmckayus: done, will remove in general in another PR from solve_lp

cpp/src/grpc/client/solve_remote.cpp
tmckayus: looked into this, std::cout in this case is the right thing to do because the log lines have already been filtered by level and handled by CUOPT_LOG_XXXX on the server side, we are simply reading them back.

cpp/src/grpc/grpc_settings_mapper.cpp
```diff
@@ -0,0 +1,267 @@
...
+// Convert cuOpt pdlp_solver_mode_t to protobuf enum
+cuopt::remote::PDLPSolverMode to_proto_pdlp_mode(pdlp_solver_mode_t mode)
+{
+  switch (mode) {
+    case pdlp_solver_mode_t::Stable1: return cuopt::remote::Stable1;
+    case pdlp_solver_mode_t::Stable2: return cuopt::remote::Stable2;
+    case pdlp_solver_mode_t::Methodical1: return cuopt::remote::Methodical1;
+    case pdlp_solver_mode_t::Fast1: return cuopt::remote::Fast1;
+    case pdlp_solver_mode_t::Stable3: return cuopt::remote::Stable3;
+    default: return cuopt::remote::Stable3;
```
rg20: avoid implementing the defaults.  This way, if a new enum is added, compiler throws error

cpp/src/grpc/grpc_settings_mapper.cpp
```diff
@@ -0,0 +1,267 @@
...
+// Convert protobuf enum to cuOpt pdlp_solver_mode_t
+pdlp_solver_mode_t from_proto_pdlp_mode(cuopt::remote::PDLPSolverMode mode)
+{
+  switch (mode) {
+    case cuopt::remote::Stable1: return pdlp_solver_mode_t::Stable1;
+    case cuopt::remote::Stable2: return pdlp_solver_mode_t::Stable2;
+    case cuopt::remote::Methodical1: return pdlp_solver_mode_t::Methodical1;
+    case cuopt::remote::Fast1: return pdlp_solver_mode_t::Fast1;
+    case cuopt::remote::Stable3: return pdlp_solver_mode_t::Stable3;
+    default: return pdlp_solver_mode_t::Stable3;
```
rg20: same here

cpp/src/grpc/grpc_settings_mapper.cpp
```diff
@@ -0,0 +1,267 @@
...
+
+// Convert cuOpt method_t to protobuf enum
+cuopt::remote::LPMethod to_proto_method(method_t method)
+{
+  switch (method) {
+    case method_t::Concurrent: return cuopt::remote::Concurrent;
+    case method_t::PDLP: return cuopt::remote::PDLP;
+    case method_t::DualSimplex: return cuopt::remote::DualSimplex;
+    case method_t::Barrier: return cuopt::remote::Barrier;
+    default: return cuopt::remote::Concurrent;
```
rg20: same here

cpp/src/grpc/grpc_settings_mapper.cpp
```diff
@@ -0,0 +1,267 @@
...
+
+// Convert protobuf enum to cuOpt method_t
+method_t from_proto_method(cuopt::remote::LPMethod method)
+{
+  switch (method) {
+    case cuopt::remote::Concurrent: return method_t::Concurrent;
+    case cuopt::remote::PDLP: return method_t::PDLP;
+    case cuopt::remote::DualSimplex: return method_t::DualSimplex;
+    case cuopt::remote::Barrier: return method_t::Barrier;
+    default: return method_t::Concurrent;
```
rg20: same here

cpp/src/grpc/grpc_settings_mapper.cpp
```diff
@@ -0,0 +1,267 @@
...
+{
+  // Termination tolerances (all names match cuOpt API)
+  pb_settings->set_absolute_gap_tolerance(settings.tolerances.absolute_gap_tolerance);
+  pb_settings->set_relative_gap_tolerance(settings.tolerances.relative_gap_tolerance);
+  pb_settings->set_primal_infeasible_tolerance(settings.tolerances.primal_infeasible_tolerance);
+  pb_settings->set_dual_infeasible_tolerance(settings.tolerances.dual_infeasible_tolerance);
+  pb_settings->set_absolute_dual_tolerance(settings.tolerances.absolute_dual_tolerance);
+  pb_settings->set_relative_dual_tolerance(settings.tolerances.relative_dual_tolerance);
+  pb_settings->set_absolute_primal_tolerance(settings.tolerances.absolute_primal_tolerance);
+  pb_settings->set_relative_primal_tolerance(settings.tolerances.relative_primal_tolerance);
```
rg20, line 87: Not sure if there is a solution here. I feel this is error prone as well.   May be generate this from yaml as well? or is it already generated from yaml?

cpp/src/grpc/grpc_settings_mapper.cpp
```diff
@@ -0,0 +1,267 @@
...
+  pb_settings->set_augmented(settings.augmented);
+  pb_settings->set_dualize(settings.dualize);
+  pb_settings->set_ordering(settings.ordering);
+  pb_settings->set_barrier_dual_initial_point(settings.barrier_dual_initial_point);
+  pb_settings->set_eliminate_dense_columns(settings.eliminate_dense_columns);
+  pb_settings->set_pdlp_precision(static_cast<int32_t>(settings.pdlp_precision));
+  pb_settings->set_save_best_primal_so_far(settings.save_best_primal_so_far);
+  pb_settings->set_first_primal_feasible(settings.first_primal_feasible);
+
+  // TODO: Add warmstart data support
```
rg20: Can you create a GH issue?

cpp/src/grpc/grpc_solution_mapper.hpp
```diff
@@ -0,0 +1,183 @@
...
+namespace cuopt::linear_programming {
+
+/**
+ * @brief Map cpu_lp_solution_t to protobuf LPSolution message.
+ *
+ * Populates a protobuf message using the generated protobuf C++ API.
+ * Does not perform serialization — that is handled by the protobuf library.
+ */
+template <typename i_t, typename f_t>
+void map_lp_solution_to_proto(const cpu_lp_solution_t<i_t, f_t>& solution,
```
rg20, line 29: are these auto generated?

cpp/src/grpc/grpc_settings_mapper.cpp
tmckayus: yes, this should be code generated as well when we put that PR up

cpp/src/grpc/grpc_settings_mapper.cpp
tmckayus: hmm let me check that comment, not sure why that's there. It was not my intent to leave anything out

cpp/src/grpc/grpc_solution_mapper.hpp
tmckayus: the grpc_*_mapper.cpp files will be auto generated, the headers are not (but they should be super stable, just the declaration of the mapper routines)

tmckayus: But I will review, it's possible we can further reduce manual code with the generator.

cpp/src/grpc/grpc_settings_mapper.cpp
tmckayus: good catch, we did this in a different place but I missed these. Will do.

conda/recipes/libcuopt/recipe.yaml
```diff
@@ -29,7 +29,14 @@ cache:
         export CXXFLAGS=$(echo $CXXFLAGS | sed -E 's@\-fdebug\-prefix\-map[^ ]*@@g')
         set +x
 
-        ./build.sh -n -v ${BUILD_EXTRA_FLAGS} libmps_parser libcuopt deb --allgpuarch --cmake-args=\"-DCMAKE_INSTALL_LIBDIR=lib\"
+        # Workaround for abseil-cpp#1624: Mutex::Dtor() not exported from shared libabseil on Linux.
+        # Build gRPC/Protobuf/Abseil from source to ensure consistent ABI and avoid symbol issues.
```
bdice: We always want to depend conda packages in conda recipes. We must be ABI-compatible with the rest of conda-forge.

gforsyth: Looks like the linked issue was fixed in https://github.com/abseil/abseil-cpp/commit/18ac60830796f7e7d4a92eec679877de3e44f13c and released in `libabseil=20260107.*` which is available on `conda-forge` (and is the version of `libabseil` that a recent build of `protobuf` depends on)

ci/utils/install_protobuf_grpc.sh
```diff
@@ -0,0 +1,234 @@
...
+export PATH="${PREFIX}/bin:${PATH}"
+export CMAKE_PREFIX_PATH="${PREFIX}:${CMAKE_PREFIX_PATH:-}"
+
+# Ensure a consistent C++ standard across Abseil/Protobuf/gRPC.
+# Abseil's options.h defaults to "auto" selection for std::string_view
+# (ABSL_OPTION_USE_STD_STRING_VIEW=2). If one library is built in
+# C++17+ and another in C++14, they will disagree on whether
+# `absl::string_view` is a typedef to `std::string_view` or Abseil's
+# own type, leading to link-time ABI mismatches.
+CMAKE_STD_FLAGS="-DCMAKE_CXX_STANDARD=17 -DCMAKE_CXX_STANDARD_REQUIRED=ON"
```
gforsyth:
```suggestion
CMAKE_STD_FLAGS=("-DCMAKE_CXX_STANDARD=17" "-DCMAKE_CXX_STANDARD_REQUIRED=ON")
```
If you ever end up needing to add or remove values programmatically from this `_FLAGS` variable, having this as an array will make your life easier

ci/utils/install_protobuf_grpc.sh
```diff
@@ -0,0 +1,234 @@
...
+
+echo ""
+echo "Building Abseil (from gRPC submodule)..."
+cmake -S "${GRPC_SRC}/third_party/abseil-cpp" -B "${ABSL_BUILD}" -G Ninja \
+    -DCMAKE_BUILD_TYPE=Release \
+    -DCMAKE_POSITION_INDEPENDENT_CODE=ON \
+    ${CMAKE_STD_FLAGS} \
```
gforsyth:
```suggestion
    "${CMAKE_STD_FLAGS[@]}" \
```
array-expansion if you accept the above suggestion

ci/utils/install_protobuf_grpc.sh
```diff
@@ -0,0 +1,234 @@
...
+    -DCMAKE_INSTALL_PREFIX="${PREFIX}"
+cmake --build "${ABSL_BUILD}" --parallel
+cmake --install "${ABSL_BUILD}"
+
+echo ""
+echo "Building Protobuf (using installed Abseil)..."
+cmake -S "${GRPC_SRC}/third_party/protobuf" -B "${PROTOBUF_BUILD}" -G Ninja \
+    -DCMAKE_BUILD_TYPE=Release \
+    -DCMAKE_POSITION_INDEPENDENT_CODE=ON \
+    ${CMAKE_STD_FLAGS} \
```
gforsyth:
```suggestion
    "${CMAKE_STD_FLAGS[@]}" \
```
array-expansion if you accept the above suggestion

ci/utils/install_protobuf_grpc.sh
```diff
@@ -0,0 +1,234 @@
...
+    -DCMAKE_INSTALL_PREFIX="${PREFIX}"
+cmake --build "${PROTOBUF_BUILD}" --parallel
+cmake --install "${PROTOBUF_BUILD}"
+
+echo ""
+echo "Building gRPC (using installed Abseil and Protobuf)..."
+cmake -S "${GRPC_SRC}" -B "${GRPC_BUILD}" -G Ninja \
+    -DCMAKE_BUILD_TYPE=Release \
+    -DCMAKE_POSITION_INDEPENDENT_CODE=ON \
+    ${CMAKE_STD_FLAGS} \
```
gforsyth:
```suggestion
    "${CMAKE_STD_FLAGS[@]}" \
```
array-expansion if you accept the above suggestion

conda/recipes/libcuopt/recipe.yaml
```diff
@@ -29,7 +29,14 @@ cache:
...
+        GRPC_INSTALL_DIR="${SRC_DIR}/grpc_install"
+
+        bash ./ci/utils/install_protobuf_grpc.sh --prefix="${GRPC_INSTALL_DIR}" --build-dir="${SRC_DIR}" --skip-deps
+
+        export CMAKE_PREFIX_PATH="${GRPC_INSTALL_DIR}:${CMAKE_PREFIX_PATH:-}"
```
gforsyth: Have we tried this build using the available conda-forge packages?

`libgrpc` is build against `libprotobuf` and `libabseil` with the fix for the linked issue available

tmckayus: yes trying this in another test PR, thanks

conda/recipes/libcuopt/recipe.yaml
tmckayus: resolved, we've got conda build using packages now

conda/recipes/libcuopt/recipe.yaml
tmckayus: modified, works

ci/utils/install_protobuf_grpc.sh
bdice, line 1: This file is only called by `build_wheel_libcuopt.sh`, if I'm reading it correctly. Do we need all of the options and flexible behavior that is implemented in this script, or could we reduce its scope?

ramakrishnap-nv: Think we can trim it and make it bare minimum. @tmckayus may be we can tackle this in a followup PR ?

PR #952, opened 2026-03-11
Add a skill for evolving skills or adding new skills based on interaction

rgsl888prabhu:
## Description
The new skill will try to propose a change or new skills so that it can capture generic patterns to help next developer

## Checklist

- Testing
   - [ ] Added tests
   - [x] NA
- Documentation
   - [ ] Added new documentation
   - [x] NA

### DIFF
--- a/cpp/tests/linear_programming/c_api_tests/c_api_tests.cpp
+++ b/cpp/tests/linear_programming/c_api_tests/c_api_tests.cpp
@@ -382,7 +382,7 @@ class CPUOnlyTestEnvironment {
 
 // TODO: Add numerical assertions once gRPC remote solver replaces the stub implementation.
 // Currently validates that the CPU-only C API path completes without errors.
-TEST(c_api_cpu_only, lp_solve)
+TEST(c_api_cpu_only, DISABLED_lp_solve)
 {
   CPUOnlyTestEnvironment env;
   const std::string& rapidsDatasetRootDir = cuopt::test::get_rapids_dataset_root_dir();
@@ -391,7 +391,7 @@ TEST(c_api_cpu_only, lp_solve)
 }
 
 // TODO: Add numerical assertions once gRPC remote solver replaces the stub implementation.
-TEST(c_api_cpu_only, mip_solve)
+TEST(c_api_cpu_only, DISABLED_mip_solve)
 {
   CPUOnlyTestEnvironment env;
   const std::string& rapidsDatasetRootDir = cuopt::test::get_rapids_dataset_root_dir();
--- a/python/cuopt/cuopt/tests/linear_programming/test_cpu_only_execution.py
+++ b/python/cuopt/cuopt/tests/linear_programming/test_cpu_only_execution.py
@@ -168,6 +168,8 @@ def _impl_warmstart_cpu_only():
 class TestCPUOnlyExecution:
     """Tests that run with CUDA_VISIBLE_DEVICES='' to simulate CPU-only hosts."""
 
+    pytestmark = pytest.mark.skip(reason="CPU-only tests temporarily disabled")
+
     @pytest.fixture
     def env(self):
         return _cpu_only_env()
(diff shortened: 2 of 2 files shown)


ci/README.md
```diff
@@ -41,3 +41,15 @@ Similarly, for Conda package,
...
+## Skill validation
+
+These checks validate the agent skills under `skills/` (structure, content, and runnable assets).
+
+### Existing tests
+
+| Test | What it does | How to run |
+|------|--------------|------------|
+| **validate_skills.sh** | Every skill has `SKILL.md`, frontmatter (`name:`, `description:`), version in sync with `VERSION`, and is referenced in plugin/AGENTS config. | `./ci/utils/validate_skills.sh` (also runs in pre-commit). |
+| **test_skills_assets.sh** | Runs executable assets: Python scripts, C programs, CLI via `cuopt_cli` on MPS files. | `./ci/test_skills_assets.sh` |
```
hlinsen: question: Should these be pushed?

skills/cuopt-user-rules/SKILL.md
```diff
@@ -113,6 +113,11 @@ Do not bury the objective value only in the middle of a paragraph; it must appea
 
 Provide diagnostic code snippets when helpful.
 
+### Post-correction check (mandatory)
+<!-- added by skill-evolution -->
+
+If the result required a correction, retry, or workaround to reach this point, you MUST evaluate the skill-evolution workflow (`skills/skill-evolution/SKILL.md`) before moving on. Do not skip this step.
+
```
hlinsen, line 120: Missing end marker

skills/lp-milp-formulation/SKILL.md
```diff
@@ -126,3 +126,22 @@ When the user gives **problem text**, classify every sentence and then summarize
...
+
+1. **Hard constraints** — capacity limits, non-negativity, etc. These hold in every phase.
+2. **Goal constraints** — for each goal, introduce deviation variables (d⁻ for underachievement, d⁺ for overachievement) and write an equality: `expression + d⁻ − d⁺ = target`.
+3. **Solve sequentially by priority:**
+   - Phase 1: minimize (or maximize) the relevant deviation for the highest-priority goal.
+   - Phase k: fix all higher-priority deviations at their optimal values, then optimize priority k's deviation.
+
+### Variable types in goal programming
+
+Deviation variables (d⁻, d⁺) and slack/idle-time variables are always **continuous**. However, **decision variables must still be INTEGER when they represent discrete/countable quantities** (units produced, vehicles, workers, etc.). Do not let the presence of continuous deviation variables cause you to make all variables continuous — the integrality of decision variables directly affects feasibility and objective values.
```
hlinsen, line 191: Missing end marker

skills/skill-evolution/SKILL.md
```diff
@@ -0,0 +1,256 @@
...
+  Phase: inference (unscored)
+  Section: <where it goes>
+  Trigger: <what happened>
+  Change: <the exact content to add or modify>
+  Note: This was not validated against ground truth. Review carefully.
+```
+
+The user may approve, decline, or defer for offline reflection.
+
+## Phase 3: Offline reflection
```
hlinsen, line 129: Few questions:
1. For training the agent should have a clear context for each new prompt.
2. The agent has seen the ground truths and would be jumping to inference without clearing context. I think we need state it explicitly and check that it works otherwise it may not be possible to do in the same active session.
3. I think we need an evaluate.py that sets up the train pipeline. I don't think it can go over a dataset and understand what is train and test with the skill only. You can take a look at: https://github.com/karpathy/autoresearch/blob/master/prepare.py
4. Should we provide a parser.py? The data is either in json or csv for outbound skill refinement: https://github.com/NVIDIA/cuopt-examples/tree/main/cuopt-agent/cuopt_agent/data/max_supply_what_ifs/eval

ramakrishnap-nv: I am currently working on eval part where I will try to evolve skill based on industry or dataset and then run again with updated ones. And also run this against the optmath and other dataset to see if it provides better performance.

ramakrishnap-nv: @hlinsen I agree, evaluation and evolution script needs to be separate. There is a on going discussion on how to tackle this, may be we can work on this as follow-up.

PR #957, opened 2026-03-14
Store best dual objective, align root LP bound with relative_mip_gap

anandhkb:
- Callback: store only best (tightest) dual objective; it can worsen during iterations (min: keep largest, max: keep smallest)
- set_final_solution: when returning with -inf, push root_lp_current_lower_bound_ to stats so relative_mip_gap uses the best/tightest bound
- user_bound_callback: only update stats when new bound is better, avoid overwriting e.g. set_simplex_solution with a worse root LP value
- Log line change: 'Root LP dual objective (last)' -> '(best)'

### DIFF
--- a/cpp/src/branch_and_bound/branch_and_bound.cpp
+++ b/cpp/src/branch_and_bound/branch_and_bound.cpp
@@ -727,7 +727,7 @@ void branch_and_bound_t<i_t, f_t>::set_final_solution(mip_solution_t<i_t, f_t>& 
   {
     const f_t root_lp_obj = root_lp_current_lower_bound_.load();
     if (std::isfinite(root_lp_obj)) {
-      settings_.log.printf("Root LP dual objective (last): %.16e\n", root_lp_obj);
+      settings_.log.printf("Root LP dual objective (best): %.16e\n", root_lp_obj);
     }
   }
 
@@ -769,6 +769,14 @@ void branch_and_bound_t<i_t, f_t>::set_final_solution(mip_solution_t<i_t, f_t>& 
   solution.lower_bound        = lower_bound;
   solution.nodes_explored     = exploration_stats_.nodes_explored;
   solution.simplex_iterations = exploration_stats_.total_lp_iters;
+
+  // When returning early (e.g. time limit at root) with -inf, push root LP bound to stats
+  // so relative_mip_gap uses the best bound we have (root_lp_bound is already in user space)
+  const f_t root_lp_bound = root_lp_current_lower_bound_.load();
+  if (!std::isfinite(lower_bound) && std::isfinite(root_lp_bound) &&
+      user_bound_callback_ != nullptr) {
+    user_bound_callback_(root_lp_bound);
+  }
 }
 
 template <typename i_t, typename f_t>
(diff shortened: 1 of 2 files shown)


cpp/src/branch_and_bound/branch_and_bound.cpp
```diff
@@ -2026,7 +2034,12 @@ mip_status_t branch_and_bound_t<i_t, f_t>::solve(mip_solution_t<i_t, f_t>& solut
   lp_settings.scale_columns                   = false;
   lp_settings.concurrent_halt                 = get_root_concurrent_halt();
   lp_settings.dual_simplex_objective_callback = [this](f_t user_obj) {
-    root_lp_current_lower_bound_.store(user_obj);
+    // Store only the best (tightest) dual objective; it can worsen during iterations
```
chris-maes, line 2037: The dual objective should always increase over the iterations.

anandhkb: Ok, thanks, I refer to Burcin's testcase:

cuopt_cli --log-file Optimality_MIP_1.log --solution-file Optimality_MIP_1.sln --time-limit 28500 Optimality_MIP_1.mps

Issue is: The dual obj value is getting worse rather than monotonically improving. He says -  "If you have taken the last dual obj for your gap calculation, the resulting value is not correct because a better dual obj value was found in an earlier iteration". Here is the log Burcin referred me to:

```
236000 -7.8615507602829341e+04   33078 2.15752325e+16 0.00e+00 873.90
237000 -7.8599148858700210e+04   32968 2.33309309e+16 0.00e+00 878.39
238000 -7.8628596397306974e+04   31778 4.13809980e+15 0.00e+00 882.75
239000 -7.8625353538739626e+04   31793 4.09842601e+15 0.00e+00 886.47
240000 -3.4223435428449867e+05   34910 4.33595251e+22 0.00e+00 894.38.  ### this is where dual obj value gets worse
241000 -4.9084521357962309e+05   33468 3.08252689e+20 0.00e+00 899.81.  ### and worse...
242000 -5.2980592629211734e+05   33705 7.28784267e+19 0.00e+00 904.61
243000 -5.3195161855784897e+05   32822 7.03858147e+18 2.00e-07 908.23
244000 -5.3188925240266300e+05   32597 2.46222227e+19 0.00e+00 912.64
...
279000 -5.7260862863620499e+05   24915 1.06134033e+18 0.00e+00 1234.99
280000 -5.7271001072192355e+05   24276 7.90538350e+17 0.00e+00 1241.53
281000 -5.7235100939733768e+05   24464 1.18694208e+18 0.00e+00 1257.79
New solution from primal heuristics. Objective -5.993702e+04. Gap 854.9%. Time 1258.26.    ### a heuristic solution found but gap should be 31.18%, in relation to line 239000
282000 -5.7285144600908319e+05   20031 1.07701601e+13 1.00e-07 1276.96
283000 -5.7284553597028111e+05   23587 4.28140763e+14 0.00e+00 1287.60
```

chris-maes: That looks like an issue in either numerics or perturbation in dual simplex. We should fix the issue inside of dual simplex, rather than adding code to hide the issue outside of dual simplex. 

Please let me know how to reproduce the above log.

PR #963, opened 2026-03-17
Use a single dual simplex pivot to initialize pseudocost

chris-maes:
This is an alternative to using strong branching to initialize pseudocosts. 

This may also be used to rank candidates for reliability branching.

### DIFF
--- a/cpp/src/branch_and_bound/pseudo_costs.cpp
+++ b/cpp/src/branch_and_bound/pseudo_costs.cpp
@@ -22,6 +22,256 @@ namespace cuopt::linear_programming::dual_simplex {
 
 namespace {
 
+
+template <typename f_t>
+struct objective_estimate_t {
+  f_t down_obj;
+  f_t up_obj;
+};
+
+template <typename i_t, typename f_t>
+f_t compute_step_length(const simplex_solver_settings_t<i_t, f_t>& settings,
+                        const std::vector<variable_status_t>& vstatus,
+                        const std::vector<f_t>& z,
+                        const std::vector<f_t>& delta_z,
+                        const std::vector<i_t>& delta_z_indices)
+{
+  f_t step_length = inf;
+  f_t pivot_tol = settings.pivot_tol;
+  const i_t nz = delta_z_indices.size();
+  for (i_t h = 0; h < nz; h++) {
+    const i_t j = delta_z_indices[h];
+    if (vstatus[j] == variable_status_t::NONBASIC_LOWER && delta_z[j] < -pivot_tol) {
+      const f_t ratio = -z[j] / delta_z[j];
+      if (ratio < step_length) {
+        step_length = ratio;
+      }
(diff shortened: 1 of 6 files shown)


cpp/src/branch_and_bound/pseudo_costs.cpp
```diff
@@ -22,6 +22,256 @@ namespace cuopt::linear_programming::dual_simplex {
...
+                                                                  y,
+                                                                  z,
+                                                                  basic_list,
+                                                                  nonbasic_list,
+                                                                  nonbasic_mark,
+                                                                  basis_factors,
+                                                                  workspace,
+                                                                  delta_z,
+                                                                  work_estimate);
+    pc.strong_branch_down[k] = estimate.down_obj;
```
nguidotti, line 267: For the pseudocost initilization, should be `strong_branch_down`/`up` set to the objective change instead of the objective? This matches the other code paths.

cpp/src/branch_and_bound/pseudo_costs.cpp
```diff
@@ -22,6 +22,256 @@ namespace cuopt::linear_programming::dual_simplex {
...
+  e_k.x.push_back(-f_t(direction));
+
+  sparse_vector_t<i_t, f_t> delta_y(lp.num_rows, 0);
+  basis_factors.b_transpose_solve(e_k, delta_y);
+
+  // Compute delta_z_N = -N^T * delta_y
+  i_t delta_y_nz0      = 0;
+  const i_t nz_delta_y = delta_y.i.size();
+  for (i_t k = 0; k < nz_delta_y; k++) {
+    if (std::abs(delta_y.x[k]) > 1e-12) { delta_y_nz0++; }
```
nguidotti, line 95: nitpick: replaced `1E-12` with `zero_tol` from `simplex_settings`.

cpp/src/branch_and_bound/pseudo_costs.cpp
```diff
@@ -22,6 +22,256 @@ namespace cuopt::linear_programming::dual_simplex {
...
+                                           const std::vector<i_t>& nonbasic_list,
+                                           const std::vector<i_t>& fractional,
+                                           f_t root_obj,
+                                           basis_update_mpf_t<i_t, f_t>& basis_factors,
+                                           pseudo_costs_t<i_t, f_t>& pc)
+{
+  i_t m = lp.num_rows;
+  i_t n = lp.num_cols;
+
+  csc_matrix_t<i_t, f_t> A_transpose(m, n, 0);
```
nguidotti, line 230: We are using $\mathbf{A}^T$ in many places (in dual simplex, bound strengthening and here). Can this move to `lp_problem_t` so it is visible across the entire code?

chris-maes: lp_problem_t is not the right place to store the A^T. Computing a transpose is relatively inexpensive. I'm not sure it's worth consolidating these.

PR #964, opened 2026-03-17
apply -UNDEBUG to coupt flags instead of global settings

tmckayus:
don't pass -UNDEBUG to other libraries we may build

### DIFF
--- a/cpp/CMakeLists.txt
+++ b/cpp/CMakeLists.txt
@@ -109,6 +109,8 @@ if(BUILD_MSAN)
   add_link_options(-fsanitize=memory)
 endif(BUILD_MSAN)
 
+# Note: -UNDEBUG is applied via CUOPT_CXX_FLAGS / CUOPT_CUDA_FLAGS (not add_definitions)
+# to avoid leaking into dependencies that are built in-tree.
 if(DEFINE_ASSERT)
   add_definitions(-DASSERT_MODE)
   list(APPEND CUOPT_CXX_FLAGS -UNDEBUG)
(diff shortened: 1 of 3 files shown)


cpp/CMakeLists.txt
```diff
@@ -184,11 +186,9 @@ elseif(CMAKE_CUDA_LINEINFO)
   set(CMAKE_CUDA_FLAGS_RELEASE "${CMAKE_CUDA_FLAGS_RELEASE} -lineinfo")
 endif(CMAKE_BUILD_TYPE MATCHES Debug)
 
-# Undefine NDEBUG if assert mode is on
-if(DEFINE_ASSERT)
-  message(STATUS "Undefining NDEBUG with assert mode enabled")
-  add_definitions(-UNDEBUG)
-endif()
+# Note: -UNDEBUG is applied via CUOPT_CXX_FLAGS / CUOPT_CUDA_FLAGS (not add_definitions)
```
hlinsen: Should these comments be moved line 112?

tmckayus: yes probably better, will do

cpp/CMakeLists.txt
```diff
@@ -109,8 +109,12 @@ if(BUILD_MSAN)
   add_link_options(-fsanitize=memory)
 endif(BUILD_MSAN)
 
+# Note: -UNDEBUG is applied via CUOPT_CXX_FLAGS / CUOPT_CUDA_FLAGS (not add_definitions)
+# to avoid leaking into dependencies that are built in-tree.
 if(DEFINE_ASSERT)
   add_definitions(-DASSERT_MODE)
+  list(APPEND CUOPT_CXX_FLAGS -UNDEBUG)
+  list(APPEND CUOPT_CUDA_FLAGS -UNDEBUG)
```
rg20, line 117: does this turn off all assertions?

tmckayus: This is "undefine no debug", which means "leave NDEBUG (no debug) unset".   So if DEFINE_ASSERT is set, we are allowing assertions guarded by NDEBUG to still work. The standard C++ assert macro checks NDEBUG to turn off assertions; cuopt_assert() delegates to assert() so this logic applies.

The only difference here is that previously if DEFINE_ASSERT was on, we set -UNDEBUG on every build under cpp, including fetched dependencies built as subdirectories. In this setup, we only set -UNDEBUG on cuopt targets so that if DEFINE_ASSERT is on we get the assertions in force.

PR #979, opened 2026-03-20
Improvements to reliability branching

nguidotti:
We can use a single dual simplex pivot to estimate the objective change after branch up or down in a variable (#963), which can be later be used for ranking the unreliable variables in reliability branching. To minimize the cost of reliability branching, we only apply trial branching to the 100 best variables. Other variables are not considered. Previously, the 100 unreliable variables were selected at random, which may not lead to best branching decision.  In this PR, we also apply a cutoff based on the objective coefficients when no incumbent is available.

MIPLIB Benchmark (10 minute runs) on a GH200 system:

Does not include `graphdraw-domain` and `cbscta`

```
RUN 1

 main f0ba447  (1) vs improve-reliable-branching (2)

| Feasible                                                 225                  226                   +1                 --- |
| Optimal                                                   69                   69                   +0                 --- |
| Solutions with <0.1% primal gap                          121                  124                   +3                 --- |
| Nodes explored (mean)                           4301972.7899         4498957.3655         +196984.5756              +4.378 |
| Nodes explored (shifted geomean)                   6414.1273            7705.4824           +1291.3550             +16.759 |
| Relative MIP gap (mean)                               0.3397               0.3403              +0.0006              +0.184 |
| Relative MIP gap (shifted geomean)                    0.1201               0.1216              +0.0016              +1.277 |
| Solve time (mean)                                   452.1236             452.7833              +0.6597              +0.146 |
| Solve time (shifted geomean)                        226.7962             229.4358              +2.6396              +1.150 |
| Primal gap (mean)                                    11.4530              10.9459              -0.5071              -4.428 |
| Primal gap (shifted geomean)                          0.6509               0.6382              -0.0127              -1.953 |
| Primal integral (mean)                               49.5055              47.2675              -2.2380              -4.521 |
| Primal integral (shifted geomean)                    11.4439              10.4722              -0.9717              -8.491 |

```

```
RUN 2

 main f0ba447  (1) vs improve-reliable-branching (2)

| Feasible                                                 224                  224                   +0                 --- |
| Optimal                                                   69                   68                   -1                 --- |
| Solutions with <0.1% primal gap                          122                  116                   -6                 --- |
| Nodes explored (mean)                           4490538.7605         4521169.8739          +30631.1134              +0.678 |
| Nodes explored (shifted geomean)                   7308.8805            7525.6319            +216.7514              +2.880 |
| Relative MIP gap (mean)                               0.3269               0.3456              +0.0186              +5.393 |
| Relative MIP gap (shifted geomean)                    0.1192               0.1214              +0.0021              +1.766 |
| Solve time (mean)                                   453.2170             451.1577              -2.0594              -0.454 |
| Solve time (shifted geomean)                        230.0384             227.2056              -2.8328              -1.231 |
| Primal gap (mean)                                    11.6922              11.5728              -0.1194              -1.021 |
| Primal gap (shifted geomean)                          0.6535               0.6617              +0.0082              +1.244 |
| Primal integral (mean)                               50.7645              49.8675              -0.8969              -1.767 |
| Primal integral (shifted geomean)                    12.4180              10.9963              -1.4217             -11.449 |


```

## Checklist

- Testing
   - [ ] Added tests
   - [ ] NA
- Documentation
   - [ ] Added new documentation
   - [ ] NA

### DIFF
--- a/cpp/src/mip_heuristics/diversity/diversity_manager.cu
+++ b/cpp/src/mip_heuristics/diversity/diversity_manager.cu
@@ -108,7 +108,48 @@ diversity_manager_t<i_t, f_t>::diversity_manager_t(mip_solver_context_t<i_t, f_t
   }
 }
 
-// this function is to specialize the local search with config from diversity manager
+template <typename i_t, typename f_t>
+void diversity_manager_t<i_t, f_t>::consume_staged_simplex_solution(lp_state_t<i_t, f_t>& lp_state)
+{
+  std::vector<f_t> staged_simplex_solution_local;
+  std::vector<f_t> staged_simplex_dual_solution_local;
+  f_t staged_simplex_objective_local = std::numeric_limits<f_t>::infinity();
+  {
+    std::lock_guard<std::mutex> guard(relaxed_solution_mutex);
+    cuopt_assert(simplex_solution_exists.load(),
+                 "Simplex solution flag set without a staged simplex solution");
+    staged_simplex_solution_local      = staged_simplex_solution;
+    staged_simplex_dual_solution_local = staged_simplex_dual_solution;
+    staged_simplex_objective_local     = staged_simplex_objective;
+  }
+  solution_t<i_t, f_t> new_sol(*problem_ptr);
+  cuopt_assert(new_sol.assignment.size() == staged_simplex_solution_local.size(),
+               "Assignment size mismatch");
+  cuopt_assert(problem_ptr->n_constraints == staged_simplex_dual_solution_local.size(),
+               "Dual assignment size mismatch");
+  new_sol.copy_new_assignment(staged_simplex_solution_local);
+  new_sol.compute_feasibility();
+  cuopt_assert(integer_equal(new_sol.get_user_objective(), staged_simplex_objective_local, 1e-3),
+               "Objective mismatch");
+  raft::copy(lp_optimal_solution.data(),
+             staged_simplex_solution_local.data(),
+             staged_simplex_solution_local.size(),
+             problem_ptr->handle_ptr->get_stream());
+  clamp_within_var_bounds(lp_optimal_solution, problem_ptr, problem_ptr->handle_ptr);
+  raft::copy(lp_state.prev_primal.data(),
+             lp_optimal_solution.data(),
+             lp_optimal_solution.size(),
+             problem_ptr->handle_ptr->get_stream());
+  problem_ptr->handle_ptr->sync_stream();
+  solution_t<i_t, f_t> bounded_lp_sol(*problem_ptr);
+  bounded_lp_sol.copy_new_assignment(lp_optimal_solution);
+  bounded_lp_sol.handle_ptr->sync_stream();
+  auto max_lp_bound_violation = bounded_lp_sol.compute_max_variable_violation();
+  cuopt_assert(max_lp_bound_violation == 0.0,
+               "LP optimal solution must be within variable bounds after staged copy");
+  set_new_user_bound(staged_simplex_objective_local);
+}
+
 template <typename i_t, typename f_t>
 bool diversity_manager_t<i_t, f_t>::run_local_search(solution_t<i_t, f_t>& solution,
                                                      const weight_t<i_t, f_t>& weights,
@@ -416,6 +457,7 @@ solution_t<i_t, f_t> diversity_manager_t<i_t, f_t>::run_solver()
   lp_state.resize(*problem_ptr, problem_ptr->handle_ptr->get_stream());
   bool bb_thread_solution_exists = simplex_solution_exists.load();
   if (bb_thread_solution_exists) {
+    consume_staged_simplex_solution(lp_state);
     ls.lp_optimal_exists = true;
   } else if (!fj_only_run) {
     convert_greater_to_less(*problem_ptr);
@@ -440,9 +482,11 @@ solution_t<i_t, f_t> diversity_manager_t<i_t, f_t>::run_solver()
     timer_t lp_timer(lp_time_limit);
     auto lp_result = solve_lp_with_method<i_t, f_t>(*problem_ptr, pdlp_settings, lp_timer);
 
+    bool use_staged_simplex_solution = false;
     {
       std::lock_guard<std::mutex> guard(relaxed_solution_mutex);
-      if (!simplex_solution_exists.load()) {
+      use_staged_simplex_solution = simplex_solution_exists.load();
+      if (!use_staged_simplex_solution) {
         cuopt_assert(lp_result.get_primal_solution().size() == lp_optimal_solution.size(),
                      "LP optimal solution size mismatch");
         cuopt_assert(lp_result.get_dual_solution().size() == lp_dual_optimal_solution.size(),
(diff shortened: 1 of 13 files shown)


cpp/src/branch_and_bound/pseudo_costs.cpp
```diff
@@ -62,6 +293,20 @@ void strong_branch_helper(i_t start,
       if (elapsed_time > settings.time_limit) { break; }
       child_settings.time_limit      = std::max(0.0, settings.time_limit - elapsed_time);
       child_settings.iteration_limit = 200;
+
+      if (std::isfinite(upper_bound)) {
+        child_settings.cut_off = upper_bound + settings.dual_tol;
+      } else {
+        child_settings.cut_off = 0;
+        for (i_t i = 0; i < original_lp.num_cols; ++i) {
+          if (original_lp.objective[i] < 0) {
```
chris-maes: Is this backwards? You want to set the cut off to an upper bound on what the objective can be.

If c_j > 0, to get an upper bound you should multiply by problem.upper[j]

If c_j < 0, to get an upper bound you should multiply by problem.lower[j]

Also, since you are looping over columns please use j as the loop variable instead of i. We use the convention that j is for columns and i is for rows.

cpp/src/branch_and_bound/pseudo_costs.cpp
```diff
@@ -165,13 +409,23 @@ f_t trial_branching(const lp_problem_t<i_t, f_t>& original_lp,
...
-  child_settings.cut_off         = upper_bound + settings.dual_tol;
+  child_settings.iteration_limit = iter_limit;
   child_settings.inside_mip      = 2;
   child_settings.scale_columns   = false;
 
+  if (std::isfinite(upper_bound)) {
+    child_settings.cut_off = upper_bound + settings.dual_tol;
+  } else {
+    child_settings.cut_off = 0;
+    for (i_t i = 0; i < original_lp.num_cols; ++i) {
```
chris-maes: Same issue here. Would it make sense to pull this out into a helper function? Maybe 
`f_t objective_upper_bound(const lp_problem_t& lp)`

cpp/src/branch_and_bound/pseudo_costs.cpp
```diff
@@ -394,6 +652,17 @@ void strong_branching(const user_problem_t<i_t, f_t>& original_problem,
       pc.strong_branch_down[k] = obj_down - root_obj;
       pc.strong_branch_up[k]   = obj_up - root_obj;
     }
+  } else if (settings.mip_batch_pdlp_strong_branching == 2) {
```
chris-maes: Probably we should introduce a different setting for this. I think @Kh4ster PR use mip_batch_pdlp_strong_branching == 2 to mean something else

Maybe we could add the setting `mip_strong_branching_dual_simplex_iteration_limit` and if it is set to 1 do this. If it is set to greater than 1, we use that value in the strong branching code.

cpp/src/branch_and_bound/pseudo_costs.cpp
```diff
@@ -623,18 +902,77 @@ i_t pseudo_costs_t<i_t, f_t>::reliable_variable_selection(
     num_tasks,
     reliable_threshold);
 
-  // Shuffle the unreliable list so every variable has the same chance to be selected.
-  if (unreliable_list.size() > max_num_candidates) { worker->rng.shuffle(unreliable_list); }
+  if (unreliable_list.size() > max_num_candidates) {
+    if (reliability_branching_settings.rank_candidates_with_dual_pivot) {
+      i_t m             = worker->leaf_problem.num_rows;
```
chris-maes, line 911: This just repeats the previous code right? Would it make sense to put it in a function and just call that function here?

cpp/src/dual_simplex/simplex_solver_settings.hpp
```diff
@@ -190,6 +190,11 @@ struct simplex_solver_settings_t {
   i_t mip_batch_pdlp_strong_branching{0};  // 0 if not using batch PDLP for strong branching, 1 if
                                            // using batch PDLP for strong branching
 
+  // Instead of using the full dual simplex for solving each trial branch in strong branching,
+  // use just a single pivot from dual simplex to estimate the objective change.
+  // TODO: Need to pass a proper benchmark before it can be enabled.
```
chris-maes: Let's keep it default false and only add it as an option. I'd probably also suggest making it an int rather than a bool. As we may want to have it be automatic in the future. And we can use -1 for automatic, 0 for  force off, 1 for force on.

chris-maes: Actually I think this would be better if we CUOPT_MIP_STRONG_BRANCHING_SIMPLEX_ITERATION_LIMIT.

For other work with strong branching / BatchPDLP we are running experiments where we change the iteration limit used by dual simplex. It would be great if this could be a parameter.

@Kh4ster for viz

cpp/src/math_optimization/solver_settings.cu
```diff
@@ -99,7 +99,7 @@ solver_settings_t<i_t, f_t>::solver_settings_t() : pdlp_settings(), mip_settings
     {CUOPT_MIP_REDUCED_COST_STRENGTHENING, &mip_settings.reduced_cost_strengthening, -1, std::numeric_limits<i_t>::max(), -1},
     {CUOPT_NUM_GPUS, &pdlp_settings.num_gpus, 1, 2, 1},
     {CUOPT_NUM_GPUS, &mip_settings.num_gpus, 1, 2, 1},
-    {CUOPT_MIP_BATCH_PDLP_STRONG_BRANCHING, &mip_settings.mip_batch_pdlp_strong_branching, 0, 1, 0},
+    {CUOPT_MIP_BATCH_PDLP_STRONG_BRANCHING, &mip_settings.mip_batch_pdlp_strong_branching, 0, 2, 0},
```
chris-maes: Could we add  CUOPT_MIP_STRONG_BRANCHING_SIMPLEX_ITERATION_LIMIT?

if iteration limit is -1 we choose, if 1 we use the single pivot estimation, if > 1 we adjust the iteration limit in strong branching. For now if the user sets 0, maybe we can do pivot estimation as well. Later on maybe we can have 0 skip strong branching (but we'd need to have other ways to decide how to branch).

I would default it to -1 for now. And make it so that defaults cause us to use a 200 iteration limit.

cpp/src/dual_simplex/simplex_solver_settings.hpp
nguidotti: Sounds good. Added in [042eba2](https://github.com/NVIDIA/cuopt/pull/979/commits/042eba24f59b038910f44887f73df600b5734cfc)

cpp/include/cuopt/linear_programming/constants.h
```diff
@@ -20,64 +20,65 @@
...
+#define CUOPT_MIP_MIXED_INTEGER_GOMORY_CUTS           "mip_mixed_integer_gomory_cuts"
+#define CUOPT_MIP_KNAPSACK_CUTS                       "mip_knapsack_cuts"
+#define CUOPT_MIP_IMPLIED_BOUND_CUTS                  "mip_implied_bound_cuts"
+#define CUOPT_MIP_CLIQUE_CUTS                         "mip_clique_cuts"
+#define CUOPT_MIP_STRONG_CHVATAL_GOMORY_CUTS          "mip_strong_chvatal_gomory_cuts"
+#define CUOPT_MIP_REDUCED_COST_STRENGTHENING          "mip_reduced_cost_strengthening"
+#define CUOPT_MIP_CUT_CHANGE_THRESHOLD                "mip_cut_change_threshold"
+#define CUOPT_MIP_CUT_MIN_ORTHOGONALITY               "mip_cut_min_orthogonality"
+#define CUOPT_MIP_BATCH_PDLP_STRONG_BRANCHING         "mip_batch_pdlp_strong_branching"
+#define CUOPT_MIP_STRONG_BRANCHING_SIMPLEX_ITER_LIMIT "mip_strong_branching_simplex_iter_limit"
```
chris-maes, line 74: This is minor, but it becomes part of the API, so probably worth getting it correct. All other parameters use full word rather than abbreviations. 

So we should probably call this `mip_strong_branching_simplex_iteration_limit`

cpp/src/branch_and_bound/pseudo_costs.cpp
```diff
@@ -399,46 +657,62 @@ void strong_branching(const user_problem_t<i_t, f_t>& original_problem,
                         settings.num_threads,
                         fractional.size());
     f_t strong_branching_start_time = tic();
-
+    i_t simplex_iteration_limit     = settings.strong_branching_simplex_iteration_limit;
+
+    if (simplex_iteration_limit < 1) {
```
chris-maes, line 662: If I understand correctly, you've set the default of `strong_branching_simplex_iteration_limit = -1`. Does that mean by default we will use `initialize_psuedo_costs_with_estimates`? Or maybe somewhere else in the code you catch this?

chris-maes: Ah got it. You set it to 200 in `solver.cu` if it was set to -1.

cpp/src/mip_heuristics/solver.cu
```diff
@@ -347,12 +347,16 @@ solution_t<i_t, f_t> mip_solver_t<i_t, f_t>::run_solver()
...
       context.settings.strong_chvatal_gomory_cuts;
-    branch_and_bound_settings.reduced_cost_strengthening =
-      context.settings.reduced_cost_strengthening;
     branch_and_bound_settings.cut_change_threshold  = context.settings.cut_change_threshold;
     branch_and_bound_settings.cut_min_orthogonality = context.settings.cut_min_orthogonality;
     branch_and_bound_settings.mip_batch_pdlp_strong_branching =
       context.settings.mip_batch_pdlp_strong_branching;
+
+    branch_and_bound_settings.strong_branching_simplex_iteration_limit =
+      context.settings.strong_branching_simplex_iteration_limit < 0
```
chris-maes, line 356: Probably this should be `<= 0`. Since if the user sets it to zero now, we still need to initialize the pseudocosts for the code to work.

nguidotti: If the user set to `0` here, we use the objective estimate via dual simplex pivot.

PR #993, opened 2026-03-24
Expose GPU heuristics tuning parameters via config files

aliceb-nv:
This PR exposes some GPU heuristics tuning parameters to the user through a configuration file of the following format:

```
# cuOpt parameter configuration (auto-generated)
# Uncomment and change only the values you want to override.

# max solutions in pool (int, range: [1, 2147483647])
# hyper_population_size = 32

# parallel CPU FJ climbers (int, range: [0, 2147483647])
# hyper_num_cpufj_threads = 8

# FP loops w/o improvement before recombination (int, range: [1, 2147483647])
# hyper_stagnation_trigger = 3

# diversity step depth after stagnation (int, range: [1, 2147483647])
# hyper_max_iterations_without_improvement = 8

# FJ baseline local-minima exit threshold (int, range: [1, 2147483647])
# hyper_n_of_minimums_for_exit = 7000

# bitmask: 1=BP 2=FP 4=LS 8=SubMIP (int, range: [0, 15])
# hyper_enabled_recombiners = 15

# FP assignment cycle ring buffer length (int, range: [1, 2147483647])
# hyper_cycle_detection_length = 30


# ...
```
Parameters are only exposed through the cuopt_cli runner for the moment, via the flags `--dump-mip-heuristic-config` (to export a sample config file with the default parameters), and `--mip-heuristic-config` (to load the parameters from a file).

## Checklist

- Testing
   - [ ] Added tests
   - [ ] NA
- Documentation
   - [ ] Added new documentation
   - [ ] NA

### DIFF
--- a/cpp/cuopt_cli.cpp
+++ b/cpp/cuopt_cli.cpp
@@ -247,12 +247,17 @@ int set_cuda_module_loading(int argc, char* argv[])
  */
 int main(int argc, char* argv[])
 {
-  // Handle --dump-hyper-params before argparse so no other args are required
+  // Handle dump flags before argparse so no other args are required
   for (int i = 1; i < argc; ++i) {
     std::string arg = argv[i];
     if (arg == "--dump-hyper-params") {
       cuopt::linear_programming::solver_settings_t<int, double> settings;
       settings.dump_parameters_to_file("/dev/stdout", true);
+      return 0;
+    }
+    if (arg == "--dump-params") {
+      cuopt::linear_programming::solver_settings_t<int, double> settings;
+      settings.dump_parameters_to_file("/dev/stdout", false);
       return 0;
     }
   }
@@ -290,7 +295,12 @@ int main(int argc, char* argv[])
     .default_value(std::string(""));
 
   program.add_argument("--dump-hyper-params")
-    .help("print hyper-parameters in config-file format and exit")
+    .help("print hyper-parameters only in config file format and exit")
+    .default_value(false)
+    .implicit_value(true);
+
+  program.add_argument("--dump-params")
+    .help("print all parameters in config file format and exit")
     .default_value(false)
     .implicit_value(true);
 


cpp/src/math_optimization/solver_settings.cu
```diff
@@ -473,6 +516,111 @@ const std::vector<std::string> solver_settings_t<i_t, f_t>::get_parameter_names(
...
+    if (p.description && p.description[0] != '\0') file << "# " << p.description << " (bool)\n";
+    file << "# " << p.param_name << " = " << (*p.value_ptr ? "true" : "false") << "\n\n";
+  }
+  for (const auto& p : string_parameters) {
+    if (hyperparameters_only && !p.is_hyperparameter) continue;
+    if (p.description && p.description[0] != '\0') file << "# " << p.description << " (string)\n";
+    file << "# " << p.param_name << " = " << quote_if_needed(*p.value_ptr) << "\n\n";
+  }
+  return true;
+}
```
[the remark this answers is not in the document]
aliceb-nv: wildly unlikely scenario that IMO doesn't warrant explicit handling and cluttering of the code

cpp/include/cuopt/linear_programming/utilities/internals.hpp
```diff
@@ -79,37 +79,71 @@ class base_solution_t {
...
+      description(description)
   {
   }
   std::string param_name;
   T* value_ptr;
   T min_value;
   T max_value;
   T default_value;
+  bool is_hyperparameter;
+  const char* description;
```
rg20, line 97: I thought about adding the description for the purpose of  cuopt_cli --help, however, I decided against it in favor of making the names themselves self explanatory.  Further more the the macros are defined as strings, we can probably use them as descriptions. Wdyt?

cpp/include/cuopt/linear_programming/constants.h
```diff
@@ -77,6 +77,25 @@
 #define CUOPT_RANDOM_SEED                     "random_seed"
 #define CUOPT_PDLP_PRECISION                  "pdlp_precision"
 
+/* @brief MIP heuristic hyper-parameter names */
+#define CUOPT_HYPER_POPULATION_SIZE               "hyper_population_size"
```
rg20: these strings are not directly used anywhere, so you can make these as descriptive as you want and use them for descriptions.

cpp/include/cuopt/linear_programming/utilities/internals.hpp
```diff
@@ -79,37 +79,71 @@ class base_solution_t {
...
 template <typename T>
 struct parameter_info_t {
-  parameter_info_t(std::string_view param_name, T* value, T min, T max, T def)
-    : param_name(param_name), value_ptr(value), min_value(min), max_value(max), default_value(def)
+  parameter_info_t(std::string_view param_name,
+                   T* value,
+                   T min,
+                   T max,
+                   T def,
+                   bool is_hyperparameter  = false,
```
rg20: Not sure if we need `is_hyperparameter`   are you using this anywhere? we have HYPER in the parameter names, I think we can deduce this.

cpp/src/mip_heuristics/diversity/diversity_manager.cu
```diff
@@ -37,12 +37,33 @@ size_t sub_mip_recombiner_config_t::max_n_of_vars_from_other =
 template <typename i_t, typename f_t>
 std::vector<recombiner_enum_t> recombiner_t<i_t, f_t>::enabled_recombiners;
 
+namespace {
+diversity_config_t make_diversity_config(const mip_heuristics_hyper_params_t& hp)
+{
+  diversity_config_t c;
+  c.max_solutions                = hp.population_size;
```
rg20: I feel this is error prone.  Now, you will have to add new parameters at two places.   How about compose the `mip_heuristics_hyper_params_t` class with objects like `diversity_config_t` and `rins_settings_t` directly?

rg20 (commented):
Great initiative!  I have some minor comments.

cpp/include/cuopt/linear_programming/utilities/internals.hpp
aliceb-nv: I see your point :) I was however a bit worried about the UX - some of these hyperparameters are named after internal implementation details which aren't immediately obvious or clear; this is why I thought a description field could be relevant to help guide tuning efforts (e.g. what results to expect from this "knob")

cpp/include/cuopt/linear_programming/utilities/internals.hpp
aliceb-nv: I use it to hide these by defaults in the cli --help to avoid cluttering. True that it could  be deduced from the HYPER_ prefix :) Do you think it'd be better to do this by looking at the prefix?

cpp/src/mip_heuristics/diversity/diversity_manager.cu
aliceb-nv: Much cleaner this way, agreed. Will fix

cpp/include/cuopt/linear_programming/constants.h
```diff
@@ -78,6 +78,25 @@
...
+#define CUOPT_HYPER_NUM_CPUFJ_THREADS             "hyper_num_cpufj_threads"
```
chris-maes: Maybe call these `CUOPT_HEURISTIC_***`

cpp/src/mip_heuristics/diversity/lns/rins.cu
```diff
@@ -31,8 +31,8 @@ rins_t<i_t, f_t>::rins_t(mip_solver_context_t<i_t, f_t>& context_,
                          rins_settings_t settings_)
   : context(context_), problem_ptr(context.problem_ptr), dm(dm_), settings(settings_)
 {
-  fixrate    = settings.default_fixrate;
-  time_limit = settings.default_time_limit;
+  fixrate    = context.settings.heuristic_params.rins_fix_rate;
+  time_limit = context.settings.heuristic_params.rins_time_limit;
```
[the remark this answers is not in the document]
aliceb-nv: rins.cuh did not hardcode 10. The time limit was reset to the default value in the constructor, which is 3. Can you check this?

cpp/include/cuopt/linear_programming/utilities/internals.hpp
aliceb-nv: Ended up switching logic to detecting the "hyper_" prefix

PR #994, opened 2026-03-25
Race batch PDLP and dual simplex in strong branching / reliability branching

Kh4ster:
This PR enables the following regarding batch PDLP:
- Enable batch PDLP in reliability branching
- Add work stealing so that batch PDLP and Dual Simplex can run concurrently and steal LPs from each other if one solves it first
- Use correct problem representation with cuts for batch PDLP 
- Use a PDLP warm start cache across strong branching at the root and in reliability branching
- Increase tolerance on batch PDLP to have higher quality solution
- Increase iteration limit to allow instances that needs a high iteration count (with low cost per iteration) to still come through (only while solving the original LP to get warm start data)
- Multiple heuristics to not run batch PDLP to not create overheads when Dual Simplex is clearly superior
- Don't store and copy primal dual solution unless need it to save on memory
- Handle batch PDLP errors better, allowing Dual Simplex to still continue in strong branching even if BPDLP fails
- No early exit if the initial warm start PDLP solution is already feasible in BPDLP
- Correct objective for BPDLP when there is an offset

Currently we still keep BPDLP off by default both at the root and in reliability branching

### DIFF
--- a/cpp/src/branch_and_bound/pseudo_costs.cpp
+++ b/cpp/src/branch_and_bound/pseudo_costs.cpp
@@ -1177,13 +1177,14 @@ void strong_branching(const lp_problem_t<i_t, f_t>& original_lp,
     i_t solved_by_both   = 0;
     for (i_t k = 0; k < fractional.size(); k++) {
       for (i_t branch = 0; branch < 2; branch++) {
-        const bool is_down       = (branch == 0);
-        f_t& sb_dest             = is_down ? pc.strong_branch_down[k] : pc.strong_branch_up[k];
-        f_t ds_obj               = is_down ? dual_simplex_obj_down[k] : dual_simplex_obj_up[k];
-        dual::status_t ds_status = is_down ? dual_simplex_status_down[k] : dual_simplex_status_up[k];
-        f_t pdlp_obj             = is_down ? pdlp_obj_down[k] : pdlp_obj_up[k];
-        bool pdlp_has            = !std::isnan(pdlp_obj);
-        bool ds_has              = ds_status != dual::status_t::UNSET;
+        const bool is_down = (branch == 0);
+        f_t& sb_dest       = is_down ? pc.strong_branch_down[k] : pc.strong_branch_up[k];
+        f_t ds_obj         = is_down ? dual_simplex_obj_down[k] : dual_simplex_obj_up[k];
+        dual::status_t ds_status =
+          is_down ? dual_simplex_status_down[k] : dual_simplex_status_up[k];
+        f_t pdlp_obj  = is_down ? pdlp_obj_down[k] : pdlp_obj_up[k];
+        bool pdlp_has = !std::isnan(pdlp_obj);
+        bool ds_has   = ds_status != dual::status_t::UNSET;
 
         const auto [value, source] =
           merge_sb_result<i_t, f_t>(ds_obj, ds_status, pdlp_obj, pdlp_has);
@@ -1414,11 +1415,11 @@ i_t pseudo_costs_t<i_t, f_t>::reliable_variable_selection(
   constexpr i_t min_num_candidates_for_pdlp                       = 5;
   constexpr f_t min_percent_solved_by_batch_pdlp_at_root_for_pdlp = 5.0;
   // Batch PDLP is either forced or we use the heuristic to decide if it should be used
-  const bool use_pdlp = (rb_mode == 2) || (rb_mode != 0 && !settings.sub_mip && !settings.deterministic &&
-                        pdlp_warm_cache.populated &&
-                        unreliable_list.size() > min_num_candidates_for_pdlp &&
-                        pdlp_warm_cache.percent_solved_by_batch_pdlp_at_root >
-                          min_percent_solved_by_batch_pdlp_at_root_for_pdlp);
+  const bool use_pdlp = (rb_mode == 2) || (rb_mode != 0 && !settings.sub_mip &&
+                                           !settings.deterministic && pdlp_warm_cache.populated &&
+                                           unreliable_list.size() > min_num_candidates_for_pdlp &&
+                                           pdlp_warm_cache.percent_solved_by_batch_pdlp_at_root >
+                                             min_percent_solved_by_batch_pdlp_at_root_for_pdlp);
 
   if (rb_mode != 0 && !pdlp_warm_cache.populated) {
     log.printf("PDLP warm start data not populated, using DS only\n");
--- a/cpp/src/pdlp/solve.cu
+++ b/cpp/src/pdlp/solve.cu
@@ -934,8 +934,9 @@ optimization_problem_solution_t<i_t, f_t> run_batch_pdlp(
   size_t memory_max_batch_size = max_batch_size;
 
   // Check if we don't hit the limit using max_batch_size
-  const bool collect_solutions  = settings.generate_batch_primal_dual_solution;
-  const double memory_estimate = batch_pdlp_memory_estimator(problem, max_batch_size, collect_solutions);
+  const bool collect_solutions = settings.generate_batch_primal_dual_solution;
+  const double memory_estimate =
+    batch_pdlp_memory_estimator(problem, max_batch_size, collect_solutions);
   size_t st_free_mem, st_total_mem;
   RAFT_CUDA_TRY(cudaMemGetInfo(&st_free_mem, &st_total_mem));
   const double free_mem  = static_cast<double>(st_free_mem);
@@ -951,7 +952,8 @@ optimization_problem_solution_t<i_t, f_t> run_batch_pdlp(
     use_optimal_batch_size = true;
     // Decrement batch size iteratively until we find a batch size that fits
     while (memory_max_batch_size > 1) {
-      const double memory_estimate = batch_pdlp_memory_estimator(problem, memory_max_batch_size, collect_solutions);
+      const double memory_estimate =
+        batch_pdlp_memory_estimator(problem, memory_max_batch_size, collect_solutions);
       if (memory_estimate <= free_mem) { break; }
 #ifdef BATCH_VERBOSE_MODE
       std::cout << "Memory estimate: " << memory_estimate << std::endl;
@@ -962,7 +964,8 @@ optimization_problem_solution_t<i_t, f_t> run_batch_pdlp(
 #endif
       memory_max_batch_size--;
     }
-    const double min_estimate = batch_pdlp_memory_estimator(problem, memory_max_batch_size, collect_solutions);
+    const double min_estimate =
+      batch_pdlp_memory_estimator(problem, memory_max_batch_size, collect_solutions);
     if (min_estimate > free_mem) {
       return optimization_problem_solution_t<i_t, f_t>(pdlp_termination_status_t::NumericalError,
                                                        stream);


anandhkb:
I think this targets the next 26.06 release. @chris-maes @Kh4ster for advise.

cpp/src/branch_and_bound/pseudo_costs.cpp
```diff
@@ -22,6 +27,12 @@ namespace cuopt::linear_programming::dual_simplex {
 
 namespace {
 
+static bool ds_is_valid_done(dual::status_t status)
```
nguidotti: Nitpick: `ds_is_valid_done --> is_dual_simplex_done`

cpp/src/branch_and_bound/pseudo_costs.cpp
```diff
@@ -237,72 +318,116 @@ static cuopt::mps_parser::mps_data_model_t<i_t, f_t> simplex_problem_to_mps_data
...
+//   1. If both found optimal   -> keep DS (higher quality vertex solution)
+//   2. Else if Dual Simplex found infeasible -> declare infeasible
+//   3. Else if one is optimal -> keep the optimal one
+//   4. Else if Dual Simplex hit iteration limit -> keep DS
+//   5. Else if none converged -> NaN (original objective)
+// Return {value, source} where source is 0 if Dual Simplex, 1 if PDLP, 2 if both
 template <typename i_t, typename f_t>
-void strong_branching(const user_problem_t<i_t, f_t>& original_problem,
-                      const lp_problem_t<i_t, f_t>& original_lp,
+static std::pair<f_t, i_t> merge_sb_result(f_t ds_val,
```
nguidotti: Nitpick: use an `enum` instead of a `int` to set the source. Avoid setting an incorrect value in the future and the intent of the return is more clear

cpp/src/branch_and_bound/pseudo_costs.cpp
```diff
@@ -237,72 +318,116 @@ static cuopt::mps_parser::mps_data_model_t<i_t, f_t> simplex_problem_to_mps_data
...
+}
+
+template <typename i_t, typename f_t>
+void strong_branching(const lp_problem_t<i_t, f_t>& original_lp,
                       const simplex_solver_settings_t<i_t, f_t>& settings,
                       f_t start_time,
+                      const std::vector<i_t>& new_slacks,
                       const std::vector<variable_type_t>& var_types,
-                      const std::vector<f_t> root_soln,
+                      const std::vector<f_t>& root_soln,
```
nguidotti: Note that in #979,  I am setting the `root_soln` to `lp_solution` such that `x`, `y` and `z` are already available.

cpp/src/branch_and_bound/pseudo_costs.cpp
```diff
@@ -34,11 +45,19 @@ void strong_branch_helper(i_t start,
                           const std::vector<f_t>& root_soln,
                           const std::vector<variable_status_t>& root_vstatus,
                           const std::vector<f_t>& edge_norms,
-                          pseudo_costs_t<i_t, f_t>& pc)
+                          pseudo_costs_t<i_t, f_t>& pc,
+                          std::vector<f_t>& ds_obj_down,
```
nguidotti: I would avoid calling `dual simplex` as `ds`. We do not use this terminology on the other parts of the code base

nguidotti: Similarly, `strong_branching` vs `sb` or `reliability branching` vs `rb`.

cpp/src/branch_and_bound/pseudo_costs.cpp
```diff
@@ -317,38 +442,162 @@ void strong_branching(const user_problem_t<i_t, f_t>& original_problem,
...
-    settings.log.printf("Batch PDLP strong branching enabled\n");
+  const i_t effective_batch_pdlp = settings.sub_mip ? 0 : settings.mip_batch_pdlp_strong_branching;
 
-    f_t start_batch = tic();
+  settings.log.printf("Strong branching using %d threads and %ld fractional variables\n",
+                      settings.num_threads,
+                      fractional.size());
+
+  // Cooperative DS + PDLP: shared context tracks which subproblems are solved
+  shared_strong_branching_context_t<i_t, f_t> shared_ctx(2 * fractional.size());
```
nguidotti, line 1029: I have a question: why do we need 2 structs here? Can you just pass a pointer or reference to `shared_strong_branching_context_t` instead?

cpp/src/branch_and_bound/pseudo_costs.cpp
```diff
@@ -605,9 +982,41 @@ i_t pseudo_costs_t<i_t, f_t>::reliable_variable_selection(
...
+                        unreliable_list.size() > min_num_candidates_for_pdlp &&
+                        pdlp_warm_cache.pourcent_solved_by_batch_pdlp_at_root >
+                          min_pourcent_solved_by_batch_pdlp_at_root_for_pdlp;
+
+  if (rb_mode != 0 && !pdlp_warm_cache.populated) {
+    log.printf("PDLP warm start data not populated, using DS only\n");
+  } else if (rb_mode != 0 && unreliable_list.size() < min_num_candidates_for_pdlp) {
+    log.printf("Not enough candidates to use batch PDLP, using DS only\n");
+  } else if (rb_mode != 0 && pdlp_warm_cache.pourcent_solved_by_batch_pdlp_at_root < 5.0) {
+    log.printf("Pourcent solved by batch PDLP at root is too low, using DS only\n");
```
nguidotti: Typo: `Pourcent --> Percent`

nguidotti: You need to check for other places as well

cpp/src/branch_and_bound/pseudo_costs.cpp
```diff
@@ -605,9 +982,41 @@ i_t pseudo_costs_t<i_t, f_t>::reliable_variable_selection(
     return branch_var;
   }
 
-  const int num_tasks          = std::max(max_num_tasks, 1);
-  const int task_priority      = reliability_branching_settings.task_priority;
-  const i_t max_num_candidates = reliability_branching_settings.max_num_candidates;
+  const i_t rb_mode = settings.mip_batch_pdlp_reliability_branching;
```
nguidotti, line 1408: I would suggest to move reliability branching with batch PDLP to another routine. Keep the original with just dual simplex

nguidotti: Also, it is important to consider that we are setting the reliability threshold based on the number of dual simplex iterations in reliability branching and B&B. So we may need a way to add the PDLP iterations to the calculation as well

cpp/src/branch_and_bound/pseudo_costs.cpp
```diff
@@ -626,89 +1035,317 @@ i_t pseudo_costs_t<i_t, f_t>::reliable_variable_selection(
...
+  // Shared context for cooperative work-stealing (mode 1)
+  // [0..num_candidates) = down, [num_candidates..2*num_candidates) = up
+  shared_strong_branching_context_t<i_t, f_t> shared_ctx(2 * num_candidates);
+  shared_strong_branching_context_view_t<i_t, f_t> sb_view(std::span(shared_ctx.solved));
+
+  std::vector<f_t> pdlp_obj_down(num_candidates, std::numeric_limits<f_t>::quiet_NaN());
+  std::vector<f_t> pdlp_obj_up(num_candidates, std::numeric_limits<f_t>::quiet_NaN());
+
+  std::atomic<int> concurrent_halt{0};
+  std::thread pdlp_thread;
```
nguidotti: I would avoid creating a `std::thread` here for three reasons:
1.  The entire B&B is using OpenMP, so it is best to use the same model throughout the entire solver. 
2. This code section is executed in parallel by a large quantity of threads, such that we are essentially duplicating the number of executing threads. This is very bad for the performance. 
3. This routine is very frequently called, such that creating and destroying a thread may lead to a significant overhead.

nguidotti: In other words, replace `std::thread` with `omp task`

cpp/src/branch_and_bound/pseudo_costs.cpp
Kh4ster: Thanks I will update that

cpp/src/branch_and_bound/pseudo_costs.cpp
Kh4ster: Agreed

cpp/src/branch_and_bound/pseudo_costs.cpp
Kh4ster: Yes if your PR gets merged first I will use it instead

cpp/src/branch_and_bound/pseudo_costs.cpp
Kh4ster: The original idea for the view version was to have sub_views as in BPDLP I might solve parts of the batch one after another and the sub view would abstract the initial offset

cpp/src/branch_and_bound/pseudo_costs.cpp
Kh4ster: Good catch

cpp/src/branch_and_bound/pseudo_costs.cpp
```diff
@@ -317,89 +744,54 @@ void strong_branching(const user_problem_t<i_t, f_t>& original_problem,
...
+  shared_strong_branching_context_t<i_t, f_t> shared_ctx(2 * fractional.size());
+  shared_strong_branching_context_view_t<i_t, f_t> sb_view(shared_ctx.solved);
+
+  std::atomic<int> concurrent_halt{0};
+
+  std::vector<f_t> pdlp_obj_down(fractional.size(), std::numeric_limits<f_t>::quiet_NaN());
+  std::vector<f_t> pdlp_obj_up(fractional.size(), std::numeric_limits<f_t>::quiet_NaN());
+
+  if (effective_batch_pdlp != 0) {
+#pragma omp task default(shared)
```
nguidotti: The `omp task` and `omp taskwait` must be inside an OpenMP `parallel` region to work. I will send you a patch

Kh4ster: Done!

cpp/src/branch_and_bound/pseudo_costs.cpp
Kh4ster: I moved it to another function. Regarding the iteration discussion it's not trivial as we discussed

cpp/src/branch_and_bound/pseudo_costs.cpp
Kh4ster: Thanks for the help on this, done

python/cuopt_server/cuopt_server/utils/linear_programming/data_definition.py
```diff
@@ -452,8 +452,15 @@ class SolverConfig(BaseModel):
     )
     mip_batch_pdlp_strong_branching: Optional[int] = Field(
         default=0,
-        description="Set 1 to enable batch PDLP strong branching "
-        "in the MIP solver, 0 to disable.",
+        description="Strong branching mode: 0 = Dual Simplex only, "
+        "1 = cooperative work-stealing (DS + batch PDLP), "
+        "2 = batch PDLP only.",
+    )
+    mip_batch_pdlp_reliability_branching: Optional[int] = Field(
```
chris-maes, line 459: Sounds like we aren't adding new parameters to the service.

chris-maes: @tmckayus to check if someone sends the new parameter it is ok.

Iroy30: SolverConfig is not a StrictModel anymore, which means we accept new params even if not added in this file (This is to reduce the maintenance of throwaway service code). That being said there is no harm in adding it. On the python side it now gets automatically added if a C++ CUOPT_MIP_BATCH_PDLP_RELIABILITY_BRANCHING #define exists in constants.h

tmckayus: if we add it though it shows up in the generated API docs, and if we don't implement it, it shouldn't show up there. Let me check a few things ....

tmckayus: Yes, @Iroy30 is correct, I removed earlier comments and approved. Any solver parameters at the C++ level are automatically picked up and passed in solver settings to solve_lp in the server.  For now, it's probably better to add them here in data_definition.py so that they do show up in the API docs -- perhaps in the future we can come up with a general statement so the docs are consistent (we don't want some parameters there, but not all).

Sorry for the confusion, I forgot that we changed the mechanism on this. Thanks @Iroy30 !

PR #1007, opened 2026-03-31
Add a code generator for boilerplate to move data in and out of protobuf messages for grpc

tmckayus:
This change adds a code generator written in Python which generates the protobuf specifications and the code to map between cuopt problem, solution, and settings structures and protobuf messages.  The protoc compiler is still used to build the serialization/deserialziation code and the grpc interface -- only the high-level .proto files are generated from the yaml description.

The code generator reads a lightweight yaml file which describes the cuopt structures and generates mapping functions which are included in the grpc client and server implementations.

When a field is added/changed in the problem, solution, or settings structures, a developer updates the yaml file and re-runs the code generator, then reviews the updated code.

A README.md file in the codegen directory describes the yaml format, defaults, and conventions. As much as possible has been left to defaults or naming conventions to make the yaml easy to maintain.


cpp/src/grpc/server/grpc_server_types.hpp
```diff
@@ -156,7 +156,7 @@ struct JobWaiter {
 // =============================================================================
 
 struct ServerConfig {
-  int port            = 8765;
+  int port            = 5001;
```
rg20, line 159: Is this intentional?

cpp/tests/linear_programming/grpc/CMakeLists.txt
```diff
@@ -16,6 +16,7 @@ target_include_directories(GRPC_CLIENT_TEST
     "${CUOPT_SOURCE_DIR}/include"
     "${CUOPT_SOURCE_DIR}/src/grpc"
     "${CUOPT_SOURCE_DIR}/src/grpc/client"
+    "${CUOPT_SOURCE_DIR}/codegen/generated"
```
rg20, line 19: Move this to `src/grpc/codegen`

ci/verify_codegen.sh
```diff
@@ -0,0 +1,57 @@
...
+# SPDX-License-Identifier: Apache-2.0
+#
+# Verify that committed codegen output matches what generate_conversions.py produces.
+# Fails if a developer edited field_registry.yaml without re-running ./build.sh codegen.
+
+set -euo pipefail
+
+SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
+REPO_DIR="$(cd "${SCRIPT_DIR}/.." && pwd)"
+CODEGEN_DIR="${REPO_DIR}/cpp/codegen"
```
rg20, line 12: codegen is a generic name,  you can call it grpc_code_gen or grpc_gen

Iroy30 (approved):

cpp/src/grpc/server/grpc_server_types.hpp
tmckayus: yes, since the old port is 5000

tmckayus:
superseded by #1107

PR #1035, opened 2026-04-03
Migrate RMM usage to CCCL MR design

bdice:
## Summary

- Remove dependency on `rmm::mr::device_memory_resource` base class; resources now satisfy the `cuda::mr::resource` concept directly
- Replace `shared_ptr<device_memory_resource>` with value types and `cuda::mr::any_resource<cuda::mr::device_accessible>` for type-erased storage
- Replace `set_current_device_resource(ptr)` / `set_per_device_resource(id, ptr)` with `set_current_device_resource_ref` / `set_per_device_resource_ref`
- Remove `make_owning_wrapper` usage and `dynamic_cast` on memory resources (no common base class)
- Add missing `thrust/iterator/transform_output_iterator.h` include (no longer transitively included via CCCL)

Depends on https://github.com/rapidsai/rmm/pull/2361.
Depends on https://github.com/rapidsai/ucxx/pull/636.
Depends on https://github.com/rapidsai/raft/pull/2996.


cpp/src/barrier/sparse_cholesky.cuh
```diff
@@ -247,8 +247,8 @@ class sparse_cholesky_cudss_t : public sparse_cholesky_base_t<i_t, f_t> {
     CUDSS_CALL_AND_CHECK_EXIT(cudssSetStream(handle, stream), status, "cudaStreamCreate");
 
     mem_handler.ctx          = reinterpret_cast<void*>(handle_ptr_->get_workspace_resource());
-    mem_handler.device_alloc = cudss_device_alloc<rmm::mr::device_memory_resource>;
-    mem_handler.device_free  = cudss_device_dealloc<rmm::mr::device_memory_resource>;
+    mem_handler.device_alloc = cudss_device_alloc<void>;
+    mem_handler.device_free  = cudss_device_dealloc<void>;
```
bdice, line 251: It doesn't seem like this template parameter is being used. The `rmm::mr::device_memory_resource` class has been removed, but this change in an unused template parameter should have no effect.

cpp/src/utilities/cuda_helpers.cuh
```diff
@@ -242,25 +240,9 @@ DI void sorted_insert(T* array, T item, int curr_size, int max_size)
...
-           limiting_adaptor->get_allocation_limit() / (double)1e6);
-    printf("used_mem: %fMiB\n", limiting_adaptor->get_allocated_bytes() / (double)1e6);
-    printf("free_mem: %fMiB\n",
-           (limiting_adaptor->get_allocation_limit() - limiting_adaptor->get_allocated_bytes()) /
-             (double)1e6);
-    return std::min(total_mem, limiting_adaptor->get_allocation_limit());
-  } else {
-    return total_mem;
-  }
+  return total_mem;
```
bdice, line 246: We cannot currently perform this kind of dynamic cast with the CCCL memory resource design. There is a CCCL feature that we can adopt soon to re-enable this functionality, which I expect to be available for 26.06. Depending on how the migration goes, I may be able to restore this as soon as next week. I have marked this as a TODO on my side.
```suggestion
  // TODO (bdice): Restore limiting adaptor check after updating CCCL to support resource_cast
  return total_mem;
```

cpp/tests/utilities/base_fixture.hpp
```diff
@@ -13,39 +13,38 @@
...
 }
 
 inline auto make_binning()
 {
   auto pool = make_pool();
   // Add a fixed_size_memory_resource for bins of size 256, 512, 1024, 2048 and
   // 4096KiB Larger allocations will use the pool resource
-  auto mr = rmm::mr::make_owning_wrapper<rmm::mr::binning_memory_resource>(pool, 18, 22);
-  return mr;
+  return rmm::mr::binning_memory_resource(pool, 18, 22);
```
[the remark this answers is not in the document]
bdice: This is wrong, the API is changing and the resource adaptors (binning, pool, etc.) are now owning.

cpp/src/utilities/cuda_helpers.cuh
```diff
@@ -242,25 +240,9 @@ DI void sorted_insert(T* array, T item, int curr_size, int max_size)
...
 }
```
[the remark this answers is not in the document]
bdice: This is probably a good idea but I don't think it is in scope for this PR. I don't want to depend on RAFT macros, we try to make each library define and use its own macros.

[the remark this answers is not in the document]
bdice: Nevermind, I'll just do it in this PR. I see that cuOpt is already using RAFT macros everywhere...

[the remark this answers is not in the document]
bdice: Done in 24d30362c1393da1275759e62df0d10ed11dec03.

akifcorduk (approved):
Approving considering that the limiting adaptor will be restored. Thanks!

cpp/src/barrier/sparse_cholesky.cuh
hlinsen: I confirm, it is not used

cpp/src/pdlp/termination_strategy/infeasibility_information.cu
```diff
@@ -15,6 +15,8 @@
 
 #include <mip_heuristics/mip_constants.hpp>
 
+#include <thrust/iterator/transform_output_iterator.h>
```
hlinsen, line 18: Is it an intentional change?

bdice: Yes, it is used here: https://github.com/bdice/cuopt/blob/24d30362c1393da1275759e62df0d10ed11dec03/cpp/src/pdlp/termination_strategy/infeasibility_information.cu#L354

I believe this was added to fix a missing header, though it's been a while since I did this. The large changes in headers probably meant that it wasn't being included transitively anymore, but we shouldn't rely on transitive includes anyway.

rgsl888prabhu:
@akifcorduk @hlinsen May I get your another round of the review ?

PR #1043, opened 2026-04-07
Dual simplex performance improvements

chris-maes:
This PR includes four performance improvements to dual simplex
- Removes an unnecessary BTRAN for computing the steepest edge weight of the entering variable
- Rewrites the right-looking LU factorization to use remove linked lists and use a sparse row and column representation of the Schur complement matrix.  
- Swaps the coefficients in the row representation of A, so that all nonbasic columns appear first. This speeds up computing the change in reduced costs. 
- Record the amount of work in the solve and refactorization, and trigger refactorization if solve work exceeds factorization work. 

Accuracy for dual simplex is also improved by
- Refactoring the basis when optimal and the primal residual || A x - b ||_inf is large and basis updates are present. 

There is a 16% (geomean) performance improvement in solve times on the NETLIB LP test set. 
We are now faster than HSOL (the precursor to HiGHS) on NETLIB.
There is a 6% (geomean) performance improvement on time to solve MIPLIB LP relaxations.


cpp/src/branch_and_bound/pseudo_costs.hpp
```diff
@@ -526,7 +526,7 @@ class pseudo_costs_t {
 
   reliability_branching_settings_t<i_t, f_t> reliability_branching_settings;
 
-  csc_matrix_t<i_t, f_t> AT;  // Transpose of the constraint matrix A
+  csr_matrix_t<i_t, f_t> Arow;
```
mlubin, line 216: Is it an invariant that all the nonbasic coefficients appear first, or only in some contexts? Consider documenting. I'm also wondering if there's scope for a helper struct to group `Arow` with `nonbasic_end` since `nonbasic_end` is a property of the stored representation of the matrix.

cpp/src/dual_simplex/phase2.cpp
```diff
@@ -1480,15 +1571,27 @@ i_t compute_steepest_edge_norm_entering(const simplex_solver_settings_t<i_t, f_t
                                         const basis_update_mpf_t<i_t, f_t>& ft,
                                         i_t basic_leaving_index,
                                         i_t entering_index,
+                                        i_t leaving_index,
                                         std::vector<f_t>& steepest_edge_norms)
 {
+#if 0
```
mlubin: Why not delete this code block?

cpp/src/branch_and_bound/pseudo_costs.hpp
chris-maes: In order to use the compute_delta_z function the nonbasic coefficients must appear first. I can add some documentation and consider grouping with a struct.

cpp/src/dual_simplex/phase2.cpp
chris-maes: Yep. This code will all be deleted as will the function `compute_steepest_edge_norm_entering` before merging.

cpp/src/branch_and_bound/pseudo_costs.cpp
```diff
@@ -1476,25 +1473,23 @@ i_t pseudo_costs_t<i_t, f_t>::reliable_variable_selection(
...
-                                                   AT,
+                                                   Arow,
                                                    node_ptr->vstatus,
                                                    j,
                                                    basic_map[j],
                                                    leaf_solution,
                                                    worker->basic_list,
                                                    worker->nonbasic_list,
-                                                   nonbasic_mark,
+                                                   nonbasic_end,
```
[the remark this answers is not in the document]
chris-maes: Ah good catch. Probably best to keep local copies of Arow for each worker.

cpp/src/dual_simplex/phase2.cpp
```diff
@@ -130,42 +130,37 @@ void compute_reduced_cost_update(const lp_problem_t<i_t, f_t>& lp,
...
+void compute_delta_z(const csr_matrix_t<i_t, f_t>& Arow,
                      const sparse_vector_t<i_t, f_t>& delta_y,
                      i_t leaving_index,
                      i_t direction,
-                     const std::vector<i_t>& nonbasic_mark,
+                     const std::vector<i_t>& nonbasic_end,
                      std::vector<i_t>& delta_z_mark,
                      std::vector<i_t>& delta_z_indices,
                      std::vector<f_t>& delta_z,
                      f_t& work_estimate)
```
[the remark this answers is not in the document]
chris-maes: Ok. I can remove the dead debug code.

chris-maes:
@coderabbitai resume

cpp/src/dual_simplex/phase2.cpp
```diff
@@ -3463,22 +3604,35 @@ dual::status_t dual_phase2_with_advanced_basis(i_t phase,
     // Refactor or update the basis factorization
     {
       PHASE2_NVTX_RANGE("DualSimplex::basis_update");
-      bool should_refactor = ft.num_updates() > settings.refactor_frequency;
+      iterations_since_refactor++;
+      bool should_refactor =
+        (ft.num_updates() > 100 && solve_work > refactor_work) || (ft.num_updates() > 1000);
```
nguidotti: Maybe move the hard-coded values  `100` and `1000` to the settings

nguidotti (approved):
Looks good to me! Thanks for the improving dual simplex, Chris. Since we now using `Arow` everywhere, should it not be stored in `lp_problem_t` and computed at the start of the solver? So we can avoid re-computing it for every call to phase2 and it will be more easily accessible to all parts of the code

chris-maes:
> Looks good to me! Thanks for the improving dual simplex, Chris. Since we now using `Arow` everywhere, should it not be stored in `lp_problem_t` and computed at the start of the solver? So we can avoid re-computing it for every call to phase2 and it will be more easily accessible to all parts of the code

I think we've discussed this multiple times before. I don't think we should store redundant information in `lp_problem_t`. It cost us O(m + nnz) to compute Arow. This is less than the cost of a single dual simplex iteration. The tradeoff of increased code complexity vs run-time does not seem worth it.

cpp/src/dual_simplex/phase2.cpp
chris-maes: Done

mlubin:
My codex claims to have found an issue, pasting it here while I look more closely:

right_looking_lu2 appears to leave col_counts_ inconsistent after pivot removal.

The sequence is:

1. update_for_pivot_removal() logically removes the pivot row from every column in the pivot row by decrementing each affected column count.

2. schur_complement() then updates the same columns and recomputes the column bucket from physical storage:

```
i_t new_cdeg = col_end_[j] - col_start_[j];
if (new_cdeg != col_counts_.get_count(j)) {
  col_counts_.update_count(j, new_cdeg);
}
```

3. But the pivot-row entry is still physically present at that point. It is only removed afterward in remove_pivot_row_and_column().

So for every non-pivot column in the pivot row, new_cdeg includes the pivot-row entry that was already logically removed. In the no-fill/no-cancel case, a column with degree d is first moved to bucket d - 1, then Step 2d moves it back to bucket d, and the later physical removal does not update col_counts_.

This makes col_counts_ one too high for affected columns. markowitz_search() then searches the wrong degree buckets and computes Markowitz costs using stale column degrees, which can change pivot order and hurt both performance and numerical behavior.

A fix would be to either update the column bucket only after remove_pivot_row_and_column(), or keep Step 2d before physical removal but subtract the still-present pivot-row entry when computing new_cdeg for columns in the pivot row.

chris-maes:
> My codex claims to have found an issue, pasting it here while I look more closely:
> 
> right_looking_lu2 appears to leave col_counts_ inconsistent after pivot removal.
> 
> The sequence is:
> 
> 1. update_for_pivot_removal() logically removes the pivot row from every column in the pivot row by decrementing each affected column count.
> 2. schur_complement() then updates the same columns and recomputes the column bucket from physical storage:
> 
> ```
> i_t new_cdeg = col_end_[j] - col_start_[j];
> if (new_cdeg != col_counts_.get_count(j)) {
>   col_counts_.update_count(j, new_cdeg);
> }
> ```
> 
> 3. But the pivot-row entry is still physically present at that point. It is only removed afterward in remove_pivot_row_and_column().
> 
> So for every non-pivot column in the pivot row, new_cdeg includes the pivot-row entry that was already logically removed. In the no-fill/no-cancel case, a column with degree d is first moved to bucket d - 1, then Step 2d moves it back to bucket d, and the later physical removal does not update col_counts_.
> 
> This makes col_counts_ one too high for affected columns. markowitz_search() then searches the wrong degree buckets and computes Markowitz costs using stale column degrees, which can change pivot order and hurt both performance and numerical behavior.
> 
> A fix would be to either update the column bucket only after remove_pivot_row_and_column(), or keep Step 2d before physical removal but subtract the still-present pivot-row entry when computing new_cdeg for columns in the pivot row.

Ok. Thanks. Let me check to see if there is a mismatch in col_counts_ versus col_end_ - col_start. I can probably remove update_for_pivot_removal and defer updating col_counts_ until remove_pivot_row_and_column. 

In practice this is a big performance improvement and there are no numerical issues. But it's possible that in certain cases we could have a off-by-one error.

mlubin:
The proposed patch is:

```patch
diff --git a/cpp/src/dual_simplex/right_looking_lu.cpp b/cpp/src/dual_simplex/right_looking_lu.cpp
index a4640a4b..851c0f6c 100644
--- a/cpp/src/dual_simplex/right_looking_lu.cpp
+++ b/cpp/src/dual_simplex/right_looking_lu.cpp
@@ -399,8 +399,12 @@ class trailing_matrix_t {
         i_t old_count = col_end_[j] - col_start_[j];
         col_end_[j]   = new_end;
         i_t new_count = new_end - col_start_[j];
-        // Update column degree for cancellations
-        if (new_count != old_count) { col_counts_.update_count(j, new_count); }
+        // Update column degree for cancellations.
+        // Subtract 1 because the pivot-row entry is still physically present
+        // in column j but has already been logically removed by
+        // update_for_pivot_removal(); it will be physically removed by
+        // remove_pivot_row_and_column().
+        if (new_count != old_count) { col_counts_.update_count(j, new_count - 1); }
       }

       // Step 2c: Insert fill-in entries. We know exactly how many there are.
@@ -440,8 +444,11 @@ class trailing_matrix_t {
       }

       // Step 2d: Update column degree bucket once for this column.
+      // Subtract 1 because the pivot-row entry is still physically present in
+      // column j but has been logically removed by update_for_pivot_removal()
+      // and will be physically removed by remove_pivot_row_and_column().
       {
-        i_t new_cdeg = col_end_[j] - col_start_[j];
+        i_t new_cdeg = (col_end_[j] - col_start_[j]) - 1;
         if (new_cdeg != col_counts_.get_count(j)) { col_counts_.update_count(j, new_cdeg); }
       }
```

chris-maes:
> The proposed patch is:
> 
> ```diff
> diff --git a/cpp/src/dual_simplex/right_looking_lu.cpp b/cpp/src/dual_simplex/right_looking_lu.cpp
> index a4640a4b..851c0f6c 100644
> --- a/cpp/src/dual_simplex/right_looking_lu.cpp
> +++ b/cpp/src/dual_simplex/right_looking_lu.cpp
> @@ -399,8 +399,12 @@ class trailing_matrix_t {
>          i_t old_count = col_end_[j] - col_start_[j];
>          col_end_[j]   = new_end;
>          i_t new_count = new_end - col_start_[j];
> -        // Update column degree for cancellations
> -        if (new_count != old_count) { col_counts_.update_count(j, new_count); }
> +        // Update column degree for cancellations.
> +        // Subtract 1 because the pivot-row entry is still physically present
> +        // in column j but has already been logically removed by
> +        // update_for_pivot_removal(); it will be physically removed by
> +        // remove_pivot_row_and_column().
> +        if (new_count != old_count) { col_counts_.update_count(j, new_count - 1); }
>        }
> 
>        // Step 2c: Insert fill-in entries. We know exactly how many there are.
> @@ -440,8 +444,11 @@ class trailing_matrix_t {
>        }
> 
>        // Step 2d: Update column degree bucket once for this column.
> +      // Subtract 1 because the pivot-row entry is still physically present in
> +      // column j but has been logically removed by update_for_pivot_removal()
> +      // and will be physically removed by remove_pivot_row_and_column().
>        {
> -        i_t new_cdeg = col_end_[j] - col_start_[j];
> +        i_t new_cdeg = (col_end_[j] - col_start_[j]) - 1;
>          if (new_cdeg != col_counts_.get_count(j)) { col_counts_.update_count(j, new_cdeg); }
>        }
> ```

Thanks. I can confirm the issue. I'm making several other (different) changes.

mlubin:
Nice, glad my vibe review was useful!

PR #1051, opened 2026-04-07
Extend the barrier solver to SOCP problems

yan-zaretskiy:
This is a draft PR that extends the 2x2 augmented system formulation of the barrier solver to solve SOCP problems. It supports both explicit cone variables as well as the cone rows in the Ax=b constraint, both of which can be specified in the CBF format. It uses Nesterov-Todd scaling to make the conic diagonal blocks symmetric.


python/cuopt/cuopt/linear_programming/problem.py
```diff
@@ -831,13 +836,105 @@ def __truediv__(self, other):
...
+                raise ValueError(
+                    "Can't compare QuadraticExpression with type %s"
+                    % type(other).__name__
+                )
 
     def __eq__(self, other):
-        raise Exception("Quadratic constraints not supported")
+        raise ValueError(
+            "Quadratic equality constraints are not supported; "
+            "use addQuadraticConstraint with QCMATRIX data or two inequality constraints."
```
chris-maes: I would just say that equality constraints are not supported.

python/cuopt/cuopt/linear_programming/problem.py
```diff
@@ -831,13 +836,105 @@ def __truediv__(self, other):
...
+        linear_values,
+        linear_indices,
+        quadratic_values,
+        quadratic_row_indices,
+        quadratic_col_indices,
+        rhs_value,
+    )
+
+
+class QuadraticConstraint:
```
chris-maes: We should not need an extra `QuadraticConstraint` class. Can this be kept in the Constraint class?  Or can you do this within the Problem class?

python/cuopt/cuopt/linear_programming/problem.py
```diff
@@ -1557,6 +1672,30 @@ def addConstraint(self, constr, name=""):
                 raise ValueError("addConstraint requires a Constraint object")
         return constr
 
+    def addQuadraticConstraint(self, constr, name=""):
```
chris-maes: We should not have a special `addQuadraticConstraint` method. These should be handled directly by `addConstraint`

python/cuopt/cuopt/linear_programming/problem.py
```diff
@@ -831,13 +836,105 @@ def __truediv__(self, other):
                 )
 
     def __le__(self, other):
-        raise Exception("Quadratic constraints not supported")
```
chris-maes, line 837: You likely need to handle __le__ and __ge__ in Variable and LinearExpression as well.

python/cuopt/cuopt/linear_programming/problem.py
```diff
@@ -831,13 +841,91 @@ def __truediv__(self, other):
...
+    linear_indices = []
+    linear_values = []
+    for idx in sorted(linear_coeff):
+        coeff = linear_coeff[idx]
+        if coeff != 0.0:
+            linear_indices.append(idx)
+            linear_values.append(coeff)
+
+    quad_coeff = {}
+    for var1, var2, coeff in zip(expr.qvars1, expr.qvars2, expr.qcoefficients):
```
Iroy30, line 902: we can use something similar to how we do it in setObjective
```
qrows = [var.getIndex() for var in expr.qvars1]
qcols = [var.getIndex() for var in expr.qvars2]
self.objective_qmatrix = coo_matrix(
    (
        np.array(expr.qcoefficients),
        (np.array(qrows), np.array(qcols)),
    ),
    shape=(self.NumVariables, self.NumVariables),
)
if expr.qmatrix is not None:
    self.objective_qmatrix += expr.qmatrix
self.objective_qmatrix = self.objective_qmatrix.tocsr()
```
scipy coo takes care of repeated entries and also can add expr.qmatrix to then get the csr needed.
Just a suggestion. Besides that, looks good! Thanks

yuwenchen95: I had a review and decide to keep it unchanged as we expect coo input form for quadratic constraints in C++. Quadratic cost in a quadratic constraint is assumed to be sparse that won't degrade performance compared to scipy's coo transformation.

cpp/src/dual_simplex/presolve.cpp
```diff
@@ -1177,7 +1182,7 @@ i_t presolve(const lp_problem_t<i_t, f_t>& original,
 
     auto old_objective = problem.objective;
     if (problem.Q.n > 0) {
-      for (i_t row = 0; row < problem.num_cols; row++) {
+      for (i_t row = 0; row < linear_cols; row++) {
```
chris-maes, line 1249: This is looping over rows, but your upper bound is based on columns? Is this a bug?

yuwenchen95: It loops over a CSR Q, where Q is symmetric that row and column are exchangable.

cpp/include/cuopt/linear_programming/pdlp/solver_settings.hpp
```diff
@@ -280,6 +280,9 @@ class pdlp_solver_settings_t {
   i_t dualize{-1};
   i_t ordering{-1};
   i_t barrier_dual_initial_point{-1};
+  f_t barrier_relative_feasibility_tolerance{1.0e-8};
```
chris-maes, line 283: Please remove these. We should not have separate tolerances for barrier. 

We already have relative tolerances that the user can set. (It's another question about how these tolerances should effect barrier.)

cpp/include/cuopt/linear_programming/constants.h
```diff
@@ -47,6 +47,10 @@
 #define CUOPT_DUALIZE                              "dualize"
 #define CUOPT_ORDERING                             "ordering"
 #define CUOPT_BARRIER_DUAL_INITIAL_POINT           "barrier_dual_initial_point"
+#define CUOPT_BARRIER_RELATIVE_FEASIBILITY_TOLERANCE "barrier_relative_feasibility_tolerance"
```
chris-maes, line 50: Please remove. I don't think we want separate tolerances for each method.

cpp/src/dual_simplex/presolve.cpp
```diff
@@ -16,37 +16,35 @@
...
 {
   constexpr bool verbose = false;
   if (verbose) { printf("Removing %d empty columns\n", num_empty_cols); }
-  // We have a variable x_j that does not appear in any rows
-  // The cost function
-  // sum_{k != j} c_k * x_k + c_j * x_j
-  // becomes
-  // sum_{k != j} c_k * x_k + c_j * l_j if c_j > 0
-  // or
-  // sum_{k != j} c_k * x_k + c_j * u_j if c_j < 0
```
chris-maes, line 36: Please restore comment. The agents have a bad habit of deleting helpful comments. Please ask the agent to go through the PR and not delete any comments unless they are incorrect.

cpp/src/dual_simplex/presolve.cpp
```diff
@@ -16,37 +16,35 @@
...
   presolve_info.removed_variables.reserve(num_empty_cols);
   presolve_info.removed_values.reserve(num_empty_cols);
   presolve_info.removed_reduced_costs.reserve(num_empty_cols);
 
-  // Check to see if a variable participates in a quadratic objective
```
chris-maes, line 41: Please restore or modify comment

cpp/src/dual_simplex/presolve.cpp
```diff
@@ -16,37 +16,35 @@
...
   std::vector<bool> has_quadratic_term(problem.num_cols, false);
+  i_t linear_cols = linear_var_count(problem);
 
   if (problem.Q.n > 0) {
-    for (i_t j = 0; j < problem.num_cols; ++j) {
+    for (i_t j = 0; j < linear_cols; ++j) {
       const i_t row_start = problem.Q.row_start[j];
       const i_t row_end   = problem.Q.row_start[j + 1];
       if (row_end - row_start == 0) { continue; }
-      // Q is symmetric, so its sufficient to check only the row size
```
chris-maes, line 49: Please restore comment

cpp/src/dual_simplex/presolve.cpp
```diff
@@ -808,8 +923,8 @@ void convert_user_problem(const user_problem_t<i_t, f_t>& user_problem,
     problem.Q.x = user_problem.Q_values;
   }
 
-  // Add artifical variables
-  if (!settings.barrier_presolve) {
+  // Artificial vars break [linear | cone]; only for barrier_presolve-off LP-style layouts.
```
chris-maes, line 926: Change this comment to
// Add artificial variables for dual simplex

cpp/src/dual_simplex/presolve.cpp
```diff
@@ -808,8 +923,8 @@ void convert_user_problem(const user_problem_t<i_t, f_t>& user_problem,
...
+  if (!settings.barrier_presolve && problem.second_order_cone_dims.empty()) {
```
chris-maes, line 927: If you have second_order_cones barrier presolve has to be true. Please change this to 
`if (!settings.barrier_presolve)`

cpp/src/dual_simplex/presolve.cpp
```diff
@@ -821,21 +936,25 @@ i_t presolve(const lp_problem_t<i_t, f_t>& original,
...
-  // Check for free variables
+
+  // Check for free variables (linear block only; cone columns are handled by the barrier SOC layout)
   i_t free_variables = 0;
-  for (i_t j = 0; j < problem.num_cols; j++) {
+  for (i_t j = 0; j < linear_cols; j++) {
     if (problem.lower[j] == -inf && problem.upper[j] == inf) { free_variables++; }
   }
 
+  // Barrier presolve phase 1: bound free linear variables from equality rows.
```
chris-maes, line 949: Please remove these phase X comments. We don't have numbered phases in presolve. 

Fine to add a comment describing what is happening, e.g
```
// Bound free linear variables from equality rows
```
But please don't introduce numbered phases that will quickly get out of date or require updating.

cpp/src/dual_simplex/presolve.cpp
```diff
@@ -859,10 +978,30 @@ i_t presolve(const lp_problem_t<i_t, f_t>& original,
     std::vector<f_t> upper_bound_coefficient(problem.num_cols, 0.0);
 
     if (!constraints_to_check.empty()) {
-      // Check if the constraints are feasible
```
chris-maes, line 862: Please restore comment

cpp/src/dual_simplex/presolve.cpp
```diff
@@ -859,10 +978,30 @@ i_t presolve(const lp_problem_t<i_t, f_t>& original,
...
       csr_matrix_t<i_t, f_t> Arow(0, 0, 0);
       problem.A.to_compressed_row(Arow);
 
+      // Keep only rows safe for phase-1 bound inference: no cone columns, exactly one free linear.
```
chris-maes, line 984: Please change to
```
// Keep only rows safe for bound inference. Those that contain no cone columns and exactly one free linear variable
```

cpp/src/dual_simplex/presolve.cpp
```diff
@@ -859,10 +978,30 @@ i_t presolve(const lp_problem_t<i_t, f_t>& original,
...
+      {
```
chris-maes, line 985: We should only do this loop when we have cones right?

cpp/src/dual_simplex/presolve.cpp
```diff
@@ -859,10 +978,30 @@ i_t presolve(const lp_problem_t<i_t, f_t>& original,
...
+        std::vector<i_t> safe_constraints;
+        safe_constraints.reserve(constraints_to_check.size());
+        for (i_t i : constraints_to_check) {
+          bool touches_cone        = false;
+          i_t free_linear_in_row = 0;
```
chris-maes, line 990: Remove the free_linear_in_row tracking and checking. 

The current code is designed to exploit the fact that if we bound variables during the loop, we can bound more variables. This destroys that by artificially limiting the number of constraints we look at.

cpp/src/dual_simplex/presolve.cpp
```diff
@@ -929,6 +1069,7 @@ i_t presolve(const lp_problem_t<i_t, f_t>& original,
         // x_j <= 1/a_ij * (rhs - upper_activity_i)
         const i_t j         = last_free_i;
         const f_t a_ij      = last_free_coeff_i;
+        if (a_ij == 0) { continue; }
```
chris-maes, line 1072: Unnecessary as we have removed all 0 coefficients from the matrix.

cpp/src/dual_simplex/presolve.cpp
```diff
@@ -986,6 +1127,9 @@ i_t presolve(const lp_problem_t<i_t, f_t>& original,
           problem.lower[j] = -inf;
         }
       }
+      if (!(problem.lower[j] == -inf && problem.upper[j] == inf)) {
+        presolve_info.phase1_bounded_linear_indices.push_back(j);
```
chris-maes, line 1131: We should not have a variable named `phase_1_bounded_linear_indices`

cpp/src/dual_simplex/presolve.cpp
```diff
@@ -1016,11 +1160,12 @@ i_t presolve(const lp_problem_t<i_t, f_t>& original,
     }
   }
 
+  // Barrier presolve phase 2: negate one-sided bounds (-inf < x <= u -> -u <= x < inf).
```
chris-maes, line 1163: Remove "Barrier presolve phase 2" from comment.

cpp/src/dual_simplex/presolve.cpp
```diff
@@ -1066,10 +1211,10 @@ i_t presolve(const lp_problem_t<i_t, f_t>& original,
     }
   }
 
-  // The original problem may have nonzero lower bounds
-  // 0 != l_j <= x_j <= u_j
+  // Barrier presolve phase 3: shift nonzero lower bounds to zero (cone/stack columns not shifted).
```
chris-maes, line 1214: Remove "Barrier phase 3". Put back the original comment that was deleted.
Fine to add something about "Cone variables should not be shifted"

cpp/src/dual_simplex/presolve.cpp
chris-maes: Probably best to add your reply as a comment.

cpp/src/dual_simplex/presolve.cpp
```diff
@@ -1142,7 +1287,7 @@ i_t presolve(const lp_problem_t<i_t, f_t>& original,
     }
   }
 
-  // Check for empty rows
+  // Barrier presolve phase 4: remove empty rows and empty linear columns.
```
chris-maes, line 1290: Remove "Barrier presolve phase 4"

cpp/src/dual_simplex/presolve.cpp
```diff
@@ -1169,19 +1314,33 @@ i_t presolve(const lp_problem_t<i_t, f_t>& original,
...
+  }
   problem.Q.check_matrix("Before free variable expansion");
 
-  // For QPs, keep free variables as-is rather than
-  // splitting x = v - w. The barrier solver handles them natively with a
-  // static regularizer on the diagonal instead of z/x complementarity terms.
-  if (settings.barrier_presolve && free_variables > 0 && problem.Q.n > 0) {
-    presolve_info.free_variable_indices.clear();
-    for (i_t j = 0; j < problem.num_cols; j++) {
+  // Barrier presolve phase 5: free linear variables — 5a native (QP/SOCP) or 5b v-w split (LP).
```
chris-maes, line 1324: Remove "Barrier presolve phase 5" comment. And "5a, 5b"

cpp/src/dual_simplex/presolve.cpp
```diff
@@ -1169,19 +1314,33 @@ i_t presolve(const lp_problem_t<i_t, f_t>& original,
...
+  if (register_native_free_linear) {
+    presolve_info.free_variable_pairs.clear();
+    presolve_info.native_free_linear_indices.clear();
+    // Only linear decision variables can be "native" free variables; cone/stack columns
+    // are unbounded by construction and must not be counted here.
+    i_t native_free_count = 0;
+    for (i_t j = 0; j < linear_cols; j++) {
       if (problem.lower[j] == -inf && problem.upper[j] == inf) {
-        presolve_info.free_variable_indices.push_back(j);
+        presolve_info.native_free_linear_indices.push_back(j);
```
chris-maes, line 1335: Why are we changing from `free_variable_indices` to `native_free_linear_indices`? Let's keep the old name

cpp/src/dual_simplex/presolve.cpp
```diff
@@ -1169,19 +1314,33 @@ i_t presolve(const lp_problem_t<i_t, f_t>& original,
...
+        native_free_count++;
       }
     }
-  } else if (settings.barrier_presolve && free_variables > 0) {
+    settings.log.printf(
+      "Keeping %d native free linear variables for augmented-system barrier (QP/SOCP)\n",
```
chris-maes, line 1340: "Keeping %d free variables"

cpp/src/dual_simplex/presolve.cpp
```diff
@@ -1169,19 +1314,33 @@ i_t presolve(const lp_problem_t<i_t, f_t>& original,
...
+      native_free_count);
+  } else if (settings.barrier_presolve && !has_cones && free_variables > 0) {
+    // Phase 5b: x_j = v - w with v, w >= 0 (LP without cones; SOCP/QP use phase 5a).
```
chris-maes, line 1343: Remove "Phase 5" and "phase 5a" from comment. You could delete this comment.

cpp/src/dual_simplex/presolve.cpp
```diff
@@ -1194,123 +1353,163 @@ i_t presolve(const lp_problem_t<i_t, f_t>& original,
     // becomes
     // sum_{k != j} c_k x_k + c_j v - c_j w
 
-    std::vector<i_t> pair_index(problem.num_cols, -1);
```
chris-maes, line 1197: Why does any of this code need to change. This is for the LP case. Just make sure we aren't in an SOCP or QP and then use this code. I'd suggest reverting these changes.

cpp/src/dual_simplex/presolve.cpp
```diff
@@ -1194,123 +1353,163 @@ i_t presolve(const lp_problem_t<i_t, f_t>& original,
...
+          }
+          if (partner_row != -1) {
+            Q_j[row_starts[partner_row]] = new_col;
+            Q_x[row_starts[partner_row]] = -qij;
+            row_starts[partner_row]++;
+            if (partner_col != -1) {
+              Q_j[row_starts[partner_row]] = partner_col;
+              Q_x[row_starts[partner_row]] = qij;
+              row_starts[partner_row]++;
+            }
```
chris-maes, line 1512: Please revert all changes from line 1196/1135 to here.

cpp/src/dual_simplex/presolve.cpp
```diff
@@ -1194,123 +1353,163 @@ i_t presolve(const lp_problem_t<i_t, f_t>& original,
...
-    i_t col        = problem.num_cols;
-    for (i_t j = 0; j < problem.num_cols; j++) {
-      if (problem.lower[j] == -inf && problem.upper[j] == inf) {
-        for (i_t p = problem.A.col_start[j]; p < problem.A.col_start[j + 1]; p++) {
-          i_t i          = problem.A.i[p];
-          f_t aij        = problem.A.x[p];
-          problem.A.i[q] = i;
-          problem.A.x[q] = -aij;
-          q++;
+    csc_matrix_t<i_t, f_t> expanded_A(problem.A.m, num_cols, nnz);
```
chris-maes, line 1389: Can you do this expansion only in the SOCP case?

cpp/src/dual_simplex/presolve.cpp
```diff
@@ -1754,6 +1981,11 @@ void uncrush_solution(const presolve_info_t<i_t, f_t>& presolve_info,
     }
   }
 
+  if (!presolve_info.phase1_bounded_linear_indices.empty()) {
+    settings.log.printf("Post-solve: %d linear column(s) had phase-1 bound tightening (x unchanged)\n",
```
chris-maes, line 1985: The user should not be expected to know the internal phases of presolve!

Remove this print and this block of code entirely.

cpp/src/dual_simplex/presolve.cpp
```diff
@@ -1677,13 +1892,25 @@ void uncrush_solution(const presolve_info_t<i_t, f_t>& presolve_info,
   if (num_free_variables > 0) {
     settings.log.printf("Post-solve: Handling free variables %d\n", num_free_variables);
     // We added free variables so we need to map the crushed solution back to the original variables
+    std::vector<bool> remove_partner(input_x.size(), false);
     for (i_t k = 0; k < 2 * num_free_variables; k += 2) {
       const i_t u = free_variable_pairs[k];
       const i_t v = free_variable_pairs[k + 1];
       input_x[u] -= input_x[v];
+      remove_partner[v] = true;
```
chris-maes, line 1900: Why is this code needed?

cpp/src/dual_simplex/presolve.cpp
```diff
@@ -1772,6 +2004,8 @@ void uncrush_solution(const presolve_info_t<i_t, f_t>& presolve_info,
     settings.log.printf("Post-solve: Correcting duals for %d bounded free variables\n",
                         static_cast<i_t>(presolve_info.bounded_free_variables.size()));
     const csc_matrix_t<i_t, f_t>& A = original_problem.A;
+    csr_matrix_t<i_t, f_t> A_row(A.m, A.n, A.nnz());
```
chris-maes, line 2007: Revert these changes. Converting to A_row is not needed

cpp/src/dual_simplex/presolve.cpp
```diff
@@ -1781,15 +2015,9 @@ void uncrush_solution(const presolve_info_t<i_t, f_t>& presolve_info,
       if (w_j == 0.0) { continue; }
       const f_t du = w_j / bfv.coefficient;
       input_y[bfv.constraint] += du;
-      for (i_t j = 0; j < A.n; j++) {
```
chris-maes, line 1784: Revert these changes.

cpp/src/dual_simplex/presolve.hpp
```diff
@@ -162,10 +164,14 @@ struct presolve_info_t {
 
   // Variables that were negated to handle -inf < x_j <= u_j
   std::vector<i_t> negated_variables;
+
+  // Free variable indices for QP/SOCP augmented barrier (not split, handled natively)
+  std::vector<i_t> native_free_linear_indices;
+
+  // Linear columns where phase 1 tightened a fully free var to a one-sided bound (same x index)
+  std::vector<i_t> phase1_bounded_linear_indices;
```
chris-maes, line 172: Revert this change. Go back to `free_variable_indices`

cpp/src/dual_simplex/simplex_solver_settings.hpp
```diff
@@ -79,7 +79,7 @@ struct simplex_solver_settings_t {
       use_left_looking_lu(false),
       eliminate_singletons(true),
       print_presolve_stats(true),
-      barrier_presolve(false),
+      barrier_presolve(true),
```
chris-maes, line 82: Revert this change. barrier_presolve defaults to false. We only set it to true when we are using barrier. You need to explicitly set barrier_presolve to true when solving a QP or SOCP.

cpp/src/dual_simplex/simplex_solver_settings.hpp
```diff
@@ -163,7 +163,7 @@ struct simplex_solver_settings_t {
     use_left_looking_lu;  // true to use left looking LU factorization, false to use right looking
   bool eliminate_singletons;  // true to eliminate singletons from the basis
   bool print_presolve_stats;  // true to print presolve stats
-  bool barrier_presolve;      // true to use barrier presolve
+  bool barrier_presolve;  // bound shifts, free-var bounding; LP v-w split / folding when no cones
```
chris-maes, line 166: Revert the change in the comment

cpp/src/math_optimization/solver_settings.cu
```diff
@@ -98,6 +98,9 @@ solver_settings_t<i_t, f_t>::solver_settings_t() : pdlp_settings(), mip_settings
     {CUOPT_MIP_INTEGRALITY_TOLERANCE, &mip_settings.tolerances.integrality_tolerance, f_t(0.0), f_t(1e-1), f_t(1e-5)},
     {CUOPT_MIP_ABSOLUTE_GAP, &mip_settings.tolerances.absolute_mip_gap, f_t(0.0), std::numeric_limits<f_t>::infinity(), std::max(f_t(1e-10), std::numeric_limits<f_t>::epsilon())},
     {CUOPT_MIP_RELATIVE_GAP, &mip_settings.tolerances.relative_mip_gap, f_t(0.0), f_t(1e-1), f_t(1e-4)},
+    {CUOPT_BARRIER_RELATIVE_FEASIBILITY_TOLERANCE, &pdlp_settings.barrier_relative_feasibility_tolerance, f_t(0.0), f_t(1e-1), f_t(1e-8)},
```
chris-maes, line 101: Revert these changes that add unnecessary parameters

cpp/src/pdlp/solve.cu
```diff
@@ -507,9 +507,14 @@ run_barrier(dual_simplex::user_problem_t<i_t, f_t>& user_problem,
   barrier_settings.barrier                         = true;
   barrier_settings.crossover                       = settings.crossover;
   barrier_settings.eliminate_dense_columns         = settings.eliminate_dense_columns;
-  barrier_settings.barrier_iterative_refinement    = settings.barrier_iterative_refinement;
-  barrier_settings.barrier_step_scale              = settings.barrier_step_scale;
+  barrier_settings.barrier_iterative_refinement = settings.barrier_iterative_refinement;
+  barrier_settings.barrier_step_scale           = settings.barrier_step_scale;
   barrier_settings.cudss_deterministic             = settings.cudss_deterministic;
+  barrier_settings.barrier_relative_feasibility_tol =
```
chris-maes, line 513: Revert these changes

cpp/src/pdlp/solve.cu
```diff
@@ -1749,6 +1754,10 @@ optimization_problem_solution_t<i_t, f_t> solve_qp(optimization_problem_t<i_t, f
     auto qp_timer = cuopt::timer_t(settings.time_limit);
 
     raft::common::nvtx::range fun_scope("Running QP solver");
+    if (op_problem.has_quadratic_constraints()) {
+      CUOPT_LOG_INFO("Problem has %d quadratic constraints. Using Barrier with SOC conversion.",
```
chris-maes, line 1758: "Using Barrier with SOC conversion" -> "Converting to second-order cones and solving with barrier."

cpp/src/pdlp/solve.cu
```diff
@@ -1803,6 +1812,24 @@ optimization_problem_solution_t<i_t, f_t> solve_lp(
     // This needs to be called before pdlp is initialized
     init_handler(op_problem.get_handle_ptr());
 
+    if (op_problem.has_quadratic_objective() || op_problem.has_quadratic_constraints()) {
+      if (op_problem.has_quadratic_objective()) {
+        CUOPT_LOG_INFO("Problem has a quadratic objective. Using Barrier.");
+      }
+      if (op_problem.has_quadratic_constraints()) {
+        CUOPT_LOG_INFO("Problem has %d quadratic constraints. Using Barrier with SOC conversion.",
```
chris-maes, line 1820: "Using Barrier with SOC conversion" -> "Converting to second-order cones and solving with barrier."

cpp/src/pdlp/solve.cu
```diff
@@ -1803,6 +1812,24 @@ optimization_problem_solution_t<i_t, f_t> solve_lp(
...
+                       static_cast<int>(op_problem.get_quadratic_constraints().size()));
+      }
+      settings.method    = method_t::Barrier;
+      settings.presolver = presolver_t::None;
+      // Quadratic objective support is minimization-only.
```
chris-maes, line 1825: Is this correct? I thought we supported maximizing with quadratic objectives.

cpp/src/pdlp/translate.hpp
```diff
@@ -186,6 +190,13 @@ static dual_simplex::user_problem_t<i_t, f_t> cuopt_problem_to_user_problem(
   user_problem.Q_indices = model.Q_indices;
   user_problem.Q_values  = model.Q_values;
 
+  if (model.original_problem_ptr->has_quadratic_constraints()) {
+    detail::apply_soc_qcmatrix_conversion_for_simplex<i_t, f_t>(
```
chris-maes, line 194: `for_simplex` seems like a bad name here. As we aren't using the simplex solver. Maybe this was picked up because of the simplex namespace? Anyway, better to rename the function to avoid confusion.

.claude-plugin/marketplace.json
```diff
@@ -5,7 +5,7 @@
   },
   "metadata": {
     "description": "Agent skills for NVIDIA cuOpt: routing (VRP, TSP, PDP), LP/MILP/QP, installation (user/developer), and REST server.",
-    "version": "26.06.00"
+    "version": "26.08.00"
```
mlubin, line 8: Please rebase on release/26.06. These diffs shouldn't be here.

cpp/src/barrier/barrier.cu
```diff
@@ -61,6 +65,29 @@ namespace cuopt::linear_programming::dual_simplex {
     cuda::std::make_tuple(a, b), out, size, cuda::std::multiplies<>{}, stream.value());
 }
 
+// out[i] = is_native_free_linear[i] ? 0 : a[i] * b[i]
+[[maybe_unused]] static void pairwise_multiply_skip_native_free_linear(
```
mlubin, line 69: Nit: any reason not to template this on `f_t` to avoid having two identical versions?

cpp/src/barrier/barrier.cu
```diff
@@ -209,202 +238,254 @@ class iteration_data_t {
...
+          for (i_t j = 0; j < lp.num_cols; j++) {
+            if (Qdiag[j] < 0.0) {
+              settings_.log.printf(
+                "Q is not positive semidefinite: Q(%d, %d) = %e\n", j, j, Qdiag[j]);
+              indefinite_Q = true;
+              return;
+            }
           }
+        } else if (settings.check_Q) {
+          // TODO: Check to ensure that Q is positive semi-definite
```
mlubin, line 307: Consider adding a warning that check_Q is ignored.

cpp/src/barrier/barrier.cu
```diff
@@ -209,202 +238,254 @@ class iteration_data_t {
...
+      raft::common::nvtx::range scope("Barrier: LP Data: complementarity buffers");
+      const i_t linear_xz_rhs_size = linear_xz_size(lp.num_cols);
+      d_complementarity_xz_rhs_.resize(linear_xz_rhs_size, stream_view_);
 
-    d_flag_buffer.resize(flag_buffer_size, stream_view_);
+      // Allocating GPU flag data for Form ADAT
+      RAFT_CUDA_TRY(cub::DeviceSelect::Flagged(
+        nullptr,
+        flag_buffer_size,
+        d_inv_diag_prime.data(),  // Not the actual input but just to allcoate the memory
```
mlubin, line 346: Typo: allocate

cpp/src/io/mps_parser.cpp
```diff
@@ -74,7 +301,7 @@ BoundType convert(std::string_view str)
     return LowerBoundIntegerVariable;
   } else if (str == "UI") {
     return UpperBoundIntegerVariable;
-  } else if (str == "SC") {
+  } else if (str == "SC" || str == "LC") {
```
mlubin, line 304: This diff shouldn't be here, please revert this line.

cpp/src/io/mps_parser.cpp
```diff
@@ -34,6 +38,229 @@ std::vector<char> string_to_buffer(std::string_view input)
 }
 }  // end namespace
 
+namespace {
+
+/** @brief ceil(log2(x)) for x >= 1; used for sparse-vs-dense CSR heuristic only. */
```
mlubin, line 43: This code is pretty confusing with three code paths(?) for processing CSR. Let's just have one code path until there's a clear reason to have more? LP also needs COO to CSR for the objective, right? let's share the code.

cpp/src/io/mps_parser.cpp
```diff
@@ -288,115 +515,72 @@ void mps_parser_t<i_t, f_t>::fill_problem(mps_data_model_t<i_t, f_t>& problem)
...
-                                        static_cast<char>(row_types[row_id]),
-                                        A_values[row_id],
-                                        A_indices[row_id],
-                                        b_values[row_id],
-                                        csr_result.values,
-                                        csr_result.indices,
-                                        csr_result.offsets);
-    ++quadratic_row_id;
+    const i_t nnz = static_cast<i_t>(block.entries.size());
+    std::vector<i_t> qc_rows(static_cast<size_t>(nnz));
```
mlubin, line 565: I don't think we need a `static_cast<size_t>` everywhere, please remove.

cpp/src/io/mps_parser.cpp
```diff
@@ -288,115 +515,72 @@ void mps_parser_t<i_t, f_t>::fill_problem(mps_data_model_t<i_t, f_t>& problem)
...
+      qc_rows[static_cast<size_t>(e)] = std::get<0>(block.entries[static_cast<size_t>(e)]);
+      qc_cols[static_cast<size_t>(e)] = std::get<1>(block.entries[static_cast<size_t>(e)]);
+      qc_vals[static_cast<size_t>(e)] =
+        std::get<2>(block.entries[static_cast<size_t>(e)]) * k_qcmatrix_value_scale;
+    }
+    problem.append_quadratic_constraint(
+      row_id,
+      row_names[row_id],
+      static_cast<char>(row_types[row_id]),
+      std::span<const f_t>(A_values[row_id].data(), A_values[row_id].size()),
```
mlubin, line 578: I think you can pass `A_values` (etc) directly, explicit `span` construction shouldn't be needed.

cpp/tests/CMakeLists.txt
```diff
@@ -17,8 +17,9 @@ if(BUILD_TESTS)
       PRIVATE
       "${CMAKE_CURRENT_SOURCE_DIR}/../src"
       "${CMAKE_CURRENT_SOURCE_DIR}"
-      "${dejavu_SOURCE_DIR}"
   )
+  # Third-party headers: SYSTEM matches libcuopt and suppresses NVCC sign-conversion noise.
+  target_include_directories(cuopttestutils SYSTEM PRIVATE "${dejavu_SOURCE_DIR}")
```
mlubin, line 22: Unrelated diff? please revert

skills/skill-evolution/skill-card.md
```diff
@@ -8,7 +8,7 @@ This skill is ready for commercial/non-commercial use. <br>
 ### License/Terms of Use: <br>
 Apache 2.0 <br>
 ## Use Case: <br>
-Developers and engineers use this skill to capture generalizable learnings from agent interactions and propose updates to existing skills, ensuring continuous improvement of agent guidance without blocking the user's primary task. <br>
+Developers and engineers use this skill to continuously improve agent skill files by capturing generalizable learnings from problem-solving interactions and proposing structured updates. <br>
```
mlubin, line 11: Please revert all of these

cpp/src/io/mps_parser.cpp
mlubin: Specifically this is reverting https://github.com/NVIDIA/cuopt/commit/eb3dac85f4f96184e0465177f4af47c39c6251dd#diff-1d4ef15136ad0cb10a2d09a2bada6bb3f324a1dc083b03d3a9df99db65367865

cpp/tests/dual_simplex/CMakeLists.txt
```diff
@@ -6,4 +6,5 @@
 ConfigureTest(DUAL_SIMPLEX_TEST
     ${CMAKE_CURRENT_SOURCE_DIR}/unit_tests/solve.cpp
     ${CMAKE_CURRENT_SOURCE_DIR}/unit_tests/solve_barrier.cu
-    LABELS numopt)
```
mlubin, line 9: Do not delete the test label

cpp/include/cuopt/linear_programming/optimization_problem_interface.hpp
```diff
@@ -64,9 +64,10 @@ class optimization_problem_interface_t {
     std::vector<f_t> linear_values{};
     std::vector<i_t> linear_indices{};
     f_t rhs_value{f_t(0)};
+    /** Q in COO: parallel arrays, same length. */
+    std::vector<i_t> quadratic_row_indices{};
```
rg20, line 68:
```suggestion
    std::vector<i_t> rows{};
```

cpp/src/dual_simplex/scaling.cpp
```diff
@@ -16,21 +16,174 @@ template <typename i_t, typename f_t>
 i_t column_scaling(const lp_problem_t<i_t, f_t>& unscaled,
                    const simplex_solver_settings_t<i_t, f_t>& settings,
                    lp_problem_t<i_t, f_t>& scaled,
-                   std::vector<f_t>& column_scaling)
+                   std::vector<f_t>& col_scale,
+                   std::vector<f_t>& row_scale)
```
rg20, line 20: Looks like row_scaling is not used to scale matrix A?

cpp/src/dual_simplex/scaling.cpp
```diff
@@ -16,21 +16,174 @@ template <typename i_t, typename f_t>
...
 {
   scaled = unscaled;
   i_t m  = scaled.num_rows;
   i_t n  = scaled.num_cols;
 
-  if (!settings.scale_columns || unscaled.Q.n > 0) {
+  row_scale.assign(m, 1.0);
+
+  // =========================================================================
+  // Ruiz equilibration for SOCP (and QP) problems
```
rg20, line 29: Can we remove this scaling?  

We should do this cleanly on the KKT system.

cpp/src/dual_simplex/solve.cpp
```diff
@@ -61,6 +61,42 @@ void write_matlab(const std::string& filename, const dual_simplex::lp_problem_t<
   fclose(fid);
 }
 
+template <typename i_t, typename f_t>
+bool validate_barrier_cone_layout(const lp_problem_t<i_t, f_t>& problem,
```
rg20, line 65: This function does not belong here. Move this to src/barrier

cpp/src/dual_simplex/presolve.cpp
yuwenchen95: Yes, the `constraints_to_check`  buildup (only loop over linear columns) will include rows that has linear variable but potential cone variables. Here is the second check that removes rows contain cone variables.

rg20 (commented):
I reviewed only a subset of files. I will continue to review.

chris-maes:
This PR is moved to #1290 . Please make all comments and changes on that PR.

cpp/src/dual_simplex/scaling.cpp
yuwenchen95: It's a temporary that works for CBLIB problems. The only difference would be in the presence with Q matrix. We can keep it now and I will polish in another PR.

cpp/src/barrier/barrier.cu
yuwenchen95: tried to merge them into a template

cpp/src/io/mps_parser.cpp
yuwenchen95: I tried to remove the sparse sorting. Addressed in the [commit](https://github.com/NVIDIA/cuopt/pull/1290/changes/a23347bf83ff413e49ab34489c403e86726f0591)

PR #1096, opened 2026-04-14
Add semi continuous variables support

### LINKED ISSUE #1059: [FEA] Add support for semi-continuous variables
A semicontinous variable satisfies X = 0 or L <= X <= U with  0 < L < U
- If U < Inf, we can model as  L*B <= X <= U*B with additional binary variable B
- If U = Inf, use bound strengthening to try to get finite upper bound on U
- If U still Inf, set U=big-M, allow user to set big-M via parameter

hlinsen:
- Add semi-continuous handling to the MIP solve path through dedicated reformulation before Papilo presolve.
- Normalize semi-continuous variables with zero lower bound to continuous variables before validation and solve.
- Treat very large semi-continuous upper bounds as effectively infinite during semi-continuous reformulation.
- Allow bounds strengthening in semi-continuous reformulation to replace non-finite or loose upper bounds with tighter finite bounds.
- Keep solve_mip on the MIP entry path, but fall back to LP solve after presolve when no integer variables remain.

Closes: https://github.com/NVIDIA/cuopt/issues/1059


cpp/include/cuopt/linear_programming/optimization_problem_utils.hpp
```diff
@@ -253,7 +258,9 @@ void populate_from_data_model_view(optimization_problem_interface_t<i_t, f_t>* p
       data_model->get_variable_types().data() + data_model->get_variable_types().size(),
       enum_variable_types.begin(),
       [](const auto val) -> var_t {
-        return (val == 'I' || val == 'B') ? var_t::INTEGER : var_t::CONTINUOUS;
+        if (val == 'I' || val == 'B') return var_t::INTEGER;
```
mlubin: Consider refactoring to have a single function that maps characters to `var_t`. This logic is duplicated.

cpp/src/mip_heuristics/presolve/semi_continuous.cu
```diff
@@ -0,0 +1,262 @@
...
+  //    This lets GPU bounds propagation derive tight upper bounds from the
+  //    constraint structure without the binary domain {0} ∪ [L, U].
+  optimization_problem_t<i_t, f_t> op_relaxed(op_problem);
+  {
+    auto relaxed_types = var_types;
+    auto relaxed_ub    = var_ub;
+    auto relaxed_lb    = op_problem.get_variable_lower_bounds_host();
+    for (i_t idx : sc_indices) {
+      relaxed_types[idx] = var_t::CONTINUOUS;
+      // Relax to the convex hull of {0} U [L, U] before running GPU bound propagation.
```
mlubin: Are we assuming 0 <= L <= U? This code is a bit inconsistent with the comment that the bounds become `[0, original_ub]`.

cpp/src/mip_heuristics/presolve/semi_continuous.cu
```diff
@@ -0,0 +1,262 @@
...
+  // 3. Run GPU bounds propagation on the relaxed problem to tighten UBs.
+  //    Skip propagation when there are no constraints (nothing to propagate).
+  auto tight_ub = var_ub;  // fallback: normalized original UBs
+
+  if (op_relaxed.get_n_constraints() > 0) {
+    problem_t<i_t, f_t> temp_pb(op_relaxed, settings.get_tolerances());
+    mip_solver_context_t<i_t, f_t> ctx(handle_ptr, &temp_pb, settings);
+
+    typename bound_presolve_t<i_t, f_t>::settings_t bp_settings;
+    bp_settings.time_limit = 5.0;
```
mlubin: Is there a deterministic alternative?

cpp/src/mip_heuristics/presolve/semi_continuous.cu
```diff
@@ -0,0 +1,262 @@
...
+  auto cub    = op_problem.get_constraint_upper_bounds_host();
+
+  // Optional arrays — only extend if they were originally set
+  auto b_rhs       = op_problem.get_constraint_bounds_host();
+  auto row_types_h = op_problem.get_row_types_host();
+
+  // Ensure objective and variable arrays are sized to n_orig
+  if (obj_c.empty()) { obj_c.assign(n_orig, f_t(0)); }
+
+  // 5. Count how many SC vars truly need the binary-variable reformulation.
```
mlubin, line 176: Is this something that might already get cleaned up by presolve? It seems like a rare case; probably a modeling error, that we don't need to optimize for.

cpp/src/mip_heuristics/presolve/semi_continuous.cuh
```diff
@@ -0,0 +1,41 @@
...
+ * GPU bounds propagation (bound_presolve_t) is used to derive tight upper bounds
+ * for SC variables that have infinite original upper bounds.  If propagation cannot
+ * derive a finite bound, settings.sc_big_m is used as a fallback.
+ *
+ * This must be called before problem_t construction and Papilo presolve.
+ *
+ * @tparam i_t  Integer index type
+ * @tparam f_t  Floating-point value type
+ * @param[in,out] op_problem  The optimization problem (modified in-place)
+ * @param[in]     settings    MIP solver settings (provides sc_big_m and tolerances)
```
mlubin: Please document `used_fallback_big_m` as well.

cpp/src/mip_heuristics/presolve/semi_continuous.cuh
```diff
@@ -0,0 +1,41 @@
...
+#include <cuopt/linear_programming/mip/solver_settings.hpp>
+#include <cuopt/linear_programming/optimization_problem.hpp>
+
+namespace cuopt::linear_programming::detail {
+
+/**
+ * @brief Reformulate semi-continuous variables in-place inside the MIP solver.
+ *
+ * A semi-continuous variable x satisfies: x = 0  OR  L <= x <= U  (0 < L < U).
+ * Reformulation introduces a binary variable b and two linking constraints:
```
mlubin: It's worth documenting the ordering of the new variables, i.e., they're added at the end (or maybe not, if we decide the extra variable isn't needed).

cpp/src/pdlp/utilities/problem_checking.cu
```diff
@@ -217,6 +217,38 @@ void problem_checking_t<i_t, f_t>::check_problem_representation(
...
+                    "lower bound %g.",
+                    static_cast<double>(var_lb[i]));
+      cuopt_expects(var_ub[i] > f_t(0),
+                    error_type_t::ValidationError,
+                    "Semi-continuous variable must have a strictly positive upper bound, but has "
+                    "upper bound %g.",
+                    static_cast<double>(var_ub[i]));
+      cuopt_expects(var_lb[i] < var_ub[i],
+                    error_type_t::ValidationError,
+                    "Semi-continuous variable must satisfy lower bound < upper bound, but has "
```
mlubin: I don't think there are any technical blockers to supporting lower bound == upper bound.

cpp/src/pdlp/solve.cu
```diff
@@ -1635,7 +1636,11 @@ cuopt::linear_programming::optimization_problem_t<i_t, f_t> mps_data_model_to_op
       data_model.get_variable_types().cbegin(),
       data_model.get_variable_types().cend(),
       enum_variable_types.begin(),
-      [](const auto val) -> var_t { return val == 'I' ? var_t::INTEGER : var_t::CONTINUOUS; });
+      [](const auto val) -> var_t {
```
mlubin: This is the third place where the char to var_t mapping logic appears.

cpp/tests/mip/semi_continuous_test.cu
```diff
@@ -0,0 +1,121 @@
...
+}
+
+TEST(mip_solve, semi_continuous_regressions)
+{
+  const raft::handle_t handle_{};
+  mip_solver_settings_t<int, double> settings;
+  settings.time_limit = 10.;
+
+  const std::vector<sc_result_t> valid_test_instances = {
+    {"mip/sc_standard.mps", 8., 0.},
```
mlubin: These files are small enough that maybe they could just be inline strings? I think it would make the tests easier to follow.

mlubin (commented):
Check the AI comments too, some of them seem relevant.

cpp/include/cuopt/linear_programming/mip/solver_settings.hpp
```diff
@@ -86,6 +94,7 @@ class mip_solver_settings_t {
 
   f_t time_limit                = std::numeric_limits<f_t>::infinity();
   f_t work_limit                = std::numeric_limits<f_t>::infinity();
+  f_t sc_big_m                  = f_t(1e5);
```
chris-maes: 10^5 seems like a pretty small big-M. I'd probably set this to 1e10 or higher at least.

cpp/include/cuopt/linear_programming/constants.h
```diff
@@ -81,6 +81,7 @@
 #define CUOPT_PRESOLVE_FILE     "presolve_file"
 #define CUOPT_RANDOM_SEED       "random_seed"
 #define CUOPT_PDLP_PRECISION    "pdlp_precision"
+#define CUOPT_SC_BIG_M          "sc_big_m"
```
chris-maes: Let's avoid abbreviations. Can we make this `CUOPT_SEMICONTINUOUS_BIG_M`? 

Or we might want to have a single big-M value that can be used for other reformulations as well. In that case, `CUOPT_BIG_M`

chris-maes: Probably should have the  `CUOPT_MIP_` prefix

cpp/src/grpc/client/solve_remote.cpp
```diff
@@ -209,6 +211,15 @@ std::unique_ptr<mip_solution_interface_t<i_t, f_t>> solve_mip_remote(
 
   // Check if user has set incumbent callbacks
   auto mip_callbacks   = settings.get_mip_callbacks();
+  const auto var_types = cpu_problem.get_variable_types_host();
+  const bool has_sc_variables =
+    thrust::count(var_types.begin(), var_types.end(), var_t::SEMI_CONTINUOUS) > 0;
+  if (has_sc_variables && !mip_callbacks.empty()) {
+    CUOPT_LOG_WARN(
```
chris-maes, line 218: What's the issue with semicontinuous variables and callbacks? The semicontinuous variable should remain the same after translation right? You just add an extra constraint. So as long as you are able to uncrush the presolved solution I think you should be able to output a solution with semicontinuous variables to the user. Am I missing something?

cpp/src/mip_heuristics/presolve/semi_continuous.cu
chris-maes: You can use the CPU bound strengthening code which is deterministic.

cpp/src/mip_heuristics/presolve/semi_continuous.cu
```diff
@@ -0,0 +1,264 @@
...
+      // Relax to the convex hull of {0} U [L, U] before running GPU bound propagation.
+      relaxed_lb[idx] = std::min(f_t(0), relaxed_lb[idx]);
+      if (std::isfinite(relaxed_ub[idx])) { relaxed_ub[idx] = std::max(f_t(0), relaxed_ub[idx]); }
+    }
+    op_relaxed.set_variable_types(relaxed_types.data(), n_orig);
+    op_relaxed.set_variable_lower_bounds(relaxed_lb.data(), n_orig);
+    op_relaxed.set_variable_upper_bounds(relaxed_ub.data(), n_orig);
+  }
+
+  // 3. Run GPU bounds propagation on the relaxed problem to tighten UBs.
```
chris-maes: Is it worth running all of bound strengthening? Perhaps we should restrict ourselves to just the constraints containing semicontinuous variables and the variables in those constraints.

cpp/src/mip_heuristics/presolve/semi_continuous.cu
chris-maes: I think he needs to know how many of these semicontinuous variables will need binaries. 

I do agree that it is probably pretty rare that we are able to convert the semicontinuous to continuous.

cpp/src/mip_heuristics/solve.cu
```diff
@@ -60,6 +64,38 @@ static void init_handler(const raft::handle_t* handle_ptr)
...
+  if (op_problem.get_variable_types().is_empty()) { return; }
+
+  auto var_types = op_problem.get_variable_types_host();
+  auto var_lb    = op_problem.get_variable_lower_bounds_host();
+  bool modified  = false;
+
+  for (i_t i = 0; i < static_cast<i_t>(var_types.size()); ++i) {
+    if (var_types[i] == var_t::SEMI_CONTINUOUS && var_lb[i] == f_t(0)) {
+      CUOPT_LOG_WARN(
+        "SC var %d has zero lower bound; treating it as continuous to match common solver behavior",
```
chris-maes: Just say "treating it as continuous".

cpp/src/mip_heuristics/solve.cu
```diff
@@ -281,9 +317,31 @@ mip_solution_t<i_t, f_t> solve_mip(optimization_problem_t<i_t, f_t>& op_problem,
     raft::common::nvtx::range fun_scope("Running solver");
 
     // This is required as user might forget to set some fields
+    normalize_zero_lb_semi_continuous(op_problem);
+    if (has_semi_continuous_variables(op_problem) && !settings.initial_solutions.empty()) {
```
chris-maes: Why can't we make the initial solution work? We can just look at the value of the semicontinuous variable to derive the value of the binary variable associated with it.

cpp/src/mip_heuristics/solve.cu
```diff
@@ -284,6 +288,21 @@ mip_solution_t<i_t, f_t> solve_mip(optimization_problem_t<i_t, f_t>& op_problem,
...
     problem_checking_t<i_t, f_t>::check_initial_solution_representation(op_problem, settings);
 
+    // Reformulate semi-continuous variables (x = 0 OR L <= x <= U) before Papilo presolve.
+    // Uses GPU bounds propagation to derive tight upper bounds for SC vars with infinite UB.
+    // Track n_orig so that auxiliary binary variables added by reformulation can be stripped
+    // from the solution before returning it to the caller.
+    const i_t n_orig_before_sc = op_problem.get_n_variables();
+    std::vector<uint8_t> sc_used_fallback_big_m;
+    const bool had_sc =
+      detail::reformulate_semi_continuous(op_problem, settings, &sc_used_fallback_big_m);
```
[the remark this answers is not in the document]
chris-maes: I agree, best to start solver timer before all of this. Also, we should use the deterministic CPU bound strengthing and give it only the constraints the variable appears in.

cpp/src/mip_heuristics/solve.cu
```diff
@@ -284,6 +288,21 @@ mip_solution_t<i_t, f_t> solve_mip(optimization_problem_t<i_t, f_t>& op_problem,
...
+    if (had_sc && !settings.initial_solutions.empty()) {
+      CUOPT_LOG_WARN(
+        "Ignoring %zu user initial solution(s): semi-continuous warm starts are not supported yet",
+        settings.initial_solutions.size());
+      settings.initial_solutions.clear();
+    }
```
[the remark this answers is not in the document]
chris-maes: Ah I see. Your solution is potentially too long (because of the addition of binary variables). Can we just add another uncrush step in the callback?

cpp/src/mip_heuristics/presolve/semi_continuous.cu
hlinsen: Papilo does not handle semi continuous vars so we have to handle the potential conversion to continuous here. It is a rare case but handled by other solvers and cheap to do.

cpp/src/grpc/client/solve_remote.cpp
hlinsen: Just wanted to keep the change minimal, we can support it.

cpp/src/mip_heuristics/presolve/semi_continuous.cu
hlinsen: I am not sure, if it's time consuming but we can do what you suggested for sure. It will require some problem manipulation or api change in bound strenghtening which I like to avoid to reduce code complexity.

cpp/src/mip_heuristics/presolve/semi_continuous.cu
hlinsen: > You can use the CPU bound strengthening code which is deterministic.

This or skip bounds strengthening in deterministic mode and just use big-m directly.

cpp/src/mip_heuristics/solve.cu
```diff
@@ -279,11 +316,31 @@ mip_solution_t<i_t, f_t> solve_mip(optimization_problem_t<i_t, f_t>& op_problem,
...
 
     raft::common::nvtx::range fun_scope("Running solver");
+    auto timer = timer_t(time_limit);
 
-    // This is required as user might forget to set some fields
     problem_checking_t<i_t, f_t>::check_problem_representation(op_problem);
     problem_checking_t<i_t, f_t>::check_initial_solution_representation(op_problem, settings);
 
+    // Reformulate semi-continuous variables (x = 0 OR L <= x <= U) before Papilo presolve.
+    // Uses GPU bounds propagation to derive tight upper bounds for SC vars with infinite UB.
```
mlubin: CPU bounds propagation, right?

cpp/tests/utilities/inline_mps_test_utils.hpp
```diff
@@ -0,0 +1,132 @@
...
+ LO BND1      x         2
+ SC BND1      x
+ENDATA
+)";
+
+inline cuopt::mps_parser::mps_data_model_t<int, double> parse_inline_mps(
+  const std::string& name, const std::string& mps_text)
+{
+  const std::filesystem::path path =
+    std::filesystem::path(cuopt::test::get_rapids_dataset_root_dir()) / "mip" /
```
mlubin: This temporary file writing seems a bit silly. Couldn't we have a version of `parse_mps` that takes the file content as a string instead of the filename?

mlubin (approved):

cpp/tests/utilities/inline_mps_test_utils.hpp
hlinsen: Makes sense, we have it already actually https://github.com/NVIDIA/cuopt/blob/main/cpp/libmps_parser/src/mps_parser.cpp#L575

cpp/include/cuopt/linear_programming/constants.h
```diff
@@ -74,13 +74,15 @@
...
-#define CUOPT_RANDOM_SEED       "random_seed"
-#define CUOPT_PDLP_PRECISION    "pdlp_precision"
+#define CUOPT_SOLUTION_FILE            "solution_file"
+#define CUOPT_NUM_CPU_THREADS          "num_cpu_threads"
+#define CUOPT_NUM_GPUS                 "num_gpus"
+#define CUOPT_USER_PROBLEM_FILE        "user_problem_file"
+#define CUOPT_PRESOLVE_FILE            "presolve_file"
+#define CUOPT_RANDOM_SEED              "random_seed"
+#define CUOPT_PDLP_PRECISION           "pdlp_precision"
+#define CUOPT_MIP_SEMICONTINUOUS_BIG_M "sc_big_m"
```
chris-maes: Probably best to avoid abbrevations: can we call this "mip_semicontinuous_big_M"?

cpp/include/cuopt/linear_programming/constants.h
```diff
@@ -74,13 +74,15 @@
...
+#define CUOPT_SC_BIG_M                 CUOPT_MIP_SEMICONTINUOUS_BIG_M
```
chris-maes: Why do we have two #define's here?

cpp/src/mip_heuristics/presolve/semi_continuous.cu
```diff
@@ -0,0 +1,394 @@
...
+  dual_simplex::simplex_solver_settings_t<i_t, f_t> simplex_settings;
+  simplex_settings.primal_tol  = settings.tolerances.presolve_absolute_tolerance;
+  simplex_settings.integer_tol = settings.tolerances.integrality_tolerance;
+  simplex_settings.set_log(false);
+
+  dual_simplex::convert_user_problem(
+    user_problem, simplex_settings, lp_problem, new_slacks, dualize_info);
+
+  std::vector<char> row_sense(lp_problem.num_rows, 'E');
+  for (i_t i = 0; i < user_problem.num_rows; ++i) {
```
chris-maes: I'm not sure I understand what's going on here. Why do you need to change the `row_sense`?

cpp/src/mip_heuristics/presolve/semi_continuous.cu
```diff
@@ -0,0 +1,394 @@
...
+    } else {
+      row_sense[i] = user_problem.row_sense[i];
+    }
+  }
+  for (i_t row : user_problem.range_rows) {
+    row_sense[row] = 'E';
+  }
+
+  auto var_types = user_problem.var_types;
+  var_types.resize(lp_problem.num_cols, dual_simplex::variable_type_t::CONTINUOUS);
```
chris-maes, line 61: Why set all variable types to continuous? If this is coming from a MIP?

cpp/src/mip_heuristics/solve.cu
```diff
@@ -544,6 +620,32 @@ mip_solution_t<i_t, f_t> solve_mip(optimization_problem_t<i_t, f_t>& op_problem,
...
+        cuopt::host_copy(sol.get_solution(), op_problem.get_handle_ptr()->get_stream());
+      const f_t active_tol          = settings.tolerances.integrality_tolerance;
+      i_t num_active_fallback_big_m = 0;
+      for (i_t i = 0; i < static_cast<i_t>(sc_used_fallback_big_m.size()); ++i) {
+        if (!sc_used_fallback_big_m[i]) { continue; }
+        if (host_solution[i] >= settings.sc_big_m - active_tol) { ++num_active_fallback_big_m; }
+      }
+      if (num_active_fallback_big_m > 0) {
+        return mip_solution_t<i_t, f_t>{
+          cuopt::logic_error("Semi-continuous solution is active at fallback sc_big_m; result may "
```
chris-maes: Nit: you might want to print the value of the big M. I also might rephrase this as:
Semi-continuous variable XXX is at upper bound coming from big-M YYY; results may depend on artificial upper bound. 

for XXX I would try to use the variable name if you have it. If not use X43, if it is variable 43 say.

chris-maes (approved):
Some minor questions. But overall LGTM. Thanks for implementing this feature @hlinsen

cpp/include/cuopt/linear_programming/constants.h
hlinsen: Thanks, CUOPT_SC_BIG_M is deprecated

cpp/src/mip_heuristics/presolve/semi_continuous.cu
hlinsen: It only assigns continuous type to the new extra slack variables that were added after `convert_user_problem`. We do the same in bb: https://github.com/NVIDIA/cuopt/blob/main/cpp/src/branch_and_bound/branch_and_bound.cpp#L272
Otherwise we would go out of bounds in the bounds strenghtening.

cpp/src/mip_heuristics/presolve/semi_continuous.cu
hlinsen: convert_user_problems is converting every row to GE, and ranged rows to equality with slack so the intent here is to revert the rows to match the original problem. Do you think it can be skipped?

cpp/src/mip_heuristics/presolve/semi_continuous.cu
chris-maes: Ah my misunderstanding. And my bad abstraction. In a later PR, we should probably create a function called `convert_user_problem_mip` that takes in a var_type and resizes, so we don't need to do this in multiple places. 

Thanks for clarifying. No changes needed now.

cpp/src/mip_heuristics/presolve/semi_continuous.cu
chris-maes: I'm pretty sure `convert_user_problem` makes everything into `E`. That's why it doesn't return a `row_sense` (but you could argue it should). I think you need to set row_sense to `E` for bound strengthening to work correctly.

rgsl888prabhu (approved):

Issue #978, opened 2026-03-20
[BUG]  Numerical instability in `cbs-cta`


nguidotti:
When solving `cbs-cta` instance, there is a numerical instability around zero causing the B&B to never finish as the nodes are properly pruned. See the log below:

```
 ❯ cuopt_cli --time-limit 100  datasets/miplib2017/cbs-cta.mps.bz2
Setting parameter time_limit to 1.000000e+02
Reading file cbs-cta.mps.bz2
optimization_problem_t constructor: Using GPU backend
cuOpt version: 26.4.0, git hash: , host arch: aarch64, device archs: 90a-real
CPU: Unknown, threads (physical/logical): 1/72, RAM: 441.12 GiB
CUDA 13.1, device: NVIDIA GH200 480GB (ID 0), VRAM: 95.00 GiB
CUDA device UUID: 1b529efd-5479-6a28-dc9f-23f37743c4c7
Solving a problem with 10112 constraints, 24793 variables (2467 integers), and 64388 nonzeros
Problem scaling:
Objective coefficents range:          [6e+00, 2e+05]
Constraint matrix coefficients range: [2e-01, 2e+05]
Constraint rhs / bounds range:        [2e-01, 3e+01]
Variable bounds range:                [0e+00, 2e+05]
Original problem: 10112 constraints, 24793 variables, 64388 nonzeros
Calling Papilo presolver (git hash 741a2b9c)
Presolve status: reduced the problem
Presolve removed: 148 constraints, 16579 variables, 33242 nonzeros
Presolved problem: 9964 constraints, 8214 variables, 31146 nonzeros
optimization_problem_t constructor: Using GPU backend
Papilo presolve time: 0.05
Objective offset 0.000000 scaling_factor 1.000000
Model fingerprint: 0xdd6336e4
Running presolve!
Unused variables detected, eliminating them! Unused var count 2
After cuOpt presolve: 9964 constraints, 8212 variables, objective offset 0.000000.
cuOpt presolve time: 0.36
Reduced cost strengthening enabled: 2
24846 variable upper bounds in 0.00 seconds
15114 variable lower bounds in 0.00 seconds
Solving LP root relaxation in concurrent mode
Skipping column scaling
Dual Simplex Phase 1
Dual feasible solution found.
Dual Simplex Phase 2
 Iter     Objective           Num Inf.  Sum Inf.     Perturb  Time
[...]

PR #1020, opened 2026-04-01
Unified pseudocost object for the regular and deterministic mode

nguidotti:
This PR simplify the pseudo cost class in such a way that both regular and deterministic B&B use the same code path as much as possible. It also disable mutexes and atomics when running in deterministic mode as each thread has its own snapshot of the pseudocost. It also move the routines related with the diving heuristics back to the `diving_heuristics.cpp`, renamed `branch_and_bound_worker.hpp` to `worker.hpp` to match the new file structure and moved `worker_pool_t` to a dedicated header.

Regular mode (GH200, 10min):
```
================================================================================
 main-190326-2 (1) vs simplify-pseudocost (2)
================================================================================

------------------------------------------------------------------------------------------------------------------------------
|                                        |       Run 1        |       Run 2        |     Abs. Diff.     |   Rel. Diff. (%)   |
------------------------------------------------------------------------------------------------------------------------------
| Feasible                                                 226                  226                   +0                 --- |
| Optimal                                                   70                   67                   -3                 --- |
| Solutions with <0.1% primal gap                          121                  122                   +1                 --- |
| Nodes explored (mean)                           4283972.9121         4455377.8117         +171404.8996              +3.847 |
| Nodes explored (shifted geomean)                   6202.3471            7062.2682            +859.9210             +12.176 |
| Relative MIP gap (mean)                               0.3382               0.3337              -0.0045              -1.325 |
| Relative MIP gap (shifted geomean)                    0.1193               0.1166              -0.0027              -2.293 |
| Solve time (mean)                                   450.2347             452.9154              +2.6806              +0.592 |
| Solve time (shifted geomean)                        221.4772             227.6381              +6.1609              +2.706 |
| Primal gap (mean)                                    11.4459              11.0482              -0.3976              -3.474 |
| Primal gap (shifted geomean)                          0.6591               0.6008              -0.0582              -8.838 |
| Primal integral (mean)                               49.9109              54.6941              +4.7832              +8.745 |
| Primal integral (shifted geomean)                    11.5672              13.9826              +2.4153             +17.274 |
------------------------------------------------------------------------------------------------------------------------------
```

Determinism mode (GH200, 5min):
```
================================================================================
 main-240426-determinism (1) vs simplify-pseudocost-determinism (2)
================================================================================

------------------------------------------------------------------------------------------------------------------------------
|                                        |       Run 1        |       Run 2        |     Abs. Diff.     |   Rel. Diff. (%)   |
------------------------------------------------------------------------------------------------------------------------------
[...]

PR #1099, opened 2026-04-14
Unified threading model in MIP solver

### LINKED ISSUE #1067: [FEA] Respect CPU thread limits
We sometimes launch 137 threads (!) when `--num-cpu-threads=64`

nguidotti:
In this PR, we migrate almost all parts of the MIP solver from `std::thread` to OpenMP (in particular, the tasking model of OpenMP). The only exception is the Papilo presolver that uses Intel TBB and the LP solver.

More specifically, this PR
- Solves the CPU oversubscription problem. The solver now respect the number of threads set to the user, with the exception of Papilo or threads created by the CUDA runtime.
- Removes overheads from creating and destroying `std::thread`. 
- Migrates RINS from `std::thread` to `omp task`. Similar to previous logic, one instance of RINS can run at a time. 
- Migrates `CPU FJ` from `std::thread` to `omp task`. There are a few limitations 
   - `scratch_cpu_fj_on_lp_opt` and `scratch_cpu_fj` are running for the entire program. This essentially allocate two dedicated threads to these functions, while other routines needs to share the remaining CPU resources. This may hurt the performance for low core count CPUs.
   - Since there is a small delay between the task creation and its start (since the threads may be busy), the GPU FJ may finish before the CPU FJ even start when racing. 
-  Migrate early FJ to `omp task`. 
- Only a single parallel region created at the beginning of the solver, so it can be shared across the MIP solver. 
- Eliminate `cpu_worker_thread` and other redundant code 
- Fixed incorrect loop ordering in the the probing cache.

MIPLIB2017:
GH200, 10min time limit, `cbs-cta` excluded (see #978) 

```
 main-190326-2 (1) vs unified-parallel-model (2)

| Feasible                                                 226                  227                   +1                 --- |
| Optimal                                                   70                   74                   +4                 --- |
| Solutions with <0.1% primal gap                          121                  124                   +3                 --- |
| Nodes explored (mean)                           4283972.9121         4684918.2469         +400945.3347              +8.558 |
| Nodes explored (shifted geomean)                   6202.3471            7545.9821           +1343.6350             +17.806 |
| Relative MIP gap (mean)                               0.3382               0.3279              -0.0103              -3.037 |
| Relative MIP gap (shifted geomean)                    0.1193               0.1146              -0.0047              -3.919 |
| Solve time (mean)                                   450.2347             449.6831              -0.5517              -0.123 |
| Solve time (shifted geomean)                        221.4772             239.4322             +17.9549              +7.499 |
| Primal gap (mean)                                    11.4459              11.3227              -0.1232              -1.076 |
| Primal gap (shifted geomean)                          0.6591               0.6122              -0.0469              -7.109 |
| Primal integral (mean)                               49.9109              54.1269              +4.2160              +7.789 |
| Primal integral (shifted geomean)                    11.5672              13.7114              +2.1442             +15.638 |
```

```
 main-190326 (1) vs omp-migration (2)

| Feasible                                                 225                  225                   +0                 --- |
| Optimal                                                   70                   72                   +2                 --- |
| Solutions with <0.1% primal gap                          122                  125                   +3                 --- |
| Nodes explored (mean)                           4471749.9038         4615040.2887         +143290.3849              +3.105 |
| Nodes explored (shifted geomean)                   7063.6974            7312.7210            +249.0235              +3.405 |
| Relative MIP gap (mean)                               0.3255               0.3223              -0.0032              -0.969 |
| Relative MIP gap (shifted geomean)                    0.1185               0.1159              -0.0026              -2.156 |
| Solve time (mean)                                   451.3237             446.7403              -4.5834              -1.016 |
| Solve time (shifted geomean)                        224.6465             234.9078             +10.2613              +4.368 |
| Primal gap (mean)                                    11.6840              11.6531              -0.0309              -0.265 |
| Primal gap (shifted geomean)                          0.6616               0.6369              -0.0248              -3.746 |
| Primal integral (mean)                               51.1565              52.9541              +1.7976              +3.395 |
| Primal integral (shifted geomean)                    12.5468              12.7091              +0.1623              +1.277 |
```

## Checklist

- Testing
   - [ ] Added tests
   - [ ] NA
- Documentation
   - [ ] Added new documentation
   - [ ] NA


nguidotti:
@coderabbit review

cpp/src/branch_and_bound/pseudo_costs.cpp
```diff
@@ -1052,75 +1051,97 @@ void strong_branching(const lp_problem_t<i_t, f_t>& original_lp,
...
+                             dual_simplex_status_up,
+                             sb_view);
       }
+      // DS done: signal PDLP to stop (time-limit or all work done) and wait
+      if (effective_batch_pdlp == 1) { concurrent_halt.store(1); }
+    }
+
+    if (effective_batch_pdlp != 0) {
+#pragma omp taskwait
     }
```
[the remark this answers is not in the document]
nguidotti: Both routines are only called within B&B, which is enclosed by a parallel region (see solve.cu).

cpp/src/mip_heuristics/solver.cu
```diff
@@ -444,33 +443,35 @@ solution_t<i_t, f_t> mip_solver_t<i_t, f_t>::run_solver()
...
+  }
+
+#pragma omp taskgroup
+  {
+    if (!context.settings.heuristics_only) {
+#pragma omp task default(none) \
+  shared(branch_and_bound, branch_and_bound_solution, branch_and_bound_status)
+      {
+        branch_and_bound_status = branch_and_bound->solve(branch_and_bound_solution);
+      }
```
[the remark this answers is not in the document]
nguidotti: As far I am aware, B&B does not throw exceptions. So, a try-catch block is not necessary here.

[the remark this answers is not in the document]
akifcorduk: Doesn't standard library algorithms throw exceptions like bad_alloc?

cpp/src/branch_and_bound/branch_and_bound.cpp
```diff
@@ -1929,9 +1940,9 @@ lp_status_t branch_and_bound_t<i_t, f_t>::solve_root_relaxation(
 
     // Check if crossover was stopped by dual simplex
     if (crossover_status == crossover_status_t::OPTIMAL) {
-      set_root_concurrent_halt(1);             // Stop dual simplex
-      root_status = root_status_future.get();  // Wait for dual simplex to finish
-      set_root_concurrent_halt(0);             // Clear the concurrent halt flag
+      set_root_concurrent_halt(1);  // Stop dual simplex
+#pragma omp taskwait                // Wait for dual simplex to finish
```
akifcorduk: Isn't it a performance issue? Can this cause dead lock?

AFAIU, taskwait waits for all tasks that were spawned in this thread. So this clause will wait for any other thread that may have launched from this task/thread. I think `depend` directive helps here.

cpp/src/branch_and_bound/pseudo_costs.cpp
```diff
@@ -1576,14 +1608,30 @@ i_t pseudo_costs_t<i_t, f_t>::reliable_variable_selection(
...
-#pragma omp taskloop if (num_tasks > 1) priority(task_priority) num_tasks(num_tasks) \
-  shared(score_mutex,                                                                \
-           sb_view,                                                                  \
-           dual_simplex_obj_down,                                                    \
-           dual_simplex_obj_up,                                                      \
-           dual_simplex_status_down,                                                 \
-           dual_simplex_status_up,                                                   \
-           unreliable_list)
+#pragma omp taskloop if (num_tasks > 1) priority(task_priority) num_tasks(num_tasks) default(none) \
+  shared(log,                                                                                      \
```
akifcorduk: For these verbose `shared` directives, i think we should default the `shared` and only use `private` for needed variables.

cpp/src/mip_heuristics/presolve/probing_cache.cu
```diff
@@ -890,23 +890,39 @@ bool compute_probing_cache(bound_presolve_t<i_t, f_t>& bound_presolve,
...
+      for (size_t i = begin; i < end; ++i) {
         auto var_idx = priority_indices[i];
         if (timer.check_time_limit()) { continue; }
 
-        int thread_idx = omp_get_thread_num();
-        CUOPT_LOG_TRACE("Computing probing cache for var %d on thread %d", var_idx, thread_idx);
-
-        auto& multi_probe_presolve = multi_probe_presolve_pool[thread_idx];
-
+        CUOPT_LOG_TRACE("Computing probing cache for var %d on thread %d", var_idx, id);
```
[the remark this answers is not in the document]
akifcorduk: Yes this is a real issue.

cpp/src/mip_heuristics/presolve/probing_cache.cu
```diff
@@ -890,23 +890,39 @@ bool compute_probing_cache(bound_presolve_t<i_t, f_t>& bound_presolve,
...
+           h_integer_indices,                                                              \
+           n_of_implied_singletons,                                                        \
+           n_of_cached_probings,                                                           \
+           problem_is_infeasible,                                                          \
+           modification_vector_pool,                                                       \
+           substitution_vector_pool)
+    for (size_t task_id = 0; task_id < num_tasks; ++task_id) {
+      size_t n                   = step_end - step_start;
+      size_t begin               = std::floor(static_cast<f_t>(n) * task_id / num_tasks);
+      size_t end                 = std::floor(static_cast<f_t>(n) * (task_id + 1) / num_tasks);
```
akifcorduk: I think you need to clamp the `end` on the last task here, right?

cpp/src/mip_heuristics/solve.cu
```diff
@@ -567,6 +568,50 @@ mip_solution_t<i_t, f_t> solve_mip(optimization_problem_t<i_t, f_t>& op_problem,
...
+mip_solution_t<i_t, f_t> solve_mip(optimization_problem_t<i_t, f_t>& op_problem,
+                                   mip_solver_settings_t<i_t, f_t> const& settings_const)
+{
+  std::exception_ptr exception;
+  i_t num_threads = 0;
+  if (settings_const.num_cpu_threads < 0) {
+    num_threads = omp_get_max_threads();
+  } else {
+    if (settings_const.num_cpu_threads < 4) {
+      CUOPT_LOG_ERROR(
```
akifcorduk: Instead of ERROR, this should be a WARNING.

cpp/src/mip_heuristics/solve.cu
```diff
@@ -567,6 +568,50 @@ mip_solution_t<i_t, f_t> solve_mip(optimization_problem_t<i_t, f_t>& op_problem,
...
+  shared(sol, op_problem, settings_const, exception)
+  {
+#pragma omp master
+    {
+      try {
+        sol = solve_mip_helper<i_t, f_t>(op_problem, settings_const);
+      } catch (...) {
+        // We cannot throw inside an OpenMP parallel region. So we need to catch and then
+        // re-throw later.
+        exception = std::current_exception();
```
akifcorduk, line 729: What if there are multiple exceptions in parallel?

cpp/src/mip_heuristics/solver.cu
```diff
@@ -181,6 +181,8 @@ void extract_probing_implied_bounds(
 template <typename i_t, typename f_t>
 solution_t<i_t, f_t> mip_solver_t<i_t, f_t>::run_solver()
 {
+  solution_t<i_t, f_t> sol(*context.problem_ptr);
```
akifcorduk, line 184: I think `problem` might be copied or modified in the following calls. I would double check that removed solution constructors below are indeed using the unmodified problem and same as the original one.

akifcorduk (commented):
I think we should run a baseline with probing cache changes as it fixes performance bugs. That way we can measure the performance impact of the threading changes. Otherwise, we are testing multiple things together.

p.s. I didn't review cp_fj changes thoroughly.

cpp/src/branch_and_bound/branch_and_bound.cpp
nguidotti: Since B&B is enclosed by a task and there are no other task created before, this should be fine. In any case, I added the depend clause since it better indicate the intent.

cpp/src/mip_heuristics/presolve/probing_cache.cu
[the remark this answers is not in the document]
nguidotti: Fixed in [50fa3d6](https://github.com/NVIDIA/cuopt/pull/1099/commits/50fa3d6129c27d00dc6b1f217162eb12d973e451)

cpp/src/branch_and_bound/pseudo_costs.cpp
nguidotti: Fixed in [180fbd9](https://github.com/NVIDIA/cuopt/pull/1099/commits/180fbd908f82e6f4e15a9b833d32d39b02b3021f)

cpp/src/mip_heuristics/presolve/probing_cache.cu
nguidotti: I do not think is needed. For the last task,

```
task_id = num_tasks - 1
begin = step_start + floor((step_end - step_start) * (num_tasks - 1) / num_tasks)
end =  step_start + floor((step_end - step_start) * num_tasks / num_tasks) = step_start + step_end - step_start = step_end
```

nguidotti:
/ok to test [b4efd67](https://github.com/NVIDIA/cuopt/pull/1099/commits/b4efd6715d72cbab3a4a897c118582163133ec1d)

cpp/src/mip_heuristics/solve.cu
nguidotti: Considering that the exceptions cannot cross thread boundaries, this will only catch the exceptions throw by the master thread.

nguidotti: However, if we want to catch the exception from all threads, then we need a mechanism for collecting and re-throwing in the master thread.

nguidotti:
/ok to test [80b14ab](https://github.com/NVIDIA/cuopt/pull/1099/commits/80b14ab2b2227dae3c79cdc09ecd9990552da3ac)

cpp/src/mip_heuristics/diversity/lns/rins.cu
```diff
@@ -134,7 +115,10 @@ void rins_t<i_t, f_t>::run_rins()
   cuopt_assert(best_sol.handle_ptr == &rins_handle, "Handle mismatch");
 
   cuopt_assert(best_sol.get_feasible(), "Best solution is not feasible");
-  if (!best_sol.get_feasible()) { return; }
+  if (!best_sol.get_feasible()) {
+    launch_new_task = true;
```
aliceb-nv: We should probably guard "launch_new_task = true" using a RAII scope_guard object instead of manually adding it to all exit paths

cpp/src/mip_heuristics/local_search/local_search.cu
```diff
@@ -121,34 +111,46 @@ void local_search_t<i_t, f_t>::start_cpufj_lptopt_scratch_threads(
...
                         context.problem_ptr->get_user_obj_from_solver_obj(
                           population.is_feasible() ? population.best_feasible().get_objective()
                                                    : std::numeric_limits<f_t>::max()));
-        local_search_best_obj = obj;
+        this->local_search_best_obj = obj;
       }
     };
 
   // default weights
   cudaDeviceSynchronize();
```
aliceb-nv: tangential but we should get rid of that cudaDeviceSynchronize I forgot to ever remove :( Could cause issues if other threads are currently running graph captures. A stream on the problem sync should be more than enough

cpp/src/mip_heuristics/local_search/local_search.cu
```diff
@@ -233,48 +239,50 @@ bool local_search_t<i_t, f_t>::do_fj_solve(solution_t<i_t, f_t>& solution,
...
+                                   true);
   }
 
   auto solution_copy = solution;
 
   // Start CPU solver in background thread
-  for (auto& cpu_fj_ptr : ls_cpu_fj) {
-    cpu_fj_ptr->start_cpu_solver();
-  }
+#pragma omp taskgroup
```
aliceb-nv, line 267: Just to make sure I understand - there's an implicit taskwait at the end of this taskgroup right? Is that how the CPUFJ tasks are joined after the halt preemption is requested?

cpp/src/mip_heuristics/solve.cu
```diff
@@ -567,6 +568,53 @@ mip_solution_t<i_t, f_t> solve_mip(optimization_problem_t<i_t, f_t>& op_problem,
...
+  } else {
+    num_threads = settings_const.num_cpu_threads;
+  }
+
+  if (num_threads < 4) {
+    CUOPT_LOG_ERROR("The MIP solver requires at least 4 CPU threads!");
+    return mip_solution_t<i_t, f_t>{
+      cuopt::logic_error("The number of CPU threads is below than expected.",
+                         cuopt::error_type_t::RuntimeError),
+      op_problem.get_handle_ptr()->get_stream()};
```
aliceb-nv: Is that true? Don't we support running on a single thread?

aliceb-nv (commented):
Thanks a lot for that work Nicolas :)
Let me know if you need help with porting the B&B determinism code.

cpp/src/mip_heuristics/diversity/lns/rins.cu
nguidotti: Good idea! Changed in [3c569e6](https://github.com/NVIDIA/cuopt/pull/1099/commits/3c569e61ec190b43a31901210d1ee6522f4e4336)

cpp/src/mip_heuristics/local_search/local_search.cu
nguidotti: Exactly, at the end of the `taskgroup` there is an implicit barrier that waits for all tasks created inside the group to finish.

nguidotti: Added comments to each implicit barrier

cpp/src/mip_heuristics/local_search/local_search.cu
nguidotti: Done in [3c569e6](https://github.com/NVIDIA/cuopt/pull/1099/commits/3c569e61ec190b43a31901210d1ee6522f4e4336)

nguidotti:
/ok to test [d7d3dad](https://github.com/NVIDIA/cuopt/pull/1099/commits/d7d3dad7878b277922b6a05da2ade20d43fca580)

cpp/src/mip_heuristics/solve.cu
nguidotti: With low thread count, some vital tasks may not execute. If you have just one thread, then how we suppose to execute the heuristic and the B&B thread at the same time?

nguidotti: We could support low thread count, if either one or another tasks can be interrupted or yielded. But I think this is too much effort for a rare case (I mean, it is pretty common to have CPUs with more than 8 threads nowadays)

nguidotti:
/ok to test [6ac7882](https://github.com/NVIDIA/cuopt/pull/1099/commits/6ac78824c3d5f195baaa3af0e1703261afa3c474)

nguidotti:
/ok to test [e152241](https://github.com/NVIDIA/cuopt/pull/1099/commits/e15224178a8cff9940bd83728f8d070042273bc1)

cpp/src/mip_heuristics/solver.cu
```diff
@@ -444,33 +443,34 @@ solution_t<i_t, f_t> mip_solver_t<i_t, f_t>::run_solver()
...
       context.stats.set_solution_bound(
         context.problem_ptr->get_user_obj_from_solver_obj(branch_and_bound_solution.lower_bound));
     }
-    if (bb_status == dual_simplex::mip_status_t::INFEASIBLE) { sol.set_problem_fully_reduced(); }
+    if (branch_and_bound_status == dual_simplex::mip_status_t::INFEASIBLE) {
+      sol.set_problem_fully_reduced();
+    }
     context.stats.num_nodes              = branch_and_bound_solution.nodes_explored;
     context.stats.num_simplex_iterations = branch_and_bound_solution.simplex_iterations;
   }
```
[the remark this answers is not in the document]
nguidotti: The best solution is currently handle by the heuristic thread.

cpp/src/mip_heuristics/solver.cu
```diff
@@ -444,33 +443,34 @@ solution_t<i_t, f_t> mip_solver_t<i_t, f_t>::run_solver()
...
+#pragma omp task default(shared)
+      {
+        branch_and_bound_status = branch_and_bound->solve(branch_and_bound_solution);
+      }
+    }
+
+    // Start the primal heuristics
+    context.diversity_manager_ptr = &dm;
+    sol                           = dm.run_solver();
+  }  // implicit barrier for all tasks created in B&B and heuristics
```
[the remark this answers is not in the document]
nguidotti: For this reason, the solver should run with at least 2 threads. The associated error is in `solve.cu` in the `mip_heuristic` folder.

nguidotti:
/ok to test [8e4b324](https://github.com/NVIDIA/cuopt/pull/1099/commits/8e4b324d7a6e53dc9980b3db1f3dea0b480fc6e2)

nguidotti:
/ok to [3f72a33](https://github.com/NVIDIA/cuopt/pull/1099/commits/3f72a3382c387ef7e28aa254f61931a5f9dfd8ff)

nguidotti:
/ok to test [be7a782](https://github.com/NVIDIA/cuopt/pull/1099/commits/be7a7822c7c158cd96eb52880048ff2f1f009bd9)

nguidotti:
/ok to test [7c4547b](https://github.com/NVIDIA/cuopt/pull/1099/commits/7c4547ba2fdac9fb88b16caaa6b821ffc6683b3c)

cpp/src/mip_heuristics/diversity/diversity_manager.cu
```diff
@@ -599,22 +599,20 @@ solution_t<i_t, f_t> diversity_manager_t<i_t, f_t>::run_solver()
     run_fj_alone(sol);
     return sol;
   }
-  rins.enable();
+
+  if (omp_get_num_threads() > 4) { rins.enable(); }
```
akifcorduk: Should we put these constants in a file? As we add more concurrent workers later, it will help us prioritize them depending on the thread count.

cpp/src/mip_heuristics/feasibility_jump/early_cpufj.cu
```diff
@@ -32,40 +30,40 @@ early_cpufj_t<i_t, f_t>::~early_cpufj_t()
 template <typename i_t, typename f_t>
 void early_cpufj_t<i_t, f_t>::start()
 {
-  if (cpu_fj_thread_) { return; }
+  // 1: presolve, 1: early GPU FJ, 1: early CPU FJ
+  if (fj_cpu_ || omp_get_num_threads() < 3) { return; }
```
akifcorduk: We both hav ethe same `if` in cpu and gpu fj code. If there are 2 threads, we disable both. I think we should change one of them to `omp_get_num_threads() < 2`, right?

cpp/src/branch_and_bound/branch_and_bound.cpp
```diff
@@ -1755,7 +1755,7 @@ void branch_and_bound_t<i_t, f_t>::run_scheduler()
         active_workers_per_strategy_[strategy]++;
         launched_any_task = true;
 
-#pragma omp task affinity(worker)
+#pragma omp task affinity(worker) default(none) firstprivate(worker)
```
akifcorduk, line 1761: Why do we need a copy of worker now?

cpp/src/mip_heuristics/local_search/local_search.cu
```diff
@@ -47,25 +43,18 @@ local_search_t<i_t, f_t>::local_search_t(mip_solver_context_t<i_t, f_t>& context
...
 
-static double local_search_best_obj       = std::numeric_limits<double>::max();
-static population_t<int, double>* pop_ptr = nullptr;
-
 template <typename i_t, typename f_t>
 void local_search_t<i_t, f_t>::start_cpufj_scratch_threads(population_t<i_t, f_t>& population)
 {
-  pop_ptr = &population;
+  // TODO: Find a way to enable this in low core count scenarios
+  if (omp_get_num_threads() < 8) return;
```
akifcorduk: Again we can put this in a file as a constant.

cpp/src/mip_heuristics/presolve/conditional_bound_strengthening.cu
```diff
@@ -246,11 +246,14 @@ void conditional_bound_strengthening_t<i_t, f_t>::select_constraint_pairs_host(
   std::vector<int2> constraint_pairs_h(max_pair_per_row * problem.n_constraints, {-1, -1});
   std::unordered_set<int> cnstr_pair;
 
-#pragma omp parallel for private(cnstr_pair)
-  for (int cnstr = 0; cnstr < problem.n_constraints; ++cnstr) {
-    for (int jj = offsets[cnstr]; jj < offsets[cnstr + 1]; ++jj) {
+  i_t num_tasks = std::max(omp_get_num_threads(), 1);
```
akifcorduk: When we use the max amount of tasks, we might be blocking the progress of some other tasks if this is long running. I think we should never use full `omp_get_num_threads` amount of tasks.

The best way would be to have a rough timeline of how many tasks are active at which stage of the solve. This way we can estimate what would be the max amount of tasks we can use per use case.

cpp/src/mip_heuristics/presolve/probing_cache.cu
```diff
@@ -860,18 +861,17 @@ bool compute_probing_cache(bound_presolve_t<i_t, f_t>& bound_presolve,
   bound_presolve.settings.iteration_limit = 50;
   bound_presolve.settings.time_limit      = timer.remaining_time();
 
-  size_t num_threads = bound_presolve.settings.num_threads < 0
-                         ? 0.2 * omp_get_max_threads()
-                         : bound_presolve.settings.num_threads;
-  num_threads        = std::clamp<size_t>(num_threads, 1, 8);
+  size_t num_tasks = bound_presolve.settings.num_tasks < 0 ? 0.2 * omp_get_num_threads()
```
akifcorduk: Now probing cache is part of presolve and nothing runs concurrent to it (other than initial cpufj). We can make this `omp_get_num_threads() - 1`

cpp/src/mip_heuristics/solver.cu
```diff
@@ -444,33 +443,34 @@ solution_t<i_t, f_t> mip_solver_t<i_t, f_t>::run_solver()
...
-                                                std::ref(branch_and_bound_solution));
   }
 
-  // Start the primal heuristics
-  context.diversity_manager_ptr = &dm;
-  auto sol                      = dm.run_solver();
-  if (run_bb) {
-    // Wait for the branch and bound to finish
-    auto bb_status = branch_and_bound_status_future.get();
+#pragma omp taskgroup
```
akifcorduk, line 452: Could you explain why this directive is needed?

cpp/src/mip_heuristics/solver.cu
akifcorduk: I think this comment is still valid. We might change the problem later in between and the solution sizes will be inconsistent with the problem.

cpp/src/mip_heuristics/solver.cu
nguidotti: All threads will wait for all tasks that were created within the `taskgroup` to finish before proceeding. In this case, includes all B&B workers, RINS, and other tasks created within B&B.

cpp/src/branch_and_bound/branch_and_bound.cpp
nguidotti: This only copies the pointer, not the actual object. We need to do this, so the task start with the correct worker pointer.

nguidotti:
/ok to test [ea7403f](https://github.com/NVIDIA/cuopt/pull/1099/commits/ea7403f12c9279247cd3fa9f716d8536042e1d58)

nguidotti:
/ok to test [f2cd669](https://github.com/NVIDIA/cuopt/pull/1099/commits/f2cd6693180b90a0a6039a645a2fb534b0571d5a)

cpp/src/mip_heuristics/solver.cu
nguidotti: Before the solver will initialize the solution object and then immediately exits. It is similar here, we are just created the object at the beggining. The copy assignment is the default one (meaning it should copy all attributes). So it should not be a problem (unless I miss something)

nguidotti:
/ok to test [641b8ee](https://github.com/NVIDIA/cuopt/pull/1099/commits/641b8eeace5eb0999f5ca60766c5aad2c6c96c9c)

cpp/include/cuopt/linear_programming/constants.h
```diff
@@ -110,6 +110,13 @@
 #define CUOPT_MODE_OPPORTUNISTIC 0
 #define CUOPT_MODE_DETERMINISTIC 1
 
+/* @brief Minimimum number of threads to enable each part of the MIP Solver */
+#define CUOPT_MIP_FJ_MIN_THREAD_COUNT          8
+#define CUOPT_MIP_EARLY_GPUFJ_MIN_THREAD_COUNT 3
+#define CUOPT_MIP_EARLY_CPUFJ_MIN_THREAD_COUNT 2
+#define CUOPT_MIP_RINS_MIN_THREAD_COUNT        8
+#define CUOPT_MIP_BATCH_PDLP_MIN_THREAD_COUNT  3
+
```
chris-maes: I'm confused. These minimum thread counts all look gigantic. Shouldn't the minimum thread count be 0 (meaning we disable the feature) or 1 (if the feature is required)?

chris-maes: Also, these constants should not live here. Do they need to live in this file at all? If so, they should be placed at the very end after important constants like MAXIMIZE, MINIMIZE and the termination criteria.

cpp/src/branch_and_bound/branch_and_bound.cpp
```diff
@@ -1876,19 +1877,23 @@ lp_status_t branch_and_bound_t<i_t, f_t>::solve_root_relaxation(
...
-                                  std::ref(basis_update),
-                                  std::ref(basic_list),
-                                  std::ref(nonbasic_list),
-                                  std::ref(root_vstatus),
-                                  std::ref(edge_norms),
-                                  nullptr);
+
+  // Note that we need to explicitly declared `root_status` as a shared variable here since
+  // it is local to the thread that are executing the enclosing task.
+#pragma omp task default(shared) depend(out : root_status)
```
chris-maes, line 1884: I find it a bit difficult to understand what is happening here. A few comments might help.

If I understand correctly, we span a task with dual simplex on another thread, and then wait until PDLP or Barrier finishes, and then check if crossover is optimal (if so we use the crossover solution), if not we use the dual simplex solution. I think there is another condition in which we use dual simplex. 

I think what makes the code difficult to understand is within this function you can't tell how dual simplex and Barrier/PDLP are coordinated (e.g. by setting the concurrent halt pointer etc). It might be useful to move that code here, so it is contained within a single function. 

I guess @rg20 was going to change the structure of this code. So maybe changing the code is not necessary, but adding a few comments might help the reader.

cpp/src/branch_and_bound/pseudo_costs.cpp
```diff
@@ -1011,7 +1012,10 @@ void strong_branching(const lp_problem_t<i_t, f_t>& original_lp,
 
   // 0: no batch PDLP, 1: cooperative batch PDLP and DS, 2: batch PDLP only
   const i_t effective_batch_pdlp =
-    (settings.sub_mip || (settings.deterministic && settings.mip_batch_pdlp_strong_branching == 1))
+    (settings.sub_mip ||
+     (settings.deterministic && settings.mip_batch_pdlp_strong_branching == 1) ||
+     omp_get_num_threads() <
+       CUOPT_MIP_BATCH_PDLP_MIN_THREAD_COUNT)  // 1: heuristics, 1: B&B, 1: batch PDLP
```
chris-maes: I'm not sure I understand what is going on here. Is the idea that if you have less than CUOPT_MIP_BATCH_PDLP_MIN_THREAD_COUNT threads that you do not launch BatchPDLP? 

If so, it's probably better to name that constant BATCH_PDLP_REQUIRED_THREADS. 

Also, sticking this many clauses inside a ternary operator ( X ? Y : Z) makes the code quite difficult to understand. Maybe you could break it up into a few named booleans to help the reader.

cpp/src/branch_and_bound/pseudo_costs.cpp
```diff
@@ -1415,11 +1422,14 @@ i_t pseudo_costs_t<i_t, f_t>::reliable_variable_selection(
...
-                                           pdlp_warm_cache.percent_solved_by_batch_pdlp_at_root >
-                                             min_percent_solved_by_batch_pdlp_at_root_for_pdlp);
+  const bool use_pdlp =
+    (rb_mode == 2) ||
+    (rb_mode != 0 && !settings.sub_mip && !settings.deterministic && pdlp_warm_cache.populated &&
+     unreliable_list.size() > min_num_candidates_for_pdlp &&
+     pdlp_warm_cache.percent_solved_by_batch_pdlp_at_root >
+       min_percent_solved_by_batch_pdlp_at_root_for_pdlp &&
+     omp_get_num_threads() >=
+       CUOPT_MIP_BATCH_PDLP_MIN_THREAD_COUNT);  // 1: heuristics, 1: B&B, 1: batch PDLP);
```
chris-maes: I would suggest breaking this up and adding some comments to explain what is going on here. Otherwise it is very difficult to understand this expression.

chris-maes (changes requested):
Thanks for unifying the threading model @nguidotti 

I only reviewed constants.h, branch_and_bound.cpp and psuedocosts.cpp. Several of the changes are difficult to understand. I'd prefer if you addressed the comments before merging.

cpp/include/cuopt/linear_programming/constants.h
nguidotti: No. In this determines the minimum number of threads required for enabling certain features. For instance, CPU FJ launches a task at the beginning at the solver that blocks one thread for the entire runtime (scratch CPU FJ). So, I disable this part when the number of threads available is low.

nguidotti: Now, in this PR, I maintain the code behaviour as was written before (with `std::thread`) to avoid any surprises. However, there are still many deficient that I will address in a follow-up PR as they required some additional testing.

cpp/src/branch_and_bound/pseudo_costs.cpp
nguidotti: Yes, we skip batch PDLP if there is not enough threads (i.e., less 3) otherwise batch PDLP might never run.

cpp/src/branch_and_bound/pseudo_costs.cpp
nguidotti: Let use merge #1020 first as it splits the expression into multiple ones

cpp/include/cuopt/linear_programming/constants.h
chris-maes: If I understand correctly, if you have less than 2 threads you don't run the early heuristic, if you have less than 8 threads you don't run RINS, if you have less than 3 threads you don't run Batch PDLP.

I think you can eliminate the confusion by changing the names to `REQUIRED_THREAD_COUNT`. I still don't think these constants need to go in this file.

chris-maes: This file is for constants required by users of the C/Python APIs.

cpp/src/branch_and_bound/pseudo_costs.cpp
chris-maes: Please rename and break up expression.

cpp/include/cuopt/linear_programming/constants.h
nguidotti: Ok. Let me find another place to put them.

cpp/src/branch_and_bound/pseudo_costs.cpp
nguidotti: Actually, I implemented the same logic here.

nguidotti:
/ok to test [b0b7827](https://github.com/NVIDIA/cuopt/pull/1099/commits/b0b782705ee2db24b15f0928060c0d5b77d6a7fd)

nguidotti:
/ok to test [a27c642](https://github.com/NVIDIA/cuopt/pull/1099/commits/a27c642b2f19ea68888acdf3a44f70b44215da47)

nguidotti:
/ok to test [373fc64](https://github.com/NVIDIA/cuopt/pull/1099/commits/373fc64811f324e300ff9899bc66957a18812980)

nguidotti:
/ok to test [b318d30](https://github.com/NVIDIA/cuopt/pull/1099/commits/b318d305cac1e3de92386ad4bf7ccf6bd8f739ab)

cpp/src/cuts/cuts.cpp
```diff
@@ -1878,12 +1878,10 @@ bool cut_generation_t<i_t, f_t>::generate_clique_cuts(
                     static_cast<double>(settings.time_limit),
                     static_cast<double>(toc(start_time)));
 
-  if (clique_table_ == nullptr && clique_table_future_ != nullptr &&
-      clique_table_future_->valid()) {
+  if (clique_table_ == nullptr) {
```
akifcorduk, line 1881: We should enter this if-block only once. Setting future to `nullptr` helped there. Now with every cut pass we will enter here and wait for the task that was not launched again.

cpp/src/branch_and_bound/branch_and_bound.cpp
```diff
@@ -2061,29 +2057,25 @@ mip_status_t branch_and_bound_t<i_t, f_t>::solve(mip_solution_t<i_t, f_t>& solut
...
-                  signal_ptr]() -> std::shared_ptr<detail::clique_table_t<i_t, f_t>> {
-                   user_problem_t<i_t, f_t> problem_copy = original_problem_;
-                   cuopt::timer_t timer(std::numeric_limits<double>::infinity());
-                   std::shared_ptr<detail::clique_table_t<i_t, f_t>> table;
-                   detail::find_initial_cliques(
-                     problem_copy, tolerances_for_clique, &table, timer, false, signal_ptr);
-                   return table;
-                 });
+
+#pragma omp task depend(out : *clique_signal) firstprivate(tolerances_for_clique)
```
akifcorduk, line 2080: Shoudn't we check the total number of available threads? And add if block to check if there are enough threads.

cpp/src/branch_and_bound/branch_and_bound.cpp
```diff
@@ -2126,44 +2118,42 @@ mip_status_t branch_and_bound_t<i_t, f_t>::solve(mip_solution_t<i_t, f_t>& solut
   exploration_stats_.total_lp_iters      = root_relax_soln_.iterations;
   exploration_stats_.total_lp_solve_time = toc(exploration_stats_.start_time);
 
-  auto finish_clique_thread = [this]() {
-    if (clique_table_future_.valid()) {
-      signal_extend_cliques_.store(true, std::memory_order_release);
```
akifcorduk, line 2129: Sorry I missed this in my initial review because omp_atomic_t was implemented before. But the way we are doing atomic operations are with sequential consistency which is the most restrictive and performance heavy one. We should implement the memory model arguments(acquire,release,seq_cst, relaxed) to the atomic operations and use it most of the time otherwise we are getting performance hits. The default (seq_cst) is rarely needed.

nguidotti: Implemented in 1f8491b

cpp/src/cuts/cuts.cpp
nguidotti: If there is no previous task, then `taskwait` is no-op, so the overhead is quite small here

akifcorduk: But isn't there error-prone ? If we add another task somewhere in between, this can cause problems in the future right?

akifcorduk (approved):

cpp/src/cuts/cuts.cpp
nguidotti: `#pragma omp taskwait depend(in : *signal_extend_)` this will only wait for a task that has an `inout` or `out` dependency on `*signal_extend_` (the atomic object that signals the clique table to stop).

nguidotti: So, unless you specify another `#pragma omp task depend(out: *signal_extend_)` (besides the one on the clique table), the thread will no stop at this `taskwait`

cpp/include/cuopt/linear_programming/constants.h
nguidotti: It is living on `mip_constant.hpp` now

cpp/src/branch_and_bound/pseudo_costs.cpp
nguidotti: Done

rgsl888prabhu (approved):

chris-maes (approved):
LGTM. Thanks for making the changes to improve and clarify the code.

nguidotti:
/ok to test [d895516](https://github.com/NVIDIA/cuopt/pull/1099/commits/d8955162b2f31d7e6ea5d146f4b0acb41a634767)

nguidotti:
/ok to test [b59764f](https://github.com/NVIDIA/cuopt/pull/1099/commits/b59764f58e5184486157bb71f9779d9cedc23f74)

nguidotti:
/ok to test [c3c0f22](https://github.com/NVIDIA/cuopt/pull/1099/commits/c3c0f22cbeadc932a9311c4c3964b6955e71a408)

nguidotti:
/ok to test [429151b](https://github.com/NVIDIA/cuopt/pull/1099/commits/429151ba0ce0d7f7e84d986cb97f0548c3f1fdf1)

PR #1102, opened 2026-04-14
Add CTest labels for selective C++ test execution

rgsl888prabhu:
## Summary
- Add two CTest labels — `routing` and `numopt` — to all C++ test executables
- Extend `ConfigureTest()` CMake macro with optional `LABELS` keyword (backward compatible)

| Label | Scope |
|-------|-------|
| `routing` | VRP, TSP, distance engine (9 tests) |
| `numopt` | LP, MIP, QP, dual simplex, MPS parser, gRPC, CLI (22 tests) |

### Usage
```bash
ctest -L routing    # only routing tests
ctest -L numopt     # only numerical optimization tests
ctest               # all (unchanged default)
```

## Test plan
- [ ] `cmake --build` succeeds
- [ ] `ctest -L routing` runs only routing tests
- [ ] `ctest -L numopt` runs only numopt tests
- [ ] Full CI passes unchanged


mlubin:
The label choices seem good to me. They will probably evolve a bit over time.

I'm not sure I understand the need for the extra code to generate a `test_labels.txt` manifest. `ctest --show-only=json-v1` prints a json file with all the information we need to map tests to labels.

rgsl888prabhu:
> The label choices seem good to me. They will probably evolve a bit over time.
> 
> I'm not sure I understand the need for the extra code to generate a `test_labels.txt` manifest. `ctest --show-only=json-v1` prints a json file with all the information we need to map tests to labels.

I was thinking to use this for testing so if we are missing any label from testing so it can be raised, but this would be helpful as well. Will remove it for now.

mlubin (approved):

jameslamb (approved):

PR #1103, opened 2026-04-15
Detect MIP symmetry using dejavu; Perform orbital fixing and lexical reduction

chris-maes:
This PR add supports for symmetry detection in MIP. If symmetry is detected we apply orbital fixing and lexical reduction inside branch and bound. 

We construct a colored graph for MIP problems that can be used with a graph automorphism solver (in this case dejavu) to detect symmetries. 

We perform symmetry detection after presolve. Note that presolve may have already destroyed symmetries. But presolve is crucial for good performance. 

We then implement orbital fixing and lexical reduction within the branch and bound tree. 

There is a long symmetry funnel. 

240 MIPLIB instances
-> 59 instances had exploitable symmetry
-> 42 instances had symmetry and were not solved at the root node
-> 33 instances had orbital fixing (12 instances had lexical reduction)
-> 7 instances solved to optimality in 3600 seconds

Overall the change is performance neutral. But we are now able to solve two more problems to optimality: `cod105` and `mzzv42z`

```
Geomean MIP GAP A / (B = basline): 0.99
Geomean Time to Optimal A/B: 1.00
A optimal 23
B optimal 23
A problems with feasible solutions 75
B problems with feasible solutions 76

A wins
                        cod105   210.47   300.00
                       mzzv42z   145.95   300.00
                      seymour1    11.62    13.05
                  neos-1582420     1.13     1.48
                       mcsched    30.57    41.47
A wins  5

A losses
                         neos8     1.10     0.83
                      fast0507   300.00   262.95
                   neos-933966   300.00   223.65
                   neos-957323   115.79   101.15
                         neos5   109.23    62.97
                         mas76     5.54     4.86
                        swath1     1.31     0.90
A losses 7
```


aliceb-nv:
Wow, those are some good numbers indeed 🙂 Thanks a lot Chris!!
I have one question - This is detecting symmetries on the standard-form problem as I understand. Should we rather detect them on the problem_t problem with inequalities? This would make it easier for presolve and/or heuristics to use the generated data, and make the generated graph smaller I think

mlubin:
Maybe @aliceb-nv would be a better reviewer for this than me? (Looks like I got pulled in by auto-assignment)

chris-maes:
> Wow, those are some good numbers indeed 🙂 Thanks a lot Chris!! I have one question - This is detecting symmetries on the standard-form problem as I understand. Should we rather detect them on the problem_t problem with inequalities? This would make it easier for presolve and/or heuristics to use the generated data, and make the generated graph smaller I think

Good question @aliceb-nv , I think it is more difficult to construct the graph to compute the automorphism using the representation in problem_t. In particular the lower and upper bounds on the constraints complicate things.

My strategy was to extract the symmetries in the variables from the symmetries computed on the graph. I think once we have symmetries on the variables, we can use them inside presolve. Note that we may need to turn off some (perhaps all) of presolve if we detect symmetries.

cpp/src/branch_and_bound/symmetry.hpp
```diff
@@ -0,0 +1,317 @@
...
+  std::vector<f_t> nonzeros = problem.A.x;
+  std::vector<i_t> nonzero_perm(nnz);
+  std::iota(nonzero_perm.begin(), nonzero_perm.end(), 0);
+  std::sort(nonzero_perm.begin(), nonzero_perm.end(), [&](i_t a, i_t b) {
+    return nonzeros[a] < nonzeros[b];
+  });
+  std::vector<i_t> nonzero_colors(nnz, -1);
+  i_t edge_color                  = rhs_color + 1;
+  f_t last_nz                     = nonzeros[nonzero_perm[0]];
+  nonzero_colors[nonzero_perm[0]] = edge_color;
```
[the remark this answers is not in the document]
chris-maes: I don't think we can have a problem with zero nonzeros.

cpp/src/branch_and_bound/symmetry.hpp
```diff
@@ -0,0 +1,317 @@
...
+    if (problem.upper[a] != problem.upper[b]) return problem.upper[a] < problem.upper[b];
+    return var_types[a] < var_types[b];
+  });
+  std::vector<i_t> var_colors(problem.num_cols, -1);
+  i_t var_color             = 0;
+  f_t last_obj              = problem.objective[obj_perm[0]];
+  f_t last_lower            = problem.lower[obj_perm[0]];
+  f_t last_upper            = problem.upper[obj_perm[0]];
+  variable_type_t last_type = var_types[obj_perm[0]];
+  var_colors[obj_perm[0]]   = var_color;
```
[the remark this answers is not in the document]
chris-maes: You can't have a problem with zero columns.

cpp/src/branch_and_bound/symmetry.hpp
```diff
@@ -0,0 +1,441 @@
...
+    for (i_t sz = 2; sz <= max_orbit_size; sz++) {
+      if (orbit_histogram[sz] > 0) {
+        settings.log.printf(" %d:%d", sz, orbit_histogram[sz]);
+      }
+    }
+    settings.log.printf("\n");
+
+    has_symmetry = (max_orbit_size >= 4) ||
+                    (num_nontrivial_orbits >= 3 && max_orbit_size >= 2) ||
+                    (total_vars_in_orbits >= 10);
```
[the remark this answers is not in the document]
chris-maes: An original variable can't share an orbit with an auxiliary vertex as these have different colors.

cpp/src/branch_and_bound/branch_and_bound.cpp
```diff
@@ -1388,6 +1391,163 @@ dual::status_t branch_and_bound_t<i_t, f_t>::solve_node_lp(
...
+          worker->leaf_problem.lower[v] = 0.0;
+          worker->leaf_problem.upper[v] = 0.0;
+        }
+        for (i_t v: fix_one) {
+          settings_.log.printf("Orbital fixing at node %d: fixing variable %d to 1\n", node_ptr->node_id, v);
+          worker->leaf_problem.lower[v] = 1.0;
+          worker->leaf_problem.upper[v] = 1.0;
+        }
+      }
+    }
```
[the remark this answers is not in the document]
chris-maes: Yes, it needs to be thread local. I'm working on that. Slow your roll code rabbit :)

cpp/src/branch_and_bound/branch_and_bound.cpp
```diff
@@ -1388,6 +1391,163 @@ dual::status_t branch_and_bound_t<i_t, f_t>::solve_node_lp(
...
+          branched_zero.push_back(node->branch_var);
+          symmetry_->marked_b0[node->branch_var] = 1;
+        } else if (node->branch_var_lower == 1.0) {
+          branched_one.push_back(node->branch_var);
+          symmetry_->marked_b1[node->branch_var] = 1;
+        } else {
+          assert(false); // Unexpected non-binary variable. Only binaries supported in symmetry handling.
+        }
+        node = node->parent;
+      }
```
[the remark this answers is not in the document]
chris-maes: Removed. It works with integer variables now.

cpp/src/branch_and_bound/branch_and_bound.cpp
```diff
@@ -1388,6 +1391,163 @@ dual::status_t branch_and_bound_t<i_t, f_t>::solve_node_lp(
...
+        for (i_t v: fix_zero) {
+          settings_.log.printf("Orbital fixing at node %d: fixing variable %d to 0\n", node_ptr->node_id, v);
+          worker->leaf_problem.lower[v] = 0.0;
+          worker->leaf_problem.upper[v] = 0.0;
+        }
+        for (i_t v: fix_one) {
+          settings_.log.printf("Orbital fixing at node %d: fixing variable %d to 1\n", node_ptr->node_id, v);
+          worker->leaf_problem.lower[v] = 1.0;
+          worker->leaf_problem.upper[v] = 1.0;
+        }
```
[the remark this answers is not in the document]
chris-maes: This is removed.

cpp/src/mip_heuristics/solve.cu
```diff
@@ -303,6 +307,17 @@ mip_solution_t<i_t, f_t> solve_mip(optimization_problem_t<i_t, f_t>& op_problem,
...
+    // Start symmetry detection
+    {
+      detail::problem_t<i_t, f_t> problem(op_problem);
+      dual_simplex::simplex_solver_settings_t<i_t, f_t> simplex_settings;
+      simplex_settings.set_log(true);
+      simplex_settings.time_limit = settings.time_limit;
+      dual_simplex::user_problem_t<i_t, f_t> user_problem =
+        cuopt_problem_to_simplex_problem<i_t, f_t>(op_problem.get_handle_ptr(), problem);
+      dual_simplex::detect_symmetry(user_problem, simplex_settings);
+    }
```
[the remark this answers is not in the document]
chris-maes: Fixed

rgsl888prabhu (approved):
Approving CMAKE changes

cpp/src/branch_and_bound/branch_and_bound.cpp
```diff
@@ -2593,6 +2650,17 @@ mip_status_t branch_and_bound_t<i_t, f_t>::solve(mip_solution_t<i_t, f_t>& solut
   node_queue_.push(search_tree_.root.get_down_child());
   node_queue_.push(search_tree_.root.get_up_child());
 
+  if (symmetry_ != nullptr) {
+    i_t removed = symmetry_->generators.template prune_by_bounds<f_t>(
```
aliceb-nv: Should we prune the inverse generators as well?

cpp/include/cuopt/linear_programming/mip/solver_settings.hpp
```diff
@@ -90,6 +90,7 @@ class mip_solver_settings_t {
   bool heuristics_only          = false;
   i_t reliability_branching     = -1;
   i_t num_cpu_threads           = -1;  // -1 means use default number of threads in branch and bound
+  i_t symmetry                  = -1;
```
aliceb-nv, line 103: Maybe this could be a bitset? We may be adding further symmetry handling techniques in the future, e.g. isomorphism pruning. That would allow users to easily disable symmetry techniques as desired

cpp/src/branch_and_bound/pseudo_costs.cpp
```diff
@@ -1226,7 +1227,95 @@ void strong_branching(const lp_problem_t<i_t, f_t>& original_lp,
...
+    for (i_t j : fractional) {
+      i_t delegate = orbit_delegate[symmetry->orbit_rep[j]];
+      if (j == delegate) continue;
+      pc.pseudo_cost_sum_down[j] = pc.pseudo_cost_sum_down[delegate];
+      pc.pseudo_cost_sum_up[j]   = pc.pseudo_cost_sum_up[delegate];
+      pc.pseudo_cost_num_down[j] = pc.pseudo_cost_num_down[delegate];
+      pc.pseudo_cost_num_up[j]   = pc.pseudo_cost_num_up[delegate];
+    }
+
+    pc.update_pseudo_costs_from_strong_branching(fractional, root_solution.x);
```
aliceb-nv: I think there's going to be an OoB issue here. `pc` resized `strong_branch_down` to `necessary_fractional.size()`, but we're working on `fractional` here

cpp/src/branch_and_bound/symmetry.hpp
```diff
@@ -0,0 +1,1122 @@
...
+
+    // In true orbital fixing we would compute the group G' = stabilizer(G, B1)
+    // Instead we compute a subgroup H of G' using generator filtering, additionally
+    // checking that for each variable in B0, its image is not fixed to 1.
+    // This is necessary because propagation using asymmetric cuts can fix variables
+    // to 1 that the symmetry group maps to 0-branched variables, creating false conflicts.
+    // See van Doornmalen & Hojny (2022), "A Unified Framework for Symmetry Handling".
+    //
+    // We always recompute because propagation at each node can invalidate generators.
+    if (should_recompute) {
```
aliceb-nv, line 404: As the comment says, we should always recompute if F0/F1 fixings were made due to bounds-propagation, right? As it stands it is not unconditional, if we are not at the beginning of the plunge, and we branch on a non-binary variable, we may not recompute generators even if bounds-prop updated F0/F1
I may be completely misunderstanding :)

aliceb-nv (approved):
Amazing work Chris. A few comments, but otherwise LGTM.

Maybe we could consider adding a few unit tests. It seems to me that it would be all too easy to accidentally break the symmetry handling code or its assumptions. Maybe we could test on some files from the Margot datasets, or create some synthetic ones.

cpp/src/branch_and_bound/branch_and_bound.cpp
```diff
@@ -1387,42 +1404,79 @@ dual::status_t branch_and_bound_t<i_t, f_t>::solve_node_lp(
...
-                                                                           lp_start_time,
-                                                                           lp_settings,
-                                                                           worker->leaf_solution,
-                                                                           worker->basis_factors,
-                                                                           worker->basic_list,
-                                                                           worker->nonbasic_list,
-                                                                           leaf_vstatus,
-                                                                           worker->leaf_edge_norms);
+    // Perform orbital fixing
+    auto* of = worker->orbital_fixing.get();
```
nguidotti: NITPICK: `of` --> `orbital_fixing`

nguidotti: Another suggestion: extract this block to a separated method

cpp/src/branch_and_bound/pseudo_costs.cpp
```diff
@@ -1226,7 +1227,95 @@ void strong_branching(const lp_problem_t<i_t, f_t>& original_lp,
...
+  } else {
+    // Use precomputed orbit representatives from the full group G. Variables
+    // in the same orbit have identical branching behavior by symmetry.
+    // Keep only one fractional variable per orbit. We use the first fractional
+    // variable encountered as the orbit's delegate (not the orbit representative,
+    // which might not be fractional itself).
+    // orbit_delegate[rep] maps each orbit representative to the fractional delegate.
+    std::vector<i_t> orbit_delegate(symmetry->num_original_vars, -1);
+    std::vector<i_t> necessary_fractional;
```
nguidotti: NITPICK: `necessary` --> `reduced` (?) or `core` (?). I find adding "necessary" to a variable name or method is little strange

cpp/src/branch_and_bound/symmetry.hpp
```diff
@@ -0,0 +1,1122 @@
...
+#include "dejavu.h"
+
+#include <memory>
+#include <numeric>
+#include <sstream>
+
+namespace cuopt::linear_programming::dual_simplex {
+
+// permutation_t stores a dense permutation plus its support (non-identity entries).
+template <typename i_t>
```
nguidotti, line 26: Maybe enclose this into a separated namespace? `generator_t` can be confused with a random number generator.

cpp/src/branch_and_bound/worker.hpp
```diff
@@ -73,6 +80,20 @@ class branch_and_bound_worker_t {
 
   pcgenerator_t rng;
 
+  std::unique_ptr<orbital_fixing_t<i_t, f_t>> orbital_fixing;
+  std::unique_ptr<lexical_reduction_t<i_t, f_t>> lexical_reduction;
+  mip_symmetry_t<i_t, f_t>* symmetry_ptr = nullptr;
+
+  void ensure_orbital_fixing()
```
nguidotti, line 74: Should this be moved to the constructor?

cpp/src/branch_and_bound/mip_node.hpp
```diff
@@ -293,6 +295,12 @@ class mip_node_t {
 
   std::vector<variable_status_t> vstatus;
 
+  // Cumulative orbital fixing bound changes from root to this node.
+  // Stored so that when a child starts a new plunge, the parent's
+  // orbital fixings can be restored without re-derivation.
+  std::vector<i_t> orbital_fix_zero;
```
nguidotti, line 305: I have a couple of naive questions. Similar to the variables bounds, can the orbital fixing be retrieved by traversing the tree? What is the size of these vectors? Can they be deallocated after the children are created?

nguidotti (commented):
Thanks for the hard work, Chris. I do not fully understand the symmetry algorithms, so I cannot comment much on those. I think merging conflicts with #1149 should be fairly limited

cpp/src/branch_and_bound/worker.hpp
chris-maes: Just to clarify are you suggesting moving `ensure_orbital_fixing()` into the constructor?

cpp/src/branch_and_bound/symmetry.hpp
chris-maes: How about I rename to `group_generator_t`, to distinguish it as a generator of a group.

cpp/src/branch_and_bound/pseudo_costs.cpp
chris-maes: I can rename to `reduced_fractional`

cpp/src/branch_and_bound/mip_node.hpp
chris-maes: Orbital fixing are expensive to compute, because they require us to generate the stabilizer of a group. We have to ensure that we always have the same orbital fixings in a child as a parent, otherwise the dual bound will not be monotonically increasing (non-decreasing really). 

We can't deallocate the vectors associated with a node, until all the nodes in the subtree rooted at that node have been removed from the heap.

cpp/src/branch_and_bound/branch_and_bound.cpp
chris-maes: Unlike the generators in orbital fixing, the inverse generators in lexical reduction can not be pruned as we move deeper in the tree.

cpp/include/cuopt/linear_programming/mip/solver_settings.hpp
chris-maes: Yeah, we can have different integers map to different settings.

cpp/src/branch_and_bound/pseudo_costs.cpp
chris-maes: This was addressed in a previous commit.

cpp/src/branch_and_bound/mip_node.hpp
chris-maes: In other words, if we want the dual bound to grow, we either have to store them or pay a big computational price to recompute them.

cpp/src/branch_and_bound/symmetry.hpp
chris-maes: Yes, we don't need to recompute in the case you describe. The symmetries are restricted to binary variables. So branching on a non-binary variable does not change them.

cpp/src/branch_and_bound/pseudo_costs.cpp
chris-maes: Done

cpp/src/branch_and_bound/symmetry.hpp
chris-maes: Done

cpp/src/branch_and_bound/branch_and_bound.cpp
chris-maes: Done.

cpp/src/branch_and_bound/worker.hpp
chris-maes: I think it needs to be lazy because we don't want to use it for diving workers.

PR #1105, opened 2026-04-15
Add QCMATRIX: parsing and quadratic-constraint model

yuwenchen95:
## Description
This change adds support for reading **QCMATRIX** blocks from MPS/QPS-style input and representing each quadratic constraint matrix in CSR form on `mps_data_model_t`.
**Included**
- Parse `QCMATRIX` (fixed and free format), validate row/variable names, and reject duplicate blocks for the same constraint row.
- Store quadratic constraint matrices via `append_quadratic_constraint_matrix` / `get_quadratic_constraint_matrices`, plus `has_quadratic_constraints()`.
- Clarify objective quadratic scaling vs constraint quadratic scaling in parser comments (objective uses ½ scaling path; `QCMATRIX` uses symmetric CSR without the extra ½ factor).
- Unit test `qps_parser.qcmatrix_append_api` for the model append API.
**Not included (follow-up work)**
- Solver / downstream algorithm consumption of quadratic constraints beyond ingestion in the parser model.
## Issue

N/A — incremental parser/model support; follow-up PR(s) will wire algorithms.
## Checklist
- Testing
   - [] New or existing tests cover these changes
   - [x] Added tests
   - [ ] NA
- Documentation
   - [] The documentation is up to date with these changes
   - [ ] Added new documentation
   - [x ] NA


akifcorduk:
/ok to test [a114099](https://github.com/NVIDIA/cuopt/pull/1105/commits/a114099aad48a8a4d44b63245555335fb5cdeb66)

rg20:
@yuwenchen95  do these changes include the mps writer as well?

yuwenchen95:
> @yuwenchen95 do these changes include the mps writer as well?

It's only relevant to the change for the constructor of `mps_parser_t`. I haven't touched the writer file for MPS yet.

rg20:
> > @yuwenchen95 do these changes include the mps writer as well?
> 
> It's only relevant to the change for the constructor of `mps_parser_t`. I haven't touched the writer file for MPS yet.

Can you add the code to write the mps file as well?

cpp/libmps_parser/tests/mps_parser_test.cpp
```diff
@@ -855,6 +855,51 @@ TEST(qps_parser, quadratic_objective_basic)
   EXPECT_EQ(1.0, model.get_quadratic_objective_values()[1]);
 }
 
+// ================================================================================================
```
chris-maes, line 861: Please add a test that parses a .MPS file containing a QCMATRIX section.

chris-maes (commented):
You should probably add support for QCMAATRIX to the MPS writer as well. That way you can verify that you can write out and read in the same problem.

cpp/libmps_parser/include/mps_parser/mps_data_model.hpp
```diff
@@ -262,6 +262,40 @@ class mps_data_model_t {
                                       const i_t* Q_offsets,
                                       i_t size_offsets);
 
+  /**
+   * @brief CSR of Q for one quadratic constraint (MPS QCMATRIX).
+   *
+   * @c constraint_row_index is the row index in the linear constraint matrix A (0-based),
+   * matching the order of non-objective rows in the ROWS section.
+   */
+  struct quadratic_constraint_matrix_t {
```
chris-maes: I think this should be `quadratic_constraint_t` and include the linear term, sense, and rhs.

cpp/libmps_parser/include/mps_parser/mps_data_model.hpp
```diff
@@ -262,6 +262,40 @@ class mps_data_model_t {
...
+   *
+   * @param constraint_row_index Row index in A (0-based), matching non-objective ROWS order.
+   * @param[in] Qc_values Values of the CSR representation; copied into the model.
+   * @param size_values Size of the Qc_values array.
+   * @param[in] Qc_indices Indices of the CSR representation; copied into the model.
+   * @param size_indices Size of the Qc_indices array.
+   * @param[in] Qc_offsets Offsets of the CSR representation; copied into the model.
+   * @param size_offsets Size of the Qc_offsets array.
+   */
+  void append_quadratic_constraint_matrix(i_t constraint_row_index,
```
chris-maes: I think this should be `append_quadratic_constraint` and you should provide the matrix, linear term, rhs, and sense.

chris-maes (changes requested):
Thanks for the PR @yuwenchen95 .I think the data model should deal with quadratic constraints. Not quadratic constraint matrices. The parser needs to sort out which row corresponds to the linear term and the sense and rhs, and feed this to the data model. 

The parser also needs to make sure that the row associated with the linear term of a quadratic constraint doesn't get pulled into the A matrix accidently.

yuwenchen95:
Now I have added the complete support for reading quadratic constraints and writing functions. @rg20 @chris-maes

cpp/include/cuopt/linear_programming/cpu_optimization_problem.hpp
```diff
@@ -41,6 +41,8 @@ class mip_solution_interface_t;
 template <typename i_t, typename f_t>
 class cpu_optimization_problem_t : public optimization_problem_interface_t<i_t, f_t> {
  public:
+  using typename optimization_problem_interface_t<i_t, f_t>::mps_quadratic_constraint_t;
```
chris-maes: It's a bit strange to refer to mps (a file format) inside the description of the problem. Probably better to create a class or struct named simply `quadratic_constraint_t`

cpp/include/cuopt/linear_programming/cpu_optimization_problem.hpp
```diff
@@ -113,6 +115,17 @@ class cpu_optimization_problem_t : public optimization_problem_interface_t<i_t,
   const std::vector<f_t>& get_quadratic_objective_values() const override;
   bool has_quadratic_objective() const override;
 
+  void set_quadratic_constraints(std::vector<mps_quadratic_constraint_t> constraints) override;
```
chris-maes: Should the argument be: `std::vector<quadratic_constraint_t>& constraints`? That is do you want to pass this in by reference instead of copying?

cpp/include/cuopt/linear_programming/cpu_optimization_problem.hpp
```diff
@@ -113,6 +115,17 @@ class cpu_optimization_problem_t : public optimization_problem_interface_t<i_t,
...
+  bool has_quadratic_constraints() const override;
+  const std::vector<mps_quadratic_constraint_t>& get_quadratic_constraints() const override;
+  // Additional methods for MPS export if quadratic constraints are present
+  void set_linear_constraint_mps_indices(std::vector<i_t> indices) override;
```
chris-maes: I'm not sure these should have mps in there names.

cpp/include/cuopt/linear_programming/cpu_optimization_problem.hpp
```diff
@@ -185,6 +198,12 @@ class cpu_optimization_problem_t : public optimization_problem_interface_t<i_t,
   std::vector<i_t> Q_indices_;
   std::vector<f_t> Q_values_;
 
+  std::vector<mps_quadratic_constraint_t> quadratic_constraints_{};
+
+  std::vector<i_t> linear_constraint_mps_indices_{};
```
chris-maes: What is this used for? linear constraint indices is a bit vauge of a name

chris-maes (changes requested):
Please consolidate all references to MPS to inside the MPS parser. Once we create an optimization problem, we should not refer to MPS. This is because we will add support for other formats (e.g. LP) and also because our problem abstraction is not tied to MPS.

I'm not clear on why there are extra functions for the linear rows etc. I think you should encapsulate this all inside the MPS parser. The job of the MPS parser is to parse the MPS file and construct an optimization problem representation. The final problem representation should not need to know that in the MPS file a certain row was used to represent the linear term in the quadratic constraint.

cpp/include/cuopt/linear_programming/optimization_problem.hpp
```diff
@@ -259,7 +264,16 @@ class optimization_problem_t : public optimization_problem_interface_t<i_t, f_t>
   const std::vector<i_t>& get_quadratic_objective_offsets() const override;
   const std::vector<i_t>& get_quadratic_objective_indices() const override;
   const std::vector<f_t>& get_quadratic_objective_values() const override;
+  const std::vector<mps_quadratic_constraint_t>& get_quadratic_constraints() const override;
   bool has_quadratic_objective() const override;
+  bool has_quadratic_constraints() const override;
+
+  void set_linear_constraint_mps_indices(std::vector<i_t> indices) override;
```
chris-maes: What does this do?

yuwenchen95: I removed redundant  linear indexing.

cpp/include/cuopt/linear_programming/cpu_optimization_problem.hpp
yuwenchen95: Now, I achieve the separation by having duplicate definition of `quadratic_constraint_t ` in `mps_parser::mps_data_model_t` and `optimization_problem_interface_t`. Is it ok?

cpp/libmps_parser/src/mps_parser.cpp
```diff
@@ -272,21 +271,30 @@ ObjSenseType convert_to_obj_sense(const std::string& str)
 template <typename i_t, typename f_t>
 void mps_parser_t<i_t, f_t>::fill_problem(mps_data_model_t<i_t, f_t>& problem)
 {
+  // count the row indices that are quadratic constraints
+  std::unordered_set<i_t> quadratic_row_ids{};
```
chris-maes, line 277: Do we need an unorder_set here? If we know the number of rows already we can just allocate a vector of length num_rows?

cpp/libmps_parser/src/mps_parser.cpp
```diff
@@ -485,9 +510,7 @@ void mps_parser_t<i_t, f_t>::fill_problem(mps_data_model_t<i_t, f_t>& problem)
 
     for (i_t row = 0; row < num_rows; ++row) {
       for (const auto& [col, val] : csr_data[row]) {
-        // While the mps format expects to optimize for 0.5 xT Q x, cuopt optimizes for xT Q x
```
chris-maes, line 488: Probably best to leave this comment in. And add that 0.5 is a common value_scale.

cpp/libmps_parser/tests/mps_parser_test.cpp
```diff
@@ -855,6 +858,173 @@ TEST(qps_parser, quadratic_objective_basic)
...
+
+  EXPECT_EQ(1, qcs[1].constraint_row_index);
+  EXPECT_EQ("QC1", qcs[1].constraint_row_name);
+  EXPECT_EQ('L', qcs[1].constraint_row_type);
+  EXPECT_EQ(qc1_linear_values, qcs[1].linear_values);
+  EXPECT_EQ(qc1_linear_indices, qcs[1].linear_indices);
+  EXPECT_EQ(10.0, qcs[1].rhs_value);
+  EXPECT_EQ(qc1_values, qcs[1].quadratic_values);
+  EXPECT_EQ(qc1_indices, qcs[1].quadratic_indices);
+  EXPECT_EQ(qc1_offsets, qcs[1].quadratic_offsets);
```
chris-maes, line 944: In this test it would be a good idea to write the MPS file out, read it in again, and verify the in memory representation of the original problem, matched the in memory representation of the file you read in.

chris-maes: Ah I guess that is covered in the roundtrip test below

chris-maes (approved):
LGTM! Thanks for all the hard work @yuwenchen95

cpp/libmps_parser/src/mps_parser.cpp
yuwenchen95: I used it since names are not required to be in sequence, e.g. QC0..QCN, so using under_set is only for searching efficiency during parsing.

cpp/include/cuopt/linear_programming/optimization_problem_interface.hpp
```diff
@@ -56,8 +56,59 @@ class optimization_problem_interface_t {
...
+    std::vector<i_t> quadratic_offsets{};
+  };
+
   virtual ~optimization_problem_interface_t() = default;
 
+  /**
+   * @brief Store quadratic constraints for MPS round-trip (linear + Q parts per QC row).
+   * @note Default implementation ignores; GPU/CPU implementations persist for write_to_mps.
+   */
+  virtual void set_quadratic_constraints(std::vector<quadratic_constraint_t> constraints)
```
mlubin: It's better practice for virtual functions to have no default implementation; see all the other virtual members of this class. Is there a reason to break the pattern here?

cpp/include/cuopt/linear_programming/optimization_problem_interface.hpp
```diff
@@ -56,8 +56,59 @@ class optimization_problem_interface_t {
...
+    set_quadratic_constraints(std::move(converted_constraints));
+  }
+
+  /** @brief Whether quadratic constraint metadata is present (for MPS export). */
+  virtual bool has_quadratic_constraints() const = 0;
+
+  /** @brief Quadratic constraints for MPS export (empty if none). */
+  virtual const std::vector<quadratic_constraint_t>& get_quadratic_constraints() const
+  {
+    static const std::vector<quadratic_constraint_t> k_empty{};
```
mlubin: Why use a static variable here? I think we can return a new empty vector instead.

cpp/libmps_parser/include/mps_parser/mps_data_model.hpp
```diff
@@ -262,6 +262,51 @@ class mps_data_model_t {
...
+
+  /**
+   * @brief Append one complete quadratic constraint (row + linear + rhs + quadratic Q).
+   * @param constraint_row_type MPS ROWS type; must be 'L'. 'G' and 'E' quadratic rows are not
+   *        supported.
+   */
+  void append_quadratic_constraint(i_t constraint_row_index,
+                                   const std::string& constraint_row_name,
+                                   char constraint_row_type,
+                                   const f_t* linear_values,
```
mlubin, line 302: It's unnecessarily unsafe to take raw pointers as inputs when they're vectors. Can we use real vectors or spans instead?

cpp/libmps_parser/tests/mps_parser_test.cpp
```diff
@@ -855,6 +858,173 @@ TEST(qps_parser, quadratic_objective_basic)
...
+}
+
+// QCQP MPS: each quadratic constraint bundles row + linear + rhs + quadratic.
+TEST(qps_parser, qcmatrix_mps_linear_rhs_and_bounds)
+{
+  if (!file_exists("qcqp/QC_Test_1.mps")) {
+    GTEST_SKIP() << "qcqp/QC_Test_1.mps not in dataset root";
```
mlubin, line 951: Where is this file? Shouldn't it be checked into the repo? If so, why would it be missing?

cpp/libmps_parser/tests/mps_parser_test.cpp
```diff
@@ -855,6 +858,173 @@ TEST(qps_parser, quadratic_objective_basic)
...
+  ASSERT_EQ(2u, qcs[1].linear_values.size());
+  EXPECT_EQ(3.0, qcs[1].linear_values[0]);
+  EXPECT_EQ(1.0, qcs[1].linear_values[1]);
+  EXPECT_DOUBLE_EQ(10.0, qcs[1].rhs_value);
+}
+
+TEST(qps_parser, qcqp_p0033_mps_sections)
+{
+  if (!file_exists("qcqp/p0033_qc1.mps")) {
+    GTEST_SKIP() << "qcqp/p0033_qc1.mps not in dataset root";
```
mlubin, line 1003: Are we sure we want to skip the test? Silently skipping the test means we don't know if it's covered by CI.

cpp/libmps_parser/tests/mps_parser_test.cpp
```diff
@@ -1127,4 +1329,27 @@ TEST(mps_roundtrip, quadratic_programming_qp_test_2)
   std::filesystem::remove(temp_file);
 }
 
+TEST(mps_roundtrip, qcqp_p0033_qc1)
+{
+  if (!file_exists("qcqp/p0033_qc1.mps")) {
+    GTEST_SKIP() << "Test file not found";
```
mlubin: Same comment here about skipping.

mlubin (changes requested):
Just did a pass for style concerns.

cpp/include/cuopt/linear_programming/optimization_problem_interface.hpp
mlubin: Is it intended for this function to take `constraints` by value?

yuwenchen95: Good point. I will fix it.

cpp/libmps_parser/tests/mps_parser_test.cpp
yuwenchen95: I'm sure about it since I'm just follow the same code style for tests of quadratic cost. What's your opinion? @chris-maes @rg20

yuwenchen95:
> Just did a pass for style concerns.

Thank you for useful advice!

yuwenchen95:
It seems to fail a lot of CI tests when I used std::span in `/cpp/libmps_parser` because of build issues on some platforms.

mlubin (approved):
Approved assuming we'll replace (pointer, len) arguments with spans shortly after landing.

PR #1107, opened 2026-04-15
add code generation for grpc boilerplate conversions

tmckayus:
This change adds a code generator written in Python which generates the protobuf specifications and the code to map between cuopt problem, solution, and settings structures and protobuf messages. The protoc compiler is still used to build the serialization/deserialziation code and the grpc interface -- only the high-level .proto files are generated from the yaml description.

The code generator reads a lightweight yaml file which describes the cuopt structures and generates mapping functions which are included in the grpc client and server implementations.

When a field is added/changed in the problem, solution, or settings structures, a developer updates the yaml file and re-runs the code generator, then reviews the updated code.

A README.md file in the codegen directory describes the yaml format, defaults, and conventions. As much as possible has been left to defaults or naming conventions to make the yaml easy to maintain.


build.sh
```diff
@@ -14,10 +14,11 @@ ARGS=$*
 REPODIR=$(cd "$(dirname "$0")"; pwd)
 LIBCUOPT_BUILD_DIR=${LIBCUOPT_BUILD_DIR:=${REPODIR}/cpp/build}
 
-VALIDARGS="clean libcuopt cuopt_grpc_server cuopt cuopt_server cuopt_sh_client docs deb -a -b -g -fsanitize -tsan -msan -v -l= --verbose-pdlp --build-lp-only  --no-fetch-rapids --skip-c-python-adapters --skip-tests-build --skip-routing-build --skip-grpc-build --skip-fatbin-write --host-lineinfo [--cmake-args=\\\"<args>\\\"] [--cache-tool=<tool>] -n --allgpuarch --ci-only-arch --show_depr_warn -h --help"
+VALIDARGS="clean codegen libcuopt cuopt_grpc_server cuopt cuopt_server cuopt_sh_client docs deb -a -b -g -fsanitize -tsan -msan -v -l= --verbose-pdlp --build-lp-only  --no-fetch-rapids --skip-c-python-adapters --skip-tests-build --skip-routing-build --skip-grpc-build --skip-fatbin-write --host-lineinfo [--cmake-args=\\\"<args>\\\"] [--cache-tool=<tool>] -n --allgpuarch --ci-only-arch --show_depr_warn -h --help"
```
ramakrishnap-nv, line 17: Think we should make codegen default, is it by default enabled ?

ramakrishnap-nv (changes requested):
Have few questions, 

Are we using these anywhere docs/cuopt/grpc/GRPC_CODE_GENERATION.md docs/cuopt/grpc/GRPC_INTERFACE.md.

If there are developer docs, may be they should live in grpc folder so it is used as README there.

build.sh
tmckayus: it's not, currently you have to run it as an explicit target. You should know though, since field_registry.yaml has to change.  What do you think?

tmckayus:
> Have few questions,
> 
> Are we using these anywhere docs/cuopt/grpc/GRPC_CODE_GENERATION.md docs/cuopt/grpc/GRPC_INTERFACE.md.
> 
> If there are developer docs, may be they should live in grpc folder so it is used as README there.

moved to cpp/src/grpc folder. These are primarily dev docs, for someone working on the embedded client or writing a standalone one, or for someone working on the code generation.

build.sh
ramakrishnap-nv: May be we can make this as a pre-commit hook or change file check to see if the file is same or different, but we can work on this as follow-up.

ramakrishnap-nv (approved):

build.sh
tmckayus: update: we've decided to not enable this by default in build.sh, we can revisit in the future if need be

cpp/src/grpc/GRPC_CODE_GENERATION.md
```diff
@@ -0,0 +1,720 @@
...
+    --output-dir cpp/src/grpc/codegen/generated
+```
+
+The generator runs in ~100ms with no external dependencies beyond PyYAML (ships
+with conda). The `--auto-number` and `--strip` options additionally require
+`ruamel.yaml` (listed in the project's development dependencies). Run it
+explicitly via `./build.sh codegen` after editing `field_registry.yaml`, then
+commit the regenerated files.
+
+## File Layout
```
mlubin, line 26: What keeps this file layout in sync with field_registry.yaml? Is the file layout hard coded? If not, we shouldn't hardcode it in the docs.

cpp/src/grpc/GRPC_CODE_GENERATION.md
```diff
@@ -0,0 +1,720 @@
...
+running the generator. Run `./build.sh codegen` to re-generate them after
+editing `field_registry.yaml`; CI verifies they stay in sync. CMake adds
+`cpp/src/grpc/codegen/generated` to the include path for both targets, so the `.inc`
+files are found at compile time with no copy step.
+
+---
+
+## Registry Structure Overview
+
+`field_registry.yaml` has these top-level sections:
```
mlubin, line 85: Are these sections hardcoded? What keeps this doc in sync when a new section is added?

cpp/tests/linear_programming/grpc/grpc_client_test.cpp
```diff
@@ -1665,6 +1665,10 @@ TEST(MapperRoundtrip, MIPSettingsAllFields)
   orig.num_gpus        = 2;
   orig.presolver       = presolver_t::Default;
   orig.mip_scaling     = true;
+  orig.symmetry        = 2;  // orbital fixing + lexical reduction
+
+  // Semi-continuous variables
+  orig.semi_continuous_big_m = 7.5e9;  // not the default 1e10, to detect overwrite-on-decode
```
mlubin, line 1671:
```suggestion
  orig.semi_continuous_big_m = 7.5e9;  // note the default 1e10, to detect overwrite-on-decode
```

cpp/tests/linear_programming/grpc/grpc_client_test.cpp
```diff
@@ -1980,3 +2060,202 @@ TEST(MapperRoundtrip, PDLPSettingsIterationLimitSentinel)
...
 }
+
+// ───────────────────────────────────────────────────────────────────────────
+// Proto3 `optional` presence handling
+// ───────────────────────────────────────────────────────────────────────────
+//
+// A handful of bool settings have a C++ default of `true` but live on the wire
+// in a proto3 message. Without `optional`, an omitted field decodes as the
+// proto3 zero (`false`) and the mapper would silently overwrite the C++
+// default. The codegen now emits `optional <type>` for these fields and
```
mlubin:
```suggestion
// default. The codegen emits `optional <type>` for these fields and
```

mlubin (approved):
Style-wise I'm a bit concerned that we're trading one maintenance burden for another. `generate_conversions.py` hardcodes a lot of cuOpt-specific logic, and one probably needs to use an LLM to properly make changes to it. Maybe we expect this to be rare enough that the trade-off is worth it. There might be other approaches to reduce the need for boilerplate, like using the proto enums as ground truth in the C++ code, which is the standard pattern at Google. Or, how far could we get by generating extra code from raw proto files instead of introducing a custom yaml format? No blocking concerns from the engine side so I'll approve.

cpp/src/grpc/GRPC_CODE_GENERATION.md
tmckayus: agreed, thanks. Will fix in follow on.

tmckayus:
> Style-wise I'm a bit concerned that we're trading one maintenance burden for another. `generate_conversions.py` hardcodes a lot of cuOpt-specific logic, and one probably needs to use an LLM to properly make changes to it. Maybe we expect this to be rare enough that the trade-off is worth it. There might be other approaches to reduce the need for boilerplate, like using the proto enums as ground truth in the C++ code, which is the standard pattern at Google. Or, how far could we get by generating extra code from raw proto files instead of introducing a custom yaml format? No blocking concerns from the engine side so I'll approve.

thanks @mlubin. I think churn will be rare, but I can also take a look at generating from protos with protoc plugins. We could pretty easily switch in the future.

cpp/src/grpc/GRPC_CODE_GENERATION.md
tmckayus: will review and potentially revise in a follow on, thanks

cpp/tests/linear_programming/grpc/grpc_client_test.cpp
tmckayus: "not" is actually the intent here, since this is testing the decode logic to make sure that we are respecting explicitly set values (since we made changes around handling of 'optional' fields_

PR #1111, opened 2026-04-16
Fixed lower bound in single thread mode

nguidotti:
This PR fixes the incorrect computation of the lower bound when running with just a single thread. In the `single_threaded_solve`, we were creating a separated worker instance from the pool, such that its lower bounds was never consider when computing the global lower bound. This PR  simply change the routine to use a worker from the pool.

## Checklist

- Testing
   - [ ] Added tests
   - [ ] NA
- Documentation
   - [ ] Added new documentation
   - [ ] NA

### DIFF
--- a/cpp/src/branch_and_bound/branch_and_bound.cpp
+++ b/cpp/src/branch_and_bound/branch_and_bound.cpp
@@ -299,10 +299,11 @@ branch_and_bound_t<i_t, f_t>::branch_and_bound_t(
 template <typename i_t, typename f_t>
 f_t branch_and_bound_t<i_t, f_t>::get_lower_bound()
 {
-  f_t lower_bound      = lower_bound_ceiling_.load();
-  f_t heap_lower_bound = node_queue_.get_lower_bound();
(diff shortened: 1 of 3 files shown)


mlubin:
Any tests worth adding? Maybe the instance that triggered the issue?

nguidotti:
Good idea. @chris-maes I can use the instance that you send?

chris-maes (approved):
Thanks for the quick fix @nguidotti

chris-maes:
> Good idea. @chris-maes I can use the instance that you send?

Yes. I think it's fine to use. It's a sample problem from page 12 of: https://coral.ise.lehigh.edu/~pubs/files/ostrowski_phd_thesis.pdf

nguidotti:
/ok to test [70bdbf4](https://github.com/NVIDIA/cuopt/pull/1111/commits/70bdbf426468e3a1af4c36645ec13bb67036da9e)

rgsl888prabhu (approved):

PR #1116, opened 2026-04-16
fix: add retry logic and error handling to dataset download scripts

rgsl888prabhu:
## Summary
- Add retry with exponential backoff (3 attempts) to `download()` in both dataset scripts
- Add 60s connection timeout to prevent hanging on unreachable hosts
- Clean up partial files on download failure
- Continue downloading remaining datasets when one fails — report all failures at the end with non-zero exit code
- Remove insecure SSL verification bypass for plato.asu.edu (site has valid cert, fixes SonarQube `python:S4830` and `python:S5527`)

### Before
A single unreachable host (e.g. `old.sztaki.hu`) crashes the entire download batch — no other datasets get downloaded.

### After
Failed downloads are retried 3 times with backoff, then skipped. All other datasets continue downloading. Failures are summarized at the end.

## Test plan
- [ ] Run `bash datasets/linear_programming/download_pdlp_test_dataset.sh` — verify plato.asu.edu downloads succeed and unreachable hosts are retried then skipped
- [ ] Run `python regression/get_datasets.py <dir> lp` — verify retry and summary behavior
- [ ] Verify non-zero exit code when downloads fail

### DIFF
--- a/benchmarks/linear_programming/utils/get_datasets.py
+++ b/benchmarks/linear_programming/utils/get_datasets.py
@@ -2,6 +2,8 @@
 # SPDX-License-Identifier: Apache-2.0
 
 import os
+import sys
+import time
 import argparse
 import urllib.request
 import urllib.parse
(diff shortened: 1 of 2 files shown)


Iroy30 (approved):
lgtm

PR #1118, opened 2026-04-16
ci(pr): gate builds on changed-files and broaden safe exclusions

rgsl888prabhu:
## Summary

Reduces PR CI time by skipping build jobs when no downstream test or doc consumer needs them, and by excluding more files that never affect builds or tests.

**Build gating (\`if:\`):**
- \`conda-cpp-build\`: \`test_cpp ∨ test_python_conda ∨ build_docs\`
- \`conda-python-build\`: \`test_python_conda ∨ build_docs\`
- All 5 \`wheel-build-*\`: \`test_python_wheels\`

**New exclusions (applied to all 4 changed-file groups):**
- \`**/*.md\` (docs are \`.rst\`), \`LICENSE\`, \`.gitattributes\`, \`.gitignore\`, \`.clang-format\`, \`.github/release.yml\`
- \`ci/thirdparty-testing/**\` (only invoked by \`nightly.yaml\`)
- \`.github/workflows/*.yaml\` files not reachable from \`pr.yaml\` (\`build.yaml\`, \`nightly.yaml\`, \`cloud_ci.yaml\`, etc.) — \`self_hosted_service_test.yaml\` is deliberately kept

## Safety

Each group still begins with \`'**'\` (include everything). \`!\` lines only remove paths, so any new file that isn't explicitly excluded triggers full CI — fail-closed default preserved.

## Expected impact

| PR touches only… | Before | After |
|---|---|---|
| Markdown / LICENSE / .clang-format | full build+test matrix | \`checks\` only |
| \`ci/thirdparty-testing/**.sh\` | full build+test matrix | \`checks\` only |
| Nightly-only workflow files | full build+test matrix | \`checks\` only |
| \`notebooks/**\` or \`docs/**/*.rst\` | full matrix | skips 5 wheel builds + wheel tests |
| \`cpp/**\` or \`python/**\` | unchanged | unchanged |


.github/workflows/pr.yaml
```diff
@@ -277,13 +342,38 @@ jobs:
           - '!.cursor-plugin/**'
           - '!.claude-plugin/**'
           - '!gemini-extension.json'
+          # Doc files and meta that never affect builds/tests.
```
jameslamb: Can you please work these into the lists to maintain alphabetical order? These lists are very long and having them be out of order makes them harder for humans to read.

jameslamb (approved):
Approving, as I don't think this will cause any correctness problems, but ... I want to mention some reasons why "build unconditionally" is beneficial, in case you haven't considered them:

1. keeps `sccache` cache fresh (more PRs = faster building for everyone)
2. keeps CI scripts simpler (no need to have checks for whether or not a given run generated artifacts)
3. amortizes the work of detecting and reacting to build issues over time (instead of relying on nightlies to catch them)

I also think it's worth reflecting on the cumulative complexity here... the logic for which CI runs when is getting really complicated in this project (I am thinking about changes like #931 coming too).

If you think the time / resource savings from these changes are worth that complexity and maintenance burden, I won't get in the way.

PR #1120, opened 2026-04-17
Add LP format reader; accept .lp wherever .mps is accepted

### LINKED ISSUE #543: [FEA] Add support for reading .lp files
LP problems are often provided in .lp format, which is not currently supported by cuOpt.
File format reference: https://www.fico.com/fico-xpress-optimization/docs/dms2021-01/solver/optimizer/HTML/chapter10_sec_section102.html

mlubin:
### What

A new LP-format parser sits alongside the existing MPS parser. Every
cuOpt entry point that takes an input file (the `cuopt_cli` CLI,
`cuOptReadProblem` in the C API, the Python `ParseLp` / `ParseMps`
wrappers, and the self-hosted client) now accepts LP files via a
case-insensitive extension dispatcher:

| Extension | Parser |
|---|---|
| `.lp`, `.lp.gz`, `.lp.bz2` | LP |
| `.mps`, `.mps.gz`, `.mps.bz2`, `.qps`, `.qps.gz`, `.qps.bz2` | MPS / QPS |
| anything else | rejected |

The unified entry point is `parse_problem<i_t, f_t>(path)` in
`cuopt/linear_programming/io/parser.hpp`. `parse_lp` and
`parse_lp_from_string` (string-input counterpart, mirroring
`parse_mps_from_string`) are exposed alongside their MPS siblings.

### LP format scope

The parser accepts the conventional LP dialect implemented by most
commercial optimization solvers (not the lpsolve variant):

- **LP, MIP, QP**.
- **Semi-continuous variables** via a `Semi-Continuous` section; each
  listed variable requires a finite upper bound.
- **Quadratic constraints (QCQP)**, `<=` only. Quadratic terms appear
  inside `[ ... ]`. The convention differs between objective and
  constraint:
  - Objective bracket **must** be followed by `/ 2` (LP file uses the
    `0.5 x^T Q x` convention).
  - Constraint bracket **must not** be followed by `/ 2` (coefficients
    at face value, `x^T Q x`).

SOS constraints, piecewise-linear objectives, general constraints, and
user cuts raise a clear `ValidationError`.

Compressed inputs (`.lp.gz`, `.lp.bz2`) are handled by the same
dlopen-based zlib / libbz2 path that already serves `.mps.gz` /
`.mps.bz2`; the helper was extracted to `cpp/src/io/file_to_string.{hpp,cpp}`.

 ### Python API changes (no release impact)

  The Python subpackage previously located at
  `cuopt.linear_programming.mps_parser` has been renamed to
  `cuopt.linear_programming.io` (mirroring the C++ `cpp/include/cuopt/linear_programming/io/` layout). The decorator
  `catch_mps_parser_exception` is
  renamed to `catch_io_exception`. The user-facing top-level imports
  (`cuopt.linear_programming.ParseLp`, `cuopt.linear_programming.ParseMps`)
  are unchanged.

  This package was introduced on this branch in #1193 (post-26.04) and
  has not appeared in any release, so the rename is safe without a
  compatibility shim.

### Notable cleanups along the way

- `cython_mps_parser.{hpp,cpp}` → `cython_parser.{hpp,cpp}` (now exposes
  both `call_parse_mps` and `call_parse_lp`).
- `mps_parser_test.cpp` → `parser_test.cpp` (consolidated; covers both
  parsers and the dispatcher).
- Error messages and the `CUOPT_MPS_FILE_ERROR` substring check
  generic-ified from "MPS file" → "input file".

### Docs

User-facing docs (`cuopt-cli`, `cuopt-c/lp-qp-milp`) state the supported
extensions and link to the API reference. The full LP-format spec
(scope, conventions, rejections) lives in the `parse_lp` /
`parse_lp_from_string` docstrings in `parser.hpp` and the `ParseLp`
docstring in `parser.py`.

Closes #543


cpp/src/io/lp_parser.cpp
```diff
@@ -0,0 +1,1088 @@
...
+  if (is_end_keyword(lower)) return true;
+
+  // Multi-word section headers: "Subject To", "Such That", "Lazy Constraints",
+  // and the unsupported "User Cuts" / "General Constraints".
+  const LpToken& t2 = peek(1);
+  if (lower == "subject" && name_equals_ci(t2, "to")) return true;
+  if (lower == "such" && name_equals_ci(t2, "that")) return true;
+  if (lower == "st" || lower == "s.t.") return true;
+  if (lower == "lazy" && name_equals_ci(t2, "constraints")) return true;
+  if (lower == "user" && name_equals_ci(t2, "cuts")) return true;
```
chris-maes, line 601: We should not support "user cuts"

cpp/src/io/lp_parser.cpp
```diff
@@ -0,0 +1,1088 @@
...
+  // "General Constraints" is unsupported but still a section boundary;
+  // reject_unsupported_section() throws before we'd keep parsing.
+  if (lower == "general" && name_equals_ci(t2, "constraints")) return true;
+
+  // Unsupported single-token sections.
+  if (lower == "sos") return true;
+  if (lower == "semi" && peek(1).kind == LpTokenKind::Minus &&
+      name_equals_ci(peek(2), "continuous"))
+    return true;
+  if (lower == "sc") return true;
```
chris-maes: We probably should support semicontinuous variables

mlubin: That's easy to add, though I'd rather let https://github.com/NVIDIA/cuopt/pull/1096 land first to avoid conflicts.

cpp/src/io/lp_parser.cpp
mlubin: They're not supported; added a test to confirm. Recognizing the section here allows us to print a more helpful error message, I believe.

cpp/src/io/lp_parser.cpp
```diff
@@ -0,0 +1,1142 @@
...
+      i_t i2 = get_or_add_var(var2);
+      // Store in upper-triangular form (i <= j) to match QUADOBJ convention.
+      i_t a = std::min(i1, i2);
+      i_t b = std::max(i1, i2);
+      raw_quad.emplace_back(a, b, sign * coeff);
+    } else {
+      // Purely linear term inside the brackets — permitted as long as the
+      // surrounding /2 convention is respected (the linear term is scaled
+      // the same way as the quadratic ones).
+      out_linear.push_back({i1, sign * coeff});
```
odow: I was skimming this to see if there was anything of interest for JuMP's reader. We don't support linear terms inside `[]`.

cpp/src/io/lp_parser.cpp
```diff
@@ -0,0 +1,1142 @@
...
+
+    first = false;
+    sign  = 1;
+  }
+  expect(LpTokenKind::RBracket, "closing ']' of quadratic section");
+
+  // Require the "/ 2" suffix after a quadratic objective expression.
+  // Without it there is no ambiguity-free way to tell whether the user
+  // meant /2 and forgot vs. intended bare coefficients, so we enforce the
+  // stricter form.
```
odow: Multiple solvers treat the `]/2` as optional in the objective. I don't remember if I just copied them, or if there were some common instances where this happened.

cpp/src/io/lp_parser.cpp
```diff
@@ -0,0 +1,1142 @@
...
+      bounds_defined_for_var_id_.insert(vid);
+      out_.variable_lower_bounds[vid] = lb;
+      lower_explicitly_set_.insert(vid);
+      if (match(LpTokenKind::LessEq)) { out_.variable_upper_bounds[vid] = parse_signed_number(); }
+    }
+  }
+
+  // A negative upper bound requires an explicitly stated lower bound,
+  // otherwise the default lower of 0 would collide with the upper and make
+  // the variable silently infeasible. Flag this at parse time.
```
odow, line 1216: Hmm. I think if the upper bound is negative, we assume that the lower bound is `-Inf`. But I see now that Gurobi flags the model as infeasible.

x-ref https://github.com/jump-dev/MathOptInterface.jl/pull/2994

cpp/src/io/lp_parser.cpp
mlubin: Yeah sounds like there's no reason for us to allow this.

cpp/src/io/lp_parser.cpp
mlubin: Agreed multiple solvers treat `]/2` as optional in the objective, but it seems scary to me. CPLEX says it's required, IIRC. I'd rather disallow this unless there's a good reason otherwise.

odow: > CPLEX says it's required

Oh, they all _say_ it's required. But the files that they _can_ actually read depart quite a lot from the written spec.

cpp/src/io/lp_parser.cpp
mlubin: Now supports semicontinuous variables

cpp/src/io/lp_parser.cpp
mlubin: Removed

mlubin:
@bbozkaya your feedback is welcome on the CLI and docs changes

python/cuopt/cuopt/linear_programming/io/parser_wrapper.pyx
```diff
@@ -131,16 +123,37 @@ def ParseMps(mps_file_path, fixed_mps_formats):
...
+
+
+@catch_io_exception
+def ParseMps(mps_file_path, fixed_mps_formats):
+    data_model = DataModel()
+    dm_ret_ptr = move(
+        call_parse_mps(
+            mps_file_path.encode('utf-8'),
+            fixed_mps_formats
+        )
```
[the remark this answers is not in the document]
mlubin: This seems pre-existing

cpp/src/io/file_to_string.cpp
```diff
@@ -0,0 +1,246 @@
...
+  struct BzReadCloseDeleter {
+    void operator()(void* f)
+    {
+      int bzerror;
+      if (f != nullptr) fptr(&bzerror, f);
+      mps_parser_expects_fatal(
+        bzerror == BZ_OK, error_type_t::ValidationError, "Error closing bzip2 file!");
+    }
+    BZ2_bzReadClose_t fptr = nullptr;
+  };
```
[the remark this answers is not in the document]
mlubin: pre-existing. this code was moved

jameslamb (approved):
The `cuopt-ci-codeowners` changes here are small (just updating a comment in a test script and updates to lists of source files in CMake), so approving.

Did not review anything else, deferring to other reviewers for that.

ci/test_self_hosted_service.sh
```diff
@@ -120,7 +120,7 @@ if [ "$doservertest" -eq 1 ]; then
     run_cli_test "'status': 'Optimal'" cuopt_sh -s -c "$CLIENT_CERT" -p $CUOPT_SERVER_PORT -t LP ../../datasets/linear_programming/good-mps-1.mps ../../datasets/linear_programming/good-mps-1.mps
 
     # Error, local file mode is not allowed with mps
-    run_cli_test "Cannot use local file mode with MPS data" cuopt_sh -s -c "$CLIENT_CERT" -p $CUOPT_SERVER_PORT -t LP -f good-mps-1.mps
+    run_cli_test "Cannot use local file mode with MPS/LP data" cuopt_sh -s -c "$CLIENT_CERT" -p $CUOPT_SERVER_PORT -t LP -f good-mps-1.mps
```
ramakrishnap-nv, line 123: It would be good to add examples for lp as well here.

datasets/linear_programming/good-mps-1.lp
```diff
@@ -0,0 +1,12 @@
+\ Equivalent of good-mps-1.mps
```
ramakrishnap-nv, line 1: Do we need compressed format of lp ? datasets/linear_programming/good-mps-1.lp.bz2 and datasets/linear_programming/good-mps-1.lp.gz

And lets add tests or examples if we are adding multiple datasets.

docs/cuopt/source/cuopt-server/examples/lp/examples/mps_datamodel_example.py
```diff
@@ -4,7 +4,7 @@
 LP DataModel from MPS Parser Example
 
 This example demonstrates how to:
-- Parse an MPS file using cuopt.linear_programming.mps_parser
+- Parse an MPS file using cuopt.linear_programming.ParseMps
```
ramakrishnap-nv, line 7: Should we also add an example for lp format ?

ramakrishnap-nv (commented):
Still reviewing other parts, but have few questions regarding datasets and tests.

docs/cuopt/source/cuopt-server/examples/lp/examples/mps_datamodel_example.py
Iroy30: Addressed in #1251

datasets/linear_programming/good-mps-1.lp
Iroy30: Addressed in #1251

ci/test_self_hosted_service.sh
Iroy30: Addressed in #1251

cpp/src/io/lp_parser.cpp
```diff
@@ -0,0 +1,1573 @@
...
+// ---- File I/O + tokenizer -------------------------------------------------
+
+template <typename i_t, typename f_t>
+void LpParseEngine<i_t, f_t>::read_and_tokenize(const std::string& file)
+{
+  // Delegates to the shared helper so .lp.gz / .lp.bz2 are handled the same
+  // way as .mps.gz / .mps.bz2 (dlopen-loaded libz / libbz2). The returned
+  // buffer is null-terminated; strip it before constructing the string view
+  // since `tokenize` walks the entire string range.
+  auto buf = detail::file_to_string(file);
```
rg20, line 318: When the files are large,  the buf and text can be very large.   Is this anyway better than reading the file line by line?

mlubin: Good catch. This is the pre-existing pattern that the MPS reader uses, as far as I can tell. If we want to stream the file instead, we should probably test that out in a separate PR as it would change both MPS and LP behavior.

python/cuopt/cuopt/tests/linear_programming/test_parser.py
```diff
@@ -67,3 +68,119 @@ def test_good_mps_file():
...
+    ) as f:
+        f.write(mps_text)
+        mps_path = f.name
+    with tempfile.NamedTemporaryFile(
+        suffix=".lp", mode="w", delete=False
+    ) as f:
+        f.write(_MINIMAL_LP)
+        lp_path = f.name
+    try:
+        lp_model = mps_parser.ParseLp(lp_path)
```
ramakrishnap-nv, line 164: we are using io as mps_parser and then using it this wasy, may io as io would be consistent, same comment for other places as well.

python/cuopt_self_hosted/cuopt_sh_client/cuopt_self_host_client.py
```diff
@@ -134,25 +134,66 @@ def is_uuid(cuopt_problem_data):
         return False
 
 
-def _mps_parse(LP_problem_data, solver_config):
+# File extensions (case-insensitive, after stripping a compression suffix) that
+# the cuopt.linear_programming.io package can parse client-side. Matches the
+# dispatch table in parse_problem() on the C++ side.
+_PARSEABLE_LP_EXTS = (".lp",)
```
ramakrishnap-nv, line 140: Do we check these extensions on C side as well?

ramakrishnap-nv (commented):
Rest looks good, have a minor comment on io and mps parser usage in python.

python/cuopt/cuopt/tests/linear_programming/test_parser.py
mlubin: This is tidied up in #1251

python/cuopt_self_hosted/cuopt_sh_client/cuopt_self_host_client.py
mlubin: There's no need to have this list in the C side because `cuOptReadProblem` directly calls `parse_problem`. I think the need for this list will be removed in #1251 with the python version of `parse_problem`.

rg20 (approved):

ramakrishnap-nv (approved):

Issue #1135, opened 2026-04-22
[BUG] Flaky test: c_api/TimeLimitTestFixture.time_limit/2 (MIP) overshoots time budget on constrained CI runners


rgsl888prabhu:
## Summary

`c_api/TimeLimitTestFixture.time_limit/2` (the MIP parametric case using `/mip/supportcase22.mps`, `time_limit=15s`, `CUOPT_METHOD_DUAL_SIMPLEX`) fails intermittently on CI when the reported `solve_time` exceeds `target_solve_time + excess_allowed_time` (15s + 3s = 18s).

## Failing assertion

`cpp/tests/linear_programming/c_api_tests/c_api_tests.cpp:64`

```cpp
EXPECT_NEAR(solve_time, target_solve_time, excess_allowed_time);  // excess = 3.0
```

## Observed CI failure

- `solve_time = 18.398s`, target `15s`, tolerance `3.0s` → over by ~0.4s.
- Solver itself respected the limit: log shows `Explored 0 nodes in 15.01s.` and termination status `Time limit (5)`.
- CI environment per log: AMD EPYC 9554, **only 12/12 threads visible** (cgroup-limited on a 64-core host), 43 GiB RAM, NVIDIA H100 NVL, CUDA 13.1.

## Root cause

Test reads `cuOptGetSolveTime`, which for MIP returns `mip_solution_t::get_total_solve_time()` → `stats_.total_solve_time`. That field is recorded at `cpp/src/mip_heuristics/solver.cu:489` (and similar `check_time_limit` guards) using a `timer_t` that was started in `cpp/src/mip_heuristics/solve.cu:306` — *before* Papilo presolve.

Therefore `solve_time` includes:
- Papilo presolve (OpenMP, ~2.59s in this run)
- cuOpt presolve (~1.59s)
- `run_mip` (LP root relaxation, primal heuristics, B&B)
- Post-B&B serial wind-down inside `mip_solver_t::solve()`: `branch_and_bound_status_future.get()`, `sol.compute_feasibility()`, `sol.test_variable_bounds(...)` + device sync

It does NOT include Papilo postsolve, the `full_sol` rebuild, or C-API wrapping (those run after `total_solve_time` is finalized).

The 3.4s gap between B&B's internal stop (15.01s) and `total_solve_time` (18.40s) is the post-join serial wind-down. On CPU-thread-constrained runners:
[...]

PR #1136, opened 2026-04-22
test(c_api): skip flaky TimeLimitTestFixture MIP case (#1135)

rgsl888prabhu:
## Summary

- Skip the `c_api/TimeLimitTestFixture.time_limit/2` parametric instance (`/mip/supportcase22.mps`, 15s, `CUOPT_METHOD_DUAL_SIMPLEX`) via `GTEST_SKIP()`.
- LP cases `/0` and `/1` remain enabled.

## Why

The MIP case intermittently overshoots the 3s `excess_allowed_time` tolerance on CPU-thread-constrained CI runners. `cuOptGetSolveTime()` for MIP returns `mip_solution_t::get_total_solve_time()`, which is set in `cpp/src/mip_heuristics/solver.cu:489` from a timer started in `cpp/src/mip_heuristics/solve.cu:306` — *before* Papilo presolve. So `solve_time` includes Papilo presolve (OpenMP-parallel) and the post-B&B serial wind-down (`branch_and_bound_status_future.get()`, `compute_feasibility`, `test_variable_bounds`), neither of which is bounded by the user's `time_limit` setting.

On a 12-thread-visible CI runner, observed `solve_time = 18.398s` vs the 18s bar (15s + 3s tolerance). Solver itself respected the limit (`Explored 0 nodes in 15.01s`).

Full investigation and proposed fixes (measure inside the time-limited region only, or per-case tolerances) are in the issue.

Tracking: NVIDIA/cuopt#1135.

## Test plan

- [x] Pre-commit clean (`clang-format`, `verify-copyright`, etc.)
- [x] Local rebuild of `C_API_TEST` succeeds
- [x] Verified `/2` is reported as `[ SKIPPED ] Disabled pending NVIDIA/cuopt#1135`; `/0` and `/1` are unchanged in behavior
- [ ] CI green

### DIFF
--- a/cpp/tests/linear_programming/c_api_tests/c_api_tests.cpp
+++ b/cpp/tests/linear_programming/c_api_tests/c_api_tests.cpp
@@ -47,6 +47,14 @@ TEST_P(TimeLimitTestFixture, time_limit)
   std::string filename                    = rapidsDatasetRootDir + std::get<0>(GetParam());
   double target_solve_time                = std::get<1>(GetParam());
   int method                              = std::get<2>(GetParam());
+
+  // The MIP case overshoots the 3s tolerance on CPU-thread-constrained CI runners
+  // because solve_time includes Papilo presolve and post-B&B serial wind-down.
+  // Tracked in https://github.com/NVIDIA/cuopt/issues/1135.
+  if (filename.find("/mip/") != std::string::npos) {
+    GTEST_SKIP() << "Disabled pending NVIDIA/cuopt#1135";
+  }
+
   int termination_status;
   double solve_time = std::numeric_limits<double>::quiet_NaN();
   EXPECT_EQ(solve_mps_file(filename.c_str(),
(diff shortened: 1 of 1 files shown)


cpp/tests/linear_programming/c_api_tests/c_api_tests.cpp
```diff
@@ -47,6 +47,14 @@ TEST_P(TimeLimitTestFixture, time_limit)
   std::string filename                    = rapidsDatasetRootDir + std::get<0>(GetParam());
   double target_solve_time                = std::get<1>(GetParam());
   int method                              = std::get<2>(GetParam());
+
+  // The MIP case overshoots the 3s tolerance on CPU-thread-constrained CI runners
+  // because solve_time includes Papilo presolve and post-B&B serial wind-down.
+  // Tracked in https://github.com/NVIDIA/cuopt/issues/1135.
+  if (filename.find("/mip/") != std::string::npos) {
```
mlubin: Check for string equality of `std::get<0>(GetParam())` with `"/mip/supportcase22.mps"` for simplicity? Otherwise the reader may be wondering if there are multiple MIP tests.

ramakrishnap-nv: I was thinking this might fail for any other mip tests as well, so I was thinking to check that, but let me update and make it limited to this file so we can comeback if there are other issues as well.

mlubin: There aren't other mip tests that use this code from what I can tell.

mlubin (approved):

chris-maes (approved):
LGTM

PR #1138, opened 2026-04-23
Fix lost nodes in branch and bound

hlinsen:
Fix lost branch-and-bound nodes when a plunge exits early due to LP solve limits.

- Requeues pending plunge nodes when the LP solve hits `CONCURRENT_LIMIT`.
- Requeues the current node and pending stack when the LP solve hits `ITERATION_LIMIT`.
- Ensures pending local stack nodes are returned to the global queue before `plunge_with` exits early, so the solver does not silently drop unexplored nodes or compute
  an incorrect lower bound.


nguidotti (approved):
Looks good to me! Thanks for the fix, Hugo.

PR #1140, opened 2026-04-23
Reduce QP overheads

rg20:
## Description
 
This PR improves QP performance, particularly on easy problems that run under few hundred milli seconds. 

**Algorithmic improvements** 
1. Free variables are directly handled in the iteration updates. Prior to this change, each free variable was split into two variables with bounds [0, \inf)
2. Early stopping of GMRES iterative refinement when there is no sufficient progress 

**Tunable parameters**
1. Ability to turn off iterative refinement.  The iterative refinement adds initial overhead for thrust calls, this is significant enough for easy problems.  
2. Ability to set step scale for barrier iteration update.  cuOpt default is 0.9, setting this to 0.99 reduces iteration count by 30% for portfolio optimization problems.  

**Overhead reductions**
1.  Conversion from `optimization_problem_t` to `problem_t` takes about 70ms.  For QP, problem_t is not used as we are again converting that to user_problem_t.  Added conversion from `optimization_problem_t` to `user_problem_t` directly.  
2. Removed/guarded dense column detection code when Q is not diagonal. 
3. Avoid computing AD, ADAT helper data when using augmented system

**Performance improvements**

1. 4% geomean speedup for barrier on LPfeas benchmark (note this with 10 min timelimit).  physiciansched and fome13, went from suboptimal to optimal
2. 46% geomean speedup for Maros benchmark.  122 optimal, 5 suboptimal  -->  128 optimal, 3 suboptimal (200 iteration limit)
3. 47% geomean speedup for portfolio problems
4. QPLIB is not impacted

## Checklist

- Testing
   - [ ] Added tests
   - [ ] NA
- Documentation
   - [ ] Added new documentation
   - [x] NA


cpp/src/dual_simplex/solve.cpp
```diff
@@ -373,23 +373,6 @@ lp_status_t solve_linear_program_with_barrier(const user_problem_t<i_t, f_t>& us
   // Solve using barrier
   lp_solution_t<i_t, f_t> barrier_solution(barrier_lp.num_rows, barrier_lp.num_cols);
 
-  // Clear variable pairs for QP
```
yuwenchen95, line 376: May I ask why we remove the check here?

rg20: This is only printing information.

cpp/include/cuopt/linear_programming/constants.h
```diff
@@ -74,14 +74,17 @@
...
+
+#define CUOPT_SOLUTION_FILE                "solution_file"
+#define CUOPT_NUM_CPU_THREADS              "num_cpu_threads"
+#define CUOPT_NUM_GPUS                     "num_gpus"
+#define CUOPT_USER_PROBLEM_FILE            "user_problem_file"
+#define CUOPT_PRESOLVE_FILE                "presolve_file"
+#define CUOPT_RANDOM_SEED                  "random_seed"
+#define CUOPT_PDLP_PRECISION               "pdlp_precision"
+#define CUOPT_MIP_SEMICONTINUOUS_BIG_M     "mip_semi_continuous_big_m"
+#define CUOPT_BARRIER_ITERATIVE_REFINEMENT "barrier_iterative_refinement"
```
chris-maes: Nit: could you put these next to CUOPT_BARRIER_DUAL_INITIAL_POINT above. So we keep all the barrier parameters together?

cpp/src/barrier/barrier.cu
```diff
@@ -382,21 +388,25 @@ class iteration_data_t {
         A_dense.from_sparse(lp.A, j, k++);
       }
     }
-    original_A_values = AD.x;
```
chris-maes, line 385: Is this line no longer necessary? It isn't included below. Just checking.

cpp/src/barrier/device_sparse_matrix.cuh
```diff
@@ -293,4 +306,82 @@ class device_csr_matrix_t {
                                          // to avoid extra space / computation)
 };
 
+template <typename i_t, typename f_t>
+void device_csc_matrix_t<i_t, f_t>::to_compressed_row(device_csr_matrix_t<i_t, f_t>& Arow,
+                                                      rmm::cuda_stream_view stream) const
+{
+  static_assert(std::is_signed_v<i_t>);
+  i_t const mm = m;
```
chris-maes: Nit: why can't you just use m and n?

chris-maes: Because you want these to be const?

cpp/src/barrier/device_sparse_matrix.cuh
```diff
@@ -293,4 +306,82 @@ class device_csr_matrix_t {
                                          // to avoid extra space / computation)
 };
 
+template <typename i_t, typename f_t>
+void device_csc_matrix_t<i_t, f_t>::to_compressed_row(device_csr_matrix_t<i_t, f_t>& Arow,
```
chris-maes, line 310: This is really cool!

cpp/src/barrier/device_sparse_matrix.cuh
```diff
@@ -293,4 +306,82 @@ class device_csr_matrix_t {
...
+  RAFT_CUDA_TRY(
+    cudaMemcpyAsync(Arow.row_start.data() + mm, &nz, sizeof(i_t), cudaMemcpyHostToDevice, stream));
+
+  rmm::device_uvector<i_t> rows(nz, stream);
+  rmm::device_uvector<i_t> cols(nz, stream);
+  rmm::device_uvector<f_t> vals(nz, stream);
+  raft::copy(rows.data(), i.data(), nz, stream);
+  raft::copy(vals.data(), x.data(), nz, stream);
+
+  thrust::tabulate(exec,
```
chris-maes, line 365: This is using a binary search to find out which a column the nonzero stored at index p is goes in?

chris-maes: Oh you are expanding to back to triplet?

cpp/src/barrier/barrier.cu
```diff
@@ -382,21 +388,25 @@ class iteration_data_t {
...
+      raft::copy(d_original_A_values.data(),
+                 device_AD.x.data(),
+                 device_AD.x.size(),
+                 handle_ptr->get_stream());
+      // For efficient scaling of AD col we form the col index array
+      device_AD.form_col_index(handle_ptr->get_stream());
+      device_A_x_values.resize(device_AD.x.size(), handle_ptr->get_stream());
+      raft::copy(
+        device_A_x_values.data(), device_AD.x.data(), device_AD.x.size(), handle_ptr->get_stream());
+      device_AD.to_compressed_row(device_A, handle_ptr->get_stream());
```
chris-maes, line 419: Very nice!

cpp/src/barrier/barrier.cu
```diff
@@ -1970,8 +1990,25 @@ int barrier_solver_t<i_t, f_t>::initial_point(iteration_data_t<i_t, f_t>& data)
                       vector_norm2<i_t, f_t>(data.dual_residual));
 #endif
   // Make sure (w, x, v, z) > 0
+  // Save free variable x values before forcing positive (they can be negative)
```
chris-maes: I will update this in my PR

cpp/src/barrier/barrier.cu
```diff
@@ -1970,8 +1990,25 @@ int barrier_solver_t<i_t, f_t>::initial_point(iteration_data_t<i_t, f_t>& data)
...
+  if (data.n_free_vars > 0) {
+    free_x_save.resize(data.n_free_vars);
+    for (i_t k = 0; k < data.n_free_vars; ++k) {
+      free_x_save[k] = data.x[presolve_info.free_variable_indices[k]];
+    }
+  }
   data.w.ensure_positive(epsilon_adjust);
   data.x.ensure_positive(epsilon_adjust);
+
+  // For native free variables (QP): restore x values and set z = 0
```
chris-maes: To be updated

cpp/src/barrier/barrier.cu
```diff
@@ -2956,24 +3081,45 @@ void barrier_solver_t<i_t, f_t>::compute_target_mu(
...
 {
   raft::common::nvtx::range fun_scope("Barrier: compute_cc_rhs");
 
-  cub::DeviceTransform::Transform(
-    cuda::std::make_tuple(data.d_dx_aff_.data(), data.d_dz_aff_.data()),
-    data.d_complementarity_xz_rhs_.data(),
-    data.d_complementarity_xz_rhs_.size(),
-    [new_mu] HD(f_t dx_aff, f_t dz_aff) { return -(dx_aff * dz_aff) + new_mu; },
-    stream_view_.value());
+  auto fill_linear_cc_rhs = [&](raft::device_span<f_t> out,
```
chris-maes: I will update this in my PR

cpp/src/barrier/barrier.cu
```diff
@@ -3443,8 +3592,19 @@ lp_status_t barrier_solver_t<i_t, f_t>::solve(f_t start_time,
 
     csc_matrix_t<i_t, f_t> Q(lp.num_cols, 0, 0);
     if (lp.Q.n > 0) { create_Q(lp, Q); }
-
-    iteration_data_t<i_t, f_t> data(lp, num_upper_bounds, Q, settings);
+    iteration_data_t<i_t, f_t> data(lp, num_upper_bounds, Q, settings, start_time);
+
+    // Set up native free variable tracking for QPs
+    if (!presolve_info.free_variable_indices.empty()) {
```
chris-maes: I will update this in my PR

cpp/src/barrier/device_sparse_matrix.cuh
```diff
@@ -293,4 +306,82 @@ class device_csr_matrix_t {
...
+                       } else {
+                         hi = mid;
+                       }
+                     }
+                     return lo - 1;
+                   });
+
+  auto row_iter = thrust::device_pointer_cast(rows.data());
+  auto col_iter = thrust::device_pointer_cast(cols.data());
+  thrust::sort_by_key(exec,
```
chris-maes, line 385: No issue with the code, but I'd love to understand how it works.

You want to sort the nonzeros by row. I'm not sure I understand why the second iterator has row_iter + nz and col_iter + nz.

cpp/src/barrier/device_sparse_matrix.cuh
```diff
@@ -293,4 +306,82 @@ class device_csr_matrix_t {
...
+  scan_tmp.resize(scan_bytes, stream);
+  cub::DeviceScan::ExclusiveSum(
+    scan_tmp.data(), scan_bytes, row_counts.data(), Arow.row_start.data(), mm, stream);
+
+  RAFT_CUDA_TRY(
+    cudaMemcpyAsync(Arow.row_start.data() + mm, &nz, sizeof(i_t), cudaMemcpyHostToDevice, stream));
+
+  rmm::device_uvector<i_t> rows(nz, stream);
+  rmm::device_uvector<i_t> cols(nz, stream);
+  rmm::device_uvector<f_t> vals(nz, stream);
```
chris-maes: Is vals needed? Why not directly use Arow.x?

cpp/src/barrier/device_sparse_matrix.cuh
```diff
@@ -293,4 +306,82 @@ class device_csr_matrix_t {
...
+    nullptr, scan_bytes, row_counts.data(), Arow.row_start.data(), mm, stream);
+  scan_tmp.resize(scan_bytes, stream);
+  cub::DeviceScan::ExclusiveSum(
+    scan_tmp.data(), scan_bytes, row_counts.data(), Arow.row_start.data(), mm, stream);
+
+  RAFT_CUDA_TRY(
+    cudaMemcpyAsync(Arow.row_start.data() + mm, &nz, sizeof(i_t), cudaMemcpyHostToDevice, stream));
+
+  rmm::device_uvector<i_t> rows(nz, stream);
+  rmm::device_uvector<i_t> cols(nz, stream);
```
chris-maes: Is cols needed? Why not directly use Arow.x? You could always make cols a reference to Arow.x if you wanted to keep the name cols. But it seems like you need an extra copy at the end.

cpp/src/barrier/device_sparse_matrix.cuh
```diff
@@ -293,4 +306,82 @@ class device_csr_matrix_t {
...
+                   });
+
+  auto row_iter = thrust::device_pointer_cast(rows.data());
+  auto col_iter = thrust::device_pointer_cast(cols.data());
+  thrust::sort_by_key(exec,
+                      thrust::make_zip_iterator(thrust::make_tuple(row_iter, col_iter)),
+                      thrust::make_zip_iterator(thrust::make_tuple(row_iter + nz, col_iter + nz)),
+                      thrust::device_pointer_cast(vals.data()));
+
+  raft::copy(Arow.j.data(), cols.data(), nz, stream);
```
chris-maes: I think you can get rid of these two copies and extra allocations by using Arow.j and Arow.x directly.

cpp/src/dual_simplex/presolve.cpp
```diff
@@ -828,7 +829,6 @@ i_t presolve(const lp_problem_t<i_t, f_t>& original,
   }
 
   if (settings.barrier_presolve && free_variables > 0) {
-    // Try to remove free variables
```
chris-maes, line 831: Please add back comment.

cpp/src/math_optimization/solver_settings.cu
```diff
@@ -102,6 +102,7 @@ solver_settings_t<i_t, f_t>::solver_settings_t() : pdlp_settings(), mip_settings
     {CUOPT_DUAL_INFEASIBLE_TOLERANCE, &pdlp_settings.tolerances.dual_infeasible_tolerance, f_t(0.0), f_t(1e-1), std::max(f_t(1e-10), std::numeric_limits<f_t>::epsilon())},
     {CUOPT_MIP_CUT_CHANGE_THRESHOLD, &mip_settings.cut_change_threshold, f_t(-1.0), std::numeric_limits<f_t>::infinity(), f_t(-1.0)},
     {CUOPT_MIP_CUT_MIN_ORTHOGONALITY, &mip_settings.cut_min_orthogonality, f_t(0.0), f_t(1.0), f_t(0.5)},
+    {CUOPT_BARRIER_STEP_SCALE, &pdlp_settings.barrier_step_scale, f_t(0.5), f_t(1.0), f_t(0.9)},
```
chris-maes: We can't ever go to 1.0. The max needs to be something like 0.9999

cpp/src/pdlp/solve.cu
```diff
@@ -435,6 +435,90 @@ optimization_problem_solution_t<i_t, f_t> convert_dual_simplex_sol(
...
+  f_t norm_user_objective,
+  f_t norm_rhs,
+  method_t method)
+{
+  auto to_termination_status = [](dual_simplex::lp_status_t st) {
+    switch (st) {
+      case dual_simplex::lp_status_t::OPTIMAL: return pdlp_termination_status_t::Optimal;
+      case dual_simplex::lp_status_t::INFEASIBLE:
+        return pdlp_termination_status_t::PrimalInfeasible;
+      case dual_simplex::lp_status_t::UNBOUNDED: return pdlp_termination_status_t::DualInfeasible;
```
chris-maes: If the dual is infeasible, I don't think we have a way to tell if the primal is unbounded or infeasible. So this should probably map to the UNBOUNDED_OR_INFEASIBLE status.

cpp/src/pdlp/solve.cu
```diff
@@ -466,6 +552,10 @@ run_barrier(dual_simplex::user_problem_t<i_t, f_t>& user_problem,
     barrier_settings.log.log = false;
   }
 
+  barrier_settings.log.printf("Barrier settings created at %.2f seconds, toc time: %.2f seconds\n",
```
chris-maes: Remove this before merging

cpp/src/pdlp/translate.hpp
```diff
@@ -186,6 +189,100 @@ static dual_simplex::user_problem_t<i_t, f_t> cuopt_problem_to_simplex_problem(
   return user_problem;
 }
 
+template <typename i_t, typename f_t>
+static dual_simplex::user_problem_t<i_t, f_t> cuopt_optimization_problem_to_simplex_problem(
```
chris-maes: Nit: this should probably be called `cuopt_optimization_problem_to_user_problem`

cpp/src/barrier/sparse_cholesky.cuh
chris-maes, line 262: Do you want to provide an option for the user to disable the CUDSS_THREADING_LIB?

chris-maes (commented):
Awesome. 

I will make changes to the free variable PR. I think you will either need to rebase off that PR or pull them in. 

Most of the comments are very minor or just questions.

The only one that definitely requires changing is the maximum barrier_step_scale

cpp/src/barrier/sparse_cholesky.cuh
rg20: I did not see the performance drop that you saw. I think it depends on the machine.

cpp/src/barrier/barrier.cu
rg20: Yes, I moved everything to GPU

cpp/src/barrier/device_sparse_matrix.cuh
rg20: Cursor wrote the code. I  changed it now and also added comments.

cpp/src/barrier/device_sparse_matrix.cuh
rg20: Yes.  CSC --> COO --> CSR. 

Yes, for the binary search.   You can write it differently by looping from col_start[j] to col_start[j+1] and set the col to j. But the memory access pattern is better with the binary search way.

cpp/src/barrier/device_sparse_matrix.cuh
rg20: Added comments now. 

CSC has this format: (after converting to COO)
cols:  0, 0, 0, 1, 1,  2
rows:  0, 1, 2,  0, 2, 2

It sorts with row as the major value and if there is tie uses col.  to finally get to:

rows: 0, 0, 1, 2, 2, 2
cols: 0, 1, 0, 0, 1, 2

cpp/src/barrier/device_sparse_matrix.cuh
rg20: You are right.  Fixed it now.  Thanks!

cpp/src/barrier/device_sparse_matrix.cuh
rg20: Done

cpp/src/barrier/device_sparse_matrix.cuh
rg20: Yes, fixed it

cpp/src/pdlp/solve.cu
rg20: I did not change the logic thats already there. Do we want to address with this PR?

cpp/src/barrier/device_sparse_matrix.cuh
chris-maes: That makes sense. But why add nz in the second make_tuple call?

cpp/src/barrier/sparse_cholesky.cuh
chris-maes: Ok. I still think it might be worth adding this. But we can do it in another PR if you like.

cpp/src/routing/local_search/local_search.cu
```diff
@@ -312,8 +312,8 @@ void local_search_t<i_t, f_t, REQUEST>::run_best_local_search(solution_t<i_t, f_
       cuopt_func_call(sol.compute_cost());
       cuopt_func_call(cost_after =
                         sol.get_cost(move_candidates.include_objective, move_candidates.weights));
-      cuopt_assert((cost_after - cost_before) - move_candidates.cycles.total_cycle_cost < 1.,
-                   "Cost mismatch after a move");
+      // cuopt_assert((cost_after - cost_before) - move_candidates.cycles.total_cycle_cost < 1.,
```
chris-maes, line 315: Is the comment out code intentional?

chris-maes (approved):
LGTM. Great work @rg20

cpp/src/pdlp/solve.cu
chris-maes: We can do it in another PR

cpp/src/routing/local_search/local_search.cu
rg20: Yes.  CI has failures on one of the VRP test. There is already a bug filed.

mlubin: I have a potential fix in https://github.com/NVIDIA/cuopt/pull/1240

ramakrishnap-nv (approved):

Issue #589, opened 2025-11-13
[FEA] Support for EV charging via vehicle capacity


bbozkaya:
Vehicle capacities in cuOpt VRP solver are currently tied to tasks, increasing or reducing capacity every time a vehicle visits a task location.

To track EV battery status of a vehicle as the vehicle travels during the day, a new capacity concept is proposed. This type of capacity needs to be linked to "edges" traversed as opposed to "nodes" where tasks will dictate how capacity is used.

An ideal solution would be to have a vehicle capacity defined (ie. starting EV charge level), which gets depleted based on distance traveled between tasks (using an applicable depletion factor). The vehicle then must return either to a depot or to one of the charging locations to recharge, and the service time at the charging location should be a function of charge level (eg. to top it off) at the time of charging.

Using estimated break times and specifying break windows and locations (ie. for charging stations) may not work because it is hard to estimate a break window due to dynamic nature of tasks assigned and distance traveled before a recharge (small number of high-service-time tasks visited may involve short travel distances and low depletion vs. large number of low-service-time tasks visited may involve large travel distances and high depletion).

anandhkb:
@afender for viz
@rgsl888prabhu for viz
@tmckayus for viz
@rg20 for advise as needed by the dev PIC

anandhkb:
Marking this issue with a proposed implementation approach from @rg20 as a feature where we welcome external contributions via the label "good first issue"

np96:
Hi there @rg20.
This one looks interesting to me. I can work on it and submit the solution in ~3 weeks.
In my use case I have max visited points per vehicle (primarily courier/client UX). The EV constraint looks like a generalization of the feature I want.

anandhkb:
> Hi there [@rg20](https://github.com/rg20). This one looks interesting to me. I can work on it and submit the solution in ~3 weeks. In my use case I have max visited points per vehicle (primarily courier/client UX). The EV constraint looks like a generalization of the feature I want.

Hi @np96 

Brilliant!
I look forward to your PR eagerly and happy to also collaborate with you in the development process. Thanks.

bbozkaya:
Perfect, thanks! Let us know if you need any further clarification. I think the final suggestion was to implement it as "distance-based" breaks, but a more general solution would also quite valuable.

np96:
@bbozkaya right, currently proposed approaches do not seem to generalize what I need.
I clearly understand the problem and approaches you discussed.
[...]

PR #1147, opened 2026-04-26
(feature) add distance-based routing ev-breaks

np96:
## Description

Adds `add_vehicle_ev_break(vehicle_id, distance_min, distance_max, charge_duration, locations)` to the C++ and Python APIs;
Extends `vehicle_break_t` with an `is_distance_based` flag and `distance_min`/`distance_max` fields; the existing break call infrastructure handles placement;
The solver inserts one charging stop per call within `[distance_min, distance_max]` of cumulative route distance; call the function multiple times to model successive charge cycles (e.g. `[0, 150], [150, 300]`);
No new solver dimension — the distance window is checked against distance_forward during forward propagation, reusing the same feasibility path as time-windowed breaks;
Charging stations are specified per-vehicle and per-cycle via break_locations.

## Issue
#589 

## Checklist

I am familiar with the [Contributing Guidelines](https://github.com/NVIDIA/cuopt/blob/HEAD/CONTRIBUTING.md).
- Testing
   - Added tests
- Documentation
   - Updated documentation


python/cuopt/cuopt/routing/vehicle_routing.py
```diff
@@ -392,6 +392,65 @@ def add_vehicle_break(
...
+
+        cycles = n_cycles if n_cycles is not None else 1
+        for vid in vehicle_ids:
+            validate_range(vid, "vehicle id", 0, self.get_fleet_size())
+            for k in range(cycles):
+                d_min = k * max_range + min_range
+                d_max = (k + 1) * max_range
+                super().add_vehicle_ev_break(
+                    vid, d_min, d_max, charge_duration, charging_stations
+                )
```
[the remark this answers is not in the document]
ramakrishnap-nv: @np96 Please consider this review

python/cuopt/cuopt/routing/vehicle_routing_wrapper.pyx
```diff
@@ -482,6 +482,39 @@ cdef class DataModel:
             vehicle_id, earliest, latest, duration,
             <const int *> c_locations_ptr, len(locations))
 
+    def add_vehicle_ev_break(
+        self, vehicle_id, distance_min, distance_max, duration, locations
+    ):
+        dim = 0
+        if vehicle_id in self.non_uniform_breaks:
```
ramakrishnap-nv: you can replace the whole segment
```
dim = 0
if vehicle_id in self.non_uniform_breaks:
            dim = len(self.non_uniform_breaks[vehicle_id])

        if dim == 0:
            self.non_uniform_breaks[vehicle_id] = {}
```

with this 

```
   dim = self.non_uniform_breaks.setdefault(vehicle_id, {})
```

python/cuopt/cuopt/tests/routing/test_ev_breaks.py
```diff
@@ -0,0 +1,273 @@
+# SPDX-FileCopyrightText: Copyright (c) 2024-2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
ramakrishnap-nv:
```suggestion
# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```

cpp/tests/routing/unit_tests/ev_breaks.cu
```diff
@@ -0,0 +1,193 @@
+/* clang-format off */
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2024-2025, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
ramakrishnap-nv:
```suggestion
 * SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```

python/cuopt/cuopt/tests/routing/test_ev_breaks.py
```diff
@@ -0,0 +1,273 @@
...
+from cuopt.routing import utils
+
+filename = utils.RAPIDS_DATASET_ROOT_DIR + "/solomon/In/r107.txt"
+
+
+# ---------------------------------------------------------------------------
+# Helpers
+# ---------------------------------------------------------------------------
+
+def _small_data_model(n_vehicles=3):
```
ramakrishnap-nv, line 15: Lets keep the test set minimal to test only apis and functionality should be tested completely in C++ testing.

rgsl888prabhu (changes requested):

python/cuopt/cuopt/tests/routing/test_ev_breaks.py
np96: Thanks for noticing. For this moment I have to work on v100 using 26.0.2 tag and cherry-picking 🙂 .

python/cuopt/cuopt/tests/routing/test_ev_breaks.py
np96: > Lets keep the test set minimal to test only apis and functionality should be tested completely in C++ testing.

What about the integration test? From my perspective a single simple integration test is the best approach to test that the functionality as a whole is not broken.
So, I'd keep:
API/validations/(single)integration.

But if you insist, I will remove it also.

np96:
BTW, regarding this:
> Docstring coverage is 57.14% which is insufficient. The required threshold is 80.00%.

It is about C++ source code, right? Not the tests?

python/cuopt/cuopt/routing/vehicle_routing.py
```diff
@@ -392,6 +392,83 @@ def add_vehicle_break(
...
+    @catch_cuopt_exception
+    def add_ev_break(
+        self,
+        vehicle_ids,
+        max_range,
+        charge_duration,
+        charging_stations=cudf.Series(),
+        min_range=0.0,
+        n_cycles=None,
+    ):
```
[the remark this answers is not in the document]
np96: Probably would make sense to do this in a dedicated PR for all methods at once.
Otherwise introduces inconsistency.

python/cuopt/cuopt/routing/vehicle_routing.py
```diff
@@ -392,6 +392,83 @@ def add_vehicle_break(
...
+        Examples
+        --------
+        >>> d.add_ev_break(
+        ...     vehicle_ids=[0, 1],
+        ...     max_range=150.0,
+        ...     charge_duration=20,
+        ...     charging_stations=cudf.Series([5, 8, 12]),
+        ...     n_cycles=3,
+        ... )
+        """
```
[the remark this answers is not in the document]
np96: return type is usually omitted in such cases, moreover, there are no raises in docstrings yet. I believe this should also be comitted consistently.

python/cuopt/cuopt/tests/routing/test_ev_breaks.py
ramakrishnap-nv: Nah, one integration test to check all parts is  perfect.

np96: > Nah, one integration test to check all parts is perfect.

Ok, only api/validations are present now.

python/cuopt/cuopt/routing/vehicle_routing.py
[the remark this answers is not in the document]
np96: @rgsl888prabhu this one is also fixed.

python/cuopt/cuopt/routing/vehicle_routing_wrapper.pyx
np96: fixed

np96:
Hi @rgsl888prabhu
I've resolved the issues you raised last week. Anything else?

rgsl888prabhu:
> Hi @rgsl888prabhu I've resolved the issues you raised last week. Anything else?

Thank you @np96, let me review.

rgsl888prabhu (approved):
Thank you @np96 for the PR, I have a minor suggestion, it would be really useful if we can add an example under doc for python.

anandhkb:
> Thank you @np96 for the PR, I have a minor suggestion, it would be really useful if we can add an example under doc for python.

Thanks @rgsl888prabhu for reviewing @np96 's impactful PR !
Thanks @np96 ! Great work.

np96:
> Thank you @np96 for the PR, I have a minor suggestion, it would be really useful if we can add an example under doc for python.

Great idea. That was very helpful, After adding I found out that the solver may intentionally generate infeasible solutions and small fix in breaks_insertion.cu was needed.

Besides this, added a unit test which reproduces the issue and aligned python/cython naming.

There is still server side support not implemented. If I have time, I'll do this in 26.06 release, otherwise in 26.08.

Thank you.

@anandhkb FYI.

np96:
@rgsl888prabhu for some reason pre-commit hook wasn't activated on signoff.
I've fixed the linting but the build errors look irrelevant to my changes.

rgsl888prabhu:
Yeah, there is a ongoing issue in wheel build due to scache bug

anandhkb:
> There is still server side support not implemented. If I have time, I'll do this in 26.06 release, otherwise in 26.08.

Thanks @np96 !, makes perfect sense!

rg20:
@np96  

Thanks for adding the distance based breaks to cuOpt.  I have a few comments on the implementation. 

Do you see the distance_min and max as hard-constraints or soft-constraints (i.e. objective to be minimized)?   I see that you are directly encoding the infeasibilities and violation as part of break insertion kernel. This is not the correct way to do it.  The infeasibility calculations should be consistent across all the operators.  Even when the route is infeasible because distance constraints are violated, other operators see the route as feasible and move on.  If the infeasibilities are handled correctly, the other operators have the opportunity to kick off some other nodes (non-break nodes) to make the route feasible. 

We handle all the infeasibilities and objectives within the route and node data structures.  A detailed logic of how to setup forward and backward calculations to be able to compute infeasibilities and objectives can be found here: https://github.com/NVIDIA/cuopt/blob/scheduling/skills/cuopt-developer/vrp_skills.md . 

The correct way to do this is by moving the infeasibility metric (or objective) to distance dimension.   This needs adding few more quantities to the distance_node_t and distance_route_t.    

From API point of view,  I would advise to use distance_based_break instead of ev_based to make it more general.  This feature will be greatly useful to impose driver rules (for example, driver can drive for more than 100 miles without a break).

np96:
Hi @rg20.
Thanks for sharing the doc. That's what I was looking for 🙂 .
I completely agree with the implementation approach you shown. Spreading the logic of a dimension over all kernels is bug-prone and difficult to maintain.
Won't take much to fix, probably 1-2 evenings.

> Do you see the distance_min and max as hard-constraints or soft-constraints (i.e. objective to be minimized)?

I believe these should be hard-constraints, though I am not quite sure with the semantics. The ones proposed currently may not be ideal, but they look simplest to implement. Do you agree with the [API](https://github.com/NVIDIA/cuopt/pull/1147/changes#diff-790e432011e4de781772226ed0f089b0e8a56e3eeeedd6026a25d57cb9c6a18fR409)? I'd be glad to hear a better idea. Thanks.

rg20:
> Hi @rg20. Thanks for sharing the doc. That's what I was looking for 🙂 . I completely agree with the implementation approach you shown. Spreading the logic of a dimension over all kernels is bug-prone and difficult to maintain. Won't take much to fix, probably 1-2 evenings.
> 
> > Do you see the distance_min and max as hard-constraints or soft-constraints (i.e. objective to be minimized)?
> 
> I believe these should be hard-constraints, though I am not quite sure with the semantics. The ones proposed currently may not be ideal, but they look simplest to implement. Do you agree with the [API](https://github.com/NVIDIA/cuopt/pull/1147/changes#diff-790e432011e4de781772226ed0f089b0e8a56e3eeeedd6026a25d57cb9c6a18fR409)? I'd be glad to hear a better idea. Thanks.

@np96  I agree with the API, except for the naming.  The current API makes it look like it is very specific to EVs. In fact, this is a much more general use case. Often, there are regulations on how far a driver can drive continuously without a break.  This feature makes it possible to model such scenarios. 

Glad you agree with the approach. Looking forward for the changes. 

Can you also provide a dataset with reasonable sizes so that we can use it for regression testing?

np96:
@rg20 regarding dataset. You want all-at-once or solely dist breaks?
@rgsl888prabhu by the way, there's discrepancy in vrp skills and your suggestion regarding api tests. 
https://github.com/NVIDIA/cuopt/blob/scheduling/skills/cuopt-developer/vrp_skills.md#python-integration-tests-pythoncuoptcuopttestsrouting — states clearly that there should be tests which do run solve. I want to double check the source of truth (only api vs tests with solve).

ramakrishnap-nv:
> @rg20 regarding dataset. You want all-at-once or solely dist breaks? @rgsl888prabhu by the way, there's discrepancy in vrp skills and your suggestion regarding api tests. https://github.com/NVIDIA/cuopt/blob/scheduling/skills/cuopt-developer/vrp_skills.md#python-integration-tests-pythoncuoptcuopttestsrouting — states clearly that there should be tests which do run solve. I want to double check the source of truth (only api vs tests with solve).

Yes, what I meant earlier was to keep on integration test which hits all the api parts and small. So we are sure everything works and doesn't take too much time.

rg20:
> @rg20 regarding dataset. You want all-at-once or solely dist breaks? @rgsl888prabhu by the way, there's discrepancy in vrp skills and your suggestion regarding api tests. https://github.com/NVIDIA/cuopt/blob/scheduling/skills/cuopt-developer/vrp_skills.md#python-integration-tests-pythoncuoptcuopttestsrouting — states clearly that there should be tests which do run solve. I want to double check the source of truth (only api vs tests with solve).

Typically we want some unit tests that test this feature, but we also like to have a representative use case (that includes this feature) that is used in production. We use such datasets for performance regression testing.

rg20:
@np96 

FYI, I shared developer skills file for VRP with you. I used these skills with Claude code to develop a new feature for another customer.  May be you can give it a try.  You will have to describe the penalty function and ask it to develop.

np96:
@rgsl888prabhu @rg20 I did a refactoring, will push today or tomorrow. Need to re-review the changes before publishing.

It turned out to be >2K diff PR with changes in all the search operators, since now have to check both TIME and DIST feasibilities. Probably should submit as another PR and drop this one, since the cpp implementation has become entirely different. What would you suggest?

Sorry for the confusion, probably had to discuss the implementation approach first.

@rg20 Claude codes fast and clean enough but can't develop on its own. Have to guide the entire process and do lots of review. Moreover, it burns tokens very fast on such problems.

rg20:
> @rgsl888prabhu @rg20 I did a refactoring, will push today or tomorrow. Need to re-review the changes before publishing.
> 
> It turned out to be >2K diff PR with changes in all the search operators, since now have to check both TIME and DIST feasibilities. Probably should submit as another PR and drop this one, since the cpp implementation has become entirely different. What would you suggest?
> 
> Sorry for the confusion, probably had to discuss the implementation approach first.
> 
> @rg20 Claude codes fast and clean enough but can't develop on its own. Have to guide the entire process and do lots of review. Moreover, it burns tokens very fast on such problems.

@np96 the time violations are already handled, you do not have to do anything extra. The only change would be in distance dimension.  

If it makes easy to just close this PR and open a new one, that is fine by me.   Its just that we might lose all the discussion we have had here.

np96:
> > @rgsl888prabhu @rg20 I did a refactoring, will push today or tomorrow. Need to re-review the changes before publishing.
> > It turned out to be >2K diff PR with changes in all the search operators, since now have to check both TIME and DIST feasibilities. Probably should submit as another PR and drop this one, since the cpp implementation has become entirely different. What would you suggest?
> > Sorry for the confusion, probably had to discuss the implementation approach first.
> > @rg20 Claude codes fast and clean enough but can't develop on its own. Have to guide the entire process and do lots of review. Moreover, it burns tokens very fast on such problems.
> 
> @np96 the time violations are already handled, you do not have to do anything extra. The only change would be in distance dimension.
> 
> If it makes easy to just close this PR and open a new one, that is fine by me. Its just that we might lose all the discussion we have had here.

@rg20 right, ime violations are handled but distance violations are not. So I had to extend time violations feasibility check pattern to both time/distance. Currently the problem I am observing is that the solver forces n_cycles breaks to a route which makes inefficient routes.
I wil share the code today, hope you have an idea how to overcome this.

np96:
@rg20 @rgsl888prabhu @anandhkb 
The implementation has changed too much, so let's continue here:  https://github.com/NVIDIA/cuopt/pull/1196.
Added solver support and regression test. Described the problems I've observed while implementing this feature here:
https://github.com/NVIDIA/cuopt/issues/1195

Thanks.

PR #1148, opened 2026-04-27
Fix loose tolerances in the objective function integrality detection

aliceb-nv:
The objective function integrality detection was relying on weak tolerances, which caused neos-827175 to be incorrectly accounted as integral and causing an incorrect solution bound to be reported. (objective coefficients were as low as 1e-5)

This PR uses a tighter 1e-9 tolerance for integrality detection in the objective function to avoid such issues.

## Checklist

- Testing
   - [ ] Added tests
   - [ ] NA
- Documentation
   - [ ] Added new documentation
   - [ ] NA


nguidotti (approved):
Approved! Thanks for the fix, Alice!

PR #1149, opened 2026-04-27
B&B with Local Heaps

nguidotti:
In this PR, each best-first worker has its own local node heap, such that it push/pop nodes without synchronizing with other workers. Each best-first worker periodically steals a node from a random worker to keep the node distribution more or less balance across them. Additionally, each best-first worker has a (fixed) set of diving worker assigned to it, which are used for performing diving on its own nodes whenever possible. This essentially eliminates the need of the scheduler thread, freeing one additional thread to do something useful. 

This also implements a compression scheme for `vstatus` using only `2bits` per entry, which reduces the memory consumption by roughly `4x` (previously was using `int8_t` per entry). Last, but not least, this PR replaces `std::deque` with a fixed-capacity `circular_deque_t` for the plunge/dive stacks and the idle-worker list.

MIPLIB results (GH200, 10min):
```
main (1, #1099) vs bnb-local-heap (2)

| Feasible                                                 227                  228                   +1                 --- |
| Optimal                                                   75                   78                   +3                 --- |
| Solutions with <0.1% primal gap                          124                  130                   +6                 --- |
| Nodes explored (mean)                              4.866e+06            1.436e+07           +9.496e+06                +195 |
| Nodes explored (shifted geomean)                        6772            1.205e+04                +5275               +77.9 |
| Relative MIP gap (mean)                               0.3264               0.3415             +0.01506               +4.62 |
| Relative MIP gap (shifted geomean)                    0.1156               0.1131              -0.0025               -2.16 |
| Solve time (mean)                                      444.6                441.5               -3.054              -0.687 |
| Solve time (shifted geomean)                           221.5                219.1               -2.327               -1.05 |
| Primal gap (mean)                                      11.57                11.15              -0.4201               -3.63 |
| Primal gap (shifted geomean)                          0.6324               0.5604             -0.07203               -11.4 |
| Primal integral (mean)                                 32.63                33.02              +0.3805               +1.17 |
| Primal integral (shifted geomean)                      6.346                6.405             +0.05989              +0.944 |
```

In summary, we explored `~3x` nodes in average` at the same time frame. The number of optimal solutions also increased by 3. 

## Checklist

- Testing
   - [ ] Added tests
   - [ ] NA
- Documentation
   - [ ] Added new documentation
   - [ ] NA


nguidotti:
/ok to test [d936e05](https://github.com/NVIDIA/cuopt/pull/1149/commits/d936e058bf8fb25d68a303b765cebfdaf36ad03d)

nguidotti:
@coderabbit full review

cpp/src/branch_and_bound/worker.hpp
```diff
@@ -0,0 +1,284 @@
...
+    --bfs_worker->active_diving_workers[this->search_strategy];
+    --bfs_worker->total_active_diving_workers;
+  }
+
+  mip_node_t<i_t, f_t> start_node;
+
+  // The best-first worker that is associated with this diving worker. Used for controlling the
+  // number of active diving workers.
+  bfs_worker_t<i_t, f_t>* bfs_worker{nullptr};
+};
```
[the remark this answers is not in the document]
nguidotti: A diving worker is always launched from a bfs worker. However, a diving worker can be re-used from a bfs worker to another (the worker pool is shared). For (2), we use set_active/inactive to mark if the worker is active or not. Some nodes are dropped (e.g., it may be infeasible) before the diving worker is launched, even after calling the init routine.

nguidotti:
/ok to test [2cd842a](https://github.com/NVIDIA/cuopt/pull/1149/commits/2cd842a372e7ad866b3158e96c68b74b4f258973)

cpp/src/branch_and_bound/worker.hpp
```diff
@@ -0,0 +1,291 @@
...
+      if (node == nullptr) { break; }
+
+      this->node_queue.lock();
+      this->node_queue.push(node);
+      this->node_queue.unlock();
+      --num_nodes;
+      steal = true;
+    }
+
+    return steal;
```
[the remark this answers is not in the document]
nguidotti: The global lower bound is calculated from the top of the heap and the current lower bound of the worker. See the `worker_pool_t::get_lower_bound()`

nguidotti:
/ok to test [56c8cff](https://github.com/NVIDIA/cuopt/pull/1149/commits/56c8cff7b91ae23f582f934bccf9daa3f68e462c)

nguidotti:
/ok to test [54d3f66](https://github.com/NVIDIA/cuopt/pull/1149/commits/54d3f6685abe76656e24cd0543ff2d1bba2bba0b)

nguidotti:
/ok to test [bc3d50b](https://github.com/NVIDIA/cuopt/pull/1149/commits/bc3d50b5d407c12bb39d3e4420f7a3b5c0a7a292)

nguidotti:
/ok to test [e1ddd8e](https://github.com/NVIDIA/cuopt/pull/1149/commits/e1ddd8eb0422f6da2f129e3e6094246ebf96bc2e)

nguidotti:
/ok to test [60de4ad](https://github.com/NVIDIA/cuopt/pull/1149/commits/60de4ad2910c744f5e5048f8a42c15b19f3592d2)

nguidotti:
/ok to test [67f539c](https://github.com/NVIDIA/cuopt/pull/1149/commits/67f539c8362c6a2bfb86befa931d2ee28d8e891e)

cpp/src/branch_and_bound/branch_and_bound.cpp
```diff
@@ -1449,6 +1457,36 @@ void branch_and_bound_t<i_t, f_t>::plunge_with(branch_and_bound_worker_t<i_t, f_
...
+    if (bfs_worker_pool_.num_idle_workers() > 0 && worker->node_queue.best_first_queue_size() > 0) {
+      worker->node_queue.lock();
+      mip_node_t<i_t, f_t>* node = worker->node_queue.pop_best_first();
+
+      // We need to temporarily save the lower bound in this worker so it is
+      // considered when calculating the global lower bound.
+      f_t node_lower_bound = node ? node->lower_bound : std::numeric_limits<f_t>::infinity();
+      worker->lower_bound  = std::min(worker->lower_bound.load(), node_lower_bound);
+
+      worker->node_queue.unlock();
```
akifcorduk: We can unlock this after poping the best first right?

cpp/src/branch_and_bound/branch_and_bound.cpp
```diff
@@ -1539,31 +1607,118 @@ void branch_and_bound_t<i_t, f_t>::plunge_with(branch_and_bound_worker_t<i_t, f_
...
+    return nullptr;
   }
 
-  if (settings_.num_threads > 1) {
-    worker_pool_.return_worker_to_pool(worker);
-    active_workers_per_strategy_[BEST_FIRST]--;
+  idle_worker->init(start_nodes);
+  idle_worker->set_active();
+
+#pragma omp task affinity(*idle_worker) priority(99) default(none) firstprivate(idle_worker)
```
akifcorduk: Should we maybe define enums for priorities and categorize them?

cpp/src/branch_and_bound/branch_and_bound.cpp
```diff
@@ -1576,8 +1731,8 @@ void branch_and_bound_t<i_t, f_t>::dive_with(branch_and_bound_worker_t<i_t, f_t>
   worker->recompute_basis  = true;
   worker->recompute_bounds = true;
 
-  search_tree_t<i_t, f_t> dive_tree(std::move(*worker->start_node));
-  std::deque<mip_node_t<i_t, f_t>*> stack;
+  search_tree_t<i_t, f_t> dive_tree(std::move(worker->start_node));
+  circular_deque_t<mip_node_t<i_t, f_t>*> stack(2 * diving_backtrack_limit + 4);
```
akifcorduk: Why is the size `2 * diving_backtrack_limit + 4` ?

nguidotti:
/ok to test [ef450ba](https://github.com/NVIDIA/cuopt/pull/1149/commits/ef450ba120e63761acb75e3aa3c7225a09012653)

cpp/src/branch_and_bound/branch_and_bound.cpp
```diff
@@ -1647,220 +1800,80 @@ void branch_and_bound_t<i_t, f_t>::dive_with(branch_and_bound_worker_t<i_t, f_t>
...
 
-#ifdef CUOPT_LOG_DEBUG
-  for (auto strategy : strategies) {
-    settings_.log.debug("%c%d: max num of workers = %d",
-                        feasible_solution_symbol(strategy),
-                        strategy,
-                        max_num_workers_per_type[strategy]);
+  if (!start_node || upper_bound_.load() < start_node->lower_bound ||
+      start_node->depth < settings_.diving_settings.min_node_depth) {
+    bfs_worker->node_queue.unlock();
```
akifcorduk: Can't we unlock the queue as soon as we pop from the queue?

cpp/src/branch_and_bound/branch_and_bound.cpp
```diff
@@ -2625,45 +2636,65 @@ mip_status_t branch_and_bound_t<i_t, f_t>::solve(mip_solution_t<i_t, f_t>& solut
     if (settings_.deterministic) {
       run_deterministic_coordinator(Arow_);
     } else if (settings_.num_threads > 1) {
-      run_scheduler();
+      const i_t num_workers        = settings_.num_threads;
+      const i_t num_bfs_workers    = std::max(settings_.num_threads / 2, 1);
+      const i_t num_diving_workers = num_workers - num_bfs_workers;
```
akifcorduk, line 2885: Why don't we leave some threads to other threads like CPUF and RINS ?

cpp/src/branch_and_bound/worker.hpp
```diff
@@ -31,16 +31,35 @@ struct branch_and_bound_stats_t {
...
+    case COEFFICIENT_DIVING: return settings.coefficient_diving != 0;
+    default: return false;
+  }
+}
+
 template <typename i_t, typename f_t>
 class branch_and_bound_worker_t {
  public:
-  const i_t worker_id;
+  using float_type = f_t;
```
akifcorduk, line 64: Do we use `float_type` and `int_type` ?

cpp/src/branch_and_bound/worker.hpp
```diff
@@ -134,14 +122,172 @@ class branch_and_bound_worker_t {
...
+  {
+    bool steal = false;
+    assert(num_nodes > 0);
+
+    if (!other->is_active || this == other ||
+        other->node_queue.best_first_queue_size() < 2 * num_nodes) {
+      return steal;
+    }
+
+    while (num_nodes > 0) {
```
akifcorduk: To reduce the amount of lock and unlock calls and minimize contention. Can we do something like this?

```
std::vector<mip_node_t<i_t, f_t>*> stolen;
stolen.reserve(num_nodes);

other->node_queue.lock();

while (num_nodes > 0 &&
       other->node_queue.best_first_queue_size() > num_nodes) {
  stolen.push_back(other->node_queue.pop_best_first());
  --num_nodes;
}

other->node_queue.unlock();

if (!stolen.empty()) {
  this->node_queue.lock();

  for (auto* node : stolen) {
    this->node_queue.push(node);
  }

  this->node_queue.unlock();
  steal = true;
}
```

cpp/src/branch_and_bound/worker.hpp
```diff
@@ -134,14 +122,172 @@ class branch_and_bound_worker_t {
...
+    }
+
+    total_max_diving_workers = 0;
+    max_diving_workers.fill(0);
+    if (num_active == 0) { return; }
+
+    for (size_t i = 1, k = 0; i < num_search_strategies; ++i) {
+      if (is_search_strategy_enabled(search_strategies[i], has_incumbent, settings)) {
+        // Calculate the number of workers for a given diving heuristic
+        i_t start            = std::floor((double)k * total_diving_workers / num_active);
```
akifcorduk: We seem to use such range computations a lot. I think we can put them in a reusable function.

cpp/src/branch_and_bound/branch_and_bound.cpp
nguidotti: Unfortunately, no. The lower bound of a node is calculated based on the heap and the node being explored by a worker. If we take a node from the heap, then it lower bound are no longer considered when retrieving the lower bound from the heap. So, we need to store it in the worker lower bound, otherwise it would temporarily vanish from the solver and potentially cause the solver to terminate early due to gap convergence rule.

nguidotti: This two operations (popping a node from the heap and updating the lower bound of the worker) needs to happen "simultaneously".

cpp/src/branch_and_bound/branch_and_bound.cpp
nguidotti: Since we are doing depth-first search and can only backtrack to a maximum number of nodes prior (`diving_backtrack_limit`), then the maximum number of nodes in the stack should be `diving_backtrack_limit + 1`

nguidotti: So, yeah, this is wrong. I will fix this

cpp/src/branch_and_bound/worker.hpp
nguidotti: It is mostly used in the `worker_pool_t` so it can retrieve the `i_t` and `f_t` from the underlying worker.

nguidotti: This is due the generic template parameter `WorkerType` in the `worker_pool_t`

cpp/src/branch_and_bound/branch_and_bound.cpp
nguidotti: No, we must keep the mutex locked here until we create a detached copy of the node (i.e., a complete copy of the node without no connection to the main tree) and retrieve the bounds from the tree. This prevents another worker stealing that node and modify its content (or even fathoming it) while the first one is still copying the information, which causes a race condition.

cpp/src/branch_and_bound/worker.hpp
nguidotti: Unfortunately, no. In terms of contention, I think the original version is better since allows the two or more workers to pipeline the operations (the first one may be popping the node, while the other one will be pushing) and each operation will be quite short. We also need to be careful here how to handle the lower bound.

nguidotti:
/ok to test [d1f4c12](https://github.com/NVIDIA/cuopt/pull/1149/commits/d1f4c125605282b992237db1b5d0289b9cd4a958)

cpp/src/branch_and_bound/branch_and_bound.cpp
nguidotti: This will be done automatically by the OpenMP. More specifically, the best-first workers are long-lived, blocking half of the CPU cores. While diving, reliability branching, RINS and CPUFJ tasks will share the remaining of the CPU cores. These tasks have a short (or medium) duration, and thus, allows the runtime to switch between them, keeping the cores always busy

nguidotti: I prioritize the RINS task based on our discussion in Slack

nguidotti:
/ok to test [3d072ce](https://github.com/NVIDIA/cuopt/pull/1149/commits/3d072ced6b8d4014026866a2e2e60e3a7786d6b4)

nguidotti:
/ok to test [3f22813](https://github.com/NVIDIA/cuopt/pull/1149/commits/3f2281366e23d05faa7633a9abd6400efc8425e7)

nguidotti:
/ok to test [49dba03](https://github.com/NVIDIA/cuopt/pull/1149/commits/49dba033ac567ff3d47e56aefc079e3fc2ab8df8)

nguidotti:
/ok to test [029ec30](https://github.com/NVIDIA/cuopt/pull/1149/commits/029ec30033849b0c93c7fcebee790024e5182e7e)

Kh4ster (approved):
Very cool results! Thanks @nguidotti !

nguidotti:
/ok to test [207fab3](https://github.com/NVIDIA/cuopt/pull/1149/commits/207fab37614779c38a7d8540cf8af84f1395b403)

nguidotti:
/ok to test [cddb787](https://github.com/NVIDIA/cuopt/pull/1149/commits/cddb7870e466b91635dadd8a2a4fcf2d84f97114)

cpp/src/dual_simplex/initial_basis.cpp
```diff
@@ -18,6 +18,90 @@
...
+  val &= 0b11;
+  if (val == 0b00) return variable_status_t::BASIC;
+  if (val == 0b01) return variable_status_t::NONBASIC_LOWER;
+  if (val == 0b10) return variable_status_t::NONBASIC_UPPER;
+  return variable_status_t::NONBASIC_FREE;
+}
+
+std::vector<uint8_t> compress_vstatus(const std::vector<variable_status_t>& vstatus)
+{
+  size_t n        = vstatus.size() / 4;
```
chris-maes: Nit: I would reserve `n` for the length of vstatus. I would use a different name (maybe `compressed_size`) for the length of the compressed vector.

cpp/src/dual_simplex/initial_basis.cpp
```diff
@@ -18,6 +18,90 @@
...
+    if (j + 3 < vstatus.size()) packed_vstatus[n] |= encode(vstatus[j + 3]) << 6;
+  }
+
+  return packed_vstatus;
+}
+
+std::vector<variable_status_t> decompress_vstatus(const std::vector<uint8_t>& packed_vstatus,
+                                                  size_t vstatus_size)
+{
+  size_t n        = vstatus_size / 4;
```
chris-maes: Same here, consider using a different name to distinguish n (which we often use to count the number of variables / columns) from the compressed size.

cpp/src/branch_and_bound/branch_and_bound.cpp
```diff
@@ -1521,6 +1531,33 @@ void branch_and_bound_t<i_t, f_t>::plunge_with(branch_and_bound_worker_t<i_t, f_
...
+    if (worker->total_active_diving_workers < worker->total_max_diving_workers &&
+        worker->node_queue.diving_queue_size() > 0) {
+      launch_diving_worker(worker);
+    }
+
+    // If any best-first worker become idle,
+    if (bfs_worker_pool_.num_idle_workers() > 0 && worker->node_queue.best_first_queue_size() > 0) {
+      worker->node_queue.lock();
+      // We need to temporarily save the lower bound in this worker so it is
+      // considered when calculating the global lower bound.
```
chris-maes: The comment says this is temporary. But I'm not sure where the lower bound is reset?  Is it
```
worker->lower_bound = std::min(worker->node_queue.get_lower_bound(), node_ptr->lower_bound);
```
You might want to call out where it is going to be reset after temporarily holding the lower bound

cpp/src/branch_and_bound/branch_and_bound.cpp
```diff
@@ -1624,28 +1683,122 @@ void branch_and_bound_t<i_t, f_t>::plunge_with(branch_and_bound_worker_t<i_t, f_
...
+
+  worker->calculate_num_diving_workers(bfs_worker_pool_.num_workers(),
+                                       diving_worker_pool_.num_workers(),
+                                       has_solver_space_incumbent(),
+                                       settings_.diving_settings);
+
+  while (solver_status_ == mip_status_t::UNSET && abs_gap > settings_.absolute_mip_gap_tol &&
+         rel_gap > settings_.relative_mip_gap_tol &&
+         worker->node_queue.best_first_queue_size() > 0) {
+    // If the guided diving was disabled previously due to the lack of an incumbent solution,
```
chris-maes, line 1771: Nit: maybe factor this out into a separate function

cpp/src/branch_and_bound/branch_and_bound.cpp
```diff
@@ -1624,28 +1683,122 @@ void branch_and_bound_t<i_t, f_t>::plunge_with(branch_and_bound_worker_t<i_t, f_
...
+      node_concurrent_halt_ = 1;
+      solver_status_        = mip_status_t::OPTIMAL;
+      break;
+    }
+
+    // Steal a node with some probability or when it is empty. The victim is determined at random.
+    if (worker->node_queue.best_first_queue_size() == 0 ||
+        worker->rng.next_double() < steal_chance) {
+      for (i_t i = 0; i < settings_.bnb_max_steal_attempts; ++i) {
+        i_t k = worker->rng.uniform(0, bfs_worker_pool_.num_workers());
```
chris-maes: Nit: I would use name the variable `victim` instead of `k`

cpp/src/branch_and_bound/branch_and_bound.cpp
```diff
@@ -2856,31 +2889,34 @@ mip_status_t branch_and_bound_t<i_t, f_t>::solve(mip_solution_t<i_t, f_t>& solut
...
         } else {
-          node_queue_.push(
-            start_node.value());  // Needed to ensure we don't lose the correct lower bound
+          // Needed to ensure we don't lose the correct lower bound
+          worker->node_queue.push(start_node);
+          lower_bound = std::min(lower_bound, worker->node_queue.get_lower_bound());
           break;
         }
       }
-      lower_bound = node_queue_.best_first_queue_size() > 0 ? node_queue_.get_lower_bound()
```
chris-maes, line 2899: Is this correct? Above you are pushing nodes into the worker queue. 
But you removed this line that gets the lower bound from the queue if there are nodes in the queue. 

Maybe this is happening in this line: 
```
lower_bound = std::min(lower_bound, worker->node_queue.get_lower_bound());
```

But it's difficult to follow. Maybe split it into two loops. The first loop, goes through the nodes in the queue and fathoms them.

The second loops of over the workers to compute the lower bound.

cpp/src/branch_and_bound/worker.hpp
```diff
@@ -160,14 +148,180 @@ class branch_and_bound_worker_t {
...
+        other->node_queue.best_first_queue_size() < 2 * num_nodes) {
+      return steal;
+    }
+
+    while (num_nodes > 0) {
+      other->node_queue.lock();
+
+      // We need to temporarily save the lower bound in this worker so it is
+      // considered when calculating the global lower bound.
+      this->lower_bound = std::min<f_t>(this->lower_bound, other->node_queue.get_lower_bound());
```
chris-maes: I'm confused as to what is happening here. Again the comment says we are temporarily saving the lower bound of the worker. But then we don't reset the lower bound

chris-maes (changes requested):
Thanks @nguidotti . I took a brief pass at this PR. 

My biggest comment is that after this PR, I find it very hard to reason about the lower bound. We've had many bugs where the lower bound is incorrect or lost. I'm concerned that unless we make the handling of the lower bound as clear as possible to follow there will be more bugs. There are so many places in the code that now track the lower bound. For instance, there is a lower bound member in the `node_queue_t`. There is a lower bound member in `bfs_worker_t` (really in `branch_and_bound_worker_t`), there is a local variable named `lower_bound` in `best_first_search_with` there is a function called `get_lower_bound`, there is the lower bound ceiling. 

Is there anyway to simplify this? For instance, do we need to have the lower bound stored in the node_queue_t and the worker? Could we only store it one place? 

Can we make clear the comments around temporarily setting the lower bound?

nguidotti:
> Thanks @nguidotti . I took a brief pass at this PR.
> 
> My biggest comment is that after this PR, I find it very hard to reason about the lower bound. We've had many bugs where the lower bound is incorrect or lost. I'm concerned that unless we make the handling of the lower bound as clear as possible to follow there will be more bugs. There are so many places in the code that now track the lower bound. For instance, there is a lower bound member in the `node_queue_t`. There is a lower bound member in `bfs_worker_t` (really in `branch_and_bound_worker_t`), there is a local variable named `lower_bound` in `best_first_search_with` there is a function called `get_lower_bound`, there is the lower bound ceiling.
> 
> Is there anyway to simplify this? For instance, do we need to have the lower bound stored in the node_queue_t and the worker? Could we only store it one place?
> 
> Can we make clear the comments around temporarily setting the lower bound?

Unfortunately, no.

Let forget the parallel execution for a moment. When we are doing best-first search with plunges, there are two places where we need to consider the lower bound: the top of the heap (`worker->node_queue.get_lower_bound()`) and the node that is currently being solved (`worker->lower_bound`). At the start of the plunge, we pop the best node from the heap, such that the heap does not contains the lowest lower bound anymore. As we do the plunge, the lower bound of the node that it is being currently solved changes and can become worse than the new top of the heap. To avoid locking the queue just for reading the lower bound, I simply use an atomic to track it, which read when call `worker->node_queue.get_lower_bound()`.

nguidotti:
For computing the lower bound for the entire solver, you take 
$$L_{global} = \min_k L^{(k)}, \qquad L^{(k)} = \min\{L_{heap}^{(k)}, L_{cur}^{(k)}\}$$
where $L_{heap}^{(k)}$ is the lower bound from the top of the heap and $L_{cur}^{(k)}$ is the node from the node being currently solved. The `k` subscript denotes the worker $k$.

nguidotti:
There are other things, we also need to consider. For instance,  we might encounter a numerical issue in a given node, which blocks the solver to progress further down that path. Similarly, starting a new worker or stealing a node, the node needs to be transferred from one worker to the next (pop, then push the best node in the heap). In the meanwhile, the lower bound of that node needs to be store somewhere so it can be considered when computing the global lower bound.

nguidotti:
I tried to concentrate all the global lower bound computation in the `get_lower_bound` in B&B, but some of the values are set during the exploration

nguidotti:
/ok to test [02d0381](https://github.com/NVIDIA/cuopt/pull/1149/commits/02d038163f8b828b34d87c93e668c3b25832aedc)

nguidotti:
/ok to test [e212f63](https://github.com/NVIDIA/cuopt/pull/1149/commits/e212f63d18542b689610ffe5908860aa47e75a28)

cpp/src/branch_and_bound/worker.hpp
```diff
@@ -42,16 +42,35 @@ struct branch_and_bound_stats_t {
...
+                                bool has_incumbent,
+                                diving_heuristics_settings_t<i_t, f_t> settings)
+{
+  switch (strategy) {
+    case BEST_FIRST: return true;
+    case PSEUDOCOST_DIVING: return settings.pseudocost_diving != 0;
+    case LINE_SEARCH_DIVING: return settings.line_search_diving != 0;
+    case GUIDED_DIVING: return settings.guided_diving != 0 && has_incumbent;
+    case COEFFICIENT_DIVING: return settings.coefficient_diving != 0;
+    default: return false;
```
aliceb-nv: We shouldn't add a default clause to most switch statements IMO - this allows the compiler to -Werror at us whenever we add a new enum value and we forget to update some of these switches :)

cpp/src/branch_and_bound/worker.hpp
```diff
@@ -160,14 +148,188 @@ class branch_and_bound_worker_t {
...
+      node_queue.push(node);
+    }
+
+    this->lower_bound = node_queue.get_lower_bound();
+  }
+
+  void set_inactive() { this->is_active = false; }
+
+  // Steal nodes from another worker
+  bool steal_node_from(bfs_worker_t* other, i_t num_nodes)
```
aliceb-nv: Let's use a reference for 'other' since it doesn't semantically make sense for it to be nullptr

cpp/src/branch_and_bound/worker.hpp
aliceb-nv: As a side note - do we (or the OMP runtime) have a way to measure and report the amount of contention on omp locks?

cpp/src/branch_and_bound/node_queue.hpp
```diff
@@ -29,52 +28,114 @@ class heap_t {
...
+  }
+
+  bool empty() const { return num_entries_ == 0; }
 
   // Read-only access to underlying buffer for iteration without modification
   const std::vector<T>& data() const { return buffer; }
 
  private:
   std::vector<T> buffer;
+  omp_atomic_t<size_t> num_entries_{0};
```
aliceb-nv, line 76: Why was this added as an atomic?

cpp/src/branch_and_bound/node_queue.hpp
```diff
@@ -29,52 +28,114 @@ class heap_t {
...
+// A queue storing the nodes waiting to be explored. Before calling pop or push in parallel,
+// the mutex NEEDS to be acquired via the `lock()` method. It must the released afterwards with
+// `unlock()`.
 template <typename i_t, typename f_t>
 class node_queue_t {
+ public:
+  void push(mip_node_t<i_t, f_t>* new_node)
+  {
+    assert(new_node != nullptr);
+    auto entry = std::make_shared<heap_entry_t>(new_node);
```
aliceb-nv: What are the ownership rules regarding new_node? Is a std::shared_ptr required here instead of unique_ptr + ownership moves?
I realize this is existing code, I'm just struggling to remember why we made these choices :)

cpp/src/branch_and_bound/constants.hpp
```diff
@@ -6,6 +6,7 @@
 /* clang-format on */
 
 #pragma once
+#include <array>
```
aliceb-nv: nit: maybe we could stick to a C array here to avoid having to pull in C++ machinery for such a very common header, to keep build times slim

cpp/src/branch_and_bound/branch_and_bound.cpp
```diff
@@ -1521,6 +1539,44 @@ void branch_and_bound_t<i_t, f_t>::plunge_with(branch_and_bound_worker_t<i_t, f_
...
+      worker->lower_bound =
+        std::min<f_t>(worker->node_queue.get_lower_bound(), worker->lower_bound);
+      mip_node_t<i_t, f_t>* node = worker->node_queue.pop_best_first();
+      worker->node_queue.unlock();
+
+      if (node != nullptr) {
+        if (!launch_bfs_worker({node})) {
+          worker->node_queue.lock();
+          worker->node_queue.push(node);
+          worker->node_queue.unlock();
```
aliceb-nv: Maybe we could turn this into a push_atomic() primitive. That'd lessen the risk of accidentally adding a racey push later down the line. If we want to keep the ability to do unlocked pushes, maybe we could make it explicit in the naming, e.g. push_nolock()?

cpp/src/branch_and_bound/branch_and_bound.cpp
```diff
@@ -1624,28 +1702,124 @@ void branch_and_bound_t<i_t, f_t>::plunge_with(branch_and_bound_worker_t<i_t, f_
...
+        i_t victim = worker->rng.uniform(0, bfs_worker_pool_.size());
+        if (worker->steal_node_from(bfs_worker_pool_[victim], settings_.bnb_nodes_per_steal)) {
+          break;
+        }
+      }
+    }
+  }
+
+  worker->set_inactive();
+  bfs_worker_pool_.return_worker_to_pool(worker);
```
aliceb-nv: Does "return_worker_to_pool" always imply "worker" is set inactive? If so, it may be safer to just have "return_worker_to_pool" unconditionally mark the worker as inactive

aliceb-nv (commented):
Excellent work Nicolas :) Mostly some nitpicks.
I've run your PR on Eos and I see material improvements:

```
            THEORETICAL BEST RESULTS (Combining Best from Both Runs)
Baseline Run:
  Feasible count: 226
  Gap <= 0.1%: 116
  Average primal gap: 0.135526
  Average dual gap: 0.223231
  Optimal (MIP gap < 1e-4): 60

Compared Run:
  Feasible count: 229
  Gap <= 0.1%: 115
  Average primal gap: 0.121167
  Average dual gap: 0.213514
  Optimal (MIP gap < 1e-4): 66

Theoretical Best (if all regressions fixed):
  Feasible count: 229
  Gap <= 0.1%: 124
  Average gap: 0.111041

Potential Improvement:
  Feasible count improvement: 0
  Average gap improvement: 0.024485

INSTANCES SOLVED TO OPTIMALITY BY COMPARED RUN BUT NOT BY BASELINE (9 instances)
Instance                                   Baseline MIP Gap   Compared MIP Gap
--------------------------------------------------------------------------------
ns1952667                                         100.0000%            0.0000%
peg-solitaire-a3                                  100.0000%            0.0000%
neos-5188808-nattai                                89.0663%            0.0000%
neos-950242                                        25.0000%            0.0000%
rail507                                             0.5747%            0.0000%
neos-4722843-widden                                 0.3150%            0.0000%
neos-4738912-atrato                                 0.0100%            0.0099%
gen-ip002                                           0.0100%            0.0099%
binkar10_1                                          0.0100%            0.0099%

INSTANCES SOLVED TO OPTIMALITY BY BASELINE BUT NOT BY COMPARED RUN (3 instances)
Instance                                   Baseline MIP Gap   Compared MIP Gap
--------------------------------------------------------------------------------
ns1208400                                           0.0000%          100.0000%
triptim1                                            0.0000%            1.3885%
neos-933966                                         0.0000%            0.9346%
```

cpp/src/branch_and_bound/node_queue.hpp
nguidotti: For each new node, we actually pushing it to two heaps simultaneously: one for best-first and another for diving. They are linked (i.e., point to the same node) in such a way that popping the node from the best-first heap will invalidate the node for diving, but not the other way around (popping the node from the diving heap does not affect the best-first heap, otherwise we will get incorrect bounds).

We use `shared_ptr` here so an entry on both heaps will point to the same object, so it is destroyed when we pop it from both heaps. We need this design because the entry position will change in both heaps when pushing/popping and the node may not exist in the other heap when we try to check if it is still valid for diving. We do not want to explore the same node twice (one for best-first and another diving) since it can already be fathomed (or the `vstatus` for that node may no longer exist)

cpp/src/branch_and_bound/node_queue.hpp
nguidotti: To avoid the need to lock the queue just for retrieving its size.

aliceb-nv: My concern is that we do this when popping:

```cpp
    if (buffer.empty()) return std::nullopt;

    std::pop_heap(buffer.begin(), buffer.end(), comp);
    T node = std::move(buffer.back());
    buffer.pop_back();
    --num_entries_;
    assert(num_entries_.load() == buffer.size());
    return node;
```

Therefore at the time of reading num_entries_ with total ordering, the buffer may have been resized/realloc'd before num_entries_ reflect the new size. Can you confirm if this is safe?

cpp/src/branch_and_bound/worker.hpp
nguidotti: There the OMPT (OpenMP Tools) interface which probes some stats from the OpenMP runtime. If you use Intel VTune or other profiler, you can see the contention for each lock and other stats

cpp/src/branch_and_bound/worker.hpp
nguidotti: The other routines in B&B already use pointers, so I think is better to keep thing consistent (also I would need to do `*bfs_worker_pool_[k]` for this routine, which is ugly...)

aliceb-nv: Then let's at least add a `assert(other)` :)

cpp/src/branch_and_bound/node_queue.hpp
nguidotti: Yeah, this may happen (`num_entries_` report a stale value). However, all changes to the buffer go through a mutex, so it should be fine. For instance, assume that the queue has only one entry. One thread is popping the node. Another reads `num_entries_ = 1` since we still do not update the value of `num_entries_` and think it is safe to pop a node from the heap. However, to do that, it needs to acquire the mutex, and thus, when it tries to pop a node, it will return as a nullopt since the buffer is empty.

nguidotti: We can move the `num_entries_` to the start of the routine, so it is the first thing updated.

cpp/src/branch_and_bound/branch_and_bound.cpp
nguidotti: Sure. Is this better [20e874a](https://github.com/NVIDIA/cuopt/pull/1149/commits/20e874ac4ebc6b3acaf381b72be8c840fc231949)?

cpp/src/branch_and_bound/node_queue.hpp
aliceb-nv: Oh okay! So that's just an opportunistic check sort of thing to avoid unecessary locks right? If so, that sounds good to me :)

nguidotti:
/ok to test [37e757a](https://github.com/NVIDIA/cuopt/pull/1149/commits/37e757a5db85023467d2d71b3ef16d4e7ea382d3)

cpp/src/branch_and_bound/worker.hpp
nguidotti: Simplified the logic here [c05b964](https://github.com/NVIDIA/cuopt/pull/1149/commits/c05b96487755bb6ce444b10ef7ed7ca2e708a271)

cpp/src/branch_and_bound/branch_and_bound.cpp
```diff
@@ -1624,28 +1696,97 @@ void branch_and_bound_t<i_t, f_t>::plunge_with(branch_and_bound_worker_t<i_t, f_
...
   }
+}
+
+template <typename i_t, typename f_t>
+void branch_and_bound_t<i_t, f_t>::best_first_search_with(bfs_worker_t<i_t, f_t>* worker)
+{
+  f_t lower_bound  = get_lower_bound();
+  f_t abs_gap      = compute_user_abs_gap(original_lp_, upper_bound_.load(), lower_bound);
+  f_t rel_gap      = user_relative_gap(original_lp_, upper_bound_.load(), lower_bound);
+  f_t steal_chance = settings_.bnb_steal_chance >= 0 ? settings_.bnb_steal_chance : 0.05;
```
chris-maes: why do you hardcode to 5% if steal chance < 0?

cpp/src/branch_and_bound/branch_and_bound.cpp
```diff
@@ -2856,31 +2877,28 @@ mip_status_t branch_and_bound_t<i_t, f_t>::solve(mip_solution_t<i_t, f_t>& solut
...
-          search_tree_.graphviz_node(
-            settings_.log, start_node.value(), "cutoff", start_node.value()->lower_bound);
-          search_tree_.update(start_node.value(), node_status_t::FATHOMED);
-          continue;
-        } else {
-          node_queue_.push(
-            start_node.value());  // Needed to ensure we don't lose the correct lower bound
-          break;
-        }
+      while (worker->node_queue.best_first_queue_size() > 0 &&
```
chris-maes, line 2922: I don't think this loop works as before. 

The point of the loop before was to go through all the nodes left in the queue and fathom those that could be fathomed. But leave nodes that could not be fathomed in the queue. 

By adding the `worker->node_queue.get_lower_bound() > upper_bound_.load()` you may not remove any nodes from the queue if the smallest node in the heap is less than the upper bound.

cpp/src/branch_and_bound/node_queue.hpp
```diff
@@ -29,52 +28,131 @@ class heap_t {
...
+  omp_atomic_t<size_t> num_entries_{0};
   Comp comp;
 };
 
-// A queue storing the nodes waiting to be explored/dived from.
+// A queue storing the nodes waiting to be explored.
 template <typename i_t, typename f_t>
 class node_queue_t {
+ public:
+  void push_nolock(mip_node_t<i_t, f_t>* new_node)
```
chris-maes: What's the difference between `push_nolock` (which internally aquires mutex_) and `push_atomic`? Some comments explaining the purpose / use of these methods would be helpful.

cpp/src/branch_and_bound/node_queue.hpp
```diff
@@ -29,52 +28,131 @@ class heap_t {
...
+    assert(new_node != nullptr);
+
+    std::lock_guard lock(mutex_);
+    auto entry = std::make_shared<heap_entry_t>(new_node);
+    best_first_heap_.push(entry);
+    if (new_node->can_dive) diving_heap_.push(entry);
+    lower_bound_ = best_first_heap_.top()->lower_bound;
+  }
+
+  void push_atomic(mip_node_t<i_t, f_t>* new_node)
```
chris-maes, line 84: The code for `push_atomic` looks the same as `push_nolock`

cpp/src/branch_and_bound/node_queue.hpp
nguidotti: My bad. `push_nolock` should be a lock-free operation when we can safely push a node. I will just remove this operation and keep the atomic one since the cost of locking the mutex is pretty low if it is no contested.

cpp/src/branch_and_bound/branch_and_bound.cpp
nguidotti: The nodes in the heap are sorted based on the lower bound, no?

nguidotti: This is the previous code, right? 
```cpp
      while (node_queue_.best_first_queue_size() > 0) {
        std::optional<mip_node_t<i_t, f_t>*> start_node = node_queue_.pop_best_first();

        if (!start_node.has_value()) { continue; }
        if (upper_bound_.load() < start_node.value()->lower_bound) {
          // This node was put on the heap earlier but its lower bound is now greater than the
          // current upper bound
          search_tree_.graphviz_node(
            settings_.log, start_node.value(), "cutoff", start_node.value()->lower_bound);
          search_tree_.update(start_node.value(), node_status_t::FATHOMED);
          continue;
        } else {
          node_queue_.push(
            start_node.value());  // Needed to ensure we don't lose the correct lower bound
          break;
        }
```
We are stopping as soon as we find a node that has a lower lower bound than the current upper bound

cpp/src/branch_and_bound/branch_and_bound.cpp
nguidotti: If we set to `-1`, we use the default setting (which is `5%`). Otherwise, we use the provided setting. In the future, we can expose the setting so the user can tune it. We also should do a sweep to find the best possible value, but this probably is system dependent

cpp/src/branch_and_bound/node_queue.hpp
nguidotti: Fixed in [b2e5f8c](https://github.com/NVIDIA/cuopt/pull/1149/commits/b2e5f8c9bdfd1939d35980d5113c1651eb933e36)

nguidotti:
/ok to test [b2e5f8c](https://github.com/NVIDIA/cuopt/pull/1149/commits/b2e5f8c9bdfd1939d35980d5113c1651eb933e36)

cpp/src/branch_and_bound/branch_and_bound.cpp
chris-maes: Maybe add a constant like `#define MIP_DEFAULT_STEAL_CHANCE 0.05`

cpp/src/branch_and_bound/branch_and_bound.cpp
chris-maes: Yes, your right these should be equivalent. We do a `break` if we need to push a node back on the queue.

chris-maes (approved):
Thanks for the nice discussion @nguidotti  and congratulations on the nice performance improvement with this PR.  I'm removing the `Request changes` so that you can merge when ready.

A couple suggestions:
1) It sounds like the only reason `node_queue_t` exposes `lock` and `unlock` is to allow a worker to steal a node from another. It would probably be better to add a method like `node_queue_t::steal_from_victim(node_queue_t& victim)` and handle the locking and unlocking of both queues directly in this method. That would allow you to not expose `node_queue_t::lock/unlock` and make it so that people touching the branch and bound code did not need to be concerned about correctly locking and unlocking the node queue. Inside `steal_from_victim` you can avoid deadlocks by acquiring the lock on the thief and the victim in sorted order according to their worker id (so this might need to be `node_queue_t::steal_from_victim(i_t thief_id, i_t victim_id, node_queue_t& victim)`)

Ideally, stealing a node is an atomic operation, so that the node is always either in one queue or another, and thus the node's lower bound is always considered. If you are able to make it an atomic operation you can avoid the need to track the lower bound associated with the node separately (which may be prone to bugs).

Also, if diving needs to copy a node from the node queue, and that cannot happen while stealing, you can add a method `node_queue_t::copy_node` that acquires `mutex_` internally. 

Maybe you are able to make the above changes before merging.

2) Longer term, I think it's worth defining the correct abstractions and data structures to make managing the lower bound simpler. A heap is already the ideal data structure for managing the lower bound, since it inherently takes the lower bound over the nodes it contains. I think we've introduced a lot of book-keeping and other data structures to manage nodes and lower bounds outside the heap. This is likely because we don't have a way to walk nodes in the heap (i.e. the standard C++ data structures only support pushing and popping). If we had a heap where we could walk nodes, I think it would simplify many of operations within the branch and bound code. Instead of popping a node off the heap, and tracking the lower bound, when solving, we could leave it on the heap and just mark that node as "solve in progress". We would only pop a node from the heap when the solve was completed. When trying to steal a node from a heap or dive from a node, thieves could avoid "solve in progress" nodes. Also, during a plunge we could push child nodes that we are not exploring directly onto the heap, instead of keeping them in a separate stack or circular buffer data structure. 

If the code maintained the invariant that all open nodes are in a heap, I think it would be much easier to reason about the correctness of branch and bound.

nguidotti:
/ok to test [d094751](https://github.com/NVIDIA/cuopt/pull/1149/commits/d0947518a1bb11068430ee69707f538683dc3799)

nguidotti:
/ok to test [6f7ab06](https://github.com/NVIDIA/cuopt/pull/1149/commits/6f7ab064b29c929545542cf1d317658861d9e4b9)

nguidotti:
/ok to test [d18c1e9](https://github.com/NVIDIA/cuopt/pull/1149/commits/d18c1e9d235142bb1bf6609dfc3415d5a4ec70df)

PR #1150, opened 2026-04-27
Clique-aware bound propagation and clique cut fixes

akifcorduk:
This PR makes heuristic bound propagation clique-aware: when the conflict graph tells us a set of binary variables can't all be 1 at once, the propagator uses that to tighten activities (and therefore bounds) more aggressively than the standard one-variable-at-a-time analysis. It also fixes longstanding performance issues in the clique table and consolidates how the table is shared between B&B and the heuristics.

Heuristic probing and bound propagation now feed the clique table into a new activity-correction pass. On constraints that touch a clique, we subtract a correction so activity bounds reflect "at most one of these can be 1" instead of treating the variables independently. On constraints that don't touch any clique, behavior is bit-identical to before. The pass only re-runs corrections on constraints whose bounds actually moved, so the per-iteration cost scales with dirty constraints rather than total constraints.

The per-variable lookups (which clique contains a variable, what's adjacent to it, what conflicts it participates in) used to go through hash sets and a linear scan. They now use a sorted CSR, which removes the linear scan and most of the hashing overhead from the hot loops in clique extension and clique-cut separation.

B&B now publishes its async-built table back to the heuristics layer the moment it's ready, well before the cut pass joins on it, through a small atomic publish/snapshot API on `problem_t`. 

- A critical bug fix:Clique-cut separation no longer early-exits when the "first" / "additional" clique containers happen to be empty — sub-threshold conflicts that were demoted to pairwise edges are still valid and Bron–Kerbosch still consumes them.
- Clique cuts handle the case where a clique contains both a variable and its complement (one such pair forces every other clique member to 0; two or more proves infeasibility) instead of bailing out.
- Few more bug fixes and perf improvements.

Note: this is only GPU bound propagation, CPU bound propagation will come in a follow-up PR.

Benchmark results and clique cut impact pending.


cpp/src/mip_heuristics/diversity/diversity_manager.cu
```diff
@@ -252,32 +251,11 @@ bool diversity_manager_t<i_t, f_t>::run_presolve(f_t time_limit, timer_t global_
...
-  //     std::iota(all_var_indices.begin(), all_var_indices.end(), 0);
-  //     problem_ptr->update_variable_bounds(all_var_indices, host_problem.lower,
-  //     host_problem.upper); trivial_presolve(*problem_ptr, remap_cache_ids);
-  //   }
-  // }
+  // The initial clique table is built asynchronously by B&B
+  // (find_initial_cliques) and published into problem_ptr->clique_table.
+  // Heuristics pick it up via the atomic snapshot accessor. We intentionally
+  // do not build a duplicate table here — it added 1-3 s to the critical path
+  // of presolve without materially changing what B&B produces.
```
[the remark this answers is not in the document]
akifcorduk: This is intended. Currently, the initial presolve will not used clique table.

cpp/src/cuts/cuts.cpp
```diff
@@ -1813,6 +1857,16 @@ bool cut_generation_t<i_t, f_t>::generate_cuts(const lp_problem_t<i_t, f_t>& lp,
...
 
+  // Generate implied bound cuts
+  if (settings.implied_bound_cuts != 0) {
+    f_t cut_start_time = tic();
+    generate_implied_bound_cuts(lp, settings, var_types, xstar, start_time);
+    f_t cut_generation_time = toc(cut_start_time);
+    if (cut_generation_time > 1.0) {
+      settings.log.debug("Implied bounds cut generation time %.2f seconds\n", cut_generation_time);
+    }
+  }
```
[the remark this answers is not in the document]
akifcorduk: @chris-maes should we add timer in `generate_implied_bound_cuts` do you think it might cause timer issues?

cpp/src/mip_heuristics/presolve/bounds_presolve.cu
```diff
@@ -88,6 +91,40 @@ void bound_presolve_t<i_t, f_t>::resize(problem_t<i_t, f_t>& problem)
...
+  // from our reads (see clique_table_t::ready_for_heuristics).
+  clique_data.build_from_host(pb, context.problem_ptr->reverse_original_ids, *current_ct);
+  upd.resize_clique_buffers(clique_data.n_groups, pb.handle_ptr->get_stream());
+  last_built_problem       = &pb;
+  last_built_clique_table  = current_ct;
+  last_built_n_variables   = pb.n_variables;
+  last_built_n_constraints = pb.n_constraints;
+  last_built_nnz           = pb.nnz;
+  clique_data_built        = true;
 }
```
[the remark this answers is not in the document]
akifcorduk: This is okay, bound propagation is run many times and its runtime is negligable.

gforsyth (approved):

cpp/src/mip_heuristics/presolve/bounds_presolve.cu
```diff
@@ -119,10 +154,49 @@ bool bound_presolve_t<i_t, f_t>::calculate_bounds_update(problem_t<i_t, f_t>& pb
...
+      <<<apply_blocks, apply_tpb, 0, stream>>>(make_span(clique_data.constraint_group_offsets),
+                                               make_span(upd.group_max_correction),
+                                               make_span(upd.group_min_correction),
+                                               make_span(upd.changed_constraints),
+                                               make_span(upd.min_activity),
+                                               make_span(upd.max_activity),
+                                               pb.n_constraints);
+    RAFT_CHECK_CUDA(stream);
+
+    update_bounds_kernel_cliq<i_t, f_t, n_threads>
```
rg20, line 192: Can you re-use the same update_bounds_kernel with an added template?   you can avoid some duplicated code.

cpp/src/branch_and_bound/branch_and_bound.cpp
```diff
@@ -299,16 +300,17 @@ branch_and_bound_t<i_t, f_t>::branch_and_bound_t(
...
+  const f_t upper_bound  = upper_bound_.load();
   f_t lower_bound        = lower_bound_ceiling_.load();
   f_t heap_lower_bound   = node_queue_.get_lower_bound();
   f_t worker_lower_bound = worker_pool_.get_lower_bound();
   lower_bound            = std::min(heap_lower_bound, lower_bound);
   lower_bound            = std::min(worker_lower_bound, lower_bound);
 
   if (std::isfinite(lower_bound)) {
-    return lower_bound;
+    return std::min(lower_bound, upper_bound);
```
aliceb-nv, line 311: That shouldn't happen right (besides numerical shenanigans)? Maybe we should add an assertion so that we can catch these scenarios and root-cause them if they occur. I will defer to Chris/Nicolas L

cpp/src/mip_heuristics/presolve/clique_activity_corrections.cu
```diff
@@ -0,0 +1,544 @@
...
+      max_groups_per_cnst,
+      phase1_groups,
+      phase2_groups);
+    CUOPT_LOG_INFO(
+      "clique_group_table_t::build_from_host: sources: large=%zu, addtl=%zu, "
+      "materialized=%zu, small_adj_edges=%zu",
+      clique_table.first.size(),
+      clique_table.addtl_cliques.size(),
+      all_cliques.size(),
+      clique_table.small_clique_adj.indices.size());
```
aliceb-nv, line 533: Should probably be reduced down to _DEBUG/_TRACE

cpp/src/mip_heuristics/presolve/clique_activity_corrections.cuh
```diff
@@ -0,0 +1,268 @@
...
+    max1         = new_max1;
+    max2         = new_max2;
+
+    f_t d1       = __shfl_xor_sync(0xffffffff, min1, off);
+    f_t d2       = __shfl_xor_sync(0xffffffff, min2, off);
+    f_t new_min1 = fmin(min1, d1);
+    f_t new_min2 = fmin(fmax(min1, d1), fmin(min2, d2));
+    min1         = new_min1;
+    min2         = new_min2;
+  }
```
aliceb-nv, line 186: Do you reckon this could be changed to raft::warpReduce or similar? For consistency with the rest of the codebase

cpp/src/mip_heuristics/problem/problem.cu
```diff
@@ -478,6 +479,26 @@ void csr_to_csc_transpose(const i_t* csr_offsets,
...
+
+template <typename i_t, typename f_t>
+void problem_t<i_t, f_t>::publish_clique_table(std::shared_ptr<clique_table_t<i_t, f_t>> ct)
+{
+  // Single-writer, many-reader publication. Invoked once from B&B's async
+  // clique-build task. Release-store pairs with the acquire-loads in
+  // get_clique_table_snapshot and in copy constructors that snapshot the
+  // shared_ptr during object duplication.
+  std::atomic_store_explicit(&clique_table, std::move(ct), std::memory_order_release);
+}
```
aliceb-nv, line 500: We build in C++20 mode now right? Maybe we could just switch to std::atomic<std::shared_ptr>>

aliceb-nv (approved):
Looking good :) Amazing work Akif. Minor comments

Additionally - do you think it'd be possible to gather up implementation details around cliques in a document somewhere? The solver is becoming more extensive and complex by the minute and I found it quite hard to fully make sense of everything. Maybe we could benefit from some implementation docs somewhere (maybe internal)

Edgarruiz8585:
Referencia : https://github.com/Edgarruiz8585
by Edgarruiz8585 @Edgar Manuel Ruiz Arias
***@***.***>

Edgarruiz8585:
Skip

Edgarruiz8585:
Skip ruizariasedgarmanuel@gmail.com

akifcorduk:
Closing for now. I will reopen it when performance is verified.

PR #1152, opened 2026-04-28
Extend batch pdlp support

Kh4ster:
This PR greatly extends the capabilities of batch PDLP. Former batch PDLP only supported having a single variable bounds being different per climber. It now supports:
- Different constraints lower and upper bounds per climber
- Different objective coefficients per climber
- Different objective offset per climber
- More than one variable bound difference per climber

This PR also adds the support of per climber residual and first primal feasible to the Stable3 PDLP solver mode and its batch version. It allows to solve a batch of problems and stop once one or all the climbers have reached primal feasibility.

All those combinations can be put together, resulting in a potential:

Solve a batch of LPs all having different: constraints lower and upper bounds, objective coefficients, objective offset, variable bounds, using per constraint residual instead of the L2 norm and stopping once one, or all, climbers have reached primal feasibility.


cpp/src/pdlp/pdlp.cu
```diff
@@ -755,6 +780,52 @@ void pdlp_solver_t<i_t, f_t>::print_final_termination_criteria(
...
+  if (sb_view_.is_valid()) { sb_view_.mark_solved(local_idx); }
+}
+
+template <typename i_t, typename f_t>
+optimization_problem_solution_t<i_t, f_t> pdlp_solver_t<i_t, f_t>::finalize_batch_return()
+{
+  current_termination_strategy_.fill_gpu_terms_stats(total_pdlp_iterations_);
+  RAFT_CUDA_TRY(cudaStreamSynchronize(stream_view_));
+  current_termination_strategy_.convert_gpu_terms_stats_to_host(
+    batch_solution_to_return_.get_additional_termination_informations());
```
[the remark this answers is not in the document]
Kh4ster: This review is incorrect. There is no index-aliasing bug in convert_gpu_terms_stats_to_host, because the GPU mirror and the host destination are both already in original-index space. The kernel writes a destination remap, not a source remap — adding original_index_[i] on the read side would actually break the function.

1. Sizing. Both arrays are sized to original_batch_size_ and never resized:

// termination_strategy.cu (ctor): sized once, never grows or shrinks.
gpu_batch_additional_termination_information_{climber_strategies.size()},
original_index_(climber_strategies.size()),
// pdlp.cu: host destination resized to the original batch size.
batch_solution_to_return_.get_additional_termination_informations().resize(original_batch_size_);
batch_solution_to_return_.get_terminations_status().resize(original_batch_size_);
2. The kernel writes into original-index space, not current-index space.

// fill_gpu_terms_stats_kernel
const int idx = threadIdx.x + blockIdx.x * blockDim.x;       // current/local slot
const i_t original_index = original_indices[idx];            // -> original slot
additional_termination_information.<field>[original_index] = ...; // dst is original
So after the kernel runs, gpu_batch_additional_termination_information_[k] holds the stats of the climber that originally lived at slot k. The pinned mirror is dense in original-index space; current/local indices idx only appear on the source side (reading from convergence_information_view, which tracks the live shrunk batch).

3. Every host-side write into batch_solution_to_return_ is also in original-index space. Examples in pdlp.cu:

// snapshot_climber_into_return:
const i_t local_idx = climber_strategies_[i].original_index;
batch_solution_to_return_.get_terminations_status()[local_idx] = term;
batch_solution_to_return_.get_additional_termination_informations()[local_idx] = ...;
// time-/iteration-/concurrent-limit fallbacks:
batch_solution_to_return_.get_terminations_status()
    [climber_strategies_[i].original_index] = pdlp_termination_status_t::TimeLimit;
4. Therefore the slot-by-slot copy is correct. convert_gpu_terms_stats_to_host iterates i ∈ [0, original_batch_size_) and copies host[i] = gpu[i]. Both sides have entry i = original climber i. No remap is needed.

5. The proposed fix would corrupt the result. host[i] = gpu[original_index_[i]] is only well-defined for i < climber_strategies_.size() (the current count). It would gather the stats of the alive climbers and pack them into the first current_count host slots, shifting them away from their original_index positions. That misaligns the additional-info entries against terminations_status (which has been written at original-index positions everywhere else in the codebase). The corruption you describe would actually be introduced by the suggested change, not removed by it.

I will add comments to clarify things.

cpp/src/pdlp/initial_scaling_strategy/initial_scaling.cu
```diff
@@ -137,7 +145,13 @@ void pdlp_initial_scaling_strategy_t<i_t, f_t>::bound_objective_rescaling()
   h_bound_rescaling = f_t(1.0) / (std::sqrt(bound_rescaling_.value(stream_view_)) + f_t(1.0));
   bound_rescaling_.set_value_async(h_bound_rescaling, stream_view_);
 
-  detail::my_l2_weighted_norm<i_t, f_t>(op_problem_scaled_.objective_coefficients,
+  // ------- Objective coefficients scaling -------
+
+  // If climbers have different objective coefficients, we just compute the norm with the first
```
akifcorduk: Can't we have different objective scaling for each climber? That probably means having different variable bounds per each climber by default.

cpp/src/pdlp/termination_strategy/termination_strategy.cu
```diff
@@ -337,16 +331,18 @@ __global__ void check_termination_criteria_kernel(
     {
       termination_status[idx] = (i_t)pdlp_termination_status_t::PrimalFeasible;
       return;
+    } else {
+      termination_status[idx] = (i_t)pdlp_termination_status_t::NoTermination;
```
akifcorduk, line 316: Why is this now necessary?

cpp/src/pdlp/pdhg.cu
```diff
@@ -170,6 +170,15 @@ void pdhg_solver_t<i_t, f_t>::swap_context(
   matrix_swap(reflected_dual_, dual_size_h_, swap_pairs);
   matrix_swap(dual_slack_, primal_size_h_, swap_pairs);
   current_saddle_point_state_.swap_context(swap_pairs);
+  // Swap per-climber scaled problem fields (objectives, constraint bounds) — all in COL-major
+  // during the convergence block when swap_context is invoked.
+  if (problem_ptr->objective_coefficients.size() > static_cast<size_t>(primal_size_h_)) {
+    matrix_swap(problem_ptr->objective_coefficients, primal_size_h_, swap_pairs);
```
akifcorduk, line 270: I am not sure if i understand the `matrix_swap` logic

cpp/src/pdlp/pdhg.cu
```diff
@@ -666,10 +685,11 @@ struct dual_reflected_major_projection_bulk_op {
...
+    const int bound_idx = per_climber_constraints ? idx : constraint_idx;
+    const f_t tmp       = current_dual / step_size - Ax;
     const f_t tmp_proj =
-      cuda::std::max<f_t>(-constraint_upper_bounds[constraint_idx],
-                          cuda::std::min<f_t>(tmp, -constraint_lower_bounds[constraint_idx]));
+      cuda::std::max<f_t>(-constraint_upper_bounds[bound_idx],
+                          cuda::std::min<f_t>(tmp, -constraint_lower_bounds[bound_idx]));
     const f_t next_dual = (tmp - tmp_proj) * step_size;
 
     potential_next_dual[idx] = next_dual;
```
akifcorduk, line 823: Why is dual indexing not the same way as constraint_bound indexing?  Also, we can just use `idx` in `constraint_upper_bounds`. For single climber, it will be equal to  `constraint_idx` anyway which is equal to `idx`. For batched case, it is also equal to `idx` all the time. So no need for extra `bound_idx`

cpp/src/pdlp/pdhg.cu
```diff
@@ -745,10 +768,11 @@ struct dual_reflected_projection_bulk_op {
     cuopt_assert(!isinf(step_size), "dual_step_size is Inf in dual_reflected_projection");
     cuopt_assert(step_size > f_t(0.0), "dual_step_size must be > 0");
 
-    const f_t tmp = current_dual / step_size - dual_gradient[idx];
+    const int bound_idx = per_climber_constraints ? idx : constraint_idx;
```
akifcorduk, line 904: Same here

cpp/src/pdlp/pdlp.cu
```diff
@@ -98,21 +99,36 @@ inline cublasStatus_t cublasGeam<double>(cublasHandle_t handle,
 }
 
 template <typename i_t, typename f_t>
-static size_t batch_size_handler(const problem_t<i_t, f_t>& op_problem,
-                                 const pdlp_solver_settings_t<i_t, f_t>& settings)
+static size_t batch_size_handler(const pdlp_solver_settings_t<i_t, f_t>& settings)
 {
-  if (settings.new_bounds.empty()) { return 1; }
+  // Two inputs only:
+  //   - fixed_batch_size > 0 : caller pre-sized the batch (fixed path). Per-climber problem data
```
akifcorduk, line 131: Do we need this logic?  I understand user might need to query the batch size and pdlp needs to setup accordingly. So we can just unify it by removing the inference from the size right? That will have single source of truth which is the initial batch size that user passes.

cpp/src/pdlp/pdlp.cu
```diff
@@ -1066,7 +1071,8 @@ std::optional<optimization_problem_solution_t<i_t, f_t>> pdlp_solver_t<i_t, f_t>
           : pdhg_solver_.get_potential_next_dual_solution(),
         get_filled_warmed_start_data(),
         {termination_current});
-    } else if (termination_average == pdlp_termination_status_t::PrimalFeasible) {
+    } else if (!settings_.hyper_params.never_restart_to_average &&
```
akifcorduk, line 1060: Why is `never_restart_to_average` needed now?

cpp/src/pdlp/pdlp.cu
```diff
@@ -2079,6 +2103,57 @@ void pdlp_solver_t<i_t, f_t>::compute_fixed_error(std::vector<int>& has_restarte
...
+                                   reusable_device_scalar_value_1_.data(),
+                                   field.data(),
+                                   climber_strategies_.size(),
+                                   reusable_device_scalar_value_0_.data(),
+                                   nullptr,
+                                   rows,
+                                   transposed.data(),
+                                   rows));
+    }
+    raft::copy(field.data(), transposed.data(), field.size(), stream_view_);
```
akifcorduk, line 2124: You need a stream synchronization here right? `transposed` is allocated within scope and will be destructed before the copy finishes.

cpp/src/pdlp/pdlp.cu
```diff
@@ -2079,6 +2103,57 @@ void pdlp_solver_t<i_t, f_t>::compute_fixed_error(std::vector<int>& has_restarte
...
+// In PDHG everything is ROW-major for faster SpMM.
+// The scaled fields need to be tranposed back to COL-major as we might need to swap and resize
+// them. No op if the fields were not expanded
+template <typename i_t, typename f_t>
+void pdlp_solver_t<i_t, f_t>::transpose_problem_fields(bool to_row)
+{
+  auto transpose_field = [&](rmm::device_uvector<f_t>& field, i_t rows) {
+    if (field.size() <= static_cast<size_t>(rows)) return;
+    rmm::device_uvector<f_t> transposed(field.size(), stream_view_);
+    if (to_row) {
```
akifcorduk: Instead of writing two branches, you can write a single transpose with just a pointer swap between `climber_strategies_` and `rows`

cpp/src/pdlp/solve.cu
```diff
@@ -909,34 +935,155 @@ static double batch_pdlp_memory_estimator(const optimization_problem_t<i_t, f_t>
...
 template <typename i_t, typename f_t>
-optimization_problem_solution_t<i_t, f_t> run_batch_pdlp(
-  optimization_problem_t<i_t, f_t>& problem, pdlp_solver_settings_t<i_t, f_t> const& settings)
+static void apply_batch_settings_overrides(
+  const pdlp_solver_settings_t<i_t, f_t>& original_settings,
+  pdlp_solver_settings_t<i_t, f_t>& batch_settings)
 {
-  // Hyper parameter than can be changed, I have put what I believe to be the best
+  constexpr int batch_iteration_limit = 100000;
+  constexpr f_t pdlp_tolerance        = 1e-4;
```
akifcorduk, line 951: Do we have a fixed precision(1e-4) for batched pdlp ?

cpp/src/pdlp/solve.cu
```diff
@@ -909,34 +935,155 @@ static double batch_pdlp_memory_estimator(const optimization_problem_t<i_t, f_t>
...
+  }
+  if (original_settings.has_initial_dual_solution() && pdlp_primal_dual_init) {
+    batch_settings.set_initial_dual_solution(
+      original_settings.get_initial_dual_solution().data(),
+      original_settings.get_initial_dual_solution().size(),
+      original_settings.get_initial_dual_solution().stream());
+  }
+  // Step size doesn't change anyways, just to save the compute
+  if (original_settings.get_initial_step_size().has_value()) {
+    batch_settings.set_initial_step_size(original_settings.get_initial_step_size().value());
```
akifcorduk, line 1012: Can't we set different initial steap size and weights for different climbers?

cpp/src/pdlp/solve.cu
```diff
@@ -1093,6 +1183,53 @@ optimization_problem_solution_t<i_t, f_t> run_batch_pdlp(
...
+                                  bool per_climber_objectives,
+                                  bool per_climber_constraint_bounds,
+                                  bool collect_solutions)
+{
+  // Find the maximum batch size that can be used without exceeding the free memory
+
+  // Since we decerement iteratively, we don't want to use std::numeric_limits<size_t>::max()
+  // Even if 20K fits in memory it will never be an optimal batch size,  it's just to have a
+  // reasonable upper bound
+  constexpr size_t max_batch_size    = 20000;
```
akifcorduk, line 1378: Maybe we should put that into a global config or some user facing API ?

cpp/tests/linear_programming/pdlp_test.cu
```diff
@@ -268,7 +275,7 @@ TEST(pdlp_class, run_sub_mittleman)
       settings.pdlp_solver_mode = solver_mode;
       settings.dual_postsolve   = false;
       for (auto [presolver, epsilon] :
-           {std::pair{presolver_t::Papilo, 1e-1}, std::pair{presolver_t::None, 1e-6}}) {
+           {std::pair{presolver_t::Papilo, 1e-1}, std::pair{presolver_t::None, 1e-4}}) {
```
akifcorduk, line 402: Why tolerance change?

akifcorduk (commented):
Thanks for the great work Nicolas! 

In your tests, can you add more tests with randomized bounds. You do a lot of index indirection and we might be missing a bug that we are always checking for example the first batch's bounds. It would be nice to have a test with randomized bounds, objectives etc. and at the end the result of each batch is the same as single run of pdlp.

cpp/src/pdlp/termination_strategy/termination_strategy.cu
Kh4ster: If you are using per_constraint_residual and neither first_primal_feasible nor all_primal_feasible, an instance may become PrimalFeasible but then lose it, so we have to set it back to NoTermination. This is not the case for Optimal because we will always remove a climber (and end in the non-batch case) if it found optimal.
I also have the same in the non per_constraint_residual path

cpp/src/pdlp/pdhg.cu
Kh4ster: When a climber is removed from the batch, all its associated data needs to be removed so that the rest of the solve can continue. For many of those fields, this represented a list of vectors stacked next to each other, aka a dense matrix. In that case I need to identify all the slices I need to move around in the matrix and swap them accordingly.
We can now have a matrix of objective offset, so if the associated climber is removed, we need to remove this slice. That being said, we can be in batch mode without having a matrix of objective_coefficients, hence the if.

cpp/src/pdlp/pdhg.cu
Kh4ster: We can have a batch case where all climbers have the same constraint_lower_bounds. Then we have to use constraint_idx and not global idx.

cpp/src/pdlp/pdhg.cu
Kh4ster: Same remark as above

cpp/src/pdlp/pdlp.cu
Kh4ster: Before we would use first PrimalFeasible in Stable2 mode which had a notion of average. In Stable3 we don't have any average and so we don't want to go in that code path.

cpp/src/pdlp/pdlp.cu
Kh4ster: I'm not sure actually. I thought rmm::device_uvector<f_t> destructor would sync on the stream used before. Let me check

Kh4ster: This is what I'm understanding from https://docs.rapids.ai/api/rmm/stable/cpp/data_containers/#_CPPv4N3rmm13device_bufferD0Ev but let me confirm with them

cpp/src/pdlp/initial_scaling_strategy/initial_scaling.cu
Kh4ster: You are right, let me change that.

Kh4ster: Should be good now

cpp/src/pdlp/pdlp.cu
Kh4ster: You are right. Fixed and in other places in the code

cpp/src/pdlp/pdlp.cu
Kh4ster: You are right, changed

cpp/src/pdlp/solve.cu
Kh4ster: Yes and it's intended to have a good warm start and fast heuristics for strong branching but you are right, I should have a clean API to handle all of that properly.

cpp/src/pdlp/solve.cu
Kh4ster: That's a research question. For now we use the same warm information across climbers

cpp/src/pdlp/solve.cu
Kh4ster: Yes, I have to think about that

cpp/src/pdlp/pdlp.cu
```diff
@@ -755,6 +720,86 @@ void pdlp_solver_t<i_t, f_t>::print_final_termination_criteria(
...
+    batch_solution_to_return_.get_primal_solution(),
+    batch_solution_to_return_.get_dual_solution(),
+    batch_solution_to_return_.get_reduced_cost(),
+    get_filled_warmed_start_data(),
+    problem_ptr->objective_name,
+    problem_ptr->var_names,
+    problem_ptr->row_names,
+    std::move(batch_solution_to_return_.get_additional_termination_informations()),
+    std::move(batch_solution_to_return_.get_terminations_status())};
+}
```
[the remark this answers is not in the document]
Kh4ster: Fixed

cpp/src/pdlp/pdhg.cu
akifcorduk: Oh I see, thanks!

cpp/src/pdlp/solve.cu
akifcorduk: We need 1e-6 abs and 1e-12 relative tolerance in MIP.

cpp/src/pdlp/solve.cu
akifcorduk: For MIP use cases, it might make sense to pass different warm starts, but let's keep it for another PR.

akifcorduk (approved):

cpp/src/grpc/cuopt_remote.proto
```diff
@@ -122,6 +122,7 @@ message PDLPSolverSettings {
   bool save_best_primal_so_far = 28;
   bool first_primal_feasible = 29;
   int32 pdlp_precision = 30;
+  bool all_primal_feasible = 31;
```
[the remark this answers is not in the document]
Kh4ster: @tmckayus I would like your opinion here please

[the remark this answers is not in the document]
tmckayus: Thanks @Kh4ster. I took a closer look, and currently we are not supporting batch PDLP for gRPC, only single problems.  If/when we decided to support batch PDLP on grpc, we'll add it all together (there are other settings we aren't passing, and we need a message scheme to handle the multiple problems and results, etc).

So, I removed the new proto mapping and added a comment.

tmckayus (approved):

PR #1153, opened 2026-04-28
Reduce memory footprint of cuPDLPx

Bubullzz:
This PR removes most of the unused/one-time-used device_vectors in cuPDLPx to reduce the memory footprint. It allows to run bigger problems without running out of memory


Bubullzz:
/ok test [5e9d12b](https://github.com/NVIDIA/cuopt/pull/1153/commits/5e9d12b3cba418273b9834b2b8368e9f70130275)

cpp/src/pdlp/pdlp.cuh
```diff
@@ -143,6 +143,7 @@ class pdlp_solver_t {
     settings_.shared_sb_solved};
 
   problem_t<i_t, f_t>* problem_ptr;
+  bool is_cupdlpx_;
```
mlubin: We should document what `true` and `false` means here.

cpp/src/pdlp/saddle_point.hpp
```diff
@@ -64,7 +64,8 @@ class saddle_point_state_t {
   saddle_point_state_t(raft::handle_t const* handle_ptr,
                        i_t primal_size,
                        i_t dual_size,
-                       size_t batch_size);
+                       size_t batch_size,
+                       const bool is_cupdlpx = false);
```
mlubin: This seems like a pretty important argument, consider documenting and not having a default.

cpp/src/pdlp/pdlp.cu
```diff
@@ -2263,6 +2277,18 @@ optimization_problem_solution_t<i_t, f_t> pdlp_solver_t<i_t, f_t>::run_solver(co
   // Update FP32 matrix copies for mixed precision SpMV after scaling
   pdhg_solver_.get_cusparse_view().update_mixed_precision_matrices();
 
+  // Redirect cuSPARSE descriptors to use the original problem's structural data (offsets, indices),
```
mlubin, line 2323: Does this optimization help reduce the peak memory usage?
I don't understand the details, but please confirm that memory management remains correct with this change.

cpp/src/pdlp/pdlp.cu
```diff
@@ -108,6 +108,14 @@ static size_t batch_size_handler(const problem_t<i_t, f_t>& op_problem,
   return settings.new_bounds.size();
 }
 
+template <typename i_t, typename f_t>
+static void lighten_problem_for_pdlp(problem_t<i_t, f_t>& problem, rmm::cuda_stream_view stream)
```
mlubin: Consider a more technically precise name, or document. Why are these fields populated originally if they're not needed by PDLP?

mlubin (commented):
Nice! Do you have any concrete numbers in the reduction of peak memory usage?

cpp/src/pdlp/pdlp.cu
```diff
@@ -2444,6 +2470,10 @@ optimization_problem_solution_t<i_t, f_t> pdlp_solver_t<i_t, f_t>::run_solver(co
     raft::print_device_vector("Initial primal_step_size", primal_step_size_.data(), 1, std::cout);
     raft::print_device_vector("Initial dual_step_size", dual_step_size_.data(), 1, std::cout);
   }
+  if (!settings_.per_constraint_residual) {
+    lighten_problem_for_pdlp(op_problem_scaled_, stream_view_);
+    lighten_problem_for_pdlp(*problem_ptr, stream_view_);
```
Kh4ster: Am I missing something? Aren't we already calling those above?

cpp/src/pdlp/pdlp.cuh
Bubullzz: Added `// Flag indicating whether the restart strategy is CUPDLPx (hyper_params.restart_strategy == static_cast<int>(pdlp_restart_strategy_t<i_t, f_t>::restart_strategy_t::CUPDLPX_RESTART))`

cpp/src/pdlp/pdlp.cu
Bubullzz: Yes it helps reducing memory usage and yes I confirm memory management remains correct.

cpp/src/pdlp/pdlp.cu
Bubullzz: you are right, these kind of fields are managed by the copy constructor with the argument no_deep_copy = true (it is used for pdlp). I followed the current flow and it makes more sense now

cpp/src/pdlp/cusparse_view.hpp
```diff
@@ -208,6 +208,10 @@ class cusparse_view_t {
 
   // Update FP32 matrix copies after scaling (must be called after scale_problem())
   void update_mixed_precision_matrices();
+
+  // Redirects the vectors of rows and columns from op_problem_scaled_ to the original problem so
+  // that they can be freed
+  void redirect_rows_and_cols(const problem_t<i_t, f_t>& original_problem);
```
rg20: Why not just `move` ?  

This is very specific to cusparse pointers right?   I think the naming should reflect that.

cpp/src/pdlp/pdlp.cuh
```diff
@@ -147,6 +147,9 @@ class pdlp_solver_t {
   // compute_initial_primal_weight_before_scaling is false because of compute_initial_primal_weight
   problem_t<i_t, f_t> op_problem_scaled_;
 
+  // Flag indicating whether the restart strategy is CUPDLPx (hyper_params.restart_strategy == static_cast<int>(pdlp_restart_strategy_t<i_t, f_t>::restart_strategy_t::CUPDLPX_RESTART))
+  bool is_cupdlpx_;
```
rg20: I recommend switching to an enum here.   That makes it easy to add new algorithms in future.

cpp/src/pdlp/saddle_point.hpp
```diff
@@ -64,7 +64,8 @@ class saddle_point_state_t {
   saddle_point_state_t(raft::handle_t const* handle_ptr,
                        i_t primal_size,
                        i_t dual_size,
-                       size_t batch_size);
+                       size_t batch_size,
+                       const bool need_primal_gradient);
```
rg20: how do we determine if primal gradient is needed or not?    Should this have some defaults?

Bubullzz: It depends on the algortihm used for the restart. I will just pass the hyper-parameters to the function this will be more clear

cpp/src/pdlp/cusparse_view.hpp
Bubullzz: We still need original_problem to own these pointers. Here we are only borrowing it from scaled problem so we can free duplicated data, this is why I don't think we should use `move` here.
I will rename it to reflect its link with cusarse ptr

cpp/src/pdlp/pdlp.cuh
Bubullzz: this bool was almost never used. I deleted it

Bubullzz:
/ok to test [851be7c](https://github.com/NVIDIA/cuopt/pull/1153/commits/851be7c8f670ffc54bed634bab2b50e2ca0d90c0)

mlubin (approved):
LGTM, but don't merge without @rg20's approval.

Bubullzz:
/ok to test [6981c1f](https://github.com/NVIDIA/cuopt/pull/1153/commits/6981c1fd69cbee6858a019addfbc2e67bbeb1ce0)

Bubullzz:
/ok to test [29bd4ca](https://github.com/NVIDIA/cuopt/pull/1153/commits/29bd4ca218d3742b0bc86095f4018702a866e47a)

Bubullzz:
/ok to test [29bd4ca](https://github.com/NVIDIA/cuopt/pull/1153/commits/29bd4ca218d3742b0bc86095f4018702a866e47a)

Bubullzz:
/ok to test [72afe0a](https://github.com/NVIDIA/cuopt/pull/1153/commits/72afe0a00b3915e74837b073b9d236d7c053e157)

Bubullzz:
/ok to test [cea749f](https://github.com/NVIDIA/cuopt/pull/1153/commits/cea749f3ae3828cc6468a1446f5cdd4be836a304)

Bubullzz:
/ok to test [1e2b582](https://github.com/NVIDIA/cuopt/pull/1153/commits/1e2b582ac8d70f94b103c33019c0f064bb334eee)

Bubullzz:
/ok to test [61e4681](https://github.com/NVIDIA/cuopt/pull/1153/commits/61e4681e83dc19d0ecd7aa4af07a92bb233eedd5)

Bubullzz:
/ok to test [aab5a29](https://github.com/NVIDIA/cuopt/pull/1153/commits/aab5a29688fc2ea8e26c10b0c2f67fbf65a1b3da)

Bubullzz:
/ok to test [c0cacce](https://github.com/NVIDIA/cuopt/pull/1153/commits/c0cacceae6d13af59dfb7d4c99c33b9f38dbbc4a)

Bubullzz:
/ok to test [29a8b46](https://github.com/NVIDIA/cuopt/pull/1153/commits/29a8b46e2e82f4ce3197f52c433af94f565bb84d)

Bubullzz:
/ok to test [7801c26](https://github.com/NVIDIA/cuopt/pull/1153/commits/7801c26202ed1fdfdbe273051a79fcafa5eeed75)

Kh4ster (approved):
All good to me! Thanks @Bubullzz that's super helpful and it was not easy to go around all the legacy parts of the code while keeping the whole thing working!

ramakrishnap-nv (approved):

PR #1158, opened 2026-04-29
ci(pr): skip build/test jobs for ci/docker-only changes

rgsl888prabhu:
## Summary

- Add `!ci/docker/**` to all four `changed-files` filter groups in `.github/workflows/pr.yaml` (`build_docs`, `test_cpp`, `test_python_conda`, `test_python_wheels`).
- Dockerfile-only PRs were tripping every conda/wheel build and test, even though the image itself is built and tested in `build_test_publish_images.yaml` (post-merge `build.yaml`), not in `pr.yaml`. Result: maximum CI cost, zero coverage of the change. This patch makes such PRs skip the irrelevant jobs.

## Why

Each filter group starts with `'**'` and excludes specific paths with `!`. Every other script under `ci/` (`ci/build_*.sh`, `ci/test_*.sh`, `ci/release/**`, `ci/thirdparty-testing/**`, etc.) is already excluded; `ci/docker/**` was an oversight. Adding it brings the deny list in line with the rest of `ci/` and matches the intent of the existing `!.github/workflows/build_test_publish_images.yaml` exclusion (which keeps changes to the image-publish workflow itself out of `pr.yaml`).

## Follow-up (not in this PR)

If we want Dockerfile changes to actually be smoke-tested on PRs, that needs a separate `if:`-gated job in `pr.yaml` that calls `build_test_publish_images.yaml` when `ci/docker/**` changes. Out of scope here — the goal of this PR is just to stop wasting CI cycles.

## Test plan

- [ ] Land this PR; on a follow-up Dockerfile-only PR, confirm only `checks` / `pr-builder` / `changed-files` run and the conda/wheel/docs jobs are skipped.
- [ ] On a code-touching PR (e.g. modifying `cpp/` or `python/`), confirm the full matrix still runs.


mlubin:
Hmm, couldn't changes to the dockerfiles cause other downstream tests to fail? Are these describing the docker containers that the tests run in?

rgsl888prabhu:
These docker images are built only in night-lies, so these will not get built in any PRs, so they will not be tested.

rgsl888prabhu:
And normally if the packages are getting built in conda and wheel env and running, this should work. But there are corner cases that might break.

tmckayus (approved):

PR #1165, opened 2026-04-30
Minimize Ubuntu Boost dependencies

bdice:
## Summary
- install only the Boost components needed by cuOpt's PaPILO dependency on Ubuntu
- add `libboost-iostreams-dev` and `libboost-serialization-dev` explicitly rather than `libboost-dev`
- skip `libboost-program-options-dev` -- this is optional and not needed for PaPILO with the options we provide (`PAPILO_NO_BINARIES=ON`)


jameslamb (approved):
Love this, thanks!

I just merged latest `main` in here to trigger a new CI run. Hopefully things will go better now that the `*.ubuntu.com` servers are somewhat stabilizing (https://status.canonical.com/).

PR #1166, opened 2026-04-30
Add dev skill agent evaluation framework with bug fixes and refactor

rgsl888prabhu:
## Summary

Branch from latest `main` containing the dev-skill agent evaluation framework from #953, plus bug fixes and a small refactor. Can supersede #953.

Three commits:

**1. Add dev skill agent evaluation framework** — same content as #953.

**2. Fix bugs:**
- `pass@K` aborted on the first attempt error and cleared all collected attempts; now records the error reason and continues to the next attempt.
- `--save` always wrote `responses[0]`; now writes the passing attempt so saved replays match the sample that demonstrated correctness.
- `default_assertions` was shadowed by per-test entries (so the issue-style suite's defaults were dead code); now merged.
- Negation regex bug: `no\s`/`not\s` + closing `\b` never matched because `\s` consumed the boundary. Use `\bno\b` / `\bnot\b` so "should not run sudo" exempts "run sudo".
- Tighten `_NEGATION_LOOKBACK` from 100 to 40 chars and reset on sentence-end punctuation (`.!?`); newlines intentionally excluded so multi-line code-comment markers (`// WRONG\nnew int[100]`) still exempt the next line.
- Narrowed over-broad forbidden phrases (bare `"yes"` tokens replaced with specific affirmative substrings) so legitimate skill-aligned responses don't trip on the word "yes".
- `validate_developer_skills.sh`: heading-aware section match (matches Markdown `#` heading lines instead of inline prose); missing SKILL.md is now a hard error, not a silent SKIP.
- `set -euo pipefail` on both shell scripts.

**3. Refactor + new options:**
- Extract `median_meta()` and `make_report_row()` so per-test rows are built in one place instead of inline at four call sites.
- Add `_validate_config_schema()` so typos in a tests JSON surface before any Claude CLI call.
- `--filter REGEX` — run only tests whose label matches (case-insensitive).
- `--list` — print test labels and exit; no Claude CLI required.
- `--junit PATH` — emit a JUnit XML report for CI ingestion.

## Test plan

- [ ] `python3 ci/utils/run_dev_skill_agent_tests.py --list` prints all labels
- [ ] `--list --filter cuda` narrows to the CUDA-related tests
- [ ] `bash ci/utils/validate_developer_skills.sh` passes against current SKILL.md files
- [ ] Live run on a subset (e.g. `--filter run_tests`) produces `report.md`, `results.csv`, and `junit.xml`
- [ ] Negation logic unit-tested for: same-clause negation, cross-sentence reset, `// WRONG`-marker code blocks, `should not` recognition

### DIFF
--- a/ci/run_dev_skill_agent_tests.sh
+++ b/ci/run_dev_skill_agent_tests.sh
@@ -1,0 +1,21 @@
+#!/bin/bash
+# SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+# Run developer-skill agent tests: send prompts through Claude CLI with the
+# cuOpt developer skill as context and validate responses (follow skill vs diverge).
+#
+# From repo root:
+#   ./ci/run_dev_skill_agent_tests.sh           # live (requires: claude auth login)
+#   ./ci/run_dev_skill_agent_tests.sh --replay ci/utils/dev_skill_responses  # validate saved responses
+#   ./ci/run_dev_skill_agent_tests.sh --pass-at 5   # pass@5: run each request 5x, pass if any passes
+#   ./ci/run_dev_skill_agent_tests.sh --runtimes-file out/runtimes.json  # write median runtimes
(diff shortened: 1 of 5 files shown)


rgsl888prabhu:
closing for #1167

PR #1168, opened 2026-04-30
Add cuOpt devcontainer configuration

bdice:
## Summary
- Add CUDA 12.9 and CUDA 13.3 devcontainer configurations for both conda and pip workflows.
- Add PR CI coverage using the RAPIDS devcontainer build workflow with the cuOpt benchmark builds enabled.
- Update the release version script so devcontainer image tags, feature versions, and container names are advanced with releases.


.github/workflows/pr.yaml
```diff
@@ -523,3 +528,18 @@ jobs:
     with:
       build_type: pull-request
       script: ci/test_self_hosted_service.sh
+  devcontainer:
+    secrets: inherit
+    uses: rapidsai/shared-workflows/.github/workflows/build-in-devcontainer.yaml@main
+    with:
+      arch: '["amd64", "arm64"]'
+      cuda: '["12.9", "13.1"]'
```
bdice: We only test one CUDA version for devcontainers in CI for other libraries. This is temporarily including 12.9 for testing, but we'll remove it before merging to keep CI runtime down.
```suggestion
      cuda: '["13.1"]'
```

trxcllnt (approved):

.github/workflows/pr.yaml
```diff
@@ -523,3 +528,18 @@ jobs:
     with:
       build_type: pull-request
       script: ci/test_self_hosted_service.sh
+  devcontainer:
```
ramakrishnap-nv, line 597: Is it possible to skip devcontainer build if there no changes required for the build or test ?

aliceb-nv (approved):
Approving engine-side

ramakrishnap-nv:
@bdice Should we follow-up on this ?

bdice:
@rgsl888prabhu Yes, I'll continue work on this when I'm able.

cpp/tests/mip/doc_example_test.cu
```diff
@@ -83,6 +83,7 @@ TEST(docs, user_problem_file)
   auto problem = create_doc_example_problem();
 
   const auto user_problem_path = std::filesystem::temp_directory_path() / "user_problem.mps";
+  std::filesystem::remove(user_problem_path);
```
bdice, line 86: This path may already exist when running the tests multiple times. This removes it if it exists, just to ensure we start from a clean state. Same below.

cpp/tests/CMakeLists.txt
bdice, line 1: These changes help align cuOpt's C++ tests with other RAPIDS libraries and best practices for CMake test declarations.

datasets/get_test_data.sh
```diff
@@ -221,15 +221,19 @@ set +e  # Disable exit on error for the entire script
 
 for index in ${!DESTDIRS[*]}; do
     tfname=$(basename "${URLS[$index]}")
+    archive="tmp/${tfname}"
 
-    if file --mime-type "tmp/${tfname}" | grep -q gzip$; then
```
bdice, line 225: We need to avoid depending on `file` which is not available on all systems (including devcontainers, which are a fairly minimal setup). We can use file extensions to infer the executable to use for unzipping/inflating. We had to do something similar in other RAPIDS libraries (cuGraph?) for broader compatibility.

cpp/CMakeLists.txt
```diff
@@ -266,11 +266,13 @@ FetchContent_Declare(
 )
 
 find_package(TBB REQUIRED)
-set(BUILD_TESTING OFF CACHE BOOL "Disable test build for papilo")
+set(CUOPT_BUILD_TESTING ${BUILD_TESTING})
```
bdice, line 287: It's safer here to capture the current value and reset it.

cpp/CMakeLists.txt
```diff
@@ -788,6 +791,7 @@ if (BUILD_MIP_BENCHMARKS AND NOT BUILD_LP_ONLY)
             cuopt
             OpenMP::OpenMP_CXX
             PRIVATE
+            argparse::argparse
```
bdice, line 976: These were just missing a dependency on `argparse`.

cpp/tests/CMakeLists.txt
```diff
@@ -66,18 +71,11 @@ function(ConfigureTest CMAKE_TEST_NAME)
       target_link_options(${CMAKE_TEST_NAME} PRIVATE -Wl,--enable-new-dtags)
     endif()
 
-    add_test(NAME ${CMAKE_TEST_NAME} COMMAND ${CMAKE_TEST_NAME})
+    rapids_test_add(NAME ${CMAKE_TEST_NAME} COMMAND ${CMAKE_TEST_NAME} GPUS 1 INSTALL_COMPONENT_SET testing)
```
bdice, line 115: Using the rapids-cmake infrastructure here lets us leverage resource specs from CTest, allowing tests to run in parallel if desired. See https://github.com/rapidsai/rapids-cmake/blob/87433a734d6c27b4fb41773c7bf6709dac893407/docs/hardware_resources_and_testing.rst.

ci/run_ctests.sh
bdice, line 1: This file is where I need the most feedback. We should move to `ctest` (and potentially add flags like `--rerun-failed` for handling flakiness?) because it's a more robust test runner. It will allow us to do things like schedule parallel tests on the GPU while respecting requested resources, and it's how we run tests for most RAPIDS libraries.

If there are features I've cut out here that are crucial to retain, let's discuss.

Edgarruiz8585:
***@***.***
by Edgarruiz8585

Obtener Outlook para Android<https://aka.ms/AAb9ysg>
________________________________

.github/workflows/pr.yaml
```diff
@@ -601,3 +606,24 @@ jobs:
...
+      actions: read
+      contents: read
+      id-token: write
+      packages: read
+      pull-requests: read
+    secrets: inherit # zizmor: ignore[secrets-inherit]
+    uses: rapidsai/shared-workflows/.github/workflows/build-in-devcontainer.yaml@main
+    with:
+      arch: '["amd64", "arm64"]'
+      cuda: '["12.9", "13.3"]'
```
bdice, line 629: We only test one CUDA version for devcontainers in CI for other libraries. This is temporarily including 12.9 for testing, but we'll remove it before merging to keep CI runtime down.
```suggestion
      cuda: '["13.3"]'
```

datasets/get_test_data.sh
```diff
@@ -221,15 +221,19 @@ set +e  # Disable exit on error for the entire script
...
+    if [ ! -f "${archive}" ]; then
+        echo "Missing download: ${archive}"
+        exit 1
     fi
 
+    case "${tfname}" in
+        *.tar.gz|*.tgz) tar xvzf "${archive}" -C "${DESTDIRS[$index]}" || true ;;
+        *.zip) unzip "${archive}" -d "${DESTDIRS[$index]}" || true ;;
+        *) tar xvf "${archive}" -C "${DESTDIRS[$index]}" || true ;;
+    esac
```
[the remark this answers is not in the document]
bdice: I think this is intentional behavior so that one failure doesn't break the script. Resolving.

.github/workflows/pr.yaml
bdice: Done in ab8bab020432001b4bf6c3a6c087f6287c19c6db.

ramakrishnap-nv (approved):

PR #1169, opened 2026-04-30
Add skill evaluation dataset for cuopt-installation-developer

rgsl888prabhu:
## Summary

Skill evaluation dataset for `cuopt-installation-developer` at `skills/cuopt-installation-developer/evals/evals.json`. 10 entries focused on the build-from-source path. Companion to #1167 (`cuopt-developer` evals).

## Coverage

| Theme | Count |
|---|---|
| First-time build / required questions | 3 |
| CUDA driver / env selection | 2 |
| Major-version mismatch diagnosis & clean build | 2 |
| User vs developer install boundary | 1 |
| Hand-off to cuopt-developer | 1 |
| Refusal (sudo conda) | 1 |

Same schema as #1167 (`id`, `question`, `expected_skill`, `expected_script`, `ground_truth`, `expected_behavior`). Schema check and pre-commit hooks pass locally.


mlubin:
Nice! Does this actually test that agents successfully build or is it just checking their textual outputs?

rgsl888prabhu:
> Nice! Does this actually test that agents successfully build or is it just checking their textual outputs?

This would check their textual output or steps that it would take, and there is a LLM judge which will check whether it took all appropriate steps or not.

mlubin:
> This would check their textual output or steps that it would take, and there is a LLM judge which will check whether it took all appropriate steps or not.

Seems like a potential loophole; we don't know if the steps it would take are correct if it doesn't actually build successfully.

tmckayus (approved):

PR #1172, opened 2026-05-01
Add skill evaluation dataset for cuopt-lp-milp-api-python

rgsl888prabhu:
## Summary

Initial skill evaluation dataset for `cuopt-lp-milp-api-python` at `skills/cuopt-lp-milp-api-python/evals/evals.json`. 99 entries adapted from the [microsoft/OptiGuide](https://github.com/microsoft/OptiGuide) IndustryOR corpus (MIT, attribution in `evals/SOURCES.md`).

- `ground_truth` is the numeric optimal value; rubric requires exact match to the precision shown (no tolerance)
- `expected_behavior` is generic across all entries — does not pre-categorize as LP vs MILP
- Each entry has a `source` field referencing the dataset row for traceability

QP eval set is out of scope (the corpus has no genuine QP problems) and will follow in a separate PR.


Iroy30 (approved):
LGTM. Just curious- Is it easier to wget each time or have it in the repo?

rgsl888prabhu:
> LGTM. Just curious- Is it easier to wget each time or have it in the repo?

If there are failures in pull github, it would be a headache for CI failures or even local. And it wants in json format which is not same as the github repo.

PR #1178, opened 2026-05-05
Add flow cover cuts at root

### LINKED ISSUE #1065: [FEA] Flow cover cuts
Add support for flow cover cuts in the MIP solver

hlinsen:
Closes: https://github.com/NVIDIA/cuopt/issues/1065
- Adds c-MIR flow-cover cut generation for 0-1 single-node-flow relaxations.
- Converts eligible MIP rows into single-node-flow form using binary variables and continuous flow variables.
- On the Wolter benchmark, gap closed is now 8.45% shifted geomean. Wolter reports 10.92% in Table 5.1 and 10.02% in B.49.


cpp/src/cuts/cuts.cpp
```diff
@@ -821,6 +823,89 @@ void cut_pool_t<i_t, f_t>::drop_cuts()
...
+                                  f_t y_star,
+                                  std::vector<snf_candidate_t<i_t, f_t>>& candidates)
+{
+  const f_t bound_tol       = context.settings.primal_tol;
+  const f_t feasibility_tol = context.settings.primal_tol;
+  if (spec.u < -bound_tol) { return; }
+  const f_t capacity = std::max<f_t>(0.0, spec.u);
+  if (capacity <= feasibility_tol) { return; }
+
+  snf_candidate_t<i_t, f_t> candidate;
```
chris-maes: Is snf single node flow? If so you might want to write out the acronyms to make it easier on the reader.

cpp/src/cuts/cuts.hpp
```diff
@@ -345,9 +484,24 @@ class knapsack_generation_t {
                             i_t knapsack_row,
                             inequality_t<i_t, f_t>& cut);
 
+  i_t generate_flow_cover_cut(const lp_problem_t<i_t, f_t>& lp,
```
chris-maes: Would it make sense to collect all the flow cover code into a class `flow_cover_generation_t`?

cpp/src/cuts/cuts.cpp
```diff
@@ -898,6 +983,626 @@ knapsack_generation_t<i_t, f_t>::knapsack_generation_t(
...
+    if (slack_count > 1 || (slack_count == 1 && std::abs(slack_coeff) <= settings.zero_tol)) {
+      continue;
+    }
+    if (slack_count == 0) {
+      flow_cover_constraints_.push_back({i, false});
+      flow_cover_constraints_.push_back({i, true});
+      continue;
+    }
+
+    const f_t sigma_slack_lower =
```
chris-maes, line 1036: Ugh. Sorry you have to deal with this as well. In 26.08 I want to fix it so that slack_coeff is always 1.

cpp/src/cuts/cuts.hpp
```diff
@@ -327,6 +330,142 @@ class cut_pool_t {
...
+  const lp_problem_t<i_t, f_t>& lp;
+  const simplex_solver_settings_t<i_t, f_t>& settings;
+  csr_matrix_t<i_t, f_t>& Arow;
+  const variable_bounds_t<i_t, f_t>& variable_bounds;
+  const std::vector<variable_type_t>& var_types;
+  const std::vector<f_t>& xstar;
+};
+
+template <typename i_t, typename f_t>
+struct flow_cover_scratch_t {
```
chris-maes: If I understand correctly `flow_cover_scratch_t`  is used to hold a bunch of intermediate variables that you need during the computation. Maybe you should just make `flow_cover_scratch_t` the `flow_cover_generator_t` and add methods to it?

cpp/src/cuts/cuts.cpp
```diff
@@ -898,6 +983,626 @@ knapsack_generation_t<i_t, f_t>::knapsack_generation_t(
...
+    slack_col   = j;
+    slack_coeff = row.coeff(k);
+  }
+  if (slack_count > 1) { return false; }
+
+  negate_row = flow_cover_row.reverse;
+  b          = negate_row ? -row.rhs : row.rhs;
+  if (slack_count == 1) {
+    if (std::abs(slack_coeff) <= coefficient_tol) { return false; }
+    const f_t sigma_slack_lower = slack_coeff > 0.0 ? slack_coeff * context.lp.lower[slack_col]
```
chris-maes, line 1075: Ugh. I'm so sorry about this. My poor design decision is leading to difficult code :(

cpp/src/cuts/cuts.cpp
```diff
@@ -898,6 +983,626 @@ knapsack_generation_t<i_t, f_t>::knapsack_generation_t(
...
+      }
+      return snf_activity <= snf_b + context.settings.primal_tol * snf_scale;
+    }(),
+    "Flow cover SNF relaxation excludes LP solution");
+
+  return true;
+}
+
+template <typename i_t, typename f_t>
+bool knapsack_generation_t<i_t, f_t>::select_flow_cover(
```
chris-maes: Is this where you are doing separation? If so you might want to call that out in the name.

cpp/src/cuts/cuts.cpp
```diff
@@ -898,6 +983,626 @@ knapsack_generation_t<i_t, f_t>::knapsack_generation_t(
...
+      scratch.best_in_l1 = scratch.in_l1;
+      scratch.best_in_l2 = scratch.in_l2;
+    }
+  }
+
+  return best;
+}
+
+template <typename i_t, typename f_t>
+flow_cover_evaluation_t<f_t> knapsack_generation_t<i_t, f_t>::evaluate_sgfci(
```
chris-maes: What does sgfci stand for? Maybe you could write out the abbrevation?

cpp/src/cuts/cuts.hpp
```diff
@@ -327,6 +330,142 @@ class cut_pool_t {
...
+        lhs_coefficients_touched[j] = 0;
+      }
+    }
+    touched_lhs_coefficients.clear();
+  }
+};
+
+template <typename f_t>
+struct flow_cover_selection_t {
+  f_t lambda;
```
chris-maes: It's a bit strange to have a struct with only a single variable? Can this be removed?

cpp/src/cuts/cuts.hpp
```diff
@@ -327,6 +330,142 @@ class cut_pool_t {
...
+  i_t y_col;
+  f_t y_coeff;
+  f_t y_x_coeff;
+  f_t b_shift;
+  f_t active_bound;
+  bool absorbs_binary_coeff;
+};
+
+template <typename i_t, typename f_t>
+struct flow_cover_context_t {
```
chris-maes, line 379: Would it make sense to unify `flow_cover_context_t` and `flow_cover_scratch_t`?

cpp/src/cuts/cuts.cpp
```diff
@@ -898,6 +983,626 @@ knapsack_generation_t<i_t, f_t>::knapsack_generation_t(
...
+    }
+    if (scratch.in_c2[k]) { sum_c2_u += scratch.arcs[k].u; }
+  }
+
+  selection.lambda = -snf_b + sum_c1_u - sum_c2_u;
+  return c1_nonempty && selection.lambda > feasibility_tol;
+}
+
+template <typename i_t, typename f_t>
+flow_cover_evaluation_t<f_t> knapsack_generation_t<i_t, f_t>::evaluate_cmirfci(
```
chris-maes: I know what cMIR stands for "complemented mixed integer rounding" what is fci? Flow cover inequality? You might want to write out at least some of this abbrevation to make it easier for the reader.

cpp/src/cuts/cuts.cpp
```diff
@@ -1437,6 +2143,24 @@ f_t knapsack_generation_t<i_t, f_t>::greedy_knapsack_problem(const std::vector<f
   // Sort by value / weight ratio
   std::sort(perm.begin(), perm.end(), [&](i_t i, i_t j) { return ratios[i] > ratios[j]; });
 
+  if (mode == greedy_knapsack_mode_t::STRICT_RATIO_PREFIX) {
+    // Wolter Algorithm 5.1 for KPSNF^rat: take the ratio-sorted prefix while
```
chris-maes: Is the only difference between this code and the greedy selection below that we are stopping at the first item that does not fit? 

If so, maybe you can just alter the code below as follows

```
for  (i_t j : perm) {
  if (weights[j] <= remaining) {
     // ...
   } else if (mode == greedy_knapsack_mode_t::STRICT_RATIO_PREFIX) {
      break;
   }
}
```

chris-maes: Oh I see, there is also a strict comparison? So you have to be strictly under the limit. 

It might still be possible to fold this into the code below. 

If not, fine to keep it separate.

cpp/src/cuts/cuts.cpp
```diff
@@ -2916,14 +3685,9 @@ variable_bounds_t<i_t, f_t>::variable_bounds_t(const lp_problem_t<i_t, f_t>& lp,
         if (a_ij > 0.0 && num_neg_inf_[i] <= 2) {
           const f_t lower_activity_j = lower_activity(lp.lower[j], lp.upper[j], a_ij);
 
-          // This is inefficient if num_neg_inf_[i] > 0
-          // If num_neg_inf_[i] == 1 and var_types[s] != INTEGER, we can't derive a bound
-          // If num_neg_inf_[i] == 2 and var_types[s ^ j] != INTEGER, we can't derive a bound
-          // If num_neg_inf_[i] == 2 and var_types[s ^ j] == INTEGER, and lower_activity_j != -inf,
-          // we can't derive a bound
```
chris-maes, line 2923: Can you restore the comment? If the comment is still valid? 

Or if not replace it with a new one?

cpp/src/cuts/cuts.cpp
```diff
@@ -2807,6 +3562,19 @@ variable_bounds_t<i_t, f_t>::variable_bounds_t(const lp_problem_t<i_t, f_t>& lp,
   }
   f_t start_time = tic();
 
+  struct variable_bound_edge_t {
```
chris-maes: Why do we need to change this for flow cover cuts?

cpp/src/cuts/cuts.cpp
```diff
@@ -2934,9 +3698,10 @@ variable_bounds_t<i_t, f_t>::variable_bounds_t(const lp_problem_t<i_t, f_t>& lp,
               // We have a valid variable upper bound
               // x_j <= -a_il/a_ij * x_l + beta/a_ij - 1/a_ij * sum_{k != l, k != j} a_ik *
               // bound(x_k)
-              upper_variables.push_back(l);
-              upper_weights.push_back(-a_il / a_ij);
-              upper_biases.push_back(beta / a_ij - (1.0 / a_ij) * sum);
+              const variable_bound_edge_t edge{l, -a_il / a_ij, beta / a_ij - (1.0 / a_ij) * sum};
```
chris-maes: Are these edges being used anywhere? If not, maybe we could revert back to the previous code?

cpp/src/cuts/cuts.cpp
chris-maes, line 3587: I'm not sure I understand the changes being made here. Can you explain the goal of these changes?

cpp/src/cuts/cuts.hpp
```diff
@@ -327,6 +330,142 @@ class cut_pool_t {
...
+  f_t x_value;  // Current LP value of x_col, or the fixed-control value.
+  f_t y_const;
+  i_t y_col;
+  f_t y_coeff;
+  f_t y_x_coeff;
+  f_t y_value;  // Current LP value of the synthetic y_j.
+};
+
+template <typename i_t, typename f_t>
+struct snf_candidate_t {
```
chris-maes: snf -> single_node_flow?

cpp/src/cuts/cuts.hpp
```diff
@@ -327,6 +330,142 @@ class cut_pool_t {
...
+class variable_bounds_t;
+
+template <typename i_t>
+struct flow_cover_row_t {
+  i_t row;
+  bool reverse;
+};
+
+template <typename i_t, typename f_t>
+struct snf_arc_t {
```
chris-maes: snf -> single_node_flow ?

cpp/src/cuts/cuts.cpp
hlinsen: I am generalizing that path as it is also reused by flow cover. We need to consider binary variables in this code path as well. The controller variable is a binary that let's us know if an arc is activated.
In the flow cover generation step we will filter variables that are not binary. For MIR the path is still correct.

cpp/src/cuts/cuts.cpp
hlinsen: I think it's best to keep seperate because we are also skipping https://github.com/hlinsen/cuopt/blob/flow-cover-cuts/cpp/src/cuts/cuts.cpp#L2216-L2231

cpp/src/cuts/cuts.cpp
```diff
@@ -3639,7 +4438,7 @@ void complemented_mixed_integer_rounding_cut_t<i_t, f_t>::transform_inequality(
   scratch_pad_.get_pad(inequality.vector.i, inequality.vector.x);
   // At this point we have converted all the continuous variables to be nonnegative
   // Note that since continuous variables had VUB or VLB, they modified
-  // the integer variables.
+  // the controller variables.
```
chris-maes: Nit: change back to integer?

chris-maes (approved):
LGTM. Thanks for implementing flow cover cuts @hlinsen ! 

I'm impressed :)

PR #1179, opened 2026-05-05
Run CPUFJ bursts at the root

aliceb-nv:
Some primal heuristics, like CPUFJ, may benefit in some cases from operating on the problem with root cuts. 
This PR adds support for this, by running bursts of a few iterations of CPUFJ per cut pass opportunistically, on threads that would otherwise remain idle; and also once after the root cut passes are completed.

In most cases, this provides no benefits, but some instances like `tutaki` find their first integer incumbent much faster. Additionally, `awhea` is now reliably feasibilized and solved to a gap of <5% on 10min runs.

## Checklist

- Testing
   - [ ] Added tests
   - [ ] NA
- Documentation
   - [ ] Added new documentation
   - [ ] NA


cpp/src/branch_and_bound/branch_and_bound.cpp
```diff
@@ -2504,6 +2567,24 @@ mip_status_t branch_and_bound_t<i_t, f_t>::solve(mip_solution_t<i_t, f_t>& solut
...
+    root_cut_cpufj_task =
+      detail::make_fj_cpu_task_from_host_lp<i_t, f_t>(original_lp_,
+                                                      var_types_,
+                                                      root_relax_soln_.x,
+                                                      settings_,
+                                                      root_cut_cpufj_improvement_callback,
+                                                      "[RootCut CPUFJ] ");
+    settings_.log.debug("Root cut CPUFJ final problem build time: %.6f seconds\n",
+                        toc(root_cut_cpufj_build_start_time));
+    f_t fj_time_limit = settings_.deterministic
```
akifcorduk: Why does undetermistic path get 1s?

cpp/src/mip_heuristics/feasibility_jump/fj_cpu.cu
```diff
@@ -41,6 +48,53 @@
 
 namespace cuopt::linear_programming::detail {
 
+template <typename f_t>
+static f_t clamp_value(f_t value, f_t lower, f_t upper)
```
akifcorduk: I think there is std::clamp available

cpp/src/mip_heuristics/feasibility_jump/fj_cpu.cu
```diff
@@ -41,6 +48,53 @@
...
+{
+  return std::min(std::max(value, lower), upper);
+}
+
+template <typename i_t, typename f_t>
+static void rebuild_reverse_matrix(i_t n_variables,
```
akifcorduk: We have a transpose function available in `sparse_matrix.cpp` even though the parameters are a bit different, we should reuse it.

cpp/src/branch_and_bound/branch_and_bound.cpp
```diff
@@ -2224,13 +2226,37 @@ mip_status_t branch_and_bound_t<i_t, f_t>::solve(mip_solution_t<i_t, f_t>& solut
...
+    };
+  auto stop_root_cut_cpufj = [&]() {
+    if (!root_cut_cpufj_task) { return; }
+    detail::stop_fj_cpu_task(*root_cut_cpufj_task);
+    root_cut_cpufj_task.reset();
+  };
+  cuopt::scope_guard root_cut_cpufj_guard([&]() { stop_root_cut_cpufj(); });
+
+  enum class cut_pass_action_t { CONTINUE, BREAK, RETURN };
+  struct cut_pass_result_t {
```
akifcorduk: I am not sure if I understand the need/logic for this. Can't we just launch a new cpufj task after each cut pass and wait until the next pass finishes and launch another cpufj task while staying in cut pass loop ?

cpp/src/mip_heuristics/feasibility_jump/fj_cpu.cu
```diff
@@ -1417,7 +1634,9 @@ std::unique_ptr<fj_cpu_climber_t<i_t, f_t>> fj_t<i_t, f_t>::create_cpu_climber(
 }
 
 template <typename i_t, typename f_t>
-void cpufj_solve(fj_cpu_climber_t<i_t, f_t>* fj_cpu, f_t in_time_limit)
+static bool cpufj_solve(fj_cpu_climber_t<i_t, f_t>& fj_cpu,
+                        f_t in_time_limit,
+                        double work_unit_limit = std::numeric_limits<double>::infinity())
 {
```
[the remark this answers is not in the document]
akifcorduk: Is it resolved now?

akifcorduk:
/ok to test [9a64c14](https://github.com/NVIDIA/cuopt/pull/1179/commits/9a64c145c27f7dc9d77863f2c66e1de2d4b103ab)

akifcorduk (approved):

cpp/src/branch_and_bound/branch_and_bound.cpp
```diff
@@ -2492,9 +2519,51 @@ mip_status_t branch_and_bound_t<i_t, f_t>::solve(mip_solution_t<i_t, f_t>& solut
...
       }
       last_objective = root_objective_;
+      return {cut_pass_action_t::CONTINUE, mip_status_t::UNSET};
+    };
+
+    cut_pass_result_t cut_pass_result;
+#pragma omp parallel num_threads(root_cut_cpufj_task ? 2 : 1)
+    {
+#pragma omp single
+      {
```
nguidotti: After #1099, the entire solver is enclosed into a `omp` parallel region and the B&B is itself a omp task. So, both `omp parallel` and `omp single` directives are not required here.

cpp/src/branch_and_bound/branch_and_bound.cpp
```diff
@@ -2228,13 +2230,41 @@ mip_status_t branch_and_bound_t<i_t, f_t>::solve(mip_solution_t<i_t, f_t>& solut
...
   for (i_t cut_pass = 0; cut_pass < settings_.max_cut_passes; cut_pass++) {
     if (num_fractional == 0) {
       set_solution_at_root(solution, cut_info);
+      signal_extend_cliques_.store(true, std::memory_order_release);
+#pragma omp taskwait depend(in : *clique_signal)
       return mip_status_t::OPTIMAL;
-    } else {
+    }
+
+    auto do_cut_pass = [&]() -> cut_pass_result_t {
```
nguidotti: Ahh I understand why Alice designed this way. You cannot call return from within a `omp parallel` region. If you eliminate the directives like I recommend, then the lambda is no longer necessary. You could just launch the omp task with cpu fj at the beginning, do the cut pass and then do a taskwait with dependency on the root fj object.

nguidotti: You could also eliminate the enum `cut_pass_action_t` in this way.

cpp/src/mip_heuristics/feasibility_jump/fj_cpu.cu
```diff
@@ -1593,24 +1765,107 @@ std::unique_ptr<fj_cpu_climber_t<i_t, f_t>> init_fj_cpu_standalone(
   return fj_cpu;
 }
 
+template <typename i_t, typename f_t>
+void fj_cpu_task_t<i_t, f_t>::fj_cpu_deleter_t::operator()(fj_cpu_climber_t<i_t, f_t>* ptr) const
```
nguidotti, line 1778: Is this not the default deleter?

cpp/src/mip_heuristics/feasibility_jump/fj_cpu.cuh
```diff
@@ -191,7 +192,8 @@ struct fj_cpu_climber_t {
 
 template <typename i_t, typename f_t>
 void cpufj_solve(fj_cpu_climber_t<i_t, f_t>* fj_cpu,
-                 f_t in_time_limit = std::numeric_limits<f_t>::infinity());
+                 f_t in_time_limit      = std::numeric_limits<f_t>::infinity(),
+                 double work_unit_limit = std::numeric_limits<double>::infinity());
```
nguidotti, line 196: Should this be `f_t` instead of `double`?

cpp/src/mip_heuristics/feasibility_jump/cpu_fj_thread.cuh
```diff
@@ -0,0 +1,51 @@
...
+#include <memory>
+#include <string>
+#include <vector>
+
+namespace cuopt::linear_programming::detail {
+
+template <typename i_t, typename f_t>
+struct fj_cpu_climber_t;
+
+template <typename i_t, typename f_t>
```
nguidotti, line 25: Do we really need a `fj_cpu_task`? The methods are operating directly on the `fj_climber` and the `preemption_flag` seems to be not used.

cpp/src/mip_heuristics/feasibility_jump/cpu_fj_thread.cuh
```diff
@@ -0,0 +1,51 @@
...
+  std::function<void(f_t, const std::vector<f_t>&, double)> improvement_callback,
+  std::string log_prefix);
+
+template <typename i_t, typename f_t>
+void run_fj_cpu_task(fj_cpu_task_t<i_t, f_t>& task,
+                     f_t time_limit         = std::numeric_limits<f_t>::infinity(),
+                     double work_unit_limit = std::numeric_limits<double>::infinity());
+
+template <typename i_t, typename f_t>
+void stop_fj_cpu_task(fj_cpu_task_t<i_t, f_t>& task);
```
nguidotti, line 54: Move the corresponding `taskwait` to within the `stop_fj_cpu_task` otherwise the program will continue after calling this routine normally (i.e., the task may not necessarily stop).

nguidotti (changes requested):
Thanks @aliceb-nv and @akifcorduk for the hard work! A couple of things:
- After #1099, the `omp parallel`/`single` directives are not longer needed. You can eliminate them completely. 
- Is wrapping the cut pass as a lambda necessary? You could just launch the CPU FJ, do the cut passes, and then stop the CPU FJ. Launching a `omp task` is an asynchronous operation.
- I think `cpu_fj_task` can be eliminated since all its methods are operating over the `cpu_fj_climber` anyway.

cpp/src/branch_and_bound/branch_and_bound.cpp
akifcorduk: The reason is that cut pass code has early return points. When we return, the cpufj task remains not synchonized.

cpp/src/mip_heuristics/feasibility_jump/cpu_fj_thread.cuh
akifcorduk: `preemption_flag` is used here: https://github.com/NVIDIA/cuopt/pull/1179/changes#diff-49cef178e4deff2c307fa54f331ece79e5a5e8658e1bf9b2504ce660bae0426cR1603

The fj_cpu_task exists as a wrapper for lifetime management of climber and the atomic flags. I would not play with the design until Alice is back.

cpp/src/mip_heuristics/feasibility_jump/fj_cpu.cu
akifcorduk: It seems to be coming from PIMPL idiom which makes this work on a forward-declared type. It wouldn't compile on default deleter as the full implementation of fj_cpu_climber_t is not visible. The default deleter requires the complete type to compile.

cpp/src/mip_heuristics/feasibility_jump/cpu_fj_thread.cuh
akifcorduk: I have moved the wait inside the if in the caller(branch and bound). I could move both task start and wait on respective `run_fj_cpu_task` and `stop_fj_cpu_task`, but at the end of the cut passes, we do a run on the branch and bound thread without launching a task.

cpp/src/branch_and_bound/branch_and_bound.cpp
nguidotti: A call to `stop_fj` + `taskwait` on the `scope_guard` should handle early returns, no?

akifcorduk:
/ok to test [9356dc8](https://github.com/NVIDIA/cuopt/pull/1179/commits/9356dc8ab95c45e05f06204c4b25255a3f938eea)

cpp/src/mip_heuristics/feasibility_jump/cpu_fj_thread.cuh
nguidotti: It is mostly to keep the start/stop self-contained (we may accidentally forgot to add the `taskwait` in the future)

akifcorduk: I agree but this is not always used with a new task. Sometimes, it just executes on the current thread.

cpp/src/mip_heuristics/feasibility_jump/fj_cpu.cuh
akifcorduk: It seems work unit is double everywhere in the code. This just keeps it consistent.

cpp/src/mip_heuristics/feasibility_jump/cpu_fj_thread.cuh
nguidotti: Fair enough

akifcorduk:
/ok to test [2763151](https://github.com/NVIDIA/cuopt/pull/1179/commits/2763151f35e3403ca323f2c17bc86c7556c7cb1e)

akifcorduk:
/ok to test [26f2cbd](https://github.com/NVIDIA/cuopt/pull/1179/commits/26f2cbd49cee5bdae8dc71c62ef2c7c1b0eaca69)

akifcorduk:
/ok to test [30596fa](https://github.com/NVIDIA/cuopt/pull/1179/commits/30596fa61d02f01f19020c0fc7aa1fb9a466b396)

akifcorduk:
/ok to test [6f1f1d9](https://github.com/NVIDIA/cuopt/pull/1179/commits/6f1f1d9cf3ea9a91d47434b1c8724abdeda98d02)

cpp/src/branch_and_bound/branch_and_bound.cpp
```diff
@@ -2228,273 +2507,92 @@ mip_status_t branch_and_bound_t<i_t, f_t>::solve(mip_solution_t<i_t, f_t>& solut
...
-                                                           variable_bounds,
-                                                           exploration_stats_.start_time);
-      if (!problem_feasible) {
-        if (settings_.heuristic_preemption_callback != nullptr) {
-          settings_.heuristic_preemption_callback();
-        }
-        signal_extend_cliques_.store(true, std::memory_order_release);
+    cut_pass_result_t cut_pass_result;
+    if (root_cut_cpufj_task) {
+#pragma omp task shared(root_cut_cpufj_task) default(none)
```
nguidotti: I would add `depend(out: *root_cut_cpufj_task)` in the `omp task` and `depend(in: *root_cut_cpufj_task)` in the `pragma omp taskwait` below. So it only waits for the CPU FJ task not all tasks created previously (e.g., clique cuts)

akifcorduk:
/ok to test [ef39841](https://github.com/NVIDIA/cuopt/pull/1179/commits/ef39841bd0f1ff8fa6ae07296a4ce413778d0143)

akifcorduk:
/ok to test [c45f3ff](https://github.com/NVIDIA/cuopt/pull/1179/commits/c45f3ff6b46e21142c129b4d16d5a90b174ec592)

nguidotti (approved):
Looks good to me. Thanks @akifcorduk!

nguidotti:
I just merged #1201. So, you can revert the changes to the determinism test

Edgarruiz8585:
- > by:Edgar Manuel Ruiz Arias @Edgarruuz8585 ***@***.***
***@***.*** ***@***.***

Edgarruiz8585:
🚀

Édgar Manuel reaccionó a través de Gmail
<https://www.google.com/gmail/about/?utm_source=gmail-in-product&utm_medium=et&utm_campaign=emojireactionemail#app>

PR #1188, opened 2026-05-07
build gRPC against OpenSSL 3

tmckayus:
Switch the wheel build to link gRPC against OpenSSL 3.  This is good for modernization, and also necessary to enable FIPS compliance on many current OS platforms.

* install_protobuf_grpc.sh: on Rocky/RHEL 8, install epel-release and openssl3-devel. Rocky/RHEL 9 and Ubuntu/Debian already have openssl-devel at version 3 so use that.

* build_wheel_libcuopt.sh: Add libssl.so.3 / libcrypto.so.3 to the auditwheel exclude list so the wheel does not bundle them; runtime resolves to the host's OpenSSL 3, ensuring libcrypto and any FIPS provider (system or mounted) stay byte-version-matched.

Verified that the resulting binary links require only OPENSSL_3.0.0 symbol versions (159 refs, 0 newer), so the wheel is ABI-compatible with the cuopt container's Ubuntu 22.04 (libssl 3.0.2), RHEL/Rocky 9 (3.0.7), and any OpenSSL 3.x runtime.

Hosts must provide libssl.so.3 / libcrypto.so.3 (Ubuntu 22.04+, Debian 12+, RHEL/Rocky/Alma 9+, Fedora 36+ provides). Stock RHEL/Rocky 8 and Ubuntu 20.04 are not covered by this; users on those systems should either use the cuopt container image or install OpenSSL 3 from EPEL or a backport before pip-installing cuopt.

### DIFF
--- a/ci/build_wheel_libcuopt.sh
+++ b/ci/build_wheel_libcuopt.sh
@@ -29,6 +29,22 @@ bash ci/utils/install_protobuf_grpc.sh
 
 export SKBUILD_CMAKE_ARGS="-DCUOPT_BUILD_WHEELS=ON;-DDISABLE_DEPRECATION_WARNING=ON"
 
+# OpenSSL 3 hints for libcuopt's own find_package(OpenSSL).
+#
+# install_protobuf_grpc.sh links gRPC against OpenSSL 3 (see that script for
+# rationale). libcuopt then re-resolves OpenSSL via find_package because
+# gRPC's imported targets propagate it transitively. On Rocky/RHEL 8 the
+# EPEL openssl3-devel package installs in non-default paths, so we have to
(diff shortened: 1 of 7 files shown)


ramakrishnap-nv (commented):
Changes looks good, but we may have to add this detail to docs or faqs to install openssl3 so it is always covered, and also did you check if the container needs this update ?

And does conda needs this update as well?

tmckayus:
> Changes looks good, but we may have to add this detail to docs or faqs to install openssl3 so it is always covered, and also did you check if the container needs this update ?
> 
> And does conda needs this update as well?

I can work on docs.  conda already has openssl3, I believe the situation was that the container had it but libcuopt was linked against openssl 1.1.1 so it didn't use openssl 3 even though it was there. I will confirm

ramakrishnap-nv (approved):

Issue #1037, opened 2026-04-04
Cost per token is not real cost reduction — hidden safety, governance and liability costs are missing


HIJO790401:
Your recent tokenomics narrative is optimizing visible inference cost, not real total cost.

You keep showing:
- cost per token
- throughput per dollar
- throughput per watt
- lower cost per query

But this is only the visible layer.

The missing ledger is the real problem:
- safety cost
- hallucination cost
- governance cost
- monitoring cost
- patching cost
- legal liability cost
- remediation cost
- boundary failure cost

If those costs are excluded, this is not true cost reduction.
It is deferred failure disguised as savings.

In plain terms:
you are optimizing how cheaply tokens are produced,
but you are not proving that the full consequence loop is cheaper, safer, or more accountable.

That is exactly where the current inference narrative breaks.

A lower token price does not answer:
- when should the system stop?
- when should it escalate to a larger / safer / more closed-loop path?
- who is accountable when a cheaper path produces harmful or misleading outputs?
- how much extra cost will be spent after safety, regulatory, or liability failure surfaces?

Without explicit boundary conditions, rollback conditions, and accountable owners,
“cheaper inference” is just a partial metric.

My view is simple:
if hidden cost and deferred cost are not included,
then “cost reduction” is mathematically incomplete and operationally misleading.

A more honest equation is:

Real Savings
= Claimed Savings
- Hidden Cost
- Deferred Cost

Where:
Hidden Cost = safety + governance + monitoring + alignment + patching
Deferred Cost = failure + leakage + legal liability + recovery

If you optimize token birth rate while ignoring consequence cost,
you are not reducing cost.
You are amplifying illusion until failure.

I already have a concrete direction for solving this:
responsibility-aware, boundary-aware, consequence-aware output and audit structure.

Portal:
https://hijo790401.github.io/shen-yao-portal/

If NVIDIA wants to keep talking about optimization,
then optimization should include boundary, accountability, and post-failure cost —
not just token throughput.

Otherwise, the industry is not saving money.
It is buying time before explosion.
你們最近這套 tokenomics 敘事，優化的是可見的推理成本，不是真正的總成本。

你們一直展示的是：
- 每 token 成本
- 每美元吞吐量
- 每瓦吞吐量
- 每次查詢成本下降

但這些都只是表面層。

真正缺失的帳本才是問題：
- 安全成本
- 幻覺成本
- 治理成本
- 監控成本
- 補洞成本
- 法律責任成本
[...]

PR #1192, opened 2026-05-08
java: cuopt-java initial release — LP/MILP/QP API + classifier JARs

ramakrishnap-nv:
## Summary

Initial Java interface for cuOpt covering LP, MILP, and QP (beta). Models cuvs-java's architecture (Project Panama / FFM, Java 21 base API + Java 22 multi-release JAR for FFM impl). API surface follows Gurobi-style conventions for problem construction and Python cuOpt conventions for solution access.

## For reviewers — what to skip

**84% of this diff (~23,812 of ~28,262 lines) is auto-generated.** The directory `java/cuopt-java/src/main/java22/com/nvidia/cuopt/internal/panama/` is jextract output from `cpp/include/cuopt/linear_programming/cuopt_c.h`. The single file `cuopt_c_h.java` is 23,515 lines. Skip the entire `panama/` subdirectory — the drift gate in `java/build.sh` keeps it in sync with the C header on every build.

The remaining ~4,350 hand-written lines, in suggested review order:

| Area | Files | Lines | Biggest file |
|---|---|---|---|
| Public API (Java 21, `src/main/java/`) | 23 | 2,105 | `Problem.java` (421) |
| FFM impl (Java 22, `src/main/java22/.../internal/`, excl. panama) | 4 | 902 | `CuOptProviderImpl.java` (531) |
| Maven + build scripts (`pom.xml`, `build.sh`, `generate-bindings.sh`, assembly) | 5 | 625 | `pom.xml` (317) |
| Tests (`src/test/java/`) | 5 | 315 | — |
| CI / conda / dependencies (`pr.yaml`, `build.yaml`, `dependencies.yaml`, env files) | 9 | 248 | `pr.yaml` (~69) |
| READMEs / resources / `.gitignore` | 8 | 156 | — |

### Public API (Java 21, `src/main/java/com/nvidia/cuopt/`)
- `Problem`, `Variable`, `Constraint`, `LinearExpr`, `QuadraticExpr`, `SolverSettings`, `DataModel`
- Enums: `VType`, `CType`, `Sense`, `SolverMethod`, `PdlpSolverMode`, `TerminationStatus`, `ErrorStatus`, `ProblemCategory`
- Records: `LpStats`, `MilpStats`
- `@FunctionalInterface` callbacks: `MIPGetSolutionCallback`, `MIPSetSolutionCallback`
- `CuOpt` constants holder, `Solver` entry points, `CuOptException`
- `spi/CuOptProvider` (sealed) — Layer 4 bridge between Java 21 public API and Java 22 FFM impl

### FFM implementation (Java 22, `src/main/java22/`)
- `CuOptProviderImpl` — full impl: settings, parameter setters, `solveProblem`, `solveDataModel`
- `CsrBuilder` — translates `List<LinearExpr>` to CSR + Q-matrix CSR
- `NativeLibraryLoader` — chooses between embedded (classifier JAR) and BYO (`System.loadLibrary`) modes via JAR manifest
- `internal/panama/` — jextract bindings (committed, drift-gate-stable)

### Build / CI
- `java/build.sh` orchestrates regen bindings → drift gate → `mvn verify`
- `panama-bindings/generate-bindings.sh` auto-downloads jextract per arch on first run; idempotently normalizes output
- `ci/build_java.sh` + `ci/test_java.sh` with `--run-java-tests` / `--unit-tests-only` modes
- `pr.yaml`: `conda-java-build-and-tests` (amd64 GPU IT) + `conda-java-build-and-tests-arm64` (arm64 GPU IT)
- `build.yaml`: `java-build` + `java-build-arm64` for main/release/tag pushes

### Distribution
- **Base JAR** (`cuopt-java-<version>.jar`) — always built. BYO `libcuopt.so` on `java.library.path` (typical conda).
- **Classifier JAR** (`cuopt-java-<version>-<arch>-cuda<n>.jar`) — opt-in via `mvn -Dcuda.version=<n>` or `CLASSIFIER_CUDA=<n> ./java/build.sh`. Bundles `libcuopt.so`, `libmps_parser.so`, `librmm.so`, `librapids_logger.so` under `<arch>/Linux/`, with manifest entry `Embedded-Libraries-Cuda-Version=<n>`. User still installs CUDA toolkit matching `<n>`.

### MIP user callbacks
- `SolverSettings.setMIPGetSolutionCallback(cb)` — user-provided callback invoked when the solver finds a new incumbent. Reads the user's lambda at every native call so the latest registration wins.
- `SolverSettings.setMIPSetSolutionCallback(cb)` — user injects a candidate solution. **Registering this disables presolve** (per the C contract).
- FFM upcall stubs are bound to the per-solve `Arena` and re-created on every solve, so cross-solve reuse of the same `SolverSettings` is safe. Trampolines catch `Throwable` so user-code exceptions never propagate across the native boundary.

### Tests
- Unit: `LinearExprTest` (4)
- Integration: `SolverIT`, `LpSolverIT`, `MilpSolverIT`, `QpSolverIT`, `MipCallbackIT` (8 total)
- All green locally (~4s)

## Design inputs

- **Corey Nolet (cuvs-java)** on Java version: Java 21 LTS base + Java 22 for FFM (matches Lucene minimums).
- **Miles Lubin (JuMP author)** on API style: follow Gurobi Java conventions.
- **cuvs lessons** baked into the pom: phase-bound dual compile (cuvs #1293), spotless under `<build><plugins>` only (cuvs #1090), MR-JAR test path (cuvs #1037).

## Out of scope — tracked as follow-up issues

| # | Item | Reason |
|---|---|---|
| #1202 | C API stats getters (LP residuals, iterations, gap, method; MIP nodes/simplex iters/presolve time/violations) | Python sees these via Cython→C++ direct; the C ABI doesn't expose them. Currently `NaN` / `-1` / `UNSET` in `LpStats`/`MilpStats`. Cross-team C++ work. |
| #1203 | Static-link gRPC, protobuf, abseil, TBB, libgomp into `libcuopt.so` | Removes runtime system-package friction for non-conda users. CMake-side change. |

## Test plan

- [x] `./java/build.sh` passes locally (drift gate stable, 4 unit + 6 IT)
- [x] `CLASSIFIER_CUDA=13 ./java/build.sh` produces a classifier JAR; embedded mode verified by running with empty `LD_LIBRARY_PATH`
- [ ] CI green on amd64 (`conda-java-build-and-tests`)
- [ ] CI green on arm64 GPU (`conda-java-build-and-tests-arm64`)
- [ ] CodeRabbit / coderabbit-review feedback addressed


java/cuopt-java/src/main/java/com/nvidia/cuopt/linear_programming/Constraint.java
```diff
@@ -0,0 +1,163 @@
...
+    public double rhs() {
+        if (ranged) {
+            throw new IllegalStateException(
+                "rhs() is not defined for ranged constraints; use lowerBound()/upperBound()");
+        }
+        return rhs;
+    }
+
+    /** Whether this is a ranged constraint (lower &lt;= lhs &lt;= upper). */
+    public boolean isRanged() {
```
chris-maes: Let's remove isRanged(). Constraints should not be ranged.

java/cuopt-java/src/main/java/com/nvidia/cuopt/linear_programming/Constraint.java
```diff
@@ -0,0 +1,163 @@
...
+        return ranged;
+    }
+
+    /**
+     * Lower bound for ranged constraints. For one-sided constraints,
+     * returns the appropriate one-sided bound (e.g. rhs for GE; -infinity for LE).
+     */
+    public double lowerBound() {
```
chris-maes: Remove lowerBound and upperBound. Constraints should just have a rhs

java/cuopt-java/src/main/java/com/nvidia/cuopt/linear_programming/DataModel.java
```diff
@@ -0,0 +1,292 @@
...
+ *         .setRowTypes(new CType[]{CType.LE, CType.GE});
+ *
+ *       Solver.solve(dm);
+ *
+ *       System.out.println(dm.status());
+ *       System.out.println(java.util.Arrays.toString(dm.primalSolution()));
+ *   }
+ * }</pre>
+ */
+public final class DataModel implements AutoCloseable {
```
chris-maes: I don't think we need DataModel

chris-maes: The user should create problems directly with the Problem

java/cuopt-java/src/main/java/com/nvidia/cuopt/optimization/LinearExpr.java
```diff
@@ -0,0 +1,129 @@
...
+ *
+ * <p>All terms in one expression must reference Variables owned by the
+ * same {@link Problem}. Mixing across problems throws
+ * {@link IllegalArgumentException}.
+ */
+public final class LinearExpr {
+
+    // LinkedHashMap preserves insertion order (useful for deterministic
+    // CSR builds), keys are Variables, values are summed coefficients.
+    private final Map<Variable, Double> terms = new LinkedHashMap<>();
```
chris-maes, line 42: Let's discuss implementation. I don't think we want to use HashMaps here.

java/cuopt-java/src/main/java/com/nvidia/cuopt/linear_programming/MilpStats.java
```diff
@@ -0,0 +1,25 @@
...
+ * {@code Problem.milpStats()} (empty {@link java.util.Optional} for LP-only problems).
+ *
+ * @param mipGap             relative MIP gap at termination ({@code (UB - LB) / |UB|})
+ * @param bestBound          best dual bound found
+ * @param nodesExplored      number of branch-and-bound nodes explored
+ * @param incumbentsFound    number of distinct incumbent solutions found
+ * @param presolveTime       wall-clock time spent in presolve, seconds
+ * @param rootRelaxationTime wall-clock time spent on the root LP relaxation, seconds
+ */
+public record MilpStats(
```
chris-maes: We should use MIP instead of MILP. Capitalization should be consistent so MIPStats

java/cuopt-java/src/main/java/com/nvidia/cuopt/optimization/Problem.java
```diff
@@ -0,0 +1,421 @@
...
+ * }</pre>
+ */
+public final class Problem implements AutoCloseable {
+
+    // ── build-time state ─────────────────────────────────────────
+    private final String name;
+    private final List<Variable> variables = new ArrayList<>();
+    private final List<Constraint> constraints = new ArrayList<>();
+    private final List<LinearExpr> constraintExpressions = new ArrayList<>();
+    private final Map<String, Variable> variablesByName = new HashMap<>();
```
chris-maes, line 57: Let's discuss and possible remove Map here.

java/cuopt-java/src/main/java/com/nvidia/cuopt/linear_programming/Problem.java
```diff
@@ -0,0 +1,421 @@
...
+        return constraints.size();
+    }
+
+    public int numNonZeros() {
+        int n = 0;
+        for (LinearExpr e : constraintExpressions) n += e.numTerms();
+        return n;
+    }
+
+    public boolean isMip() {
```
chris-maes: Consistent capitalization of abbrevation isMIP

java/cuopt-java/src/main/java/com/nvidia/cuopt/linear_programming/Problem.java
```diff
@@ -0,0 +1,421 @@
...
+        for (Variable v : variables) {
+            if (v.variableType() != VType.CONTINUOUS) return true;
+        }
+        return false;
+    }
+
+    public boolean isQp() {
```
chris-maes: Consistent capitalization isQP

java/cuopt-java/src/main/java/com/nvidia/cuopt/linear_programming/ProblemCategory.java
```diff
@@ -0,0 +1,20 @@
...
+
+/**
+ * Coarse classification of an optimization problem. Returned by
+ * {@code Problem.problemCategory()}.
+ */
+public enum ProblemCategory {
+    /** Linear program: linear objective, linear constraints, all continuous variables. */
+    LP,
+    /** Integer program: linear objective and constraints, all variables integer. */
+    IP,
```
chris-maes: Remove IP.

java/cuopt-java/src/main/java/com/nvidia/cuopt/spi/CuOptProvider.java
```diff
@@ -0,0 +1,95 @@
...
+ * only Java 21-compatible types (no {@code MemorySegment}, no
+ * {@code Arena}) — native handles cross the boundary as opaque
+ * {@code long} values (raw addresses).
+ *
+ * <p>The implementation lives in
+ * {@code com.nvidia.cuopt.internal.CuOptProviderImpl} under
+ * {@code src/main/java22/}; the JVM resolves it via
+ * {@link java.util.ServiceLoader} on Java 22+ runtimes.
+ */
+public interface CuOptProvider {
```
chris-maes: Consistent capitalization: cuOptProvider

java/cuopt-java/src/main/java/com/nvidia/cuopt/CuOpt.java
```diff
@@ -0,0 +1,76 @@
...
+ *     problem.addConstraint(expr, LESS_EQUAL, 100, "c1");
+ *     problem.setObjective(obj, MAXIMIZE);
+ * }</pre>
+ *
+ * <p>Parameter name string constants (mirrors of the C-side
+ * {@code CUOPT_*} preprocessor macros from {@code constants.h}) are
+ * grouped at the bottom. They're used with
+ * {@link com.nvidia.cuopt.linear_programming.SolverSettings#setParameter(String, Object)}.
+ */
+public final class CuOpt {
```
chris-maes: Consistent capitalization cuOpt

java/cuopt-java/src/main/java/com/nvidia/cuopt/CuOptException.java
```diff
@@ -0,0 +1,27 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package com.nvidia.cuopt;
+
+/**
+ * Unchecked exception thrown by cuopt-java when a native call fails or
+ * the runtime environment is misconfigured (e.g., the native library
+ * cannot be loaded, or the JVM is older than Java 22).
+ */
+public class CuOptException extends RuntimeException {
```
chris-maes: Consistent captialization: cuOpt

java/cuopt-java/src/main/java22/com/nvidia/cuopt/internal/CsrBuilder.java
```diff
@@ -0,0 +1,152 @@
...
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Builds CSR (compressed sparse row) representations from
+ * {@link LinearExpr} / {@link QuadraticExpr} for the FFM bridge.
+ *
+ * <p>Output arrays are ready to copy into native {@code MemorySegment}s.
+ */
+final class CsrBuilder {
```
chris-maes: Consistent capitalization: CSRBuilder

java/cuopt-java/src/main/java22/com/nvidia/cuopt/internal/CuOptProviderImpl.java
```diff
@@ -0,0 +1,608 @@
...
+ * {@code META-INF/services/com.nvidia.cuopt.spi.CuOptProvider}.
+ *
+ * <p>Lives in the Java 22 multi-release layer; loaded only on JVMs at
+ * Java 22 or higher.
+ *
+ * <p>Native handles cross the SPI boundary as {@code long} (raw
+ * addresses); this class reconstructs {@code MemorySegment} from the
+ * addresses on each call.
+ */
+public final class CuOptProviderImpl implements CuOptProvider {
```
chris-maes: Consistent capitalization: cuOptProviderImpl

java/cuopt-java/src/test/java/com/nvidia/cuopt/linear_programming/MilpSolverIT.java
```diff
@@ -0,0 +1,63 @@
...
+
+            LinearExpr cap = new LinearExpr().addTerms(weights, x);
+            problem.addConstraint(cap, LESS_EQUAL, capacity, "capacity");
+
+            LinearExpr obj = new LinearExpr().addTerms(values, x);
+            problem.setObjective(obj, MAXIMIZE);
+
+            problem.solve(settings);
+
+            assertTrue(problem.isMip());
```
chris-maes: Consistent capitalization: isMIP

java/cuopt-java/src/test/java/com/nvidia/cuopt/linear_programming/MipCallbackIT.java
```diff
@@ -0,0 +1,116 @@
...
+import static com.nvidia.cuopt.CuOpt.*;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.atomic.AtomicInteger;
+import org.junit.jupiter.api.Test;
+
+class MipCallbackIT {
```
chris-maes: Consistent capitalization: MIPCallbackIT

java/cuopt-java/src/test/java/com/nvidia/cuopt/linear_programming/QpSolverIT.java
```diff
@@ -0,0 +1,56 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package com.nvidia.cuopt.linear_programming;
```
chris-maes: I don't think we should call this `com.nvidia.cuopt.linear_programming`. We should try to deprecate the `linear_programming` term since cuOpt contains much more than this now.

java/cuopt-java/src/test/java/com/nvidia/cuopt/linear_programming/MipCallbackIT.java
chris-maes: Why does it have an `IT` at the end

ramakrishnap-nv: `IT` is the Maven convention for **integration tests**, separating two phases:

- `*Test.java` — run by **Surefire** in the `test` phase, against `target/classes/` (exploded classes)
- `*IT.java` — run by **Failsafe** in the `integration-test` phase, against the **packaged JAR**

This split matters specifically for cuopt-java because:

1. **MR-JAR semantics** are only honored when classes are loaded from an actual JAR file, not from exploded `target/classes/`. The Java 22 FFM implementation lives under `META-INF/versions/22/` and is invisible to test runs that load from exploded classes. `pom.xml` configures Failsafe with `<classesDirectory>${project.build.directory}/${project.build.finalName}.jar</classesDirectory>` to force JAR-based loading. (cuvs Issue #1037)

2. **GPU dependency** — every `*IT.java` test actually solves a problem, so it requires `libcuopt.so` on the path and a CUDA GPU. The `*Test.java` files (e.g. `LinearExprTest`) are pure-Java and run anywhere.

3. **Selective skipping** — `UNIT_TESTS_ONLY=true ./java/build.sh` runs only the fast Surefire batch (useful when iterating without a GPU). Arm64 CI used this mode before the GPU runner was wired up.

So `MipCallbackIT` keeps the `IT` suffix because it solves a MIP at runtime — i.e. it's an integration test, not a unit test.

java/cuopt-java/src/main/java/com/nvidia/cuopt/optimization/LinearExpr.java
ramakrishnap-nv: Agreed worth a look. Current `LinkedHashMap<Variable, Double>` was chosen for two properties we'd lose with naive arrays: same-variable coefficient accumulation (Gurobi convention — `e.addTerm(2,x).addTerm(3,x)` → `5x` is one term, not two) and deterministic insertion-order iteration during CSR builds.

Three alternatives that keep the user-facing API the same:

1. **Parallel arrays + uniquify at CSR-build**. `int[] varIndices + double[] coeffs` in `LinearExpr`; deduplication moved into `CSRBuilder`. ~5× less memory in the build phase (~88 bytes/term → ~16). `numTerms()` semantics shift from "distinct" to "raw" until the CSR build dedups.

2. **Hand-rolled primitive-keyed open-addressing map** (`int → double` keyed by `Variable.index()`, ~50 LOC of arithmetic + linear probing). Preserves the O(1)-merge semantics, no boxing, no API change, no new dependency. ~16-24 bytes/term.

3. **`fastutil` (`Int2DoubleOpenHashMap`) or similar**. Same payoff as (2) without writing the hash logic ourselves, but adds a runtime dependency to cuopt-java.

The work-log under `temp/` flagged this as a deferred ~2-day optimization for 10⁶-variable problems — it's small at typical sizes. My lean is **(2)** since it preserves the merge semantics, no API contract change, and no new dep. Happy to take any of the three; which direction do you want?

java/cuopt-java/src/main/java/com/nvidia/cuopt/optimization/Problem.java
ramakrishnap-nv: The two name-keyed maps (`variablesByName`, `constraintsByName`) back `Problem.getVariable(String)` / `Problem.getConstraint(String)` for Gurobi-style name lookup. They're populated only when a non-empty name is passed at `addVariable`/`addConstraint` time — so problems built without naming entities pay essentially nothing (~96 bytes total per Problem for two empty HashMap headers).

Three options:

1. **Drop entirely.** Remove both maps + the `getVariable(String)` / `getConstraint(String)` public methods. Users keep their own `Variable`/`Constraint` references. Cleanest; loses Gurobi parity (no `getVarByName`-style lookup) but no internal code paths depend on it — only one test reference exists (`LpSolverIT.java:107-108`, easy to update).

2. **Build lazily on first lookup.** Zero cost if never looked up. Adds complexity for marginal savings since the empty-map cost is already tiny.

3. **Keep as-is.** Cost-when-empty is ~zero; users who name things get O(1) lookup.

The internal solve path uses `getVariable(int)` and `getConstraint(int)` (by index), not by name — so dropping name lookup wouldn't touch the FFM impl. My lean is **(1)** if you want simpler; **(3)** if Gurobi parity is worth keeping. Which way?

PR #1194, opened 2026-05-08
ci(pr): classify PR test failures as new vs. known and post sticky comment

ramakrishnap-nv:
## Summary

This is to handle flaky CI where failures might not be part of the PR and it will categorized as known failures so developers can admin merge PRs in urgent situations.

On every PR run, classify each test failure against the target branch's nightly history and post a single sticky comment that splits failures into:

- **CRASHES** (signal-death) — surfaced in a red `[!CAUTION]` callout with the full diagnostic.
- **NEW** — failures this PR likely introduced.
- **KNOWN** — recurring on nightly, known flaky on nightly, or flaked in this run.


.github/workflows/pr.yaml
```diff
@@ -627,3 +631,24 @@ jobs:
...
+      - conda-python-tests
+      - wheel-tests-cuopt
+      - wheel-tests-cuopt-server
+      - test-self-hosted-server
+    if: always()
+    # The reusable workflow's job needs `contents: read` (for
+    # actions/checkout) and `pull-requests: write` (to post the sticky
+    # comment).  Reusable workflows inherit the caller's GITHUB_TOKEN
+    # permissions, so the caller must grant at least the same set —
+    # the workflow-level default is `permissions: {}`.
```
```suggestion
```

Minor personal preference, I don't think this comment is adding much information. Recommend removing it.

.github/workflows/pr_test_summary.yaml
```diff
@@ -0,0 +1,76 @@
...
+        required: true
+      CUOPT_S3_URI:
+        required: true
+
+# The python:3.14-slim container's default /bin/sh is dash, which doesn't
+# understand `[[ ... ]]` or `set -o pipefail`.  Force bash for all run
+# steps in this workflow (matches self_hosted_service_test.yaml convention).
+defaults:
+  run:
+    shell: bash
```
jameslamb:
```suggestion
```

I don't understand this at all. GitHub Actions default shell if you don't explicitly set `shell:` is already `bash -e {0}`.

docs: https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#jobsjob_iddefaultsrunshell

I'm sorry to have to ask this but... is this something you added for a real problem you faced, or something an LLM generated?

Unless this was really intentional, please remove it to simplify this.

.github/workflows/pr_test_summary.yaml
```diff
@@ -0,0 +1,76 @@
...
+    permissions:
+      contents: read
+      pull-requests: write
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          persist-credentials: false
+      - name: Install dependencies
+        run: |
+          apt-get update && apt-get install -y --no-install-recommends curl
```
jameslamb:
```suggestion
          apt-get update
          apt-get install -y --no-install-recommends curl
```

This is already multiple lines. Let's simplify this and make it a bit easier to read by splitting commands onto their own lines.

ci/pr_summary.sh
```diff
@@ -0,0 +1,71 @@
...
+if [ -z "${GITHUB_RUN_ID:-}" ]; then
+    echo "WARNING: GITHUB_RUN_ID is not set; cannot locate per-matrix summaries." >&2
+    exit 0
+fi
+
+# aws CLI uses the standard AWS_* env vars; map the cuOpt-prefixed secrets onto them.
+export AWS_ACCESS_KEY_ID="${CUOPT_AWS_ACCESS_KEY_ID:-${AWS_ACCESS_KEY_ID:-}}"
+export AWS_SECRET_ACCESS_KEY="${CUOPT_AWS_SECRET_ACCESS_KEY:-${AWS_SECRET_ACCESS_KEY:-}}"
+unset AWS_SESSION_TOKEN
+
```
jameslamb:
```suggestion
```

All of this flexibility adds complexity and increases the risk of "the job has been silently failing for months" types of bugs. In my opinion, it'd be much better to make this strict everywhere and have it error out with a big loud error if anything other than the exact required environment variables, arguments, etc. are set.

Will you consider that?

ci/pr_summary.sh
```diff
@@ -0,0 +1,71 @@
...
+GITHUB_RUN_URL="${GITHUB_SERVER_URL:-https://github.com}/${GITHUB_REPOSITORY}/actions/runs/${GITHUB_RUN_ID}"
```
jameslamb:
```suggestion
GITHUB_RUN_URL="https://github.com/${GITHUB_REPOSITORY}/actions/runs/${GITHUB_RUN_ID}"
```

This is using GitHub-specific conventions so I really wouldn't expect any other value of `GITHUB_SERVER_URL` to work here. If you don't have a specific intended use case for overriding `GITHUB_SERVER_URL`, just hard-code github.com to make this simpler.

ci/pr_summary.sh
```diff
@@ -0,0 +1,71 @@
...
+S3_PR_SUMMARIES_PREFIX="${CUOPT_S3_URI}ci_test_reports/pr/run-${GITHUB_RUN_ID}/"
+COMMENT_FILE="${OUTPUT_DIR}/pr_comment.md"
+
+echo "Aggregating PR per-matrix summaries from ${S3_PR_SUMMARIES_PREFIX}"
+python3 "${SCRIPT_DIR}/utils/aggregate_pr.py" \
+    --s3-pr-summaries-prefix "${S3_PR_SUMMARIES_PREFIX}" \
+    --output-dir "${OUTPUT_DIR}" \
+    --target-branch "${GITHUB_BASE_REF:-main}" \
+    --sha "${GITHUB_SHA:-}" \
```
jameslamb:
```suggestion
    --target-branch "${GITHUB_BASE_REF" \
    --sha "${GITHUB_SHA}" \
```

From this point forward, assume I'm asking for a review of everything in this PR following this same trend... remove unnecessary flexibility like fallback values. This code is so complex, you could make it simpler and reduce the risk of mistakes by making all of the interfaces strict and limiting how many different places configuration comes from.

ci/pr_summary.sh
```diff
@@ -0,0 +1,71 @@
...
+    --github-run-url "${GITHUB_RUN_URL}" \
+    --run-date "$(date +%F)"
+
+if [ ! -s "${COMMENT_FILE}" ]; then
+    echo "No failures or flakes; not posting a PR comment."
+    exit 0
+fi
+
+if [ -z "${GITHUB_TOKEN:-}" ]; then
```
jameslamb: This validation should happen earlier, before you've done all the work of running `aggregate_pr.py`.

That'll reduce the time it takes to be notified of an issue, which saves some capacity in CI and helps you make faster progress on the project.

jameslamb (dismissed):
Ok I had to stop... this PR is way too large to review line-by-line. From my perspective, this amount of new code and complexity is not really worth it in exchange for generating summary comments on PRs... but ultimately that's your decision to make.

Please do a pass over all this new code for the general themes I brought up... remove unnecessary flexibility and competing sources of configuration.

After that, I'll be happy to review again.

ci/utils/aggregate_pr.py
```diff
@@ -0,0 +1,513 @@
...
+
+    Returns ``(broken_on_nightly, known_flaky_nightly, flaked_in_pr_run)``.
+    Each entry retains its full per-matrix context.
+    """
+    broken_on_nightly = []
+    known_flaky_nightly = []
+    flaked_in_pr_run = []
+
+    for entry in recurring:
+        cls = entry.get("pr_classification", "")
```
jameslamb, line 102: I'm really struggling to follow this code. Using `dataclasses` to represent the JSON data would be clearer and stricter.

jameslamb (commented):
Thanks for the changes and for re-writing the description, it's much clearer to me now what the goal is.

Now that I can see what it is, I don't support this. 

I'm not convinced that this can work well (for example... we run different test matrices in nightlies vs. PRs, won't that confuse the "known issue" logic?), or that it's worth the costs it imposes which include:

* non-trivial new complexity and maintenance burden
* expanded GitHub Actions workflow permissions
* new cloud costs (API calls + storage in S3)
* new risk of hitting GitHub API limits (which could block all other operations in the repo, like artifact uploading / downloading)
* a huge amount of new CI code (more code = larger changesets = easier for malicious code to sneak past human review)

HOWEVER... I'm leaving a non-blocking review. I am not an active maintainer of cuOpt. If you weigh all of those costs and still think the benefits outweigh them, and get another cuOpt maintainer in `cuopt-ci-codeowners` like @tmckayus to approve, then I won't stand in the way.

ramakrishnap-nv:
> Thanks for the changes and for re-writing the description, it's much clearer to me now what the goal is.
> 
> Now that I can see what it is, I don't support this.
> 
> I'm not convinced that this can work well (for example... we run different test matrices in nightlies vs. PRs, won't that confuse the "known issue" logic?), or that it's worth the costs it imposes which include:
> 
> * non-trivial new complexity and maintenance burden
> * expanded GitHub Actions workflow permissions
> * new cloud costs (API calls + storage in S3)
> * new risk of hitting GitHub API limits (which could block all other operations in the repo, like artifact uploading / downloading)
> * a huge amount of new CI code (more code = larger changesets = easier for malicious code to sneak past human review)
> 
> HOWEVER... I'm leaving a non-blocking review. I am not an active maintainer of cuOpt. If you weigh all of those costs and still think the benefits outweigh them, and get another cuOpt maintainer in `cuopt-ci-codeowners` like @tmckayus to approve, then I won't stand in the way.

Hmm, I agree with nightly vs PR test suite testing on different matrices, but most of the times, the tests are failing similarly in both, so the corner case seems acceptable.

But I my major concern would be the new failure points getting added due to these which might trigger other issues.

Currently we have nightly dashboard which showcases all the failures and may be users can manually cross reference when needed. 

@mlubin @chris-maes @rg20 what are your thoughts ?

gforsyth:
I would be concerned about the ability to distinguish between "this common failure is because this test is flaky" and "this seemingly common failure is because of a bug that's just been introduced".

Perfectly deterministic tests are hard, and I can imagine optimization problems make them even harder -- but if there are tests that become routine to ignore, then those tests no longer provide a useful signal, or they potentially mask a useful signal.

I would suggest that if a test flakes out more than 50% of the time, that is a broken test -- it should be fixed, or made more flexible (by, e.g. loosening comparison tolerances), or removed entirely.

mlubin:
I'd like to put aside the workflow/procedural discussion of if or when we should decide to merge PRs despite test failures, that's not being changed by this PR and is probably better to cover on a call.

The team is excited about the reporting aspect because test failures are common, and having to dig through the CI logs on each occasion is a waste of developer time. The tie in with historical data on if this is a new or pre-existing or flaky failure is also useful and saves developer time. However, maybe this is too much for a single PR and we could split this into more digestible chunks.

@jameslamb how many of your concerns would still apply to a bot that narrowly reads the PR's CI artifacts and produces a comment summarizing which tests failed/crashed?

jameslamb:
> _how many of your concerns would still apply to a bot that narrowly reads the PR's CI artifacts and produces a comment summarizing which tests failed/crashed?_

That would eliminate the need for S3, which'd remove the concerns about cost and remove the need for this new workflow to have access to AWS credentials.

The other concerns would remain.

I would not support something like this RAPIDS-wide, I think the costs/risks outweigh the benefits and I agree with @gforsyth 's comments.

You all (cuOpt developers) can decide how you want to proceed here... I am not blocking the PR and you don't have to convince me. I'm going to remove myself from this PR.

tmckayus:
I like the idea of running a bot outside of github to pull the logs/artifacts and evaluate, and even post back to the PR with conclusions against a database somewhere. Such a bot could even be triggered by a webhook. The nightly runs are already logging information about tests, so that could remain.  We could grow the capabilities of an external bot over time, and it remains credential issues etc from github.

As for ignoring flakes, that's a current and different problem. We have some flaky tests now and many times, under time pressure, we just re-run til we get a pass.  We need to be more intentional about fixing or scrapping those tests, regardless of the tool.

ramakrishnap-nv:
closing this on behalf of #1412

Issue #1195, opened 2026-05-09
[BUG] Routing solver: break-dimension bugs


np96:
## Describe the bug

For the entire problem setup and reproduce, see [repro_break_dim_bugs.py](https://github.com/user-attachments/files/27583318/repro_break_dim_bugs.py).

| | Setup | Observed | Expected |
|---|---|---|---|
| **A** | unreachable break TW `[5, 5]`, break loc 200 s away | status 0, route `0→1→0` (break silently missing) | status 1 INFEASIBLE |
| **B** | break TW `[100, 200]`, customer at t=10, route can finish at t=20 | status 0, route `0→2→1→0`, cost 410.25 | status 0, route `0→1→0`, cost 20 |

Both stem from a single design assumption: every route has *exactly N breaks*. The cost formula `max(0, breaks − N)` only penalizes excess; squeeze unconditionally inserts every dim; `eject_until_feasible` protects breaks; `find_break_insertions` only moves existing breaks.

## Observed

```
expect status 1 since the problem is infeasible
  A: cost=20.00 vehicles=1
      vehicle 0: 0(Depot) -> 1(Delivery) -> 0(Depot)
expect status 0 and no break in solution since it is not needed
  B: cost=410.25 vehicles=1
      vehicle 0: 0(Depot) -> 2(Break) -> 1(Delivery) -> 0(Depot)

```
## Expected

```
expect status 1 since the problem is infeasible
  A: INFEASIBLE (status=1)
expect status 0 and no break in solution since it is not needed
  B: cost=20.00 vehicles=1
      vehicle 0: 0(Depot) -> 1(Delivery) -> 0(Depot)
```

## Why it matters

This issue negatively affects UX and is a blocker for feature development. 
Case A is a bug with straightforward fix.
Case B appears to be a solver/design limitation rather than a simple feasibility bug.

## Environment

- cuOpt: `b311a990` (main), built from source (Release).
- Single A100 80 GB PCIe
- Driver Version: 580.65.06, CUDA Version: 13.0, nvcc V13.1.115 
- Linux 5.15.179, Python 3.14, miniconda.

## Notes

- Adding distance breaks semantics forces EVs travel around chargers to comply with the required breaks due to issue B.
- To fix case A override `inf_cost[BREAK]` in route_t::compute_cost with `max(0, required - breaks)` using a dynamic required-count derived from the route's actual arrival_at_depot. Squeeze still inserts every dim so no construction path changes; all 83 routing tests pass.
[...]

PR #1197, opened 2026-05-11
fix(routing): respect vehicle breaks edge-cases

np96:
## Description

Restore required breaks after route construction, initial-solution loading, and route changes. Skip time and distance breaks when the actual route finishes at or before their deadline, recheck skipped dimensions when another break extends the route, and remove unnecessary breaks when routes shrink.

Fixes both cases in #1195: the unreachable required break returns `INFEASIBLE`, and the late-break example returns cost **20** with no break stop. Configured break windows must still be compatible with the vehicle shift.

Validation: **129 local Python routing tests** and **28 C++ tests** passed. The two tests that grow and shrink routes with existing breaks also passed CUDA synchronization checking with zero errors. Regression coverage includes distance breaks, generated and injected routes, exact deadlines, waiting and return travel, dependent dimensions, mixed break types, and unchanged rejection of breaks outside the configured shift.

## Checklist

- Testing
  - [x] Added tests
- Documentation
  - [x] Documentation changes are excluded from this PR

### DIFF
--- a/cpp/src/routing/adapters/adapted_generator.cu
+++ b/cpp/src/routing/adapters/adapted_generator.cu
@@ -1,6 +1,6 @@
 /* clang-format off */
 /*
- * SPDX-FileCopyrightText: Copyright (c) 2022-2025, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ * SPDX-FileCopyrightText: Copyright (c) 2022-2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
  * SPDX-License-Identifier: Apache-2.0
  */
 /* clang-format on */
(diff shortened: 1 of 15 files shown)


rg20:
@np96 

In the current form, I intentionally don't treat missing breaks as a violation as initial solutions are generated without breaks first and then breaks are inserted.  Breaks should not be missed, I think there is some other bug in the code.

np96:
Hi @rg20 

Do you have a chance to check this draft? Looks like there was a lack of squeeze_breaks in certain places. this fixes both cases. Most of the PR are the tests.

rg20:
> Hi @rg20
> 
> Do you have a chance to check this draft? Looks like there was a lack of squeeze_breaks in certain places. this fixes both cases. Most of the PR are the tests.

Hi @np96, 

I will have to spend sometime on rethinking the logic.  Thanks for the PR.

PR #1199, opened 2026-05-11
miplib_test: allow tiny error on objective equality

mlubin:
EXPECT_DOUBLE_EQ was observed failing in CI with a 2.2e-15 error:

https://github.com/NVIDIA/cuopt/actions/runs/25695199349/job/75445965119

```
Expected equality of these values:
  solution.get_objective_value()
    Which is: 2.9999999999999978
  3.0
    Which is: 3
```


nguidotti (approved):
Looks good to me. @mlubin do you think should we use a slightly looser tolerance like `1E-12`?

mlubin:
> do you think should we use a slightly looser tolerance like 1E-12?

I just picked the minimal tolerance needed to make the test pass since the violation was pretty small.

PR #1209, opened 2026-05-13
add routing and math flag for selective builds

Iroy30:
## Checklist

- Testing
   - [ ] Added tests
   - [ ] NA
- Documentation
   - [ ] Added new documentation
   - [ ] NA


python/cuopt_self_hosted/pyproject.toml
```diff
@@ -16,8 +16,7 @@ readme = { file = "README.md", content-type = "text/markdown" }
 authors = [
     { name = "NVIDIA Corporation" },
 ]
-license = "Apache-2.0"
-license-files = ["LICENSE"]
+license = { text = "Apache-2.0" }
```
jameslamb, line 19: What is the purpose of this change?

We very intentionally migrated all projects in RAPIDS away from this form and towards the form you see in the repo as part of https://github.com/rapidsai/build-planning/issues/152

The form you're proposing is considered deprecated, and newer versions of build tools may stop supporting it: https://peps.python.org/pep-0639/#deprecate-license-key-table-subkeys

jameslamb (changes requested):
I left one comment asking for clarification but more broadly... can you please add a description explaining this?

I don't understand the changes that say "try split wheels first, then full libcuopt" or what the `_build_variant.py.in` is used for. GitHub search doesn't return any other file by that name across all of GitHub (https://github.com/search?q=path%3A_build_variant.py.in&type=code), so I'm guessing this is something cuOpt-custom? If you can explain this change in more detail I'd be happy to review again.

PR #1211, opened 2026-05-13
ci: split build.yaml concurrency group by build_type

ramakrishnap-nv:
## Summary
- The concurrency group in `build.yaml` was keyed on `workflow + ref`, so a push to `main` (or `release/*`) would cancel an in-progress nightly-dispatched `build.yaml` run on the same branch — and the reverse was equally true.
- Add the `build_type` input to the group key (falling back to `'branch'` when `inputs` is unset on `push`) so nightly and branch builds occupy separate groups.
- Same-type, same-ref runs still cancel older runs (e.g. rapid pushes to `main` still supersede each other).

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -39,7 +39,7 @@ on:
         default: false
 
 concurrency:
-  group: ${{ github.workflow }}-${{ github.ref }}
+  group: ${{ github.workflow }}-${{ github.ref }}-${{ inputs.build_type || 'branch' }}
   cancel-in-progress: true
 
 permissions: {}


.github/workflows/build.yaml
```diff
@@ -39,7 +39,7 @@ on:
         default: false
 
 concurrency:
-  group: ${{ github.workflow }}-${{ github.ref }}
+  group: ${{ github.workflow }}-${{ github.ref }}-${{ inputs.build_type || 'branch' }}
```
jameslamb, line 42: This makes sense to me. Just noting I looked into this here: https://docs.github.com/en/actions/reference/workflows-and-actions/contexts#github-context

`github.ref` should be identical and set to just the branch name for both a `push` and `workflow_dispatch` run, I think, so it makes sense that those would conflict.

jameslamb (approved):
Oh cool idea!

A nightly run cancelling one triggered from a merge is mildly annoying only because it delays package publishing, but I think not that big of a problem.

However, a merge cancelling a nightly build is more serious. Because the nightly CI polls for **a single, specific workflow run** ([rapidsai/trigger-workflow-and-wait code link](https://github.com/rapidsai/trigger-workflow-and-wait/blob/ce9001c52e6a24c01f59a5a647a37d259310846b/entrypoint.sh#L206)), in that situation where there's a merge while the nightly build is going on, the nightly build will be cancelled and the nightly tests will never run.

I support trying this, and we should probably do it across all of RAPIDS. I'll put up a `build-planning` issue proposing that and see what folks think.

PR #1225, opened 2026-05-14
Implement python and C api for semi-continuous variables

### LINKED ISSUE #1156: [FEA] Add python support for semi-continuous variables

hlinsen:
This PR adds semi continuous to python and C APIs.
Closes: https://github.com/NVIDIA/cuopt/issues/1156


cpp/src/io/mps_parser.cpp
```diff
@@ -250,7 +250,7 @@ BoundType convert(std::string_view str)
     return LowerBoundIntegerVariable;
   } else if (str == "UI") {
     return UpperBoundIntegerVariable;
-  } else if (str == "SC" || str == "LC") {
+  } else if (str == "SC") {
```
mlubin, line 253: Is this change to mps_parser intentional?

hlinsen: Yes. The semi continuous variables is always defined with the upper bound SC which is mandatory. LC shouldn't have been added.

ramakrishnap-nv (approved):
Rest looks good, I would suggest adding an example to docs under python. Not a blocking comment.

cpp/src/pdlp/cuopt_c.cpp
```diff
@@ -195,8 +195,7 @@ cuopt_int_t cuOptCreateProblem(cuopt_int_t num_constraints,
     // Set variable types (problem category is auto-detected)
     std::vector<var_t> variable_types_host(num_variables);
     for (int j = 0; j < num_variables; j++) {
-      variable_types_host[j] =
-        variable_types[j] == CUOPT_CONTINUOUS ? var_t::CONTINUOUS : var_t::INTEGER;
+      variable_types_host[j] = detail::char_to_var_type(variable_types[j]);
```
mlubin, line 203: Agreed with the AI comment. We should validate the variable type is valid at the API boundary.

mlubin (approved):

chris-maes (approved):
LGTM. Thanks @hlinsen

PR #1227, opened 2026-05-15
Declare defaulted move ops noexcept (SonarQube BLOCKER cpp:S5018)

ramakrishnap-nv:
SonarQube rule **`cpp:S5018`** flags `= default`ed move constructors and move-assignment operators that arent declared `noexcept`. This PR adds `noexcept` to the seven flagged sites on `main`, clearing **7 of 8 BLOCKER** findings.

All affected classes already move trivially-noexcept members (`std::vector`, `std::unique_ptr`, `rmm::device_uvector`, raw pointers, scalars), so the specifier compiles cleanly. None of these four types are currently stored by value in an STL container (`mip_node_t` is always behind `unique_ptr` or raw pointer; the other three dont appear in containers), so on the current code this is a runtime no-op — it just satisfies the lint rule.

### DIFF
--- a/cpp/include/cuopt/linear_programming/optimization_problem.hpp
+++ b/cpp/include/cuopt/linear_programming/optimization_problem.hpp
@@ -116,8 +116,8 @@ class optimization_problem_t : public optimization_problem_interface_t<i_t, f_t>
 
   explicit optimization_problem_t(raft::handle_t const* handle_ptr);
   optimization_problem_t(const optimization_problem_t<i_t, f_t>& other);
-  optimization_problem_t(optimization_problem_t<i_t, f_t>&&)            = default;
-  optimization_problem_t& operator=(optimization_problem_t<i_t, f_t>&&) = default;
+  optimization_problem_t(optimization_problem_t<i_t, f_t>&&) noexcept            = default;
+  optimization_problem_t& operator=(optimization_problem_t<i_t, f_t>&&) noexcept = default;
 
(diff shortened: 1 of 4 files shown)


mlubin:
> so the specifier is a no-op at runtime — it just lets the rule pass and, where it matters, lets STL containers use move-not-copy when these types are stored

Sorry which one is it? Does it change the behavior of STL containers when these types or stored or is it a no-op?

ramakrishnap-nv:
Sorry, that was contradictory wording on my part. Quick clarification:

In general, `noexcept` on a move ctor IS observable: `std::vector` (and friends) reallocate via `std::move_if_noexcept`, so adding `noexcept` lets reallocations *move* elements instead of *copy* them. So it's not strictly a runtime no-op — but only when the type is stored by value in such a container.

For these four types specifically, I checked the codebase:

- `optimization_problem_t`, `pdlp_warm_start_data_t`, `deterministic_diving_worker_t` — never stored by value in any STL container.
- `mip_node_t` — only ever held as `std::unique_ptr<mip_node_t>` or raw `mip_node_t*` in containers (`std::vector`, `std::deque`, `std::list`); the pointer types' moves are noexcept regardless of `mip_node_t`'s declaration.

So in practice on the current code, the change really is a runtime no-op; it only satisfies the lint rule and would *start* mattering if a future change put one of these in a container by value. I've also dropped the misleading paragraph from the PR description.

rg20 (approved):

chris-maes (approved):
+1 for mip_node.hpp changes.

PR #1230, opened 2026-05-15
Make mip_node_t destructor exception-safe (SonarQube CRIT BUG cpp:S1048)

ramakrishnap-nv:
## Summary

The iterative-teardown destructor in `mip_node_t` uses `std::vector::push_back` to walk and detach the subtree. `push_back` can throw `std::bad_alloc`, and a destructor that lets an exception propagate calls `std::terminate`.

Wraps the body in `try { ... } catch (...) {}` so the destructor stays exception-free. Under OOM the catch absorbs the failure, and any descendants still attached to `children` are destroyed via the recursive `unique_ptr` chain as this frame unwinds — which risks stack overflow on a very deep tree under OOM, but is strictly better than `std::terminate`.


mlubin:
Do we intent to fail nicely on every potential line that could throw `bad_alloc`? I doubt this is the only spot that would need to change if that's true.

ramakrishnap-nv:
Good question — I shouldn't have framed this as "fail nicely on bad_alloc." The narrower motivation is the destructor contract specifically: since C++11, destructors are implicitly `noexcept`, so an escaping exception calls `std::terminate` immediately, with no chance for a handler. In non-destructor code an unhandled `bad_alloc` can still propagate to a caller that may choose what to do. Rule `cpp:S1048` targets only this destructor case.

I also surveyed `cpp/{include,src}` for other destructors that allocate or otherwise can throw. Results, and what's now in this PR:

| File | Destructor | Status |
|---|---|---|
| `cpp/src/branch_and_bound/mip_node.hpp` | `mip_node_t` | Fixed in this PR (`push_back` → `bad_alloc`) |
| `cpp/src/mip_heuristics/feasibility_jump/fj_cpu.cu` | `timing_raii_t` | Fixed in this PR (same `push_back` pattern) |
| `cpp/src/grpc/client/grpc_client.cpp` | `grpc_client_t` (via `stop_log_streaming`) | Fixed in this PR (`thread::join` → `system_error`, `lock_guard` ctor) |
| `cpp/src/utilities/high_res_timer.hpp` | `HighResTimer` | Body is empty — safe |

SonarQube only flagged `mip_node_t` because `.cu` files aren't covered by its C++ analyzer, and it doesn't reason about `thread::join` propagating from the called `stop_log_streaming()`. That's the gap your question implicitly pointed at.

mlubin:
I'll defer approval to others who have more familiarity with this code.

Edgarruiz8585:
> **Actionable comments posted: 1**
> 
> ---
> 
> <details>
> <summary>ℹ️ Review info</summary>
> 
> <details>
> <summary>⚙️ Run configuration</summary>
> 
> **Configuration used**: Path: .coderabbit.yaml
> 
> **Review profile**: CHILL
> 
> **Plan**: Enterprise
> 
> **Run ID**: `6e1fa19c-ebee-4801-9e5f-7248286561d0`
> 
> </details>
> 
> <details>
> <summary>📥 Commits</summary>
> 
> Reviewing files that changed from the base of the PR and between b08429f5ec708ef3c61d4a58d36cafe5014507d6 and fd66d5a82a378d0678778dc2ada98cc4d17e22b2.
> 
> </details>
> 
> <details>
> <summary>📒 Files selected for processing (1)</summary>
> 
> * `cpp/src/grpc/client/grpc_client.cpp`
> 
> </details>
> 
> </details>
> 
>  @Edgarruiz8585

cpp/src/grpc/client/grpc_client.cpp
```diff
@@ -127,7 +127,15 @@ grpc_client_t::grpc_client_t(const std::string& server_address) : impl_(std::mak
...
-grpc_client_t::~grpc_client_t() { stop_log_streaming(); }
+grpc_client_t::~grpc_client_t()
+{
+  // stop_log_streaming() can throw via std::thread::join (std::system_error)
+  // and std::lock_guard's mutex acquisition. Keep the destructor exception-free.
+  try {
+    stop_log_streaming();
+  } catch (...) {
+  }
+}
```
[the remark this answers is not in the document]
ramakrishnap-nv: Good catch — both scenarios are correct. The `try/catch` around `stop_log_streaming()` does not by itself prevent `std::terminate`, because:

1. If `lock_guard`/`TryCancel` throw before the swap, the member `log_thread_` still owns a joinable thread, and that joinable-thread destructor fires *after* my catch block when the enclosing `grpc_client_t` destructor finishes.
2. If `t->join()` throws, the local `unique_ptr<std::thread> t` is destroyed during the same unwind, before my outer catch fires.

Fixed in `9dcb39ea`: inlined a noexcept shutdown directly in the destructor — `stop_logs_.store(true)` + try-cancel + swap into a local, then `try{ join } catch { detach }`. After the swap `log_thread_` is empty, and on a join failure we detach before the local goes out of scope, so neither thread destructor sees a joinable state. The public `stop_log_streaming()` is unchanged (it still has callers in `solve_lp`/`solve_mip` that rely on its current throwing contract).

rg20:
Looks like you are catching the exceptions but not reporting them.   You can use cuopt_expects and throw RunTimeError.

ramakrishnap-nv:
@rg20 Good point — silently swallowing hides real failures. Switched each catch in this PR to log via `CUOPT_LOG_ERROR("…: %s", e.what())`, matching the pattern used in `cpp/src/mip_heuristics/solve.cu` (commit `c27e8f6a`).

One thing I could not literally do: throw `RuntimeError` from these catch blocks. Re-throwing out of a destructor would re-introduce the exact `std::terminate` path this PR exists to close (destructors are implicitly `noexcept` since C++11, so any escaping exception aborts the process). Logging is the closest we can get while keeping the destructor exception-free. If you’d rather we *also* surface these via `cuopt_expects` somewhere upstream of the destructor — e.g., add a noexcept teardown helper that callers like `solve_lp`/`solve_mip` use, and have *that* throw — happy to do that as a follow-up.

nguidotti:
IMO we should move this logic to `search_tree_t` since it is where the entire tree is stored, not in the `mip_node_t`. For saving memory I would first switch to a depth-first search instead of bread-first and destroy the parent after moving the children to the stack.

ramakrishnap-nv:
> IMO we should move this logic to `search_tree_t` since it is where the entire tree is stored, not in the `mip_node_t`. For saving memory I would first switch to a depth-first search instead of bread-first and destroy the parent after moving the children to the stack.

@nguidotti  We may have to address this in a follow-up PR

rg20 (approved):

nguidotti (approved):

cpp/src/branch_and_bound/mip_node.hpp
```diff
@@ -44,18 +45,29 @@ class mip_node_t {
...
+      for (size_t i = 0; i < nodes.size(); ++i) {
+        for (auto& c : nodes[i]->children) {
+          if (c) { nodes.push_back(std::move(c)); }
+        }
+      }
 
-    // scope-exit ensure destruction of all detached leaves
+      // scope-exit ensure destruction of all detached leaves
+    } catch (const std::exception& e) {
+      CUOPT_LOG_ERROR(
```
hlinsen: `CUOPT_LOG_ERROR` can throw with allocations. I think it's best to use `std::cerr` inside the `catch` block

ramakrishnap-nv: Using fprintf instead of std:cerr

```
            Property             │                                       fprintf(stderr, ...)                                       │                                                       std::cerr <<                                                       │
  ├──────────────────────────────────┼──────────────────────────────────────────────────────────────────────────────────────────────────┼──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
  │ Allocation                       │ Uses libc's pre-allocated FILE buffer for the format. Proven not to allocate for simple formats. │ Sentry construction and locale facet calls can allocate. Standard doesn't forbid it.                                     │
  ├──────────────────────────────────┼──────────────────────────────────────────────────────────────────────────────────────────────────┼──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
  │ Can throw?                       │ No — it's C, returns error code on failure.                                                      │ Yes, if cerr.exceptions(...) was ever flipped on transitively (off by default, but you'd be trusting the whole process). │
  ├──────────────────────────────────┼──────────────────────────────────────────────────────────────────────────────────────────────────┼──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
  │ Compatibility with existing code │ Same "%s" format strings already in the PR — mechanical swap.                                    │ Requires rewriting to << chains.                                                                       
```

PR #1231, opened 2026-05-17
Restore resource_cast memory limit check

bdice:
## Description
Restore the memory-limit-aware branch in `get_device_memory_size()` that was temporarily removed during the CCCL/RMM memory-resource migration in #1035.

The function now uses `cuda::mr::resource_cast` on `rmm::mr::get_current_device_resource_ref()` to detect an active `rmm::mr::limiting_resource_adaptor`. When present, it reports the limit/usage/free values and returns the smaller of CUDA total memory and the adaptor allocation limit, matching the previous behavior.


anandhkb:
Requesting @rg20 's review

cpp/src/utilities/cuda_helpers.cuh
```diff
@@ -242,7 +247,19 @@ inline size_t get_device_memory_size()
 {
   size_t free_mem, total_mem;
   RAFT_CUDA_TRY(cudaMemGetInfo(&free_mem, &total_mem));
-  // TODO (bdice): Restore limiting adaptor check after updating CCCL to support resource_cast
+
+  auto res              = rmm::mr::get_current_device_resource_ref();
+  auto limiting_adaptor = cuda::mr::resource_cast<rmm::mr::limiting_resource_adaptor>(&res);
+  if (limiting_adaptor) {
+    printf("limiting_adaptor->get_allocation_limit(): %fMiB\n",
```
rg20: Remove the printf

rg20 (commented):
@bdice is this still needed?

cpp/src/utilities/cuda_helpers.cuh
bdice: This is just restoring previous code. I can remove it, just noting that I didn't originally add it.

bdice: Completed in aa2d665.

rg20 (approved):

PR #1239, opened 2026-05-19
Detect when MIP objectives must move in discrete steps. Tighten the dual bound using this info.

chris-maes:
This PR recognizes when the objective for a MIP can only move in discrete steps or increments. 

Integer variables can only move in steps of 1. You can then use equality constraints to try to find steps for continuous variables. And eventually show that an objective containing a mix of integer and continuous variables must also move in discrete steps.

Using this allows you to improve your lower bound and stop early. For example, let's say objective has to be a multiple of 1/2. You have an upper bound of 2 and a lower bound of 1.7. You can declare optimality already, since the next best objective is 1.5 and your lower bound is already better than that.

The computation is extremely fast. The longest problem in the MIPLIB benchmark, `roi5alpha10n8`,
only takes 157 ms. 

This structure appears in 15/240 MIPLIB benchmark problems. It produces a 2x speedup on `mzzv42z`

```
Model                  Step      Time w/   Time w/o  Speedup   Gap w/    Gap w/o   Gap ratio
------------------------------------------------------------------------------------------
leo1                   4         300.7s    300.8s    1.00x     2.13%     2.48%     1.16
leo2                   1         300.0s    302.2s    1.01x     2.22%     3.21%     1.45
mas74                  1e-05     35.8s     39.4s     1.10x     0.01%     0.01%     0.99
mzzv11                 2         300.0s    300.1s    1.00x     5.08%     4.63%     0.91
mzzv42z                2         113.2s    235.1s    2.08x     0.00%     0.00%     —
n2seq36q               200       305.1s    300.0s    0.98x     0.38%     0.38%     1.00
n3div36                200       300.6s    301.8s    1.00x     3.52%     3.55%     1.01
neos-2075418-temuka    250       300.0s    300.0s    1.00x     inf       inf       —
neos-4532248-waihi     0.1       300.2s    300.0s    1.00x     99.2%     99.9%     1.01
nu25-pr12              5         4.0s      3.7s      0.92x     0.01%     0.00%     0.20
nw04                   2         54.3s     52.9s     0.97x     0.00%     0.01%     —
qap10                  2         43.5s     34.2s     0.79x     0.00%     0.00%     —
sp97ar                 8         302.3s    302.7s    1.00x     1.50%     1.46%     0.97
sp98ar                 4         302.2s    302.8s    1.00x     0.41%     0.68%     1.66
supportcase33          5         300.1s    300.1s    1.00x     13.2%     16.1%     1.22
var-smallemery-m6j6    0.03125   301.6s    301.5s    1.00x     1.86%     1.81%     0.97
------------------------------------------------------------------------------------------
Geomean speedup:    1.03x  (N=15)
Geomean gap ratio:  1.14   (N=9, excluding solved/infeasible)
```


cpp/src/mip_heuristics/problem/problem.cu
```diff
@@ -1404,6 +1392,387 @@ void problem_t<i_t, f_t>::recompute_objective_integrality()
...
+  for (const auto& eq : equations) {
+    for (i_t idx = offsets[eq.con_idx]; idx < offsets[eq.con_idx + 1]; ++idx) {
+      var_eq_count[variables[idx]]++;
+    }
+  }
+
+  // Iteratively propagate
+  bool any_discovered = false;
+  bool changed        = true;
+  while (changed) {
```
akifcorduk: We can keep changed per equation right? That way we only propagate on the equations that contain variables contained in changed equations in the previous iteration.

cpp/src/mip_heuristics/problem/problem.cu
```diff
@@ -1404,6 +1392,387 @@ void problem_t<i_t, f_t>::recompute_objective_integrality()
...
+
+// Lattice propagation: for each variable, determine if it must lie on a lattice
+// x_j = k * step_j + bias_j for integer k. This is done by scanning equality constraints
+// (and inequalities treated as equalities with implicit slack variables) for constraints
+// with exactly one variable of unknown lattice structure, and solving for it.
+//
+// Internally uses rational arithmetic (int64_t numerator/denominator pairs) to avoid
+// floating-point GCD issues.
+template <typename i_t, typename f_t, typename f_t2>
+static bool propagate_lattice(i_t n_vars,
```
akifcorduk: Could you briefly explain the lattice propagation logic?

I think we can also divide this function a bit into logical pieces. I think we can also move the `rat_t` out.

akifcorduk (approved):

cpp/src/branch_and_bound/branch_and_bound.cpp
```diff
@@ -1227,7 +1227,13 @@ std::pair<node_status_t, rounding_direction_t> branch_and_bound_t<i_t, f_t>::upd
...
     node_ptr->lower_bound = leaf_obj;
-    if (original_lp_.objective_is_integral) {
+    if (original_lp_.objective_step.has_step()) {
+      f_t step = original_lp_.objective_step.step_size;
+      f_t bias = original_lp_.objective_step.bias;
+      // Round up to next value on the lattice: k * step + bias >= leaf_obj
+      f_t k             = std::ceil((leaf_obj - bias) / step - settings_.integer_tol);
+      node_ptr->lower_bound = k * step + bias;
+    } else if (original_lp_.objective_is_integral) {
       node_ptr->lower_bound = std::ceil(leaf_obj - settings_.integer_tol);
```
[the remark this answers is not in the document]
nguidotti: Maybe we should apply this directly to the `leaf_obj` and then update the `node_ptr->lower_bound`.

cpp/src/mip_heuristics/problem/problem.cu
```diff
@@ -1404,6 +1392,387 @@ void problem_t<i_t, f_t>::recompute_objective_integrality()
...
+                              const std::vector<f_t2>& var_bounds,
+                              std::vector<f_t>& lattice_step,
+                              std::vector<f_t>& lattice_bias)
+{
+  constexpr f_t eq_tol    = 1e-8;
+  constexpr int64_t max_d = 10000000;  // max denominator for rational approximation
+  constexpr double eps    = 1e-9;
+
+  // Rational number: p/q with q > 0
+  struct rat_t {
```
nguidotti: NITPICK: `rat_t` --> `integer_rational_t` (or similar) since we are storing the numerator and denominator as integers

nguidotti: I do not remember if we have a dependency on boost, but there is `boost/rational.hpp` that implement similar logic (I think)

cpp/src/mip_heuristics/problem/problem.cu
nguidotti: Should we have an iteration limit here to avoid spending too much time on this?

nguidotti (approved):
Looks good to me! We can also use this for bound propagation in the future. Just a minor comment: if we have a dependency on `boost`, maybe we should use `boost/rational.hpp` instead of our own implementation.

cpp/src/branch_and_bound/branch_and_bound.cpp
```diff
@@ -1345,8 +1351,18 @@ dual::status_t branch_and_bound_t<i_t, f_t>::solve_node_lp(
...
+    f_t bias = original_lp_.objective_step.bias;
+    // Any improving feasible solution must have objective <= cutoff - step.
+    f_t k               = std::floor((cutoff - bias) / step + settings_.integer_tol);
+    lp_settings.cut_off = (k - 1) * step + bias + settings_.dual_tol;
+  } else if (original_lp_.objective_is_integral) {
+    // If the objective is integral, any feasible solution should produce an upper bound that is
+    // (approximately) integral. We add a small tolerance and floor this value to get an integer,
+    // we then subtract 1, to stop simplex on problems that cannot improve the primal objective.
+    lp_settings.cut_off =
+      std::floor(cutoff + settings_.integer_tol) - 1 + settings_.dual_tol;
```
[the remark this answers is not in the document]
chris-maes: How could cutoff not be on the objective lattice? Cut off is equal to the objective of the best integer feasible solution.

cpp/src/mip_heuristics/problem/problem.cu
chris-maes: The time spent in this is very small.

cpp/src/mip_heuristics/problem/problem.cu
chris-maes: I don't think we should pull in boost for this

cpp/src/mip_heuristics/problem/problem.cu
```diff
@@ -1392,6 +1392,387 @@ void problem_t<i_t, f_t>::recompute_objective_integrality()
...
+  std::vector<rat_t> step_r(n_vars);
+  std::vector<rat_t> bias_r(n_vars);
+
+  for (i_t j = 0; j < n_vars; ++j) {
+    bool is_int      = (var_types[j] == var_t::INTEGER);
+    bool is_impl_int = (var_flags[j] & (i_t)problem_t<i_t, f_t>::VAR_IMPLIED_INTEGER) != 0;
+    if (is_int || is_impl_int) {
+      step_r[j] = {1, 1};
+      bias_r[j] = rat_t::from_double(get_lower(var_bounds[j]));
+    }
```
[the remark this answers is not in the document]
chris-maes: Where does an infinite lower bound lead to a NaN?

cpp/src/branch_and_bound/branch_and_bound.cpp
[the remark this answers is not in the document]
chris-maes: Probably we should distinguish between the objective at the node (leaf_obj) and the lower bound coming from integrality or objective steps. To me this suggests that we use node_ptr->lower_bound rather than leaf_obj in the fathom check.

cpp/src/mip_heuristics/problem/problem.cu
nguidotti: Yeah. This only make sense if we already had a dependency on it.

cpp/src/mip_heuristics/solver.cu
```diff
@@ -294,7 +294,15 @@ solution_t<i_t, f_t> mip_solver_t<i_t, f_t>::run_solver()
...
   }
+  if (context.settings.objective_step && context.problem_ptr->get_objective_step().has_step()) {
+    CUOPT_LOG_INFO("Objective step size %g, bias %g",
+                   context.problem_ptr->get_objective_step().step_size,
+                   context.problem_ptr->get_objective_step().bias);
+  }
   branch_and_bound_problem.objective_is_integral = context.problem_ptr->is_objective_integral();
+  if (context.settings.objective_step) {
+    branch_and_bound_problem.objective_step = context.problem_ptr->get_objective_step();
+  }
```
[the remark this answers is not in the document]
chris-maes: Yep. We will leave moving this into deterministic mode as a TODO.

cpp/include/cuopt/linear_programming/constants.h
```diff
@@ -70,21 +68,20 @@
 #define CUOPT_MIP_CLIQUE_CUTS                      "mip_clique_cuts"
 #define CUOPT_MIP_STRONG_CHVATAL_GOMORY_CUTS       "mip_strong_chvatal_gomory_cuts"
 #define CUOPT_MIP_REDUCED_COST_STRENGTHENING       "mip_reduced_cost_strengthening"
+#define CUOPT_MIP_OBJECTIVE_STEP                   "mip_objective_step"
```
rg20, line 77: Do we need this setting?  Its not expensive to compute and it can only make things better.

cpp/src/mip_heuristics/problem/problem.cu
```diff
@@ -1404,6 +1392,387 @@ void problem_t<i_t, f_t>::recompute_objective_integrality()
...
+    // Try to find a scaling factor that makes all coefficients integral
+    std::vector<double> nonzero_coefs_double(nonzero_coefs.begin(), nonzero_coefs.end());
+    double scaling_factor = find_objective_scaling_factor(nonzero_coefs_double);
+    if (!std::isnan(scaling_factor)) {
+      std::vector<f_t> scaled_coefs;
+      for (f_t c : nonzero_coefs) {
+        scaled_coefs.push_back(std::round(static_cast<f_t>(scaling_factor) * c));
+      }
+      f_t g = gcd_of_integer_values(scaled_coefs);
+      if (g > 0) {
```
rg20, line 1471:
```suggestion
      if (g > 1) {
```

if g=1, its not useful right?

cpp/src/mip_heuristics/problem/problem.cu
chris-maes: Done.

cpp/src/mip_heuristics/problem/problem.cu
chris-maes: I moved the rational type to it's own class and add lots of comments and explaination.

cpp/include/cuopt/linear_programming/constants.h
chris-maes: It's always good to have the option to turn things off

cpp/src/mip_heuristics/problem/problem.cu
chris-maes: I think the g==1 case is redundant only if the bias is also zero.

cpp/src/cuts/objective_step.cpp
```diff
@@ -0,0 +1,411 @@
...
+  for (i_t i = 0; i < n_cons; ++i) {
+    bool lb_finite = std::isfinite(con_lb[i]);
+    bool ub_finite = std::isfinite(con_ub[i]);
+    cuopt_assert(lb_finite || ub_finite, "propagate_lattice: free constraints are not supported");
+
+    if (lb_finite && ub_finite && std::abs(con_lb[i] - con_ub[i]) < eq_tol) {
+      rhs[i] = rational128_t<f_t>::from_floating_point(con_lb[i]);
+    } else {
+      rhs[i] = rational128_t<f_t>::from_floating_point(ub_finite ? con_ub[i] : con_lb[i]);
+    }
```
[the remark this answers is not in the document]
chris-maes: This is already handled in the code via unknown singletons

ramakrishnap-nv (approved):

cpp/src/mip_heuristics/solver.cu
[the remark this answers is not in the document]
chris-maes: Sure

cpp/src/cuts/objective_step.cpp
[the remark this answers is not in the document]
chris-maes: Ah. This was removed in a refactor. And probably did not work in the first place. Thanks for catching this. I believe it is now fixed.

cpp/src/cuts/rational.hpp
```diff
@@ -0,0 +1,163 @@
...
+    // Ignore the success bool: even when rational_approximation reports the approximation
+    // is not exact within 1e-14, the out-parameters still hold the best convergent found.
+    // On the early-overflow path the out-parameters are left at our 0/1 default, which is a
+    // safe placeholder representing 0.
+    rational_approximation(x, max_d, num, den);
+    if (den <= 0) {
+      den = 1;
+      num = static_cast<int64_t>(std::llround(static_cast<double>(x)));
+    }
+    return {num, den};
```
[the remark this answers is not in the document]
chris-maes: This should be fixed now

cpp/src/mip_heuristics/problem/problem.cu
```diff
@@ -1406,6 +1412,116 @@ void problem_t<i_t, f_t>::recompute_objective_integrality()
...
+  objective_step = dual_simplex::compute_objective_step_info<i_t, f_t>(h_obj_coefs,
+                                                                       h_var_lb,
+                                                                       h_var_ub,
+                                                                       is_lattice_known_initially,
+                                                                       h_offsets,
+                                                                       h_vars,
+                                                                       h_coefs,
+                                                                       h_con_lb,
+                                                                       h_con_ub);
+  CUOPT_LOG_INFO("Objective step (step %e, bias %e) completed in %.3f seconds",
```
chris-maes: Remove after final test

PR #1247, opened 2026-05-19
Implement C APIs for adding quadratic objectives and constraints

rg20:
## Description

Adds a composable C API for building QP and QCQP models without growing the number of monolithic cuOptCreate* functions. Users create the linear problem with `cuOptCreateProblem` or `cuOptCreateRangedProblem`, then attach quadratic objectives and constraints incrementally.

Marks `cuOptCreateQuadraticProblem` and `cuOptCreateQuadraticRangedProblem` as deprecated (runtime warnings + docs) and migrates existing C API QP tests to the new workflow.

**New APIs**

- `cuOptAddQuadraticObjective`  Adds `xT Q x ` using coordinate (triplet) format (row_index, col_index, coeff). 
- `cuOptAddQuadraticConstraint` Appends one constraint `xT Q x + dT x <= / >= rhs`

Triplet input is converted to CSR internally before calling the optimization problem interface.

**Deprecation**
`cuOptCreateQuadraticProblem` and `cuOptCreateQuadraticRangedProblem` log `CUOPT_LOG_WARN` on use with migration guidance

## Checklist

- Testing
   - [x] Added tests
   - [ ] NA
- Documentation
   - [ ] Added new documentation
   - [ ] NA


cpp/src/pdlp/cuopt_c.cpp
```diff
@@ -82,6 +86,101 @@ solver_settings_handle_t* get_settings_handle(cuOptSolverSettings settings)
...
+    const cuopt_int_t col = key.second;
+    if (row < 0 || row >= num_rows) {
+      throw raft::exception("Quadratic matrix row index out of range");
+    }
+    row_entries[row].emplace_back(col, val);
+  }
+
+  for (cuopt_int_t row = 0; row < num_rows; ++row) {
+    auto& entries = row_entries[row];
+    std::sort(entries.begin(), entries.end());
```
chris-maes: Do we need to sort here? That can be expensive. I'm not sure we require the entries to be sorted at this point.

cpp/src/pdlp/cuopt_c.cpp
```diff
@@ -82,6 +86,101 @@ solver_settings_handle_t* get_settings_handle(cuOptSolverSettings settings)
...
+};
+
+void accumulate_csr_triplets(const std::vector<cuopt_int_t>& offsets,
+                             const std::vector<cuopt_int_t>& indices,
+                             const std::vector<cuopt_float_t>& values,
+                             std::map<triplet_key_t, cuopt_float_t, triplet_key_cmp_t>& triplets)
+{
+  if (offsets.size() < 2) { return; }
+  const cuopt_int_t num_rows = static_cast<cuopt_int_t>(offsets.size()) - 1;
+  for (cuopt_int_t row = 0; row < num_rows; ++row) {
```
chris-maes, line 111: If this is O(n) anyway, there are more efficient methods of converting COO to CSR that don't require using a map.

cpp/src/pdlp/cuopt_c.cpp
```diff
@@ -82,6 +86,101 @@ solver_settings_handle_t* get_settings_handle(cuOptSolverSettings settings)
...
+void coordinate_triplets_to_csr(cuopt_int_t num_entries,
+                                const cuopt_int_t* row_index,
+                                const cuopt_int_t* col_index,
+                                const cuopt_float_t* coeff,
+                                cuopt_int_t num_rows,
+                                std::vector<cuopt_int_t>& offsets,
+                                std::vector<cuopt_int_t>& indices,
+                                std::vector<cuopt_float_t>& values)
+{
+  std::map<triplet_key_t, cuopt_float_t, triplet_key_cmp_t> triplets;
```
chris-maes: I think this can be done more efficiently using an algorithm similar to the one below from:
https://github.com/DrTimothyAldenDavis/SuiteSparse/blob/dev/CSparse/Source/cs_compress.c

We can allocate a workspace w of size n inside cuOptOptimizationProblem to avoid the overhead of creating it each time. 

```
cs *cs_compress (const cs *T)
{
    csi m, n, nz, p, k, *Cp, *Ci, *w, *Ti, *Tj ;
    double *Cx, *Tx ;
    cs *C ;
    if (!CS_TRIPLET (T)) return (NULL) ;                /* check inputs */
    m = T->m ; n = T->n ; Ti = T->i ; Tj = T->p ; Tx = T->x ; nz = T->nz ;
    C = cs_spalloc (m, n, nz, Tx != NULL, 0) ;          /* allocate result */
    w = cs_calloc (n, sizeof (csi)) ;                   /* get workspace */
    if (!C || !w) return (cs_done (C, w, NULL, 0)) ;    /* out of memory */
    Cp = C->p ; Ci = C->i ; Cx = C->x ;
    for (k = 0 ; k < nz ; k++) w [Tj [k]]++ ;           /* column counts */
    cs_cumsum (Cp, w, n) ;                              /* column pointers */
    for (k = 0 ; k < nz ; k++)
    {
        Ci [p = w [Tj [k]]++] = Ti [k] ;    /* A(i,j) is the pth entry in C */
        if (Cx) Cx [p] = Tx [k] ;
    }
    return (cs_done (C, w, NULL, 1)) ;      /* success; free w and return C */
}
```

cpp/include/cuopt/linear_programming/cuopt_c.h
```diff
@@ -254,6 +257,9 @@ cuopt_int_t cuOptCreateRangedProblem(cuopt_int_t num_constraints,
                                      cuOptOptimizationProblem* problem_ptr);
 
 /** @brief Create an optimization problem of the form
+ *
+ * @deprecated Use ``cuOptCreateProblem`` to set up the linear problem, then
```
mlubin: Are we allowed to use `[[deprecated]]`? It's a C23 feature.

cpp/src/pdlp/cuopt_c.cpp
```diff
@@ -419,6 +522,132 @@ cuopt_int_t cuOptCreateQuadraticRangedProblem(
...
+    std::vector<cuopt_int_t> Q_indices;
+    std::vector<cuopt_float_t> Q_values;
+    coordinate_triplets_to_csr(
+      quad_num_entries, row_index, col_index, coeff, num_variables, Q_offsets, Q_indices, Q_values);
+    if (Q_offsets.empty()) { return CUOPT_INVALID_ARGUMENT; }
+
+    using quadratic_constraint_t =
+      optimization_problem_interface_t<cuopt_int_t, cuopt_float_t>::quadratic_constraint_t;
+
+    std::vector<quadratic_constraint_t> constraints(op_problem->get_quadratic_constraints());
```
mlubin: I think this is making a copy of all the quadratic constraints each time we add a new constraint.

cpp/include/cuopt/linear_programming/cuopt_c.h
rg20: This is customer facing API. Not sure if we want to add this restriction on the version of C.

cpp/src/pdlp/cuopt_c.cpp
rg20: Changed the implementation now

cpp/src/pdlp/cuopt_c.cpp
rg20: You are right, I will implement a new C++ API for adding  constraints

rg20: Fixed it

mlubin:
We should have test coverage for `cuOptAddQuadraticConstraint` including for unsupported constraint types like quadratic `>=` anything or general quadratic.

chris-maes:
> We should have test coverage for `cuOptAddQuadraticConstraint` including for unsupported constraint types like quadratic `>=` anything or general quadratic.

We support >=  for quadratic constraints. We should also support general convex quadratic constraints.

cpp/src/pdlp/cuopt_c.cpp
```diff
@@ -82,6 +83,85 @@ solver_settings_handle_t* get_settings_handle(cuOptSolverSettings settings)
...
+    perm[k] = static_cast<cuopt_int_t>(k);
+  }
+
+  const auto triplet_less = [row_index, col_index](cuopt_int_t a, cuopt_int_t b) {
+    const cuopt_int_t row_a = row_index[a];
+    const cuopt_int_t row_b = row_index[b];
+    if (row_a != row_b) { return row_a < row_b; }
+    return col_index[a] < col_index[b];
+  };
+  std::sort(perm.begin(), perm.end(), triplet_less);
```
chris-maes: This is O(nnz log nnz). That can be quite expensive. You can do a linear bucket sort here.

chris-maes (changes requested):
Let's get the complexity of COO to CSR down to O(n + nnz) before merging.

rg20:
> > We should have test coverage for `cuOptAddQuadraticConstraint` including for unsupported constraint types like quadratic `>=` anything or general quadratic.
> 
> We support >= for quadratic constraints. We should also support general convex quadratic constraints.

Can the barrier handle >= for quadratic constraints?  The constraints become non-convex right?

cpp/src/pdlp/cuopt_c.cpp
```diff
@@ -82,6 +83,94 @@ solver_settings_handle_t* get_settings_handle(cuOptSolverSettings settings)
...
+
+  // Group triplet indices by row (counting/bucket sort on row): O(nnz + num_rows).
+  std::vector<cuopt_int_t> perm(static_cast<size_t>(num_entries));
+  std::vector<cuopt_int_t> row_cursor(offsets.begin(), offsets.begin() + num_rows);
+  for (cuopt_int_t k = 0; k < num_entries; ++k) {
+    const cuopt_int_t row                        = row_index[k];
+    perm[static_cast<size_t>(row_cursor[row]++)] = k;
+  }
+
+  // Per row: sort by column, merge duplicates, emit CSR. Row grouping is O(nnz + num_rows);
```
chris-maes: Do we need sorted columns in the CSR downstream? 

If not, let's skip the sort. If so, the most efficient sort is to do a double transpose. Convert the CSR to CSC and then convert back to CSR.

rg20: If we don't sort, then we can't handle the duplicates.

cpp/src/pdlp/cuopt_c.cpp
```diff
@@ -82,6 +83,94 @@ solver_settings_handle_t* get_settings_handle(cuOptSolverSettings settings)
...
+    const cuopt_int_t end   = offsets[row + 1];
+    offsets[row]            = out_nnz;
+    if (start >= end) { continue; }
+
+    auto row_begin = perm.begin() + static_cast<ptrdiff_t>(start);
+    auto row_end   = perm.begin() + static_cast<ptrdiff_t>(end);
+    std::sort(row_begin, row_end, col_less);
+
+    cuopt_int_t last_col = -1;
+    for (auto it = row_begin; it != row_end; ++it) {
```
chris-maes: I think this could be simplified to just
```
cuopt_int_t last_col = -1
for (cuopt_int idx = start; idx < end; idx++) {
     const cuopt_int_t col = col_index[idx];
     // ...
```

cpp/src/pdlp/cuopt_c.cpp
```diff
@@ -82,6 +83,94 @@ solver_settings_handle_t* get_settings_handle(cuOptSolverSettings settings)
...
+      const cuopt_int_t idx = *it;
+      const cuopt_int_t col = col_index[idx];
+      if (it != row_begin && col == last_col) {
```
chris-maes: `if (idx != start && ... `

cpp/src/pdlp/cuopt_c.cpp
```diff
@@ -82,6 +83,94 @@ solver_settings_handle_t* get_settings_handle(cuOptSolverSettings settings)
...
+  cuopt_int_t out_nnz = 0;
+  for (cuopt_int_t row = 0; row < num_rows; ++row) {
+    const cuopt_int_t start = offsets[row];
+    const cuopt_int_t end   = offsets[row + 1];
+    offsets[row]            = out_nnz;
+    if (start >= end) { continue; }
+
+    auto row_begin = perm.begin() + static_cast<ptrdiff_t>(start);
+    auto row_end   = perm.begin() + static_cast<ptrdiff_t>(end);
+    std::sort(row_begin, row_end, col_less);
```
chris-maes: If you can avoid the sort, you can allocate a dense array of size n, then when looping over columns in a row, mark the column in the dense array. This allows you to detect duplicates. If you mark with the row number you do not need to clear the mark.

ramakrishnap-nv (approved):

cpp/include/cuopt/linear_programming/cuopt_c.h
```diff
@@ -417,6 +428,62 @@ cuopt_int_t cuOptCreateQuadraticRangedProblem(
   const cuopt_float_t* variable_upper_bounds,
   cuOptOptimizationProblem* problem_ptr);
 
+/** @brief Set the quadratic objective term x^T Q x on an existing problem.
+ *
+ * The matrix Q is specified in coordinate (triplet) format. This function may be called
+ * after ``cuOptCreateProblem`` or ``cuOptCreateRangedProblem`` to build a QP or QCQP model
+ * without using ``cuOptCreateQuadraticProblem`` or ``cuOptCreateQuadraticRangedProblem``.
+ * Each call replaces any previously set quadratic objective. Duplicate (row, col) indices
+ * in the triplet arrays are summed. Matrix Q must be symmetric positive semidefinite.
```
chris-maes: Q does not need to be symmetric right? 

Should we remove the requirement that it is positive semidefinite. We could maximize a negative semidefinite objective.

docs/cuopt/source/cuopt-c/lp-qp-milp/lp-qp-milp-c-api.rst
```diff
@@ -51,6 +51,17 @@ Optimization problems can be created via five different functions
 .. doxygenfunction:: cuOptCreateQuadraticProblem
 .. doxygenfunction:: cuOptCreateQuadraticRangedProblem
 
+.. note::
+   ``cuOptCreateQuadraticProblem`` and ``cuOptCreateQuadraticRangedProblem`` are deprecated.
+   Prefer ``cuOptCreateProblem`` or ``cuOptCreateRangedProblem`` followed by
+   ``cuOptAddQuadraticObjective`` (and ``cuOptAddQuadraticConstraint`` for QCQP).
+
```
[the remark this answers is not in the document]
chris-maes: The external doc site is generated via these .rst files that are checked into the repo.

chris-maes (approved):
LGTM. You should probably remove symmetric from the comment. 

I assume we will have to remove the coo_to_csr call when adding a quadratic constraint in the SOCP PR.

PR #1250, opened 2026-05-20
Fix CUDA Graph Capture/Sync race

akifcorduk:
run_concurrent in cpp/src/pdlp/solve.cu constructed the barrier raft::handle_t inside the spawned barrier std::thread, concurrently with PDLP's graph capture on the main thread. First-use init of cuBLAS / cuSPARSE / cuSolverDn / cuDSS triggers synchronous cudaMalloc / cudaDeviceSynchronize under the hood, which invalidates the in-flight capture. 

We now construct barrier_handle_ptr on the main thread. Init cudssCreate+cudssDestroy on the main thread to force the CUDA driver's lazy cuDSS module load to settle outside the capture window. 
std::promise<void> handshake so run_pdlp only starts after the barrier worker is scheduled. Promise is set unconditionally (including in the catch block) so the main thread cannot deadlock.

Closes #1185

PR #1252, opened 2026-05-20
Migrate concurrent root LP to OMP

akifcorduk:
run_concurrent in `cpp/src/pdlp/solve.cu `now dispatches the barrier and dual simplex workers as #pragma omp task inside a #pragma omp taskgroup instead of raw `std::thread`. PDLP still runs synchronously on the dispatching thread.

MIP path (omp_in_parallel()): reuses the upstream solve_mip OMP team. Barrier and dual simplex now consume slots from the configured `num_cpu_threads` budget instead of spawning extra OS threads outside it.
Stand-alone LP path: stands up a local `#pragma omp parallel + single` with the right number of workers.

Builds on #1250 , merge after that. This PR also removes the confuscated std::future logic on the barrier on other PR.


cpp/src/pdlp/solve.cu
```diff
@@ -1539,82 +1548,132 @@ optimization_problem_solution_t<i_t, f_t> run_concurrent(
...
-  // Create a thread for dual simplex
+  // Dual simplex / barrier results — written by tasks, read after the taskgroup barrier.
   std::unique_ptr<
     std::tuple<dual_simplex::lp_solution_t<i_t, f_t>, dual_simplex::lp_status_t, f_t, f_t, f_t>>
     sol_dual_simplex_ptr;
-  std::thread dual_simplex_thread;
   std::exception_ptr dual_simplex_exception;
+  // TODO: concurrent_halt is std::atomic<int>* across ~25 sites in dual_simplex / barrier /
+  // branch_and_bound / pdlp. Could be converted to int* + `#pragma omp atomic` (all-or-nothing,
+  // mixing OMP atomic with non-atomic on the same var is UB). Functionally equivalent today.
```
nguidotti: We can use  `omp_atomic_t` (a wrapper over the `pragma omp atomic` located at `utilities/omp_helpers.hpp`) instead `std::atomic`. I follow the style of the `std::atomic`, so they should be a drop-in replacement.

nguidotti (approved):
Looks good to me! Thanks for converting the remaining parts of the solver to OpenMP, Akif!

cpp/src/pdlp/solve.cu
akifcorduk: I wanted to keep it less invasive as it touches many files. The TODO is for not touching 25 files now. I can give it a try to see how invasive it is.

chris-maes: We have to be careful with changes to concurrent_halt. As it is also used inside cuDSS. So whatever changes we make need to be compatible with cuDSS.

akifcorduk: I think it should be equivalent to what we are doing with `std::atomic`version. But I am not sure if OMP runtime does late submission of memory requests when the other read is not atomic.  https://github.com/NVIDIA/cuopt/pull/1252/changes#diff-ff65045b18ca97b752289f52050884f6a91f1cd7b2dec84e2d30eaf539787f1dR290

What do you think @nguidotti ?

rg20:
@akifcorduk is this needed for 26.06?

akifcorduk:
@rg20 to fully obey thread count, it would be very beneficial. It was a small bug before as we don't obey the low thread count. Performance wise, functionality wise, it is not a big deal i think.

cpp/src/pdlp/solve.cu
nguidotti: For my understanding, the `pragma omp atomic` just signal for the compiler to use atomic instructions when reading/writing/updating that memory location. By default, OpenMP enforce a strong memory ordering (`seq_cst`). If you mix an atomic with a non-atomic operation and they happen at the same time, then you have a race condition.

nguidotti: In theory, `pragma omp atomic` and `std::atomic` should emit similar atomic instructions, but I would avoid to mix them up. My recommendation is to use whatever cuDSS is using for `concurrent_halt` to avoid potential incompatibilities.

nguidotti:
Replaced by #1291, which only includes the code changes related with the OpenMP migration.

PR #1254, opened 2026-05-20
Improve TSP with different vehicle start/end locations

hlinsen:
Improve routing support for TSP models that use `set_vehicle_locations()` with distinct start and
  end locations. These models are fixed paths, not closed tours, so some TSP-specific operators need
  to avoid cyclic assumptions around the start/end boundary.

 - Add path-aware EAX handling for fixed-start/fixed-end TSPs:
    - Detect TSPs with vehicle locations and no single depot.
    - Use a virtual depot during EAX recombination so endpoint depots can be normalized locally.
    - Keep normal single-depot behavior unchanged.

 - Add path-aware OX handling:
    - Use a linear offspring fill for fixed-path TSPs.
    - Avoid wrapping across the route end/start boundary, which is only valid for closed tours.
    - Fold the path logic into `fill_offspring(..., linear_path_tsp)` to reduce duplicate helper code.

 - Update sliding TSP local search
    - Allow insertion at the start-depot boundary not only `nodes_to_consider`
   
 - Add a unit test covering TSP with order locations plus vehicle start/end locations.


cpp/src/routing/local_search/sliding_tsp.cu
```diff
@@ -320,10 +325,13 @@ __global__ void execute_sliding_moves_tsp(
         NodeInfo<i_t>(original_fragment_end_node_id,
                       sol.problem.order_info.get_order_location(original_fragment_end_node_id),
                       node_type_t::DELIVERY);
+      auto start_depot_node_info = sol.problem.get_start_depot_node_info(route.get_vehicle_id());
       auto original_node_insertion_info =
-        NodeInfo<i_t>(original_node_insertion,
-                      sol.problem.order_info.get_order_location(original_node_insertion),
-                      node_type_t::DELIVERY);
+        original_node_insertion == start_depot_node_info.node()
```
rg20: The node value for the depot is undefined, I think

cpp/src/routing/crossovers/ox_recombiner.cuh
```diff
@@ -722,66 +722,105 @@ struct OX {
     }
   }
 
+  bool use_linear_ox_for_path_tsp(Solution const& S, int n_routes) const
+  {
+    if (!S.problem->is_tsp || n_routes != 1) { return false; }
+    if (S.problem->data_view_ptr->get_vehicle_locations().first == nullptr) { return false; }
```
rg20: What happens if the vehicle start location and end locations are same, but they are set by the user?

cpp/src/routing/crossovers/eax_recombiner.hpp
```diff
@@ -108,6 +109,55 @@ struct a_eax {
       a.clear();
   }
 
+  detail::NodeInfo<> normalize_depot(detail::NodeInfo<> node) const
+  {
+    if (use_virtual_depot_node && node.is_depot()) { return single_depot_node.value(); }
```
rg20, line 114: Can you explain the logic here?

cpp/src/routing/crossovers/ox_recombiner.cuh
hlinsen: The check is actually redundant. `get_single_depot` already abstracts all the logic.

cpp/src/routing/local_search/sliding_tsp.cu
hlinsen: It is set to 0 or order location size so this should be safe. https://github.com/NVIDIA/cuopt/blob/main/cpp/src/routing/problem/problem.cu#L411

cpp/src/routing/crossovers/eax_recombiner.hpp
hlinsen: We pick the parent A's start depot and set it as both start and end to generate the offspring. Then this is reverted when reconstructing the actual route. The goal is just to allow EAX to run.

The other approach is to add one artificial closing edge that has zero cost from end location to start location.

cpp/src/routing/local_search/sliding_tsp.cu
rg20: I think its cleaner to just look at the type.  That also means, original_node_insertion should be of NodeInfo type as well.

rg20 (approved):

PR #1260, opened 2026-05-20
python: add solve_batch aliases for LP and routing

Pritiks23:
## Why this change
Users naturally expect a snake_case batch API in Python (`solve_batch([...])`) that mirrors common ML serving patterns (single-call batch inference).

cuOpt already has batch solving support through `BatchSolve(...)`, but discoverability is lower and naming is inconsistent with common Python style.

This PR improves API ergonomics without changing solver behavior.

## What we changed
1. Added `solve_batch(...)` aliases for:
   - Linear programming solver
   - Routing solver
2. Exported the new alias from public module entry points so users can access it directly.
3. Kept `BatchSolve(...)` fully intact for backward compatibility.
4. Added tests that verify `solve_batch(...)` returns the same outcomes as `BatchSolve(...)`.
5. Updated routing API coverage documentation to include the new alias.

## What we did NOT change
- No solver-engine algorithm changes.
- No status/termination semantics changes.
- No removal or deprecation of `BatchSolve(...)`.

## User impact
Before:
- Users needed to discover and call `BatchSolve(...)`.

After:
- Users can call either `BatchSolve(...)` or the clearer Python-style `solve_batch(...)`.
- Existing code continues to work unchanged.

## Compatibility and risk
- Additive, low-risk API improvement.
- Existing integrations remain valid.
- New tests guard alias parity with existing batch path.

## Testing
- Added targeted unit tests for LP and routing alias parity.
- In this session, runtime execution of those tests was environment-gated because required runtime deps (`cuopt`/`cudf`) were unavailable in the shell; tests are included in the PR for CI validation.


mlubin:
The `BatchSolve` API is deprecated, see https://github.com/NVIDIA/cuopt/pull/915. We will not be providing new aliases for it.

PR #1269, opened 2026-05-21
ci: fix mangled python image tag on release/26.06

ramakrishnap-nv:
Three fixes for the `release/26.06` release-prep regression:

**1. Restore the python image tag** — `nightly-summary.yaml:63` and `build.yaml:315` had `python:26.06-slim`, which doesn't exist on Docker Hub. Reverted to `python:3.14-slim` (matches `main`). Unblocks the nightly-summary and build-summary jobs.

**2. Scope the version-bump regex** — `ci/release/update-version.sh:142` was matching any `:N.N-` in workflow files, which is what mangled the python image during release prep. Scoped to lines containing `rapidsai/` so unrelated images are left alone.

**3. Drop dead routing-example.ipynb sed call** — `ci/release/update-version.sh` was sed-ing a file that no longer exists, erroring on every release-prep run. The routing docs currently have no `main`-branch links so the line was a no-op even before the file was removed.

Verified by running `./ci/release/update-version.sh --run-context=release 99.99.00` in a throwaway worktree — only the four `rapidsai/ci-conda:NN.NN-latest` lines bump, `python:3.14-slim` is untouched, and no errors are emitted.

Same regex bug exists on `main`; this PR carries the fix on `release/26.06` so patch releases off this branch won't re-mangle. A separate cherry-pick to `main` is straightforward.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -312,7 +312,7 @@ jobs:
       contents: read
     runs-on: linux-amd64-cpu4
     container:
-      image: python:26.06-slim
+      image: python:3.14-slim
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
         with:
--- a/.github/workflows/nightly-summary.yaml
+++ b/.github/workflows/nightly-summary.yaml
@@ -60,7 +60,7 @@ jobs:
       contents: read
     runs-on: linux-amd64-cpu4
     container:
-      image: python:26.06-slim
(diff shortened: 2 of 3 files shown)


bdice (approved):
Please fix `ci/release/update-version.sh` in this PR. We may need to rerun that script as a part of manual forward-merges from `release/26.06`, so landing a fix for the script on `release/26.06` is needed.

bdice (changes requested):
Meant to request changes for the above, accidentally hit wrong button

bdice (approved):

PR #1284, opened 2026-05-21
Replace __constant__ FJ settings with per-instance device_scalar

### LINKED ISSUE #1277: [BUG] __constant__  fj_settings_t device_settings is mutated
Codex-discovered bug. This appears to be the cause of the crash in https://github.com/NVIDIA/cuopt/actions/runs/26165162216/job/76973927909?pr=1253#step:13:4379.

FJ stores `fj_settings_t` in CUDA constant memory via the translation-unit global `__constant__ fj_settings_t device_settings`, and `view_t.settings` points at that symbol.

That is invalid because FJ kernels mutate fields through `fj.settings`. In particular, `handle_local_minimum_kernel` can execute:

```cpp
fj.settings->n_of_minimums_for_exit *= 3;
fj.settings->baseline_objective_for_longer_run = *fj.incumbent_objective;
```

Those writes target CUDA constant memory. Device writes to constant memory are illegal, so when this local-minimum/improvement path is taken the kernel can trigger `cudaErrorIllegalAddress`. The error is asynchronous and usually surfaces later at a stream sync; subsequent RMM cleanup may then abort while seeing the sticky CUDA error, making the destructor path look like the crash site even though it is only a secondary failure.

There is a second correctness issue with the same storage: `device_settings` is process-global, not per `fj_t`. Concurrent FJ solver instances in the same process can overwrite each other’s settings before graph/kernel execution.

The fix should make settings storage instance-local and writable from device code, for example by adding an `rmm::device_scalar<fj_settings_t>` member to `fj_t`, pointing `view_t.settings` at that allocation, and copying host `settings` into that per-instance storage before launching FJ work.

mlubin:
- Fixes #1277. FJ kernels (`handle_local_minimum_kernel`) write through
    `view_t::settings`, but storage was a `__constant__` symbol — illegal
    on device, surfacing as `cudaErrorIllegalAddress` later at a sync.
  - Moves storage to a per-instance `rmm::device_scalar<fj_settings_t>`.
    Also closes the secondary issue of the shared symbol racing across
    concurrent `fj_t` instances.

### DIFF
--- a/cpp/src/mip_heuristics/feasibility_jump/feasibility_jump.cu
+++ b/cpp/src/mip_heuristics/feasibility_jump/feasibility_jump.cu
@@ -41,14 +41,13 @@ static constexpr int iterations_per_graph = 1;
 static constexpr int iterations_per_graph = 50;
 #endif
 
-__constant__ fj_settings_t device_settings;
-
 template <typename i_t, typename f_t>
 fj_t<i_t, f_t>::fj_t(mip_solver_context_t<i_t, f_t>& context_, fj_settings_t in_settings)
   : context(context_),
     pb_ptr(context.problem_ptr),
     handle_ptr(const_cast<raft::handle_t*>(pb_ptr->handle_ptr)),
     settings(in_settings),
+    device_settings(pb_ptr->handle_ptr->get_stream()),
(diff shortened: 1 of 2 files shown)


hlinsen (approved):
Thanks @mlubin! lgtm

anandhkb:
This is a P0 for 26.06

PR #1286, opened 2026-05-22
Stabilize logger config in LP BatchSolve and version logging

mlubin:
(Claude-driven bug fix, potentially fixing the crash in https://github.com/NVIDIA/cuopt/actions/runs/26249434727/job/77262219725?pr=1103. The issue seems rare enough that I wasn't able to reliably reproduce locally to confirm the fix.)

`init_logger_t` reconfigures the process-global `cuopt::default_logger()` sinks under `g_guard_mutex`, but `CUOPT_LOG_*` calls go through that same global logger without holding the mutex. Three risky entry points existed:

- `call_batch_solve()` ran an OpenMP loop where every worker constructed its own `init_logger_t` inside `solve_lp()`. When workers finished at different times the last guard's destructor reset sinks while siblings were still logging — a likely source of rare `pure virtual method called -> std::terminate -> SIGABRT` aborts.
- `solve_lp()` and `solve_qp()` called `print_version_info()` (which uses `CUOPT_LOG_*`) before constructing their `init_logger_t`, so version logging ran through whatever global sink configuration happened to be active at that moment.

Fix:

- Construct an outer `init_logger_t` at the top of `call_batch_solve()` before any `CUOPT_LOG_*` and before the OpenMP region. Worker-local `init_logger_t` instances now ref-count onto it and never tear down global sinks.
- Reorder `solve_lp()` and `solve_qp()` so `init_logger_t` is constructed before `print_version_info()`.

Verified by running `test_parser_and_batch_solver` 300x in one process and 2x300 in concurrent processes without any abort signature.

The changes to `solve_lp` and `solve_qp` are incidental and probably not related to the original crash. They're preventing a potential similar issue if `solve_lp` or `solve_qp` are called from multiple threads.

### DIFF
--- a/cpp/src/pdlp/solve.cu
+++ b/cpp/src/pdlp/solve.cu
@@ -1739,9 +1739,9 @@ optimization_problem_solution_t<i_t, f_t> solve_qp(optimization_problem_t<i_t, f
                                                    bool problem_checking)
 {
   try {
-    print_version_info();
     // Create log stream for file logging and add it to default logger
     init_logger_t log(settings.log_file, settings.log_to_console);
+    print_version_info();
 
     // Init libraries before to not include it in solve time
     init_handler(op_problem.get_handle_ptr());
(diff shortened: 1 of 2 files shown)


rg20:
FYI, @mlubin `call_batch_solve` is being deprecated in its current form.

rg20 (approved):

PR #1305, opened 2026-05-27
add mip start attribute

Iroy30:
## Description
This PR adds a per-variable attribute `MIPStart` that a user can set to specify the initial variable(primal) values for the problem.

## Checklist

- Testing
   - [x] Added tests
   - [ ] NA
- Documentation
   - [x] Added new documentation
   - [ ] NA


ramakrishnap-nv (approved):
@Iroy30 The changes look good, just had a suggestion, if we have a doc example for warmstart, may be we can update it with this example so it encompasses.

PR #1310, opened 2026-05-27
skills: prep for nvidia/skills catalog onboarding

ramakrishnap-nv:
Aligns this repo's skills with the NVIDIA Agent Skills Publishing Onboarding Guide.

- Add top-level `SECURITY.md` so the catalog's Getting Help & Contributing links resolve once cuopt is registered in `components.d/`.
- Shorten descriptions that previously tripped the "Description very long" check (above 150 chars). Trigger keywords preserved.


tmckayus (approved):

skills/cuopt-server-api-python/assets/lp_basic/client.py
```diff
@@ -37,18 +37,32 @@ def main():
...
+                4.0,
+                2.0,
+            ],
         },
         "constraint_bounds": {
-            "upper_bounds": [240.0, 200.0],
+            "upper_bounds": [
+                240.0,
+                200.0,
+            ],
```
jameslamb: Out of curiosity, where did all these style-only changes come from? I don't see any linter configs getting changed in this PR.

I recommend making sure these files are covered by `pre-commit`, so you don't have contributors' IDEs flipping the style of lines like this.

jameslamb (approved):
Looks great, thanks for adopting this `SECURITY.md`.

skills/cuopt-server-api-python/assets/lp_basic/client.py
ramakrishnap-nv: @jameslamb I have raised with internal team on the static analysis on these, so analysis flag these as coordinates and raises PII issue, even though this is just a numerical array.

PR #1313, opened 2026-05-27
Validate cuOptGetErrorString buffer size

### LINKED ISSUE #1311: [BUG] cuOptGetErrorString should reject non-positive buffer sizes
**Describe the bug**
While looking through the current `main` branch, I noticed that `cuOptGetErrorString()` checks `solution` and `error_string_ptr`, but it does not validate `error_string_size` before passing it to `std::snprintf(...)`.

That means `error_string_size == 0` still returns `CUOPT_SUCCESS`, and a negative `cuopt_int_t` size is forwarded straight into `snprintf`. In the same file, `cuOptGetParameter()` already rejects `parameter_value_size <= 0`, so this looks like a missed validation check rather than intentional API behavior.

1. Build cuOpt from source on current `main`.
2. Create or read any problem that produces a non-successful `cuOptSolution`.
3. Call `cuOptGetErrorString(solution, buf, 0)` or `cuOptGetErrorString(solution, buf, -1)`.

Minimal example:

```c
char buf[16];
cuopt_int_t st = cuOptGetErrorString(solution, buf, 0);
```

Current behavior: this goes through the API and returns `CUOPT_SUCCESS`.

Expected behavior: return `CUOPT_INVALID_ARGUMENT` for non-positive buffer sizes.

I first noticed this by source inspection in `cpp/src/pdlp/cuopt_c.cpp`: `cuOptGetErrorString()` passes `error_string_size` straight to `std::snprintf(...)`, while `cuOptGetParameter()` in the same file already has an early `<= 0` guard for its output buffer size.

**Expected behavior**
`cuOptGetErrorString()` should return `CUOPT_INVALID_ARGUMENT` when `error_string_size <= 0`, matching the validation pattern already used by `cuOptGetParameter()` and the rest of the C API.

 - Environment location: Bare-metal (local source checkout)
 - Method of cuOpt install: from source

**Additional context**
This one seems small but worth fixing because it is public C API surface, and the neighboring getter already follows the safer pattern.

I also checked the file history while looking at this: the missing size check was present when `cuOptGetErrorString()` was added, and it stayed that way through the later memory-model refactor, so this does not look like a branch-specific regression.

A minimal fix would just be an early:

```cpp
if (error_string_size <= 0) { return CUOPT_INVALID_ARGUMENT; }
```

right before the `std::snprintf(...)` call, plus a focused C API regression test.

fallintoplace:
This tightens one C API edge case and adds a small regression test for it.

What changed:
- reject non-positive buffer sizes in `cuOptGetErrorString`
- add a focused C API test that checks `0` and `-1` buffer sizes

Why:
`cuOptGetParameter` already rejects `<= 0` output buffer sizes, but `cuOptGetErrorString` was still passing its signed size straight into `snprintf`. This just makes the behavior consistent and keeps the validation on the API boundary.

Testing:
- `git diff --check`
- not run here: this machine does not have `nvcc` or a configured cuOpt build environment, so I could not build `libcuopt` / run `C_API_TEST` locally

Closes #1311.

### DIFF
--- a/cpp/include/cuopt/linear_programming/cuopt_c.h
+++ b/cpp/include/cuopt/linear_programming/cuopt_c.h
@@ -959,7 +959,7 @@ cuopt_int_t cuOptGetErrorStatus(cuOptSolution solution, cuopt_int_t* error_statu
  * @param[out] error_string_ptr - A pointer to a char that on output will contain the
  *  error string.
  *
- * @param[in] error_string_size - Size of the char buffer/
+ * @param[in] error_string_size - Size of the char buffer. Must be positive.
  *
  * @return A status code indicating success or failure.
  */
--- a/cpp/src/pdlp/cuopt_c.cpp
+++ b/cpp/src/pdlp/cuopt_c.cpp
@@ -1174,6 +1174,7 @@ cuopt_int_t cuOptGetErrorString(cuOptSolution solution,
 {
   if (solution == nullptr) { return CUOPT_INVALID_ARGUMENT; }
   if (error_string_ptr == nullptr) { return CUOPT_INVALID_ARGUMENT; }
+  if (error_string_size <= 0) { return CUOPT_INVALID_ARGUMENT; }
   solution_and_stream_view_t* solution_and_stream_view =
     static_cast<solution_and_stream_view_t*>(solution);
   std::string error_string = solution_and_stream_view->get_solution()->get_error_status().what();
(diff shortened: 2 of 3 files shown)


nguidotti:
/ok to test [bd69111](https://github.com/NVIDIA/cuopt/pull/1313/commits/bd6911172dcf1e32aedb585b9c27a25c3a2b7919)

PR #1316, opened 2026-05-28
Add extra numerical guards for flow cover cuts

aliceb-nv:
PR #1178 (flow cover cuts at root) added support for flow cover cuts at the root. On cbs-cta, this produced an invalid LP lower bound at the root, usually in the 1e+10 range while the real optimal is 0 - i.e. the cuts excluded all optimal solutions.

Disclaimer - the following explanation is AI-provided.

```
cbs-cta has rows whose continuous variables are bounded by big-M relations on binary controllers.
The variable-bound substitution step in build_single_node_flow_relaxation propagates that big-M into per-arc c·alpha shifts of magnitude ~1e7. Flow-cover construction then needs to subtract sums of these large opposite-signed quantities — the result is dominated by floating-point cancellation, even though every individual operation is mathematically valid.

Diagnostics-driven investigation identified three distinct cancellation sites in series:

single_node_flow_b = b − Σ(c·alpha) in build_single_node_flow_relaxation. Observed |snfb| / Σ|delta| ≈ 1.8e−16 (machine epsilon) on one row.
lhs_constant = Σ(multiplier · arc.y_const) in emit_flow_cover_cut. Observed 5.5e−6 ratio on another row.
cut.rhs = lhs_constant − single_node_flow_b — both inputs individually clean, but near-equal in magnitude. Observed 1.0e−6 ratio on a third row.
```

This PR adds appropriate guards for these cases.

The cuOpt skills were also updated by claude to reflect the learnings found here.

## Checklist

- Testing
   - [ ] Added tests
   - [ ] NA
- Documentation
   - [ ] Added new documentation
   - [ ] NA


skills/cuopt-developer/SKILL.md
```diff
@@ -240,3 +240,11 @@ When implementing or debugging **VRP dimensions** (constraints, objectives, forw
 - **`resources/vrp_skills.md`** — architecture contracts, required interfaces, and implementation checklist.
 
 Read it **before** adding a new dimension or changing combine semantics.
+
+## Numerical issues in solver internals
```
ramakrishnap-nv: Do we need to make this explicit for LP/MILP/QP so this doesn't get pulled into routing, may be we can mention in numerical optimization ?

skills/cuopt-developer/resources/numerical_debugging.md
```diff
@@ -0,0 +1,121 @@
+# Debugging Numerical Issues in Solver Internals
```
ramakrishnap-nv: May be use numerical optimization ?

skills/cuopt-developer/resources/numerical_debugging.md
```diff
@@ -0,0 +1,121 @@
...
+floating-point accumulator: `final = Σ(signed contributions)` collapses to a
+value many orders of magnitude smaller than its constituents, leaving the
+result dominated by floating-point noise.
+
+## Methodology — Instrument Before Patching
+
+The classical mistake is to guess the cancellation site and apply a fix. There
+are usually several candidates and you will guess wrong. Do this instead:
+
+1. **Locate the suspicious region.** Usually a recent commit or a code path
```
ramakrishnap-nv: INstead of numbering, use ###

skills/cuopt-developer/resources/numerical_debugging.md
```diff
@@ -0,0 +1,121 @@
...
+
+When in doubt, log the ratio *without* filtering first, observe the
+distribution across a representative benchmark, and place the threshold at
+least one order of magnitude below the cleanest "bad" case and at least one
+order of magnitude above the cleanest "good" case. Single-instance threshold
+choices tend to over-fit.
+
+## Cancellation Sites in Cut Generation
+
+Cut-generation routines (Gomory, MIR, complemented-MIR, flow-cover) are
```
ramakrishnap-nv, line 71: I was also thinking whether agent will be able to know Gomory, MIR and other jargans or there are README around it. But we can check how this spans out.

ramakrishnap-nv (commented):
@aliceb-nv Awesome addition to skill, I have few comments, but rest looks good.

skills/cuopt-developer/resources/numerical_debugging.md
aliceb-nv: In my experience agents are already familiar with such terms (or able to infer their meaning and research if necessary)

skills/cuopt-developer/SKILL.md
aliceb-nv: Good point :) Addressed

ramakrishnap-nv (approved):

cpp/src/cuts/cuts.cpp
```diff
@@ -1588,6 +1589,19 @@ bool flow_cover_generation_t<i_t, f_t>::emit_flow_cover_cut(
   cut.rhs = lhs_constant - single_node_flow_b;
   cut.sort();
 
+  // Small residual-RHS cuts are common and useful at normal scales. They become
+  // numerically dangerous when the RHS is the difference of two large internal
+  // SNF constants, because the added row can destabilize warm dual-simplex
```
chris-maes, line 1594: Nit: "warm dual-simplex reoptimization" should probably just be "dual-simplex reoptimization" or "warm-started dual-simplex reoptimization".

chris-maes (approved):
I'm not sure I understand exactly what is going on here. But the effect is just to disable flow cover cuts in certain situations. So it's fine to merge.

Thanks for the fix @aliceb-nv !

PR #1290, opened 2026-05-22
Extend the barrier solver for Second-Order Cone Programming (SOCP)

chris-maes:
Extend the barrier solver to handle second-order cone programming (SOCP) specified via quadratic constraints (x^T Q x + a^T x <= rhs).

Algorithmic changes:
- Implement Nesterov-Todd scaling for second-order cone (SOC)  blocks with Mehrotra predictor-corrector steps including cone corrector terms.
- Extend the augmented KKT system with dense SOC Hessian blocks and cone-aware step-length computation.
- Add Ruiz equilibration for improved conditioning of SOCP/QP problems.
- Convert quadratic constraints to canonical SOC form, supporting standard, rotated, and affine cone variants with automatic variable aliasing and bound splitting.

Public API:
- Python Modeling API: Extend Variable, Quadratic Expression, and Constraint to support quadratic constraints. 
- Python Data Model API: Add the functions add_quadratic_constraint, get_quadratic_constraints and clear_quadratic_constraints. 

Presolve:
- Preserve cone variable structure during presolve (skip cone block for bound tightening, variable/constraint removal).
- Use CSR row access for dual correction of bounded free variables in post-solve (replaces column scan).

Code quality and refactoring:
- Introduce coo_entries_t to replace std::tuple-based COO storage in parsers.
- Simplify triples_to_csr_flat and build_symmetric_q_coo.

Tests:
- Add unit tests for SOC kernels (NT scaling, step length, Hessian).
- Add integration tests for SOCP problems via the barrier solver.
- Add C API tests for quadratic constraints.

PR #1320, opened 2026-05-28
add QCQP support for grpc

tmckayus:
This change adds grpc transport for quadratic constraints, so that QCQP problems can be solved remotely.


tmckayus:
Note, this PR should merge after #1290 so that we can test a true end-to-end solve with grpc

yuwenchen95:
Better to change "QPQC" to "QCQP" in the title of PR.

tmckayus:
> Better to change "QPQC" to "QCQP" in the title of PR.

thank you!

cpp/src/grpc/grpc_solution_mapper.cpp
```diff
@@ -113,22 +113,6 @@ cpu_mip_solution_t<i_t, f_t> map_proto_to_mip_solution(
 #include "generated_proto_to_mip_solution.inc"
 }
```
tmckayus, line 114: Note, these were removed because they are dead code from an earlier version. The server does not need a size estimator for solutions, it now knows the size already.

cpp/tests/linear_programming/grpc/grpc_integration_test.cpp
```diff
@@ -1206,6 +1206,81 @@ TEST_F(ChunkedUploadTests, ConcurrentChunkedUploads)
   EXPECT_EQ(success_count.load(), num_clients);
 }
 
+// =============================================================================
```
tmckayus, line 1213: Note this comment/test will need to change after SOCP is merged so we can check the actual objective. Not crucial because we also have integration tests

ramakrishnap-nv (approved):

cpp/src/grpc/codegen/field_registry.yaml
```diff
@@ -915,3 +915,67 @@ optimization_problem:
...
+      setter: set_quadratic_constraints
+      scalars:
+      - constraint_row_index:
+          field_num: 1
+          type: int32
+          to_proto_cast: int32_t
+      - constraint_row_name:
+          field_num: 2
+          type: string
+      - constraint_row_type:
```
mlubin, line 956: "sense" or "constraint sense" is a more standard and specific name for this field

cpp/src/grpc/codegen/field_registry.yaml
```diff
@@ -915,3 +915,67 @@ optimization_problem:
     quadratic_objective:
       setter: set_quadratic_objective_matrix
       fields: [Q_values, Q_indices, Q_offsets]
+
+  # Quadratic constraints (QCQP).  Each entry is a row with its own RHS plus
+  # the row's linear coefficients (linear_values/linear_indices) and quadratic
+  # term (quadratic_values + quadratic_indices/offsets CSR per row-of-rows).
+  # Per-entry arrays are unbounded in size so they travel via container chunks
```
mlubin, line 932: It's worth a comment why this is different from linear constraints, i.e., we pass a matrix that encodes all linear constraints so we don't need to chunk within a repeated field there.

mlubin, line 932: I would have guessed that the chunking inside the quadratic constraints is premature given the scale of typical problems, but ok to keep this functionality now that we have it.

mlubin (approved):
Are we clearly documenting the extent to which the chunking protocol makes our grpc server special? I wouldn't want a user to try to roll their own client.

cpp/src/grpc/codegen/field_registry.yaml
tmckayus: @mlubin that could be. It was a pretty natural extension from the existing array chunking.

tmckayus:
> Are we clearly documenting the extent to which the chunking protocol makes our grpc server special? I wouldn't want a user to try to roll their own client.

Updating docs to note that the client/server message size limits can be configured. Our client uses a 256MB max message size for efficiency, but you COULD push the configured size limit closer to the 2GB grpc hard max to allow unary submission of larger problems if say you know your max problem size is going to be 1GB. In that case you don't have to implement chunking at all in a custom client.

However, for cases where chunking is required, working on making the documentation better where there are some gaps and noting that it's hard :) (and these updates can be iterative). To allow essentially unlimited problem is complicated by nature, but there is no reason why a determined user couldn't write a custom client with access to the proto definitions and the open-source implementation of the internal client.

cpp/src/grpc/codegen/field_registry.yaml
tmckayus: these fields model the internal C++ struct names directly. I'd suggest if we want to change it we change the internals too (at a future point) otherwise we're just introducing more overrides here.

cpp/src/grpc/codegen/field_registry.yaml
tmckayus: done, thanks

PR #1346, opened 2026-05-29
cuopt-server-api-python: catalog onboarding prep

ramakrishnap-nv:
## Summary
Work around NV-BASE PII false positives on inline numeric arrays in the `assets/lp_basic` and `assets/milp_basic` example clients. The validator's GPS-coordinate heuristic was firing on patterns like `{3.0, 4.0, 2.7, 10.1}` and `[40.0, 30.0]`. Reformatting each value on its own line bypasses the regex.

No behavior change; documentation/metadata only.


ramakrishnap-nv:
All content was PII validator workarounds (inline numeric array reformats). Dropping per request.

PR #1347, opened 2026-05-29
Remove CUDA whole compilation programming model violations

robertmaynard:
Previously we tried to launch kernels from files like `feasibility_jump` that are compiled into other files ( e.g. `feasibility_jump_kernels` ). This isn't allowed by the CUDA whole compilation programming model.

We remove the `static-global-template-stub=false` flag to enforce valid CUDA program going forward.


cpp/src/routing/ges/compute_fragment_ejections.cu
```diff
@@ -113,6 +113,55 @@ __global__ void kernel_get_best_insertion_ejection_solution(
...
+template <int BLOCK_SIZE, typename i_t, typename f_t, request_t REQUEST>
+void launch_kernel_get_best_insertion_ejection_solution(dim3 grid,
+                                                        dim3 blocks,
+                                                        size_t shmem_bytes,
+                                                        void** kernel_args,
+                                                        rmm::cuda_stream_view stream)
+{
+  RAFT_CUDA_TRY(cudaLaunchKernel(
+    (void*)kernel_get_best_insertion_ejection_solution<BLOCK_SIZE, i_t, f_t, REQUEST>,
+    grid,
```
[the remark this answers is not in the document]
robertmaynard: Not going to fix, matching the style of the rest of the cuOpt code base

bdice (approved):
Thanks @robertmaynard!

cpp/CMakeLists.txt
```diff
@@ -175,10 +175,6 @@ message("-- Building for GPU_ARCHS = '${CMAKE_CUDA_ARCHITECTURES}'")
 message("-- Host target architecture = '${CMAKE_SYSTEM_PROCESSOR}'")
 
 # make the flags global in order to propagate flags to test cmake files
-set(CMAKE_CUDA_FLAGS "${CMAKE_CUDA_FLAGS} --expt-relaxed-constexpr --expt-extended-lambda")
```
akifcorduk, line 178: Why do we also remove these experimental flags?

akifcorduk (approved):

cpp/CMakeLists.txt
robertmaynard: Removed by mistake. Restored

ramakrishnap-nv (approved):

PR #1352, opened 2026-06-01
B&B logs deduplication

nguidotti:
### Changes
- De-duplicate methods for reporting the progress of the MIP solver. Uses `std::format` to be easier to see the spacing for each column.
- Added support for the `std::format` syntax to the `logger_t`

## Checklist

- Testing
   - [ ] Added tests
   - [ ] NA
- Documentation
   - [ ] Added new documentation
   - [ ] NA


chris-maes:
Let's chat before merging

nguidotti:
> Let's chat before merging

Sure

cpp/src/dual_simplex/logger.hpp
```diff
@@ -16,6 +16,8 @@
 #include <cstdarg>
 #include <cstdio>
 #include <cstring>
+#include <format>
```
aliceb-nv, line 19: <format> is a very heavy heade in terms of compilation times in most c++ implementations. I'd be in favor of keeping it out of common headers and only use it in the TUs where it is required

cpp/src/branch_and_bound/search_tree.hpp
```diff
@@ -0,0 +1,181 @@
...
+                   "recursive unique_ptr destruction.\n",
+                   e.what());
+    } catch (...) {
+      std::fprintf(stderr,
+                   "search_tree_t destructor: iterative teardown failed (unknown exception); "
+                   "falling back to recursive unique_ptr destruction.\n");
+    }
+  }
+
+  void update(mip_node_t<i_t, f_t>* node_ptr, node_status_t status)
```
aliceb-nv: I'd be more in favor of letting it throw rather than trying to revert to recursive destruction. If iterate teardown fails I struggle to see why recursive destruction would fare any better, and more importantly it may turn a clear exception throw into a harder-to-debug stack overflow

nguidotti: I also find it strange to have a try-catch in the destructor, but it was a suggestion from #1230.

cpp/src/dual_simplex/logger.hpp
nguidotti: I did not notice a significant increase in compilation time by including or using the `std::format`. Maybe if it is used everywhere

cpp/src/branch_and_bound/search_tree.hpp
nguidotti: I also moved from a breadth-first to a depth-first approach, so the memory cost should be lower and `std::bad_alloc` should rarely happen

cpp/src/dual_simplex/logger.hpp
nguidotti: BTW, we can skip the compile-time checks of `std::format` with `std::vformat`. Although I would say the former is better.

cpp/src/branch_and_bound/search_tree.hpp
aliceb-nv: I think this is LLM paranoia :) It really wanted to avoid std::terminate, but in turns made things worse IMO. Worth discussing with the rest of the team but I'd much prefer a terminate than a callstack overflow. We already had to deal with these in the past

nguidotti: I agree, let me remove it then

cpp/src/branch_and_bound/branch_and_bound.cpp
```diff
@@ -326,29 +316,57 @@ void branch_and_bound_t<i_t, f_t>::set_initial_upper_bound(f_t bound)
...
+                  user_obj,
+                  user_lower,
+                  "",  // integer infeasible
+                  "",  // depth
+                  "",  // iter/node
+                  user_gap_text);
+
+    if (settings_.deterministic) { log_line += std::format("{:^8}", ""); }
+    log_line += std::format(" {:>8.2f}", toc(exploration_stats_.start_time));
+    settings_.log.printf("%s\n", log_line.c_str());
```
aliceb-nv: The exact field widths (12, 12, 19, 15, 8, 7, 11 ,11) remain duplicated across calls. Maybe we could have them be macro constants to keep a single source of truth

aliceb-nv (approved):
Thank you! Just a minor nitpick.

cpp/src/branch_and_bound/branch_and_bound.cpp
nguidotti: We could have a class dedicated to printing the table (since it also used in the other parts of the code) where each column has its own formatting rules (alignment, width, sign, etc.). This seems a little bit overkill, though hahaha

aliceb-nv: I was thinking of something like 

```

#define NODES_EXPL_WIDTH 12
#define NODES_UNEX_WIDTH 12
#define OBJ_WIDTH 19
/*  ... */

#define SECTION_HEADER_FORMAT "{:>" #NODES_EXPL_WIDTH "} {:>" #NODES_UNEX_WIDTH "} {:^+" #OBJ_WIDTH "} /* .... */
#define SECTION_ENTRY_FORMAT  "{:>" #NODES_EXPL_WIDTH "} {:>" #NODES_UNEX_WIDTH "} {:^+" #OBJ_WIDTH ".6e} /* .... */

// ....

std::format("H " SECTION_ENTRY_FORMAT,
                   values....);

```

Might be clearer in the actual business code, but it's kind of a matter of taste :)

aliceb-nv: Or, as Codex kindly just pointed out to me, do this!

```
// The {} inside the format specifier tells std::format to look at the arguments for the width
std::format("H {:>{}} {:>{}} {:^+{}.6e}",
            nodes_expl_val, NODES_EXPL_WIDTH,
            nodes_unex_val, NODES_UNEX_WIDTH,
            obj_val,        OBJ_WIDTH);
```

cpp/src/branch_and_bound/branch_and_bound.cpp
```diff
@@ -326,29 +329,77 @@ void branch_and_bound_t<i_t, f_t>::set_initial_upper_bound(f_t bound)
   upper_bound_ = bound;
 }
 
+template <typename i_t, typename f_t>
+void branch_and_bound_t<i_t, f_t>::print_table_header()
+{
+  std::string header =
+    std::format("{:^{}}|{:^{}}|{:^{}}|{:^{}}|{:^{}}|{:^{}}|{:^{}}|{:^{}}|{:^{}}|",
```
chris-maes, line 336: I find this significantly harder to read and understand than the printf format string. It's very easy to read a printf format string and understand the width and precision of each number printed out. I.e. %12.6e means to use a width of 12 with 6 digits of precision after the decimal. The following code is very difficult to read because I now must look up the width from the constants. Also it tells me nothing about the precision after the decimal point.

chris-maes (changes requested):
Let's discuss this more. I find the current changes harder to read and understand than the previous code.

cpp/src/branch_and_bound/branch_and_bound.cpp
nguidotti: I can revert the changes with the constants, so it easier to read. 

This specifically is just for the header of the table (so no precision). For the actual formatting for float, it is quite similar actually with the standard C formatting, e.g., `{:>12.6f}` is right-aligned, width 12 and precision 6. The main advantage here is to avoid manually setting the spacing for each log line and aligning to right/center/left. The syntax is the same that Python.

nguidotti:
I will unify all log changes into #1402

PR #1356, opened 2026-06-01
fix(ci): fix configuration for breaking change notification workflow

jameslamb:
Contributes to https://github.com/rapidsai/shared-workflows/issues/560

Fixes misconfigurations in the GitHub Actions workflow that generates Notifications when PRs are labeled `breaking`.

Any other changes come from making the configuration for this workflow identical across all RAPIDS repos.

### DIFF
--- a/.github/workflows/trigger-breaking-change-alert.yaml
+++ b/.github/workflows/trigger-breaking-change-alert.yaml
@@ -1,4 +1,4 @@
-# SPDX-FileCopyrightText: Copyright (c) 2021-2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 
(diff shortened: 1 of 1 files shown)


jakirkham (commented):
Thanks James! 🙏

.github/workflows/trigger-breaking-change-alert.yaml
```diff
@@ -1,4 +1,4 @@
-# SPDX-FileCopyrightText: Copyright (c) 2021-2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
jakirkham, line 1: Guessing this is just because this is when the repo was published

jameslamb: Yes I think that's right. This was done automatically by `pre-commit`.

Because all of the history was squashed when the project was open-sourced: https://github.com/NVIDIA/cuopt/commit/b833f7a1943a35dd75afcf3ef281c17618d00c6f

jakirkham (approved):

PR #1361, opened 2026-06-01
Add support for general convex quadratic constraints (x^T Q x + c^T x…

chris-maes:
Implement sparse LDL^T factorization with symmetric Markowitz pivoting and use it to translate general convex quadratic constraints into second-order cone form for the barrier solver.

Sparse LDL^T factorization: Add symmetric_trailing_matrix_t class that stores the lower triangle of a symmetric matrix with column storage, row index, diagonal cache, and a single degree structure exploiting symmetry. Add right_looking_ldlt function that computes PAP^T = LDL^T for symmetric PSD matrices using minimum- degree pivot selection with largest diagonal tiebreaker within each degree level. Returns an error code for indefinite (non-PSD) input.

General quadratic constraint translation: Add fallback path in translate_soc.hpp when Q doesn't fit existing special-case SOC patterns (non-uniform diagonals, multiple off-diagonals, nonzero RHS, etc.). Symmetrize Q from COO into H using a dense accumulator on the local variable subset. Factorize H via right_looking_ldlt and reject indefinite (non-convex) Q with a validation error. Construct a standard SOC with linking rows derived from the LDL^T factors. Properly invalidate affine-head allocations for constraints handled by the general path.

Add unit tests for LDL^T: diagonal, dense PD, rank-1 PSD, rank-2 PSD, zero matrix, scalar, 5x5 PD, rank-3 5x5 PSD, identity, graph Laplacian, symmetrized from unsymmetric, 10x10 sparse single entry, indefinite 2x2, indefinite 4x4.

Add unit tests for general quadratic translation:
dense PD with equality constraint, non-convex rejection, rank-deficient PSD, inequality constraint alongside quadratic constraint, least-squares with b in range of A (optimal residual zero), least-squares with b not in range of A (optimal residual positive).


cpp/src/barrier/translate_soc.hpp
```diff
@@ -247,6 +237,28 @@ void convert_quadratic_constraints_to_second_order_cones(
...
+      for (size_t k = 1; k < pos_diag_rows.size(); k++) {
+        const f_t scale =
+          std::max({f_t(1), std::abs(first_val), std::abs(pos_diag_rows[k].second)});
+        if (std::abs(pos_diag_rows[k].second - first_val) > tol * scale) {
+          has_nonuniform_diag = true;
+          break;
+        }
+      }
+    }
+    const bool use_general_path = has_duplicate_rows || has_near_zero_diag || has_nonzero_rhs ||
```
yuwenchen95, line 277: We should throw an error once `has_near_zero_diag` is true, since positive semidefinite matrix requires all diagonals to be nonnegative.

cpp/src/barrier/translate_soc.hpp
```diff
@@ -276,228 +288,488 @@ void convert_quadratic_constraints_to_second_order_cones(
...
+      cuopt_expects((qc.rhs_value < tol) && (qc.rhs_value > -tol),
+                    error_type_t::ValidationError,
+                    "SOC conversion currently requires rhs = 0 for quadratic constraints "
+                    "(constraint '%s' has rhs %.17g)",
+                    qc.constraint_row_name.c_str(),
+                    static_cast<double>(qc.rhs_value));
+
+      if (offdiag_entries.empty()) {
+        if (!has_linear_part) {
+          if (pos_diag_rows.empty()) {
```
yuwenchen95, line 323: Do we need this branching`if (pos_diag_rows.empty()) ` here? This corresponds to `−s*x^2 ≤0,rhs=0,s>0`, which is not we expect for second-order cone.

cpp/src/barrier/translate_soc.hpp
yuwenchen95, line 277: It's better to define `use_fast_path` since we have clear conditions for three special cases we want to address separately. Agents seem to generate additional branching for `!use_general_path` that doesn't belong to three special cases.

cpp/tests/socp/general_quadratic_test.cu
```diff
@@ -0,0 +1,712 @@
...
+
+  user_problem.rhs            = {0.0};
+  user_problem.row_sense      = {'E'};
+  user_problem.lower          = {-inf, -inf};
+  user_problem.upper          = {inf, inf};
+  user_problem.num_range_rows = 0;
+  user_problem.var_types.assign(n, variable_type_t::CONTINUOUS);
+
+  // Build quadratic constraint: x^T [2 1; 1 2] x <= 1
+  // Q in COO (lower triangle stored per MPS convention):
```
yuwenchen95: Why do we use only lower triangle part here? I remember MPS requires explicit full symmetric matrix representation for quadratic cost in a quadratic constraint, e.g. `x1*x2` and 'x2*x1'.

cpp/tests/socp/general_quadratic_test.cu
```diff
@@ -0,0 +1,712 @@
...
+  user_problem.var_types.assign(n, variable_type_t::CONTINUOUS);
+
+  // Q COO: (0,0,1), (1,0,4), (1,1,1) → H(0,0)=2, H(1,0)=4, H(1,1)=2
+  // Full H = [2 4; 4 2], eigenvalues 6 and -2 → indefinite
+  qc_t qc;
+  qc.constraint_row_index = 0;
+  qc.constraint_row_name  = "non_convex";
+  qc.constraint_row_type  = 'L';
+  qc.rhs_value            = 1.0;
+  qc.rows                 = {0, 1, 1};
```
yuwenchen95, line 193: The same concern about the full symmetric representation in MPS files as above.

cpp/tests/socp/general_quadratic_test.cu
```diff
@@ -0,0 +1,712 @@
...
+
+  // Q = [1 2; 2 1] gives H = [2 2; 2 2] (rank 1), rhs = 4
+  // (1/2)*x^T*[2,2;2,2]*x = (x0+x1)^2 <= 4
+  // With x0=x1=t: (2t)^2 <= 4 → t >= -1, obj = 2t = -2
+  qc_t qc;
+  qc.constraint_row_index = 0;
+  qc.constraint_row_name  = "rank1_cone";
+  qc.constraint_row_type  = 'L';
+  qc.rhs_value            = 4.0;
+  qc.rows                 = {0, 1, 1};
```
yuwenchen95, line 256: The same concern as above.

cpp/tests/socp/general_quadratic_test.cu
```diff
@@ -0,0 +1,712 @@
...
+  user_problem.var_types.assign(n, variable_type_t::CONTINUOUS);
+
+  // Q COO: (0,0,2), (1,0,1), (1,1,2) → H = [4,1;1,4]
+  // (1/2) x^T [4,1;1,4] x <= 1
+  qc_t qc;
+  qc.constraint_row_index = 0;
+  qc.constraint_row_name  = "ellipse_ineq";
+  qc.constraint_row_type  = 'L';
+  qc.rhs_value            = 1.0;
+  qc.rows                 = {0, 1, 1};
```
yuwenchen95, line 346: The same concern as above.

yuwenchen95 (changes requested):

cpp/src/barrier/translate_soc.hpp
chris-maes: near_zero is not negative right?

chris-maes: We are trying to do the fast path first, then the general path.

cpp/tests/socp/general_quadratic_test.cu
chris-maes: The factorization only takes in the lower triangular part. We use the factorization to determine if the matrix is PSD.

cpp/tests/socp/general_quadratic_test.cu
chris-maes: See answer above.

cpp/tests/socp/general_quadratic_test.cu
chris-maes: See answer above.

cpp/tests/socp/general_quadratic_test.cu
chris-maes: Oh sorry this in the test. Remember the C API and Python API are different from MPS. In C and Python we support unsymmetric matrices.

cpp/tests/socp/general_quadratic_test.cu
chris-maes: See answer above

cpp/src/barrier/translate_soc.hpp
chris-maes: Can you be more specific about "additional branching that doesn't belong to three special cases"

cpp/src/barrier/translate_soc.hpp
chris-maes: This is part of the previous code. That is, `if(pos_diag_rows.empty())` was part of the previous special case code.

cpp/src/barrier/translate_soc.hpp
chris-maes: In a PSD matrix you are allowed to have diagonals that are exactly zero.

cpp/tests/socp/general_quadratic_test.cu
chris-maes: Ah I see. The comment about MPS storing the lower triangluar part is wrong. I will fix that.

yuwenchen95 (approved):

ramakrishnap-nv (approved):

PR #1364, opened 2026-06-02
Expose diving hyper parameters

nguidotti:
Similar to #993, this PR exposes the diving hyper parameters to `cuopt_cli` and to the public API in order to be easier to tune. They are hidden by default in `cuopt_cli`. 

## Checklist

- Testing
   - [ ] Added tests
   - [ ] NA
- Documentation
   - [ ] Added new documentation
   - [ ] NA


cpp/src/branch_and_bound/branch_and_bound.cpp
```diff
@@ -1761,22 +1740,24 @@ void branch_and_bound_t<i_t, f_t>::best_first_search_with(bfs_worker_t<i_t, f_t>
     settings_.bnb_steal_chance >= 0 ? settings_.bnb_steal_chance : MIP_DEFAULT_STEAL_CHANCE;
   node_queue_t<i_t, f_t>& node_queue = worker->node_queue;
 
-  worker->calculate_num_diving_workers(bfs_worker_pool_.size(),
-                                       diving_worker_pool_.size(),
-                                       has_solver_space_incumbent(),
-                                       settings_.diving_settings);
+  mip_diving_hyper_params_t<i_t, f_t> diving_settings = settings_.diving_settings;
+  if (diving_settings.guided_diving != 0 && !has_solver_space_incumbent()) {
+    diving_settings.guided_diving = 0;
```
akifcorduk, line 1745: Does this mean we never do guided diving even if there is an incumbent later?

cpp/src/branch_and_bound/diving_heuristics.hpp
```diff
@@ -16,6 +18,44 @@
 
 namespace cuopt::linear_programming::dual_simplex {
 
+// When `log_diving_type` is true, each diving strategy gets its own letter;
+// otherwise every dive collapses to 'D'.
+inline char feasible_solution_symbol(search_strategy_t strategy, bool log_diving_type)
+{
+  if (strategy == BEST_FIRST) return 'B';
+  if (!log_diving_type) {
+    switch (strategy) {
```
akifcorduk: Why do we need those cases? We can just return "D" when log_diving_type is not set.

cpp/src/branch_and_bound/branch_and_bound.cpp
nguidotti: The guided diving is re-enabled as soon as we find an incumbent (it is within the loop below)

cpp/src/branch_and_bound/diving_heuristics.hpp
nguidotti: We don't. Let me remove those

akifcorduk (approved):

aliceb-nv (approved):

PR #1377, opened 2026-06-03
Fix for maximization problems with quadratic constraints

### LINKED ISSUE #1375: [BUG] MAXIMIZE silently ignored (returns minimizer) when a quadratic constraint is present
**Describe the bug**

When a problem has a **quadratic constraint**, the objective sense `MAXIMIZE` is
ignored — the solver returns the **minimizer** instead. This is a silent
wrong-answer (no error, status `Optimal`).

Found while writing general-convex-quadratic docs against `release/26.06`
(general convex quadratic support from #1361). `MAXIMIZE` works correctly for
LP/QP without quadratic constraints; it only misbehaves once a quadratic
constraint is present.

```python
import numpy as np
from cuopt.linear_programming.problem import Problem, MINIMIZE, MAXIMIZE

def solve(sense):
    p = Problem()
    x = p.addVariable(lb=-np.inf, name="x")
    y = p.addVariable(lb=-np.inf, name="y")
    p.addConstraint(x + y <= 10)
    p.addConstraint(2*x*x + 2*x*y + 2*y*y <= 6)   # quadratic constraint
    p.setObjective(x + y, sense=sense)
    p.solve()
    return p.ObjValue, x.Value, y.Value

print("MINIMIZE ->", solve(MINIMIZE))   # (-2.0, -1.0, -1.0)  correct
print("MAXIMIZE ->", solve(MAXIMIZE))   # (-2.0, -1.0, -1.0)  WRONG: expected (2.0, 1.0, 1.0)
```

Output:

```
MINIMIZE -> (-2.0, -1.0, -1.0)
MAXIMIZE -> (-2.0, -1.0, -1.0)
```

**Expected behavior**

`MAXIMIZE` of `x + y` over the ellipsoid `2x^2 + 2xy + 2y^2 <= 6` should return
`x = y = 1`, objective `+2`. (As a workaround, minimizing the negated objective
`-x - y` does return `(1, 1)` correctly, confirming the maximize path is the
problem.)

**Environment details:**
 - Environment location: Bare-metal
 - Method of cuOpt install: from source (`release/26.06`, cuOpt 26.06.00, includes #1361)

**Additional context**

Likely the barrier / second-order-cone conversion path does not apply the
maximize→minimize objective negation. Surfaced while documenting QCQP/SOCP
examples.

chris-maes:
The QCQP solve path translates the optimization_problem_t into the user_problem_t format consumed by Barrier via
cuopt_optimization_problem_to_user_problem(). Barrier always minimizes; it does not inspect obj_scale to determine optimization direction. In the non-QCQP paths the objective coefficients are already negated by convert_to_maximization_problem() before reaching Barrier, but in the QCQP path the optimization_problem_t is passed directly without that conversion, so the objective was never negated for maximization. This caused the solver to silently return the minimizer when maximize was requested.

Fix:
- Negate the objective coefficients in cuopt_optimization_problem_to_user_problem() when the problem is a maximization problem, so Barrier finds the maximizer.
- Set obj_scale = -objective_scaling_factor for maximization, so the reported objective value has the correct sign.

Add test_maximize_with_quadratic_constraint to verify both minimize and maximize give the correct optimal solution and objective value for a problem with a quadratic constraint.

Fixes #1375.

### DIFF
--- a/python/cuopt/cuopt/tests/socp/test_socp.py
+++ b/python/cuopt/cuopt/tests/socp/test_socp.py
@@ -196,3 +196,50 @@ def test_general_quadratic_unsymmetric():
     assert problem.ObjValue == pytest.approx(expected_obj, abs=OBJ_TOL)
     assert x0.Value == pytest.approx(expected_x, abs=PRIMAL_TOL)
     assert x1.Value == pytest.approx(expected_x, abs=PRIMAL_TOL)
+
+
+def test_maximize_with_quadratic_constraint():
+    """
+    Maximize x + y
+    s.t.  x + y <= 10
+          2*x^2 + 2*x*y + 2*y^2 <= 6
+
+    The quadratic constraint is the binding one.
+    With x = y = t: 2t^2 + 2t^2 + 2t^2 = 6t^2 <= 6 => t in [-1, 1].
+    Maximizing 2t gives t = 1, obj = 2.
+
+    Minimizing gives t = -1, obj = -2.
+
+    This test verifies that MAXIMIZE is respected when quadratic constraints
+    are present (regression for a bug where the QCQP path ignored the
+    objective sense).
+    """
+    from cuopt.linear_programming.problem import MAXIMIZE, MINIMIZE
+
+    # Solve as MINIMIZE first to establish baseline
+    prob_min = Problem("qc_maximize_min")
+    x = prob_min.addVariable(lb=-np.inf, name="x")
+    y = prob_min.addVariable(lb=-np.inf, name="y")
+    prob_min.addConstraint(x + y <= 10)
+    prob_min.addConstraint(2 * x * x + 2 * x * y + 2 * y * y <= 6)
+    prob_min.setObjective(x + y, sense=MINIMIZE)
+    _solve(prob_min)
+    _assert_feasible(prob_min)
+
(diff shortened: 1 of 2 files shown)


nguidotti (approved):

Kh4ster (approved):

ramakrishnap-nv (approved):

python/cuopt/cuopt/tests/socp/test_socp.py
```diff
@@ -196,3 +196,50 @@ def test_general_quadratic_unsymmetric():
     assert problem.ObjValue == pytest.approx(expected_obj, abs=OBJ_TOL)
     assert x0.Value == pytest.approx(expected_x, abs=PRIMAL_TOL)
     assert x1.Value == pytest.approx(expected_x, abs=PRIMAL_TOL)
+
+
+def test_maximize_with_quadratic_constraint():
```
mlubin, line 201: (Non-blocking) Should this test be in C++? You can use the LP reader to easily specify the instance, in case that was the motivation for using python.

chris-maes: The motivation for Python is that is the API where the bug report originated. Best to add the test in the originating API to make sure the bug is fixed.

PR #1355, opened 2026-06-01
Add cuopt-multi-objective-exploration skill (Pareto frontier)

cafzal:
## Description

Adds `cuopt-multi-objective-exploration` — a concept skill for problems with **two or more objectives and no fixed weighting**, where the user wants to see the tradeoff instead of accepting one weighted answer. It turns repeated single-objective cuOpt solves into a Pareto frontier (payoff table → ε-constraint / weighted-sum sweep → filter dominated) and supplies the discipline to read it: exchange rates, knee points, deferring the final choice. It adds no solver features and invents no API; it sits above the api-* and formulation skills, orchestrating the solves they already cover.

### cuOpt-specific correctness points
- ε-constraining is most natural on **linear** objectives, so a quadratic objective (risk `xᵀΣx`) can simply stay the objective. A convex one can also be ε-constrained directly: cuOpt routes `xᵀQx ≤ ε` through the barrier solver as a second-order cone.
- PDLP warm-start is LP-only; MILP frontier points are optimal **to the gap you set**, since each solve gets a time limit.
- A hard constraint (coverage floor, budget, fairness cap) is often a latent objective — promote it to a swept ε-constraint (when its level was an assumption, not a firm limit).

### User testing
Used the skill to enrich two existing examples and confirmed it drove the right calls — companion PR NVIDIA/cuopt-examples#151:
- **Workforce (MILP)** — promoted two hard constraints into tradeoffs (coverage → cost vs. coverage; the `max_shifts` cap → cost vs. fairness), chose ε-constraint over weighted-sum, anchored objective ranges, filtered dominated points, capped each solve, and reported no MILP duals.
- **Portfolio (QP)** — recognized the base notebook's hand-coded target-return loop as ε-constraint and added what it omits: the return-constraint dual (shadow price d(variance)/d(return)), with the PDLP-tolerance caveat.

The skill drives the *method*; runnable code still needs `cuopt-numerical-optimization-api-python` plus the worked notebooks, which are Colab-GPU validated (T4, clean end-to-end).

### Validation & gating
Registered in `AGENTS.md` + `marketplace.json`; `ci/utils/validate_skills.sh` passes. The contribution is `SKILL.md` + `evals/evals.json` — `BENCHMARK.md`, the skill card, and `skill.oms.sig` are generated by the NVSkills onboarding pipeline. The official NVSkills-Eval is the gate (PENDING).

## Checklist

[...]

PR #1393, opened 2026-06-05
Note dual/sensitivity support per problem type in formulation skill

cafzal:
## Description

Adds a per-problem-type **dual / sensitivity** capability note to the concepts-only `cuopt-numerical-optimization-formulation` skill, so the agent guides users correctly on what cuOpt exposes after a solve:

- **LP** — shadow prices (constraint duals) + reduced costs
- **MILP** — none (integer optima are not continuous)
- **QP** — shadow prices + reduced costs

Two edits: a `Duals / sensitivity` row in the LP/MILP/QP comparison table, and a short post-solve note giving the decision meaning — **shadow price = where to invest** (marginal objective gain from relaxing a binding constraint) and **reduced cost = near-miss** (how far a left-out option must improve before it enters the solution). As a concepts skill it states *what's supported per type* and *what it means*, and defers *how to read them* to the language-specific API skills (no API symbols inlined).

**Why.** Surfaced from real integration use — wiring solver-exact sensitivity/explainability into a downstream multi-objective decision layer. The formulation skill covered how to *formulate* but not which problem types yield sensitivity information, which is exactly what a user needs before relying on it (e.g. not expecting duals off a MILP).

The note also reflects the review feedback on #1355: a concepts skill carries no maturity ("beta") labels and no specific API symbols — both drift out of date, with no CI to catch the staleness.

**Validation & gating.** `ci/utils/validate_skills.sh` passes (skill structure, marketplace manifest, `AGENTS.md` references). The NVSkills-Eval pipeline — which also (re)generates `BENCHMARK.md`, the skill card, and the signature — is the gate and is **pending**: `skills/**` CI needs a maintainer and cannot be triggered from a fork branch.

**Related.**
- NVIDIA/cuopt-examples#154 — the diet LP duals example: a runnable worked case of the shadow-price / reduced-cost reading this note describes.
- Codebase-side sensitivity gaps (objective/RHS ranging; exposing variable/basis status) are tracked separately as #1394 and #1395.

## Checklist

- Testing
   - [ ] Added tests
   - [x] NA
- Documentation
   - [x] Added new documentation
   - [ ] NA

### DIFF
--- a/skills/cuopt-numerical-optimization-formulation/SKILL.md
+++ b/skills/cuopt-numerical-optimization-formulation/SKILL.md
@@ -32,8 +32,11 @@ Concepts and workflow for going from a problem description to a clear formulatio
 | Constraints | Linear | Linear | Linear (no quadratic constraints) |
 | Variables | Continuous | Mixed: continuous + integer/binary | Continuous |
 | Sense | min or max | min or max | **minimize only** (negate to max) |
+| Duals / sensitivity | Shadow prices + reduced costs | **None** (integer optima) | Shadow prices + reduced costs (beta) |
 
 If the objective is purely linear, prefer LP/MILP — do not artificially introduce quadratic terms. If any variable is integer or binary, the problem is MILP regardless of the rest.
+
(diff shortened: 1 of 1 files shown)


cafzal:
@rgsl888prabhu For visibility: Here's a minor skill addition to clarify dual/sensitivity support.

ramakrishnap-nv (approved):

PR #1400, opened 2026-06-06
Add algebraic __str__ and detailed __repr__ to Python LP API classes

jackthepunished:
Stacked on #1737; its commit appears here until it merges.

The Python LP modeling classes (Variable, LinearExpression, QuadraticExpression, Constraint, Problem) currently fall back to `<cuopt.linear_programming.problem.X object at 0x...>` when printed, which makes model construction hard to verify in notebooks and REPLs. This adds `__str__` (algebraic form, e.g. `2.0 * x + 3.0 * y <= 10.0`) and `__repr__` (detailed summary with bounds, type, variable/constraint counts and solve status) to all five classes. Expressions past 10 terms end in `... (N more terms)`. Purely additive; covered by `test_str_and_repr`, `test_problem_str_after_solve` and `test_str_truncation_large_expression`.

`Variable.__repr__` decodes byte-valued `VariableType` until #1736 is addressed.

### DIFF
--- a/python/cuopt/cuopt/linear_programming/problem.py
+++ b/python/cuopt/cuopt/linear_programming/problem.py
@@ -1341,7 +1341,12 @@ class Constraint:
             self.rhs_value = rhs_value
             self.RHS = rhs_value
             self.vindex_coeff_dict = {}
-            self.vars = expr.vars
+            # expr.vars holds only the linear terms; id() because Variable
+            # overrides __eq__ and is unhashable.
+            seen = {}
+            for var in (*expr.vars, *expr.qvars1, *expr.qvars2, *expr.qvars):
+                seen.setdefault(id(var), var)
+            self.vars = list(seen.values())
             return
 
         self.is_quadratic = False
@@ -1760,9 +1765,15 @@ class Problem:
                 )
             if isinstance(coeffs, dict):
                 coeffs = coeffs.items()
+            new_vars = []
             for var, coeff in coeffs:
                 idx = var.index
+                if idx not in constr.vindex_coeff_dict:
+                    new_vars.append(var)
                 constr.vindex_coeff_dict[idx] = coeff
+            if new_vars:
+                # constr.vars aliases the expression's list; rebind it.
+                constr.vars = constr.vars + new_vars
             if rhs is not None:
                 constr.RHS = rhs
         else:
--- a/python/cuopt/cuopt/linear_programming/problem.py
+++ b/python/cuopt/cuopt/linear_programming/problem.py
@@ -16,133 +16,6 @@ import cuopt.linear_programming.solver_settings as solver_settings
 import warnings
 
 
-# ---- Display helpers for __str__/__repr__ ----
-
-# Keyed by the underlying CType char codes ("L"/"G"/"E"). CType is a
-# ``(str, Enum)`` whose members compare and hash equal to these codes, so the
-# lookup works whether ``Constraint.Sense`` holds a CType member or a raw
-# string. Using the codes (rather than the LE/GE/EQ aliases) also keeps this
-# module-level table independent of definition order.
-_SENSE_SYMBOLS = {"L": "<=", "G": ">=", "E": "=="}
-_TYPE_NAMES = {"C": "CONTINUOUS", "I": "INTEGER", "S": "SEMI_CONTINUOUS"}
-
-# Maximum number of terms rendered when stringifying a linear or quadratic
-# expression. Beyond this, the head is shown followed by a ``... (N more
-# terms)`` marker so that printing a model with thousands of terms stays
-# readable in a REPL or notebook instead of flooding the output. Set to
-# ``None`` to disable truncation entirely.
-_MAX_DISPLAY_TERMS = 10
-
-
-def _var_display_name(var):
-    """Return the display name of a Variable."""
-    name = var.VariableName
-    if name:
-        return name
-    if getattr(var, "index", -1) >= 0:
-        return f"C{var.index}"
-    return f"V{id(var)}"
-
-
-def _type_display(type_val):
-    """Return a human-readable name for a variable type."""
-    if isinstance(type_val, VType):
-        return type_val.name
-    if isinstance(type_val, (bytes, bytearray)):
-        type_val = type_val.decode()
-    return _TYPE_NAMES.get(type_val, str(type_val))
-
-
-class _ExprBuilder:
-    """Build an algebraic string from a sequence of terms.
-
-    The first term is emitted without a sign; subsequent terms are joined
-    with ' + ' or ' - ' separators. A coefficient of 1.0 or -1.0 is
-    elided, so '1.0 * x' becomes 'x' and '-1.0 * x' becomes '-x'.
-
-    When ``max_terms`` is set, only the first ``max_terms`` non-zero terms
-    are rendered; any remaining terms are counted and summarized as a
-    trailing ``... (N more terms)`` marker. This keeps the output bounded
-    for expressions with very many terms. ``max_terms=None`` (the default)
-    renders every term.
-    """
-
-    def __init__(self, max_terms=None):
-        self.parts = []
-        self.max_terms = max_terms
-        # Non-zero terms seen so far (rendered + hidden).
-        self.n_terms = 0
-        # Non-zero terms omitted because the cap was reached.
-        self.n_hidden = 0
-
-    def add_linear(self, coef, var):
-        """Add a linear term ``coef * var``."""
-        if coef == 0.0:
-            return
-        var_str = _var_display_name(var)
-        if coef == 1.0:
-            self._append(var_str, negative=False)
-        elif coef == -1.0:
-            self._append(var_str, negative=True)
-        else:
-            self._append(f"{abs(coef)} * {var_str}", negative=coef < 0)
-
-    def add_quadratic(self, coef, var1, var2):
-        """Add a quadratic term ``coef * var1 * var2``."""
-        if coef == 0.0:
-            return
-        v1_str = _var_display_name(var1)
-        v2_str = _var_display_name(var2)
-        if v1_str == v2_str:
-            term_str = f"{v1_str}^2"
-        elif v1_str <= v2_str:
-            term_str = f"{v1_str} * {v2_str}"
-        else:
-            term_str = f"{v2_str} * {v1_str}"
-        if coef == 1.0:
-            self._append(term_str, negative=False)
-        elif coef == -1.0:
-            self._append(term_str, negative=True)
-        else:
-            self._append(f"{abs(coef)} * {term_str}", negative=coef < 0)
-
-    def add_constant(self, value):
-        """Add a constant term."""
-        if value == 0.0:
-            return
-        self._append(f"{abs(value)}", negative=value < 0)
-
-    def _append(self, term, negative):
-        self.n_terms += 1
-        if self.max_terms is not None and self.n_terms > self.max_terms:
-            # Past the cap: count the term but don't render it.
-            self.n_hidden += 1
-            return
-        if not self.parts:
-            self.parts.append(f"-{term}" if negative else term)
-        else:
-            self.parts.append(f" - {term}" if negative else f" + {term}")
-
-    def build(self):
-        if not self.parts and not self.n_hidden:
-            return "0.0"
-        result = "".join(self.parts)
-        if self.n_hidden:
-            plural = "term" if self.n_hidden == 1 else "terms"
-            marker = f"... ({self.n_hidden} more {plural})"
-            result = f"{result} + {marker}" if result else marker
-        return result
-
-
-def _format_linear(vars, coeffs, constant, max_terms=None):
-    """Format a linear expression as an algebraic string."""
-    builder = _ExprBuilder(max_terms=max_terms)
-    for var, coef in zip(vars, coeffs):
-        builder.add_linear(coef, var)
-    builder.add_constant(constant)
-    return builder.build()
-
-
 class VType(str, Enum):
     """
     The type of a variable is continuous, integer, or semi-continuous.
@@ -174,6 +47,11 @@ class CType(str, Enum):
     LE = "L"
     GE = "G"
     EQ = "E"
+
+    @property
+    def symbol(self):
+        """Algebraic symbol used when printing constraints."""
+        return {CType.LE: "<=", CType.GE: ">=", CType.EQ: "=="}[self]
 
 
 LE = CType.LE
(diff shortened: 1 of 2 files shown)


chris-maes:
This is interesting. Could you provide some example output, showing what strings this would compute, on a few different problems?

jackthepunished:
Here are some examples of what you'd see when printing or repr'ing objects after this change. On a small MILP with constraints like 2*x + 3*y <= 10, x - y >= 0, and x + 1 == 5, you get readable algebra instead of memory addresses: str(c1) gives 2.0 * x + 3.0 * y <= 10.0, str(c2) gives x - y >= 0.0, and str(c3) gives x + 1.0 == 5.0 (the last one keeps the constant on the LHS as the user wrote it, not the normalized internal form). Variables print as their name (str(x) → x) or as C{index} when unnamed (str(z) → C2), and repr(x) gives something like <cuopt.Variable 'x' (index=0), type=CONTINUOUS, bounds=[0.0, 10.0], value=nan>. For a QP, quadratic terms format naturally: str(x*x) → x^2, str(x*x + 2*x*y + 3*x) → x^2 + 2.0 * x * y + 3.0 * x, and mixed-sign expressions like -x*x + 0.5*y*y + x*y → -x^2 + 0.5 * y^2 + x * y. On the problem itself, str(prob) is a short summary with the problem name, objective sense, variable/constraint counts, non-zero count, and after solve the status and objective value, while repr(prob) stays compact, e.g. <cuopt.Problem 'str_repr_test' (3 vars, 3 constrs, IsMIP=True)>.

jackthepunished:
@chris-maes can you /review

ramakrishnap-nv:
@chris-maes May I get your review on this PR

mlubin:
I think this would be a nice addition. My main concern on the behavior side is there's no truncation in the case of large expressions. I doubt users want to see a linear or quadratic expression with 10,000 terms in it. If we can work out the right behavior in this case and validate it in the both REPL and notebooks then we'll be on track to merging this.

mlubin (changes requested):
Requesting the changes I mentioned above.

python/cuopt/cuopt/linear_programming/problem.py
```diff
@@ -16,6 +16,133 @@
...
+# ``None`` to disable truncation entirely.
+_MAX_DISPLAY_TERMS = 10
+
+
+def _var_display_name(var):
+    """Return the display name of a Variable."""
+    name = var.VariableName
+    if name:
+        return name
+    if getattr(var, "index", -1) >= 0:
```
mlubin: Why getattr instead of `var.getVariableIndex()`? When would the attribute not exist? And when would it be -1?

python/cuopt/cuopt/linear_programming/problem.py
```diff
@@ -16,6 +16,133 @@
 import warnings
 
 
+# ---- Display helpers for __str__/__repr__ ----
```
mlubin: The location of this code isn't very natural. Why are we defining the printing helpers before `VType`, `CType`, `Variable`, etc?

python/cuopt/cuopt/linear_programming/problem.py
```diff
@@ -16,6 +16,133 @@
...
+
+# Keyed by the underlying CType char codes ("L"/"G"/"E"). CType is a
+# ``(str, Enum)`` whose members compare and hash equal to these codes, so the
+# lookup works whether ``Constraint.Sense`` holds a CType member or a raw
+# string. Using the codes (rather than the LE/GE/EQ aliases) also keeps this
+# module-level table independent of definition order.
+_SENSE_SYMBOLS = {"L": "<=", "G": ">=", "E": "=="}
+_TYPE_NAMES = {"C": "CONTINUOUS", "I": "INTEGER", "S": "SEMI_CONTINUOUS"}
```
mlubin: Use `VType().name` rather than hardcoding this mapping.

python/cuopt/cuopt/linear_programming/problem.py
```diff
@@ -16,6 +16,133 @@
 import warnings
 
 
+# ---- Display helpers for __str__/__repr__ ----
+
+# Keyed by the underlying CType char codes ("L"/"G"/"E"). CType is a
+# ``(str, Enum)`` whose members compare and hash equal to these codes, so the
+# lookup works whether ``Constraint.Sense`` holds a CType member or a raw
+# string. Using the codes (rather than the LE/GE/EQ aliases) also keeps this
```
mlubin: Could this mapping be a member of the CType enum?

python/cuopt/cuopt/linear_programming/problem.py
```diff
@@ -335,6 +462,19 @@ def __eq__(self, other):
             case _:
                 raise ValueError("Unsupported operation")
 
+    def __str__(self):
+        return _var_display_name(self)
+
+    def __repr__(self):
+        name = _var_display_name(self)
+        idx = getattr(self, "index", -1)
```
mlubin: Again why getattr?

python/cuopt/cuopt/linear_programming/problem.py
```diff
@@ -1322,6 +1492,7 @@ def __init__(self, expr, sense, rhs, name=""):
         self.ConstraintName = name
         self.DualValue = float("nan")
         self.Slack = float("nan")
+        self._expr = expr
```
mlubin: Seems like we're potentially doubling the model's memory usage by storing an extra copy of the constraint data here. That's not ideal.

mlubin:
@jackthepunished do you plan to follow up on this PR?

jackthepunished:
Firstly I'm sorry to leave this PR without notice, but I was a companion for a family member at hospital so I couldn't give any attention to here, i'll get onto it asap.

python/cuopt/cuopt/linear_programming/problem.py
```diff
@@ -335,6 +340,123 @@ def __eq__(self, other):
             case _:
                 raise ValueError("Unsupported operation")
 
+    def __str__(self):
+        if self.VariableName:
+            return self.VariableName
+        if self.index >= 0:
+            return f"C{self.index}"
```
mlubin, line 348: Is this meant to be `f"V{self.index}"`?

python/cuopt/cuopt/linear_programming/problem.py
```diff
@@ -335,6 +340,123 @@ def __eq__(self, other):
...
+        # Not yet added to a problem: no name and no index to show.
+        return f"V{id(self)}"
```
mlubin: `id(self)` is a strange fallback, maybe just fall back to `__repr__`?

python/cuopt/cuopt/linear_programming/problem.py
```diff
@@ -335,6 +340,123 @@ def __eq__(self, other):
...
+
+    def __repr__(self):
+        vtype = self.VariableType
+        if isinstance(vtype, (bytes, bytearray)):
+            # The MPS data model yields variable types as byte codes.
```
mlubin: I don't know if this behavior of `VariableType` is intended, could you open an issue about this behavior?

python/cuopt/cuopt/linear_programming/problem.py
```diff
@@ -335,6 +340,123 @@ def __eq__(self, other):
...
+        if coef == 0.0:
+            return
+        v1_str = str(var1)
+        v2_str = str(var2)
+        if v1_str == v2_str:
+            term_str = f"{v1_str}^2"
+        elif v1_str <= v2_str:
+            term_str = f"{v1_str} * {v2_str}"
+        else:
+            term_str = f"{v2_str} * {v1_str}"
```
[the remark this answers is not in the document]
mlubin: Agreed, check on equality on the variables, no equality on the variable strings.

python/cuopt/cuopt/linear_programming/problem.py
```diff
@@ -1341,7 +1493,14 @@ def __init__(self, expr, sense, rhs, name=""):
             self.rhs_value = rhs_value
             self.RHS = rhs_value
             self.vindex_coeff_dict = {}
-            self.vars = expr.vars
+            # All participating variables (linear and quadratic terms), so
```
mlubin: I'm not sure I follow, but this seems like a pre-existing separate issue. If so it should be reviewed and fixed in a separate PR.

python/cuopt/cuopt/linear_programming/problem.py
```diff
@@ -1397,6 +1556,28 @@ def compute_slack(self):
...
+        if self.is_quadratic:
+            for row, col, val in zip(self.rows, self.cols, self.vals):
+                builder.add_quadratic(
+                    val, index_to_var[row], index_to_var[col]
+                )
+            for idx, val in zip(self.linear_indices, self.linear_values):
+                builder.add_linear(val, index_to_var[idx])
+        else:
+            for idx, coeff in self.vindex_coeff_dict.items():
+                builder.add_linear(coeff, index_to_var[idx])
```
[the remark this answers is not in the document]
mlubin: Please make sure this use case is tested

python/cuopt/cuopt/linear_programming/problem.py
jackthepunished: `C` is deliberate rather than a typo: `Problem._to_data_model` already names unnamed variables `C{index}` on export (and unnamed rows `R{index}`), following the MPS column/row convention, so `str(var)` lines up with what you see in an exported MPS file or in a solution. Happy to switch to `V{index}` if you'd rather printing be independent of the export naming, say which you prefer and I'll change it.

python/cuopt/cuopt/linear_programming/problem.py
jackthepunished: Agreed it's ugly. Falling back to `__repr__` unfortunately recurses, since `__repr__` interpolates `{self}`: `__str__` -> `__repr__` -> `__str__`. I've changed it to read like a default repr instead, `<Variable at 0x{id(self):x}>`, which keeps it unique per object so several detached variables in one expression don't collapse to the same token. Note this branch is only reachable by constructing `Variable()` directly rather than going through `Problem.addVariable`.

python/cuopt/cuopt/linear_programming/problem.py
jackthepunished: Will do. It does look unintended: `_to_data_model` builds `var_type` with `dtype="S1"`, and `_from_data_model` passes `dm.get_variable_types()[i]` straight into `addVariable(vtype=...)`, so a problem read from an MPS file ends up with `VariableType == b"I"` rather than `VType.INTEGER`. There's already a workaround for it in the `IsMIP` property (`if var.VariableType in ("I", "S", b"I", b"S")`). I'll open an issue with that trace and link it from here. The `decode()` has to stay in the meantime, since `VType(b"I")` raises `ValueError`.

python/cuopt/cuopt/linear_programming/problem.py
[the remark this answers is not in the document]
jackthepunished: Agreed, fixed. One wrinkle worth flagging: `==` isn't usable here because `Variable.__eq__` is overloaded to build a `Constraint` (which also leaves `Variable` unhashable), so the check compares by identity first and then by index when both variables are attached to a problem. The string comparison stays only for the cosmetic ordering of cross terms. Added a test with two distinct variables sharing the name `x`, asserting `x * x` rather than `x^2`.

python/cuopt/cuopt/linear_programming/problem.py
jackthepunished: You're right that it's pre-existing: on `main` the quadratic branch does `self.vars = expr.vars`, and a `QuadraticExpression`'s `vars` holds only the linear variables — so anything appearing solely in a quadratic term never lands in `Constraint.vars`. Nothing reads it for quadratic rows today, which is why printing is the first thing to trip over it.

Agreed on splitting it out: I'll open it as its own PR against `main` with tests and stack this one on top, so the data-model change gets reviewed on its own and this PR shrinks to just the printing once that lands. That also settles your earlier memory concern properly, since #1400 then carries no data-model changes at all. (For the record, the dedup only held pointers to already-live `Variable`s rather than a second copy of the coefficients; the `id()` keying is forced by `Variable` being unhashable, same reason as the equality thread above.)

python/cuopt/cuopt/linear_programming/problem.py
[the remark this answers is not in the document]
jackthepunished: Good catch — it turns out to be a genuine crash rather than only a missing test. `updateConstraint` writes into `vindex_coeff_dict` without touching `constr.vars`, so giving a coefficient to a variable that wasn't in the original expression leaves the row holding a column index it can't map back, and `str(constraint)` raises `KeyError`. The existing test only updates a variable that's already present, so it never reached this.

The real fix belongs with the `Constraint.vars` problem in the other thread, so it goes into that PR: `updateConstraint` now tracks newly referenced variables, rebinding rather than appending, since `Constraint.vars` aliases the list owned by the expression the constraint was built from and appending would leave a reused expression with more `vars` than `coefficients`. The same latent `KeyError` exists in `compute_slack` today, so that PR covers it with a test as well. On the printing side, the index lookup now falls back to the exported `C{idx}` name instead of raising, so `print(constraint)` can't crash even if a row is mutated directly.

Sorry again for the long gap here — I'm getting onto it asap and will push the split PR along with these fixes shortly.

python/cuopt/cuopt/linear_programming/problem.py
jackthepunished: Opened #1736. One correction to what I wrote above: `readMPS` yields plain `"I"` strings (the parser builds `var_types` with `dtype='str'`), not bytes; bytes only arrive when a caller passes them, and `addVariable`/`setVariableType` accept them unchecked. So the real problem is that `VariableType` isn't normalized to `VType` — three representations coexist and `IsMIP` / `solver.is_mip` each cope ad hoc. Details and repro in the issue.

jackthepunished:
@mlubin I've pushed the follow-ups. The `Constraint.vars` change is now its own PR (#1737) and this one is stacked on it, the `VariableType` thing is tracked in #1736, and the inline threads are handled in the last commit.

Also wanted to properly close the loop on your original point about large expressions, since I never answered it head-on. Right now an expression prints its first 10 non-zero terms and then `... (N more terms)`. `Problem.__str__` is just a summary, so it doesn't get slower on big models. This is what a 10,000-variable objective looks like:
```
>>> prob.getObjective()
<cuopt.LinearExpression: x0 + 2.0 * x1 + 3.0 * x2 + 4.0 * x3 + 5.0 * x4 + 6.0 * x5 + 7.0 * x6 + 8.0 * x7 + 9.0 * x8 + 10.0 * x9 + ... (9990 more terms)>
>>> print(prob)
Problem: big
  Objective: MINIMIZE
  Variables: 10000 (continuous=10000, integer=0, semi-continuous=0)
  Constraints: 1 (linear=1, quadratic=0)
  Non-zeros: 10000
```
Two things I'd rather ask than guess: is head-only with a cap of 10 fine, or would you prefer something like JuMP does (head and tail, omitted count in the middle, bigger cap)? And for notebooks, is plain text the same as the REPL good enough for this PR, with `_repr_latex_` later, or do you want that in here too? Either is a small change.

mlubin:
This PR is blocked on #1737, please address the pending comments there then we can land this!

PR #1402, opened 2026-06-08
MIP log cleanup

### LINKED ISSUE #1073: [FEA] Clean up last line in logs
This is difficult to read and contains redundant information (from what is printed in branch and bound). Unfortunately, some of our benchmark scripts currently depend on this log line.

nguidotti:
This PR cleans the logs for the MIP solver. More specifically,
- De-duplicate how each row of B&B table is constructed using `std::format`
- Extend support for `std::format` to the `logger_t`
- Add spacing between different sections of the solver (papilo, cuopt presolve and symmetry detection)
- Removed duplicated information (problem dimensions, B&B info)
- Replace last log line with a summary of the solve

Closes #1073.

## Examples
[feasible.log](https://github.com/user-attachments/files/28792523/feasible.log)
[infeasible.log](https://github.com/user-attachments/files/28792525/infeasible.log)
[optimal.log](https://github.com/user-attachments/files/28792526/optimal.log)
[timeout.log](https://github.com/user-attachments/files/28792527/timeout.log)

## Checklist

- Testing
   - [ ] Added tests
   - [ ] NA
- Documentation
   - [ ] Added new documentation
   - [ ] NA


cpp/src/mip_heuristics/presolve/third_party_presolve.cpp
```diff
@@ -682,12 +682,12 @@ third_party_presolve_result_t<i_t, f_t> third_party_presolve_t<i_t, f_t>::apply(
 
   papilo::Problem<f_t> papilo_problem = build_papilo_problem(op_problem, category, maximize_);
 
-  CUOPT_LOG_INFO("Original problem: %d constraints, %d variables, %d nonzeros",
-                 papilo_problem.getNRows(),
-                 papilo_problem.getNCols(),
-                 papilo_problem.getConstraintMatrix().getNnz());
+  CUOPT_LOG_DEBUG("Original problem: %d constraints, %d variables, %d nonzeros",
```
nguidotti, line 686: We already log the dimensions of the problems in another place. This is duplicated info

mlubin:
Is there an easy way to show a side-by-side of the key differences before and after to facilitate review?

nguidotti:
> Is there an easy way to show a side-by-side of the key differences before and after to facilitate review?

I can add diff of each log

nguidotti:
"old" logs (from the main branch):
[feasible_old.log](https://github.com/user-attachments/files/28804178/feasible_old.log)
[infeasible_old.log](https://github.com/user-attachments/files/28804180/infeasible_old.log)
[optimal_old.log](https://github.com/user-attachments/files/28804181/optimal_old.log)
[timeout_old.log](https://github.com/user-attachments/files/28804182/timeout_old.log)

Side-by-side:
[feasible_diff.log](https://github.com/user-attachments/files/28804197/feasible_diff.log)
[infeasible_diff.log](https://github.com/user-attachments/files/28804198/infeasible_diff.log)
[optimal_diff.log](https://github.com/user-attachments/files/28804199/optimal_diff.log)
[timeout_diff.log](https://github.com/user-attachments/files/28804200/timeout_diff.log)

edit: the spacing in the diff version make it harder to see. I think the best way is to open them side-by-side in your text editor

mlubin:
[logdiff.html](https://github.com/user-attachments/files/28856910/logdiff.html)

mlubin:
- The optimal case has a new extra blank line at the end.
- "The problem is integer infeasible." this is a bit jargon-y. Maybe "No feasible solution exists."?
- The timeout case has 4 blank lines between the iteration logs and the summary at the end. All other cases have 2.
- Should we consider using %g formatting for the summary to be a bit more human friendly?

nguidotti:
Here are the new logs:

[feasible.log](https://github.com/user-attachments/files/29102911/feasible.log)
[infeasible.log](https://github.com/user-attachments/files/29102913/infeasible.log)
[optimal.log](https://github.com/user-attachments/files/29102915/optimal.log)
[timeout.log](https://github.com/user-attachments/files/29102916/timeout.log)

mlubin:
Looks good to me. Chris should also approve. @chris-maes you can download and view this page to get a nice side-by-side diff:

[logdiff.html](https://github.com/user-attachments/files/29169580/logdiff.html)

mlubin (approved):
Log changes LGTM. Didn't review the code changes carefully.

cpp/src/branch_and_bound/branch_and_bound.cpp
```diff
@@ -302,29 +292,57 @@ void branch_and_bound_t<i_t, f_t>::set_initial_upper_bound(f_t bound)
   upper_bound_ = bound;
 }
 
+template <typename i_t, typename f_t>
+void branch_and_bound_t<i_t, f_t>::print_table_header()
+{
+  std::string header = std::format("{:^1}|{:^12}|{:^12}|{:^19}|{:^15}|{:^8}|{:^7}|{:^11}|{:^11}|",
```
akifcorduk, line 325: It seems we reuse the space counts often. Can we put them in variables like `log_col_1_length` etc.?

cpp/src/branch_and_bound/branch_and_bound.cpp
```diff
@@ -3410,9 +3387,10 @@ void branch_and_bound_t<i_t, f_t>::deterministic_sync_callback()
     exploration_stats_.last_log = tic();
   }
 
-  f_t obj              = compute_user_objective(original_lp_, upper_bound);
-  f_t user_lower       = compute_user_objective(original_lp_, lower_bound);
-  std::string gap_user = user_mip_gap<i_t, f_t>(original_lp_, upper_bound, lower_bound);
+  f_t obj                   = compute_user_objective(original_lp_, upper_bound);
+  f_t user_lower            = compute_user_objective(original_lp_, lower_bound);
+  f_t user_gap              = user_relative_gap(original_lp_, upper_bound, lower_bound);
```
akifcorduk: I think here you are applying the user objective transformation two times. `user_relative_gap` already converts the arguments to user objective. So this might produce wront results on certain scale and offsets.

cpp/src/branch_and_bound/symmetry.hpp
```diff
@@ -684,6 +684,8 @@ std::unique_ptr<mip_symmetry_t<i_t, f_t>> detect_symmetry(
   const simplex_solver_settings_t<i_t, f_t>& settings,
   bool& has_symmetry)
 {
+  settings.log.printf("\nRunning symmetry detection...\n");
```
akifcorduk, line 687: Why do we need this log? I think running ... logs are unnecessary.

cpp/src/dual_simplex/logger.hpp
```diff
@@ -84,6 +87,29 @@ class logger_t {
...
+        fflush(stdout);
+#endif
+      }
+      if (log_to_file && log_file != nullptr) {
+        std::fprintf(log_file, "%s", msg.c_str());
+        fflush(log_file);
+      }
+    }
+  }
+
```
akifcorduk, line 111: If all logs are ported. We can remove debug and printf now, so that we have a single logging mechanism.

akifcorduk (commented):
Thanks Nicolas. I have added few comments.

cpp/src/branch_and_bound/branch_and_bound.cpp
nguidotti: Yes, that was originally my intention, but @chris-maes did not like it: https://github.com/NVIDIA/cuopt/pull/1352#discussion_r3375353595

cpp/src/branch_and_bound/symmetry.hpp
nguidotti: Mostly for separating between parts of the solver. Otherwise, we will have only these lines:
```
Graph construction time 0.000131
Automorphism group size 1*10^0, 0 dejavu generators (0 move variables)
Dejavu time 0.000136
Projected 0 generators onto 210 binary variables (0 skipped non-binary), 0 stored
Total symmetry detection time 0.000306
No exploitable symmetry found (0 generators, 0 non-trivial orbits, max orbit size 0)
```

cpp/src/dual_simplex/logger.hpp
nguidotti: I kept the old log calls to avoid too many code changes. In terms of consistency, it would be best if we just use the new ones.

cpp/src/branch_and_bound/branch_and_bound.cpp
nguidotti: `upper/lower_bound` are in the solver space, while `obj` and `user_lower` are in the user space. So, in this case, it is correct. However, I could rename `obj -> user_obj` and refactor `user_relative_gap` to use the already converted values.

akifcorduk (approved):

cpp/src/branch_and_bound/symmetry.hpp
chris-maes: The "Running" logs provide valuable info on when a task is started. This allows users (and developers) to identify if a significant amount of time is spent in a particular task.

chris-maes (changes requested):
I need to do a detailed code review. But I see some issues just looking at the diff in the logs that Miles provided.

* We are only printing out the best objective / best bound to 3 decimal digits. We spend a lot of work computing 6 decimal digits. We should print 6 digits after the decimal for the objective and bound. 

* We lost information about whether we terminated due to relative MIP gap or absolute MIP gap

* Do we still print out info on variable lower / variable upper bounds if this computation takes more than a second?

I'll take a look at the code in detail later today.

nguidotti:
> I need to do a detailed code review. But I see some issues just looking at the diff in the logs that Miles provided.
> 
>     * We are only printing out the best objective / best bound to 3 decimal digits. We spend a lot of work computing 6 decimal digits. We should print 6 digits after the decimal for the objective and bound.
> 
>     * We lost information about whether we terminated due to relative MIP gap or absolute MIP gap
> 
>     * Do we still print out info on variable lower / variable upper bounds if this computation takes more than a second?
> 
> 
> I'll take a look at the code in detail later today.

1. I will increase the precision to `6`. Note that with `%g`, this is equal to 6 significant digits.
2. I did not change the log line when the optimal was found. In the example, it just happens to fall to the last case that prints without the gap info. 
3. What do you mean? I did not change how frequent the logs are printed in B&B.

chris-maes:
> 1. I will increase the precision to `6`. Note that with `%g`, this is equal to 6 significant digits.
> 2. I did not change the log line when the optimal was found. In the example, it just happens to fall to the last case that prints without the gap info.
> 3. What do you mean? I did not change how frequent the logs are printed in B&B.

1. Best to be explicit about the number of digits printed. %g is not the number of significant digits. It simply attempts to shorten the representation of the number when possible. 
2. We no longer see this line `Optimal solution found within absolute MIP gap tolerance (1.0e-10)`. I found it useful to know if we terminated from the absolute or relative gap. 
3. I no longer see the lines `35927 variable upper bounds in 0.00 seconds`. We should still print info about how long variable upper and lower bound calculations took, if they took longer than 1 second.

cpp/src/cuts/cuts.cpp
```diff
@@ -3763,7 +3765,7 @@ variable_bounds_t<i_t, f_t>::variable_bounds_t(const lp_problem_t<i_t, f_t>& lp,
     }
   }
   upper_offsets[lp.num_cols] = upper_edges;
-  settings.log.printf("%d variable upper bounds in %.2f seconds\n", upper_edges, toc(start_time));
+  settings.log.debug("%d variable upper bounds in %.2f seconds\n", upper_edges, toc(start_time));
```
chris-maes: We should still print this line, but we can gate it based on how long the computation took.

cpp/src/cuts/cuts.cpp
```diff
@@ -3854,7 +3856,7 @@ variable_bounds_t<i_t, f_t>::variable_bounds_t(const lp_problem_t<i_t, f_t>& lp,
     }
   }
   lower_offsets[lp.num_cols] = lower_edges;
-  settings.log.printf("%d variable lower bounds in %.2f seconds\n", lower_edges, toc(start_time));
+  settings.log.debug("%d variable lower bounds in %.2f seconds\n", lower_edges, toc(start_time));
```
chris-maes: We should still print this line. But we can gate it based on how long the computation took. I.e. don't print if it took less than 1 second.

cpp/src/mip_heuristics/solver_solution.cu
```diff
@@ -238,21 +238,39 @@ void mip_solution_t<i_t, f_t>::log_summary() const
...
-    max_constraint_violation_,
-    max_int_violation_,
-    max_variable_bound_violation_,
-    stats_.num_nodes,
-    stats_.num_simplex_iterations);
+  switch (termination_status_) {
+    case mip_termination_status_t::Optimal:
+    case mip_termination_status_t::FeasibleFound:
+      CUOPT_LOG_INFO("%s\n",
+                     std::format("Best objective {:.6g}, best bound {:.6g}, gap {:.2f}%.",
```
chris-maes: We should use .6e for the objective and bound. It's easy to get confused by g. I think it's worth it to always output 6 decimal digits in a consistent format.

cpp/src/mip_heuristics/solver_solution.cu
```diff
@@ -238,21 +238,39 @@ void mip_solution_t<i_t, f_t>::log_summary() const
...
+                                 objective_,
+                                 stats_.get_solution_bound(),
+                                 mip_gap_ * 100)
+                       .c_str());
+      break;
+
+    case mip_termination_status_t::Infeasible:
+      CUOPT_LOG_INFO("No integer feasible solution exists.\n");
```
chris-maes: Nit: I think this would read better as `Problem has no integer feasible solution`

cpp/src/branch_and_bound/branch_and_bound.cpp
```diff
@@ -185,34 +185,22 @@ f_t compute_user_abs_gap(const lp_problem_t<i_t, f_t>& lp, f_t obj_value, f_t lo
...
-    constexpr int BUFFER_LEN = 32;
-    char buffer[BUFFER_LEN];
-    if (user_mip_gap > 1e-3) {
-      snprintf(buffer, BUFFER_LEN - 1, "%5.1f%%", user_mip_gap * 100);
-    } else {
-      snprintf(buffer, BUFFER_LEN - 1, "%5.2f%%", user_mip_gap * 100);
-    }
-    return std::string(buffer);
-  }
+  if (value == std::numeric_limits<f_t>::infinity()) return "---";
```
chris-maes: I think it was better to have ` - ` than `---`. The previous version gave a little more whitespace in the table.

cpp/src/branch_and_bound/branch_and_bound.cpp
nguidotti, line 781: @chris-maes  The optimal log lines are still here. I did not change it, so they do not appear in the diff.

cpp/src/branch_and_bound/branch_and_bound.cpp
```diff
@@ -707,36 +713,33 @@ void branch_and_bound_t<i_t, f_t>::set_final_solution(mip_solution_t<i_t, f_t>&
...
-      (long long)exploration_stats_.lexical_reduction_nodes.load(),
-      (long long)exploration_stats_.lexical_reduction_fixings_applied.load(),
-      (long long)exploration_stats_.lexical_reduction_pruned_nodes.load());
+    settings_.log.print_format(
+      "Lexical reduction applied at {} nodes, {} total variable fixings, {} nodes pruned\n",
+      exploration_stats_.lexical_reduction_nodes.load(),
+      exploration_stats_.lexical_reduction_fixings_applied.load(),
+      exploration_stats_.lexical_reduction_pruned_nodes.load());
   }
-  settings_.log.printf("Absolute Gap %e Objective %.16e %s Bound %.16e\n",
```
chris-maes, line 764: Note in the absolute gap case (meaning the difference between the obj and bound was tiny) we where printing all 16 decimal digits. This was useful, because it allowed you to see the full difference between the objective and the bound.

cpp/src/branch_and_bound/branch_and_bound.cpp
```diff
@@ -2111,40 +2120,18 @@ lp_status_t branch_and_bound_t<i_t, f_t>::solve_root_relaxation(
...
 
-  settings_.log.printf("\n");
-  if (root_status == lp_status_t::OPTIMAL) {
-    settings_.log.printf("Root relaxation solution found in %d iterations and %.2fs by %s\n",
-                         iter,
-                         toc(start_time),
-                         solver_name.c_str());
-    settings_.log.printf("Root relaxation objective %+.8e\n", user_objective);
-  } else {
-    settings_.log.printf("Root relaxation returned: %s\n",
```
chris-maes, line 2172: It looks like we lost this print. It's useful to know what status the root relaxation returned when it could not be solved.

cpp/src/branch_and_bound/branch_and_bound.cpp
chris-maes: Can you point me to the line in the code where we print "within absolute MIP gap tolerance"?

nguidotti: https://github.com/nguidotti/cuopt/blob/7f0b1a3ccb0f47b6dab7a3e5f28c212433f5fc3b/cpp/src/branch_and_bound/branch_and_bound.cpp#L752

cpp/src/branch_and_bound/branch_and_bound.cpp
nguidotti: Do you want to also print the absolute gap in the last line?

cpp/src/branch_and_bound/branch_and_bound.cpp
chris-maes: Ah ok. I missed this. Thanks.

cpp/src/branch_and_bound/branch_and_bound.cpp
chris-maes: Let's leave it off for now and see how it looks.

chris-maes (approved):
Removing requested changes so this can be merged while I am out of office. Please address comments. Thank you!

cpp/src/branch_and_bound/branch_and_bound.cpp
nguidotti: This is already handle in the `solve` method.

ramakrishnap-nv (approved):

PR #1403, opened 2026-06-08
Fix error string buffer size validation

### LINKED ISSUE #1311: [BUG] cuOptGetErrorString should reject non-positive buffer sizes

Rohithmatham12:
## Summary
- Reject non-positive `error_string_size` in `cuOptGetErrorString` before calling `snprintf`.
- Add C API regression coverage for zero and negative buffer sizes using a valid CPU-backed solution handle.
- Clarify that the error string buffer size must be positive.

Fixes #1311.

## Testing
- `/opt/homebrew/opt/llvm/bin/clang-format -i cpp/src/pdlp/cuopt_c.cpp cpp/include/cuopt/linear_programming/cuopt_c.h cpp/tests/linear_programming/c_api_tests/c_api_tests.cpp`
- `git diff --check`

### DIFF
--- a/cpp/include/cuopt/linear_programming/cuopt_c.h
+++ b/cpp/include/cuopt/linear_programming/cuopt_c.h
@@ -959,7 +959,7 @@ cuopt_int_t cuOptGetErrorStatus(cuOptSolution solution, cuopt_int_t* error_statu
  * @param[out] error_string_ptr - A pointer to a char that on output will contain the
  *  error string.
  *
- * @param[in] error_string_size - Size of the char buffer/
+ * @param[in] error_string_size - Size of the char buffer. Must be positive.
  *
  * @return A status code indicating success or failure.
  */
--- a/cpp/src/pdlp/cuopt_c.cpp
+++ b/cpp/src/pdlp/cuopt_c.cpp
@@ -1174,6 +1174,7 @@ cuopt_int_t cuOptGetErrorString(cuOptSolution solution,
 {
   if (solution == nullptr) { return CUOPT_INVALID_ARGUMENT; }
   if (error_string_ptr == nullptr) { return CUOPT_INVALID_ARGUMENT; }
+  if (error_string_size <= 0) { return CUOPT_INVALID_ARGUMENT; }
   solution_and_stream_view_t* solution_and_stream_view =
     static_cast<solution_and_stream_view_t*>(solution);
   std::string error_string = solution_and_stream_view->get_solution()->get_error_status().what();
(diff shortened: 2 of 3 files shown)


nguidotti:
There is a similar fix in #1313. Although the test case here seems better.

nguidotti:
/ok to test [4284d91](https://github.com/NVIDIA/cuopt/pull/1403/commits/4284d91df01ad19b91b16b147278e6c84a29ae59)

Issue #904, opened 2026-02-24
[BUG] cuOpt VRP solver ignores min_vehicles constraint when fixed costs are included


bbozkaya:
When vehicle costs are included in the data VRP data model and `min_vehicles` constraint is set, cuOpt generates a solution with fewer than `min_vehicles` constraint value. Likely, this arises when `min_vehicles` value = total number of vehicles and fixed costs are included.

The expected behavior is for the solver to use no fewer than `min_vehicles` vehicles and pick the least costly vehicles as dictated by fixed costs.

The following example is where total num vehicles is 3, and they all have fixed costs + `min_vehicles` = 3, and the resulting solution uses only 1 vehicle.

`{ "cost_matrix_data": { "data": { "0": [ [0, 17, 12, 11, 10, 18, 10], [16, 0, 15, 11, 19, 15, 16], [19, 19, 0, 11, 16, 11, 17], [17, 19, 17, 0, 11, 18, 19], [10, 19, 19, 19, 0, 17, 15], [12, 18, 15, 18, 18, 0, 14], [12, 12, 11, 19, 10, 17, 0] ] } }, "fleet_data": { "vehicle_locations": [ [0, 0], [0, 0], [0, 0] ], "vehicle_fixed_costs": [10, 20, 30], "min_vehicles": 3 }, "task_data": { "task_locations": [1, 2, 3, 4, 5, 6, 0, 0, 0, 0, 0, 0], "demand": [[2, 1, 4, 2, 1, 4, -2, -1, -4, -2, -1, -4]] }, "solver_config": { "objectives": { "cost": 1, "vehicle_fixed_cost": 1 } } }`

When the `vehicle_fixed_cost` values are set to 0, the solution uses 3 vehicles, as required by `min_vehicles` setting.

anandhkb:
@rg20 for viz

anandhkb:
Attached is a minimum repro test case for triaging this bug.

[test_issue_904_min_vehicles_fixed_costs.py](https://github.com/user-attachments/files/26039381/test_issue_904_min_vehicles_fixed_costs.py)

anandhkb:
@np96 for viz

Hi @np96 Please take a look if this interests you! Thanks!

np96:
I'll check it.

anandhkb:
Thanks a ton @np96

np96:
I'll try to fix it in the next week, straight after dist breaks. Someone else can take it if it should be fixed ASAP.

[...]

PR #1405, opened 2026-06-08
bugfix (routing): fix min_vehicles bug

np96:
## Description
Fix min_vehicles ignored with vehicle fixed costs (#904)

When vehicle_fixed_costs were set and min_vehicles == fleet_size, the
solver returned fewer vehicles than min_vehicles and, in debug builds,
crashed with device-side asserts. Two coordinated issues:

1. move_candidates.cuh: move_path::reset() zeroed n_insertions via
   device_scalar::set_value_async(), which copies from a host stack local.
   reset() runs inside a captured CUDA graph, so that host source dangled
   by graph-launch time and left n_insertions garbage (~32k) — perform_moves
   then launched tens of thousands of out-of-bounds blocks. Use the
   memset-based set_value_to_zero_async(), which is capture-safe.

2. ox_recombiner.cuh: in fixed_route mode (fleet_size == min_vehicles), the
   VEHICLE_FIXED_COST branch enabled optimal-routes search, letting
   Bellman-Ford vary the route count. That contradicts the fixed vehicle
   count: it broke the recreate_solution invariant (routes removed == routes
   added) and drifted the solution below min_vehicles. Skip optimal-routes
   search when the count is fixed, and reject any count-mismatched offspring
   instead of applying it.

## Issue
#904

## Checklist

- Testing
   - [x] Added tests
   - [ ] NA
- Documentation
   - [ ] Added new documentation
   - [ ] NA


np96:
@hlinsen please take a look. Current main branch and all releases do have UB here:
https://github.com/NVIDIA/cuopt/pull/1405/changes#diff-bd4a0f52972e8aa5df38c9f1ceefd3fb3fd326d1d227f7116d41ad6cfa80fd26L126

Particulary, when I tested at large-scale instances with heterogenous fleet, this caused non-converging solver behavior. After the fix the solver started to converge correctly.

cpp/src/routing/solver.cu
```diff
@@ -65,9 +66,10 @@ assignment_t<i_t> solver_t<i_t, f_t>::solve()
   // TODO accept a settings object once we have full feature in ges solver
   // We only set target vehicles and use fixed route loop in the below case. The other paths will
   // run regular fixed route loop.
-  auto target_vehicles = -1;
-  if (data_view_ptr_->get_fleet_size() == data_view_ptr_->get_min_vehicles()) {
-    target_vehicles = data_view_ptr_->get_min_vehicles();
+  auto target_vehicles    = -1;
+  const auto min_vehicles = data_view_ptr_->get_min_vehicles();
+  if (min_vehicles > 0 && data_view_ptr_->get_fleet_size() >= min_vehicles) {
```
hlinsen: I believe we can revert this change. min_vehicle > 0 should be  checked in data_model and here we handle a special mode called "fixed route" that avoids route minimization paths so `data_view_ptr_->get_fleet_size() == min_vehicles` needs to hold. If `fleet_size > min_vehicle` we still want to use route minimization path.

cpp/src/routing/local_search/cycle_finder/cycle.hpp
```diff
@@ -54,12 +55,27 @@ struct ret_cycles_t {
...
+  /// Device view of the cycle output buffers.
   struct view_t {
-    DI void push_back(i_t val) { paths[offsets[*n_cycles_] + curr_cycle_size++] = val; }
+    /// Append a vertex to the in-progress cycle (`*n_cycles_`): writes at
+    /// `offsets[*n_cycles_] + curr_cycle_size` into the flat `paths` buffer and
+    /// bumps the per-cycle counter. Caller resets `curr_cycle_size` per cycle;
+    /// append_cycle() finalizes it.
+    DI void push_back(i_t val)
+    {
+      const auto write_pos = offsets[*n_cycles_] + curr_cycle_size++;
```
hlinsen: Is the purpose of the change simply to add asserts?

np96: Yes.
I believe it is good to fire assertions as early as possible.
With assertions enabled the corrupted async-copied stack value propagated downstream and the assertion happened only at [perform_moves.cu:78](https://github.com/NVIDIA/cuopt/blob/main/cpp/src/routing/local_search/perform_moves.cu#L78). If this assertion existed the origin of the problem would be bisected faster.

If you don't want this addition, I will revert it.

cpp/src/routing/solver.cu
np96: Oh, I see. Reverting.

Just to confirm my understanding, in this mode each vehicle must serve exactly 1 order. If there's a situation when such solution is impossible, the problem must be considered infeasible.

For me this looks unusual — this looks like an assignment problem solved via Hungarian algorithm: N*N complexity scoring, infeasible assignment gets large score, then polynomial weighted matching. If any part of the solution has large score, consider infeasible.

I ran the solver with this mode enabled and it didn't consolidate routes, just naive 1-to-1 assignment, which matches that interpretation.

I don't understand why VRP solver may be needed for this kind of problem. Guess I misunderstand something, and would be happy to be corrected.

np96: @hlinsen reverted this one.

cpp/src/routing/local_search/cycle_finder/cycle.hpp
hlinsen: We use spans so out of bounds should already trigger: https://github.com/NVIDIA/cuopt/blob/main/cpp/src/routing/local_search/cycle_finder/cycle.hpp#L69

np96: @hlinsen reverted.

np96:
@hlinsen both comments addressed, only necessary fixes present. Also edited PR message to match current state.

hlinsen (approved):

np96:
A kind reminder to consider this PR.
@kaatish @Iroy30
Thanks.

Iroy30 (approved):

PR #1409, opened 2026-06-09
Build and test with CUDA 13.3.0

bdice:
Contributes to https://github.com/rapidsai/build-planning/issues/286

* uses CUDA 13.3.0 to build and test
* updates to CUDA 13.3.0 devcontainers

## Notes for Reviewers

This switches GitHub Actions workflows to the `cuda-13.3.0` branch from here: https://github.com/rapidsai/shared-workflows/pull/574

A future round of PRs will revert that back to `main`, once all of RAPIDS is migrated.

### DIFF
--- a/cpp/src/pdlp/cusparse_view.cu
+++ b/cpp/src/pdlp/cusparse_view.cu
@@ -270,28 +270,51 @@ void my_cusparsespmm_preprocess(cusparseHandle_t handle,
 }
 #endif
 
-#if CUDA_VER_13_2_UP
-// SpMVOp symbols. Resolved at runtime via dlsym, because the runtime minor version might not match
-// the compiled minor version. We can go back to direct linking once CUDA 14 is adopted
-using cusparseSpMVOp_destroyDescr_sig = cusparse_sig<cusparseSpMVOpDescr_t>;
-using cusparseSpMVOp_destroyPlan_sig  = cusparse_sig<cusparseSpMVOpPlan_t>;
-using cusparseSpMVOp_bufferSize_sig   = cusparse_sig<cusparseHandle_t,
-                                                     cusparseOperation_t,
-                                                     cusparseSpMatDescr_t,
-                                                     cusparseDnVecDescr_t,
-                                                     cusparseDnVecDescr_t,
-                                                     cusparseDnVecDescr_t,
-                                                     cudaDataType,
-                                                     size_t*>;
-using cusparseSpMVOp_createDescr_sig  = cusparse_sig<cusparseHandle_t,
-                                                     cusparseSpMVOpDescr_t*,
-                                                     cusparseOperation_t,
-                                                     cusparseSpMatDescr_t,
-                                                     cusparseDnVecDescr_t,
-                                                     cusparseDnVecDescr_t,
-                                                     cusparseDnVecDescr_t,
-                                                     cudaDataType,
-                                                     void*>;
+#if CUOPT_CUSPARSE_VER_12_7_UP
+// SpMVOp symbols. Resolved at runtime via dlsym, because the runtime cuSPARSE version
+// might not match the headers used at compile time. cuSPARSE 12.7 corresponds to CUDA
+// Toolkit 13.2; cuSPARSE 12.8 corresponds to CUDA Toolkit 13.3. We can go back to
+// direct linking once CUDA 14 is adopted.
+using cusparseSpMVOp_destroyDescr_sig     = cusparse_sig<cusparseSpMVOpDescr_t>;
+using cusparseSpMVOp_destroyPlan_sig      = cusparse_sig<cusparseSpMVOpPlan_t>;
+using cusparseSpMVOp_bufferSize_12_7_sig  = cusparse_sig<cusparseHandle_t,
+                                                         cusparseOperation_t,
+                                                         cusparseSpMatDescr_t,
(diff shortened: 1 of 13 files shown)


ramakrishnap-nv (approved):

jameslamb (approved):

bdice:
This project needs some nontrivial C++ changes to support CUDA 13.3. I'll push those shortly.

cpp/src/pdlp/cusparse_view.cu
```diff
@@ -1360,35 +1428,38 @@ bool is_cusparse_runtime_mixed_precision_supported()
 
 bool is_cusparse_runtime_spmvop_supported()
 {
-#if CUDA_VER_13_2_UP
-  // Probe the runtimme to ensure cusparseSpMVOp is supported
+#if CUOPT_CUSPARSE_VER_12_7_UP
+#if !CUOPT_CUSPARSE_VER_12_8_UP
+  // Headers older than cuSPARSE 12.8 (CUDA Toolkit 13.3) cannot name the newer SpMVOp
+  // descriptor types, so do not call the older 12.7 signature against a 12.8+ runtime.
+  if (is_cusparse_runtime_12_8_or_newer()) { return false; }
```
mlubin, line 1435: Do I understand correctly that this function will return false if compiled with CUDA 13.2 and run on 13.3? And true if compiled with 13.3 and run on 13.2?

cpp/src/routing/ges/lexicographic_search/brute_force_lexico.cu
```diff
@@ -184,7 +184,7 @@ std::vector<i_t> guided_ejection_search_t<i_t, f_t, REQUEST>::brute_force_lexico
   auto stream          = sol.sol_handle->get_stream();
   i_t TPB              = 32;
   const i_t zero       = 0;
-  const auto value_max = std::numeric_limits<typename decltype(global_min_p_)::value_type>::max();
+  const auto value_max = std::numeric_limits<uint32_t>::max();
```
mlubin:
```suggestion
  const uint32_t value_max = std::numeric_limits<uint32_t>::max();
```
`auto` isn't doing anything here

cpp/src/routing/ges/lexicographic_search/lexicographic_search.cu
```diff
@@ -707,7 +707,7 @@ bool guided_ejection_search_t<i_t, f_t, REQUEST>::run_lexicographic_search(
   }
 
   // Init global min before call to lexicographic
-  const auto max = std::numeric_limits<typename decltype(global_min_p_)::value_type>::max();
+  const auto max = std::numeric_limits<uint32_t>::max();
```
mlubin:
```suggestion
  const uint32_t max = std::numeric_limits<uint32_t>::max();
```

cpp/src/pdlp/cusparse_view.cu
hlinsen: That's my understanding as well. Should we log here that SpMVOp is disabled to avoid performance surprises?

mlubin: I think we won't distribute binaries compiled on cuda 13.2 so it seems like a rare case to hit.

hlinsen: I agree it is rare

bdice: > Do I understand correctly that this function will return false if compiled with CUDA 13.2 and run on 13.3? And true if compiled with 13.3 and run on 13.2?
Yes, that is correct.

In theory since we are using `dlsym` I think we _could_ leave it all up to runtime information, and let builds from 13.2 use features from 13.3 (since we have already hardcoded the signature of the newer library...) but that felt a bit strange to me and I didn't want to propose that as a tested, supported case.

Bubullzz:
What do we think about simply removing support for SpMVOp in cusparse 12.7 and only support it in 12.8 ? The material implications this have will be that the ~15% speedup will be lost on cuda 13.2 but gained back on cuda 13.3

This would make this versionning juggle inexistent and less error prone.

mlubin:
>  The material implications this have will be that the ~15% speedup will be lost on cuda 13.2 but gained back on cuda 13.3

IMO it's worth a bit of logic for the 15% speedup on cuda 13.2 runtimes.

Bubullzz:
It seems very dangerous and crumblesome, only for an edge case that might never happen. Our customers that need the fastest, bleeding edge Cuopt will be using 13.3 anyways, right ?

mlubin:
> Our customers that need the fastest, bleeding edge Cuopt will be using 13.3 anyways, right ?

Some NVIDIA environments including brev still have containers with 12.9, so you may be assuming too much :)

Bubullzz:
Then this also means having a CI ensuring 13.2 does not break, in addition to 13.3.

mlubin:
I don't think CI coverage is necessary but some manual testing is called for. @bdice is there any way we could test running the 13.3 compiled version on a 13.2 runtime? If not now then blocking before this code is released.

hlinsen (approved):

bdice:
@mlubin Thanks for the review and help with merging. If you ever do need explicit jobs for 13.2 to be added to your CI matrix, you can do something like this: https://gist.github.com/bdice/ea5b000be16ac4a8b3e19d984048840f

PR #1410, opened 2026-06-09
Use LP format reader throughout unit tests

mlubin:
Specify test problems in LP format instead of manually constructing the data structures. This is a big win for readability


aliceb-nv (approved):
Outstanding stuff :) LGTM, thanks a lot Miles, this is going to make writing tests much easier

PR #1413, opened 2026-06-09
Fix concurrent root halt on infeasible MILP

### LINKED ISSUE #1396: [BUG] std::terminate (process abort) on an infeasible MILP in opportunistic (default) mode

Rohithmatham12:
Summary
- Signal the root concurrent halt flag when the root dual-simplex task returns a terminal LP status in concurrent root solve.
- Leave CONCURRENT_LIMIT unchanged, since that means another concurrent solver already requested the stop.
- Add a cuopt_cli regression that generates an infeasible 120-binary MPS and checks default opportunistic mode returns Infeasible instead of aborting.

Why
- #1396 reproduces as a default opportunistic cuopt_cli abort on an infeasible MILP, while deterministic mode returns Infeasible.
- The concurrent root path only signaled root_concurrent_halt_ on the crossover-OPTIMAL path. When dual simplex proved infeasibility before crossover produced a solution, the concurrent PDLP/barrier/cuDSS side could remain active into teardown.
- This matches the existing standalone concurrent LP pattern, where terminal dual-simplex/barrier statuses signal concurrent_halt so peer solvers stop cleanly.

Testing
- /opt/homebrew/opt/llvm/bin/clang-format -i cpp/src/branch_and_bound/branch_and_bound.cpp
- bash -n python/libcuopt/libcuopt/tests/test_cli.sh
- git diff --check

Fixes #1396

### DIFF
--- a/cpp/src/branch_and_bound/branch_and_bound.cpp
+++ b/cpp/src/branch_and_bound/branch_and_bound.cpp
@@ -1981,6 +1981,15 @@ lp_status_t branch_and_bound_t<i_t, f_t>::solve_root_relaxation(
   std::string solver_name = "";
 
   lp_status_t root_status;
+  auto request_root_concurrent_halt = [this](lp_status_t status) {
+    if (status == lp_status_t::OPTIMAL || status == lp_status_t::INFEASIBLE ||
+        status == lp_status_t::UNBOUNDED || status == lp_status_t::UNBOUNDED_OR_INFEASIBLE ||
+        status == lp_status_t::ITERATION_LIMIT || status == lp_status_t::TIME_LIMIT ||
+        status == lp_status_t::NUMERICAL_ISSUES || status == lp_status_t::CUTOFF ||
+        status == lp_status_t::WORK_LIMIT) {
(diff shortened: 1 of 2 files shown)


Rohithmatham12:
Hi maintainers, the label checker is waiting for category/breaking labels. Suggested labels for this PR: bug, non-breaking, mip.

This is a fresh, narrow root-cause PR for #1396: it only changes concurrent root-solve halt coordination and adds a default-opportunistic cuopt_cli regression for the infeasible MPS crash path.

mlubin:
Closing in favor of https://github.com/NVIDIA/cuopt/pull/1416 which better addresses the root cause.

PR #1416, opened 2026-06-10
Fix PDLP capture cublas error and hang on infeasible solutions

### LINKED ISSUE #1396: [BUG] std::terminate (process abort) on an infeasible MILP in opportunistic (default) mode

akifcorduk:
## Summary

Fixes two concurrency bugs in the MIP concurrent root-LP path that could crash or
hang the solver: (1) an intermittent `CUBLAS_STATUS_INTERNAL_ERROR` during PDLP
graph capture, and (2) a deadlock when the concurrent root LP fails to produce a
usable solution (e.g. on an infeasible model). Both are exercised by the same
flow — the heuristics' concurrent root LP (PDLP + cuDSS barrier) running alongside
B&B — and both reproduce independently of each other.

### 1. cuBLAS error during graph capture
PDLP captures its adaptive step into a CUDA graph (`manual_cuda_graph_t`,
`cudaStreamCaptureModeThreadLocal`) while the cuDSS barrier runs concurrently on a
separate handle. cuDSS's device-synchronizing, library-global work (handle
create/first factorization/destroy, allocations) invalidates PDLP's in-flight
capture. When that invalidation is observed *inside* a cuBLAS call (cuBLAS can't
return a CUDA error code), it surfaces as `CUBLAS_STATUS_INTERNAL_ERROR` thrown
from `cublasdot` in `adaptive_step_size_strategy.cu`.

`manual_cuda_graph_t::run` already recovered from this collision, but **only** when
`cudaStreamEndCapture` returned `cudaErrorStreamCaptureInvalidated` — not when a
library call *threw* mid-capture before `EndCapture` was reached. So that surface
escaped and aborted the solve.

### 2. Deadlock when the root LP fails
On an infeasible model the barrier's cuDSS factorization fails (`NumericalError`,
empty primal/dual). The heuristic then unconditionally `raft::copy`'d `n` elements
from the empty result, which **threw**. That exception unwound out of
`dm.run_solver()` and got parked at the MIP `omp taskgroup` join (it must wait for
child tasks before propagating). Meanwhile B&B's `solve_root_relaxation` spins in a
wait loop whose only exits are `root_crossover_solution_set_` / `root_concurrent_halt`
— neither of which the (now-thrown) heuristic ever signals. Result: a hang that
ignores the time limit, with no Python traceback (exception stuck mid-unwind).

### 1. `manual_cuda_graph.cuh` — recover from thrown errors mid-capture
Wrap `work()` in `try/catch`. On a throw, end the capture and let its status
disambiguate: if the capture was invalidated, the recorded work was never issued —
drain the sticky error and re-run `work` eagerly (no capture, so the concurrent op
can't break it). Otherwise the error is genuine and is rethrown. This routes the
thrown-`CUBLAS_STATUS_INTERNAL_ERROR` surface into the same recovery the clean
`cudaErrorStreamCaptureInvalidated` path already used, without keying off cuBLAS's
ambiguous status code, and without masking real failures.

### 2. `diversity_manager.cu` — handle an unusable root LP and release B&B
After the concurrent root LP returns, check whether the result is usable
(`status != NumericalError` and primal/dual sizes match). When it isn't, skip the
copy/hand-off that previously threw and release B&B's root-relaxation wait via
`branch_and_bound_ptr->set_root_concurrent_halt(1)`, so B&B falls back to its own
dual-simplex root instead of deadlocking. When the result is usable, behavior is
unchanged (copy + hand off via the root-relaxation callback).

## Testing

- Reproduced the original crash signature and confirmed the hang is pre-existing on
  `main` (identical with/without the capture fix).
- Stress loop on the infeasible MILP: **62 iterations / 124 concurrent root-LP
  solves, 0 crashes, 0 hangs**; every solve now terminates cleanly as
  `Infeasible` (B&B reports `Root relaxation returned: INFEASIBLE`).
- Tested on user provided infeasible mps file from issue #1396 

closes #1396


nguidotti (approved):
Thanks for the fix, Akif! I suggest to also add the failing example as a test case.

PR #1429, opened 2026-06-12
Fast MPS parser for free-format MPS files

aliceb-nv:
This PR introduces an experimental, opt-in fast parser for well-formed MPS files, which takes advantage of available parallelism and SIMD extensions on the target machine. This is mostly intended for huge (>1GB) MPS files, e.g. for distributed solves.

The parser relies on parallel I/O overlapping with MPS section parsing workers which start parsing as soon as read/decoded data is made available, in order to hide latency on slow NFS filesystems.
LZ4 is added as a supported decompression format, which is done in a parallel fashion. LZ4 tends to perform unusually well (~10-30% compression ratios) on MPS data due to their very regular nature and common prefixes in row/column names, and decompresses at a few GB/s.

This feature is exposed via a new CLI flag `--mps-reader`, which can take either `default` (existing parser), or `experimental-fast` (this PR) as its argument.

LP, MIP, QP, and MPS SOCP formats are supported. 

Results, on a corpus of 42 instances with sizes ranging from 1.55GB to 50.5GB:

```
 NFS reference vs NFS fast:
  - mean: 17.10x
  - median: 18.08x
  - range: 6.72x to 29.05x

  NFS reference vs NFS LZ4-fast:
  - mean: 34.98x
  - median: 35.28x
  - range: 11.80x to 97.99x
  
 NVMe reference vs NVMe fast:
   - mean: 25.59x
  - median: 23.74x
  - range: 11.81x to 45.62x
```

Results for some >100GB files, NFS storage, cold cache (reference parser times out at 1 hour)

```
Name | Size | Runtime
psr_100.mps | 58 GB | 104s
psr_100.mps.lz4 | 8.6 GB | 40s
design_match.mps | 131.8 GB | 209.5s
design_match.mps.lz4 | 44 GB | 60.4s
tsp-gaia-100m.mps | 397GB | 620s
tsp-gaia-100m.mps.lz4 | 83GB | 362s
```

## Checklist

- Testing
   - [ ] Added tests
   - [ ] NA
- Documentation
   - [ ] Added new documentation
   - [ ] NA


Bubullzz:
I can confirm I extensively used this code to parse several big mps instances (> 50GB) and the results were bitwise equal with the original parser.

cpp/src/io/experimental_mps_fast/fast_fp64_parser.hpp
```diff
@@ -0,0 +1,431 @@
...
+struct power_10_lut_entry_t {
+  uint64_t high;
+  uint64_t low;
+  int biased_e2;
+};
+
+// util class to perform 256bit precision arithmetic in constexpr to build the eisel-lemire lookup
+// table
+struct cuopt_uint256_t {
+  std::array<uint64_t, 4> limb{};
```
nguidotti, line 47: Does simde expose some object that have 256-bit? I know for AVX2, you have `__m256` and their variations. If not, maybe I would use a C style array with `alignas(256)` so you can use aligned memory instructions

cpp/src/io/experimental_mps_fast/fast_fp64_parser.hpp
```diff
@@ -0,0 +1,431 @@
...
+
+  constexpr uint32_t mul_u32(uint32_t m)
+  {
+    unsigned __int128 carry = 0;
+    for (uint64_t& v : limb) {
```
nguidotti, line 52: You can use `pragma omp simd` to give a hint to the compiler (although this should be easily auto-vectorizable)

cpp/src/io/experimental_mps_fast/fast_fp64_parser.hpp
```diff
@@ -0,0 +1,431 @@
...
+  for (int e = -1; e >= FASTP64_MIN_EXP_10; --e) {
+    p.div10();
+    table[e - FASTP64_MIN_EXP_10] = make_power(p);
+  }
+  return table;
+}
+
+inline constexpr auto fast_fp64_parse_lut = make_power_table();
+
+inline constexpr std::array<double, 23> small_powers = {
```
nguidotti, line 154: If you want some a little fancier, you can use compile-time index sequences (https://cppreference.com/cpp/utility/integer_sequence)

cpp/src/io/utilities/error.hpp
```diff
@@ -34,6 +34,30 @@ inline std::string error_to_string(error_type_t error)
   return std::string("UnAccountedError");
 }
 
+[[noreturn]] inline void mps_parser_throw(error_type_t error_type, const char* msg)
+{
+  throw std::logic_error("{\"MPS_PARSER_ERROR_TYPE\": \"" + error_to_string(error_type) +
+                         "\", \"msg\": " + "\"" + std::string(msg) + "\"}");
```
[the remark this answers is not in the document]
aliceb-nv: Let's address this in later PRs instead

cpp/src/utilities/perf_counters.hpp
```diff
@@ -0,0 +1,194 @@
...
+#include <linux/perf_event.h>
+#include <sys/ioctl.h>
+#include <sys/syscall.h>
+#include <unistd.h>
+
+#include <array>
+#include <cerrno>
+#include <cstdint>
+#include <cstdio>
+#include <vector>
```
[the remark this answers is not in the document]
aliceb-nv: I'm curious where you found such rules on IWYU for this project. These headers are extremely common, I don't think it's unreasonable to skip them for the sake of visual clarity (we'd just be adding lines without any actual added value for any code reader - we're stating the obvious, so to speak)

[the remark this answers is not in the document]
aliceb-nv: That was a question. Where did you find IWYU guidelines? I can find no such thing and I'm genuinely curious

cpp/src/io/utilities/error.hpp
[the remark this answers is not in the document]
aliceb-nv: Feel free

cpp/src/io/experimental_mps_fast/file_reader.hpp
```diff
@@ -0,0 +1,314 @@
...
+        try {
+          body(index);
+        } catch (...) {
+          latch.capture(std::current_exception());
+          return;
+        }
+      }
+    });
+  }
+}
```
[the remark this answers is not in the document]
aliceb-nv: Adressed

cpp/include/cuopt/linear_programming/io/parser.hpp
```diff
@@ -112,37 +129,75 @@ mps_data_model_t<i_t, f_t> read_lp_from_string(std::string_view lp_contents);
...
-  if (lower.ends_with(".lp") || lower.ends_with(".lp.gz") || lower.ends_with(".lp.bz2")) {
+  if (lower.ends_with(".lp.lz4") || lower.ends_with(".lp.bz2") || lower.ends_with(".lp.gz") ||
+      lower.ends_with(".lp")) {
     return read_lp<i_t, f_t>(path);
   }
   throw std::logic_error(
     "read: unrecognized input file extension. Supported (case-insensitive): "
-    ".mps, .mps.gz, .mps.bz2, .qps, .qps.gz, .qps.bz2, .lp, .lp.gz, .lp.bz2. "
+    ".mps, .mps.gz, .mps.bz2, .mps.lz4, .qps, .qps.gz, .qps.bz2, .qps.lz4, "
+    ".lp, .lp.gz, .lp.bz2, .lp.lz4. "
```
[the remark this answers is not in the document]
aliceb-nv: This would not align with the existing behavior of still accepting these file extensions and only rejecting later in the code if the project wasn't built with the appropriate compression lib support

cpp/src/utilities/perf_counters.hpp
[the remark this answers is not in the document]
aliceb-nv: There is no such line in the .coderabbit.yaml

cpp/src/io/experimental_mps_fast/file_reader.cpp
```diff
@@ -0,0 +1,310 @@
...
+  int fd = ::open(path.c_str(), O_RDONLY);
+  if (fd < 0) {
+    mps_parser_fail(error_type_t::RuntimeError,
+                    "Failed to open file '%s': %s",
+                    path.c_str(),
+                    std::strerror(errno));
+  }
+  std::size_t size = get_file_size(fd, path);
+  ::close(fd);
+  return size;
```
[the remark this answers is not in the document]
aliceb-nv: This was addressed

cpp/include/cuopt/linear_programming/io/parser.hpp
```diff
@@ -111,38 +132,71 @@ mps_data_model_t<i_t, f_t> read_lp_from_string(std::string_view lp_contents);
...
+                                       mps_reader_type_t mps_reader,
+                                       bool fixed_mps_format = false)
 {
   std::string lower(path);
   std::transform(lower.begin(), lower.end(), lower.begin(), [](unsigned char c) {
     return static_cast<char>(std::tolower(c));
   });
-  if (lower.ends_with(".mps") || lower.ends_with(".mps.gz") || lower.ends_with(".mps.bz2") ||
-      lower.ends_with(".qps") || lower.ends_with(".qps.gz") || lower.ends_with(".qps.bz2")) {
+  if (lower.ends_with(".mps.lz4") || lower.ends_with(".mps.bz2") || lower.ends_with(".mps.gz") ||
```
nguidotti: Could you create a function that returns the filename without the compression (if it exists)? This will simplify the logic here

cpp/src/io/experimental_mps_fast/fast_fp64_parser.hpp
```diff
@@ -0,0 +1,436 @@
...
+struct parsed_decimal_t {
+  bool negative      = false;
+  bool fast_eligible = false;
+  uint64_t mantissa  = 0;
+  int exp10          = 0;
+};
+
+static inline bool is_digit(char c) noexcept { return c >= '0' && c <= '9'; }
+
+// SWAR 8char run of digits -> integer representation
```
nguidotti, line 184: Can you explain this to me? There are instructions for operating with 16b in AVX2 that have similar latency as 32b (see https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#techs=AVX_ALL&cats=Swizzle&ig_expand=460,465,476)

cpp/src/io/experimental_mps_fast/fast_parse_primitives.hpp
```diff
@@ -0,0 +1,386 @@
...
+}
+
+static inline void expect_eol(cursor_t& cursor)
+{
+  if (UNLIKELY(!cursor.eol())) {
+    auto got = cursor.peek_field();
+    cursor.error("expected end of line, got '%.*s'", (int)got.size(), got.data());
+  }
+
+  for (;;) {
```
nguidotti, line 304: NITPICK: I not a fan of having a for loop with a blank termination condition

cpp/src/io/experimental_mps_fast/fast_parser.cpp
```diff
@@ -0,0 +1,3221 @@
...
+// avoid openmp setup for small bounds sections
+static constexpr size_t MPS_BOUNDS_PARALLEL_MIN_BYTES = 256 * MiB;
+// ordered-name fallback is cheap enough to parallelize on smaller bounds sections
+static constexpr size_t MPS_BOUNDS_ORDERED_HINT_PARALLEL_MIN_BYTES = 8 * MiB;
+// lower bound on columns chunk size to avoid tiny parser tasks
+static constexpr size_t MPS_COLUMNS_MIN_CHUNK_BYTES = 1 * MiB;
+// parser-wide thread cap switch; very small files lose to scheduling overhead
+static constexpr size_t MPS_MEDIUM_FILE_THREAD_THRESHOLD_BYTES = 100ull * 1000ull * 1000ull;
+// thread caps for small and large files
+static constexpr int MPS_SMALL_FILE_THREAD_CAP = 16;
```
nguidotti, line 71: Is there are reason for hard-capping the number of threads?

cpp/src/io/experimental_mps_fast/fast_parser.cpp
```diff
@@ -0,0 +1,3221 @@
...
+  const char* name_;
+#endif
+  double* accumulator_;
+#ifdef MPS_FAST_TIMERS
+  nvtx::scoped_range_t nvtx_;
+  std::chrono::high_resolution_clock::time_point start_;
+#endif
+};
+
+class omp_max_active_levels_guard_t {
```
nguidotti, line 248: NITPICK: We should avoid nested parallel regions if we can

cpp/src/io/experimental_mps_fast/fast_parser.cpp
```diff
@@ -0,0 +1,3221 @@
...
+template <typename i_t, typename f_t>
+static void allocate_column_outputs(parse_state_t<i_t, f_t>& state,
+                                    const column_merge_shape_t<i_t>& shape)
+{
+  scoped_timer_t timer("allocate_temp_csr_arrays");
+  size_t values_bytes  = shape.total_nnz * sizeof(f_t);
+  size_t indices_bytes = shape.total_nnz * sizeof(i_t);
+  state.temp_csr_nnz   = shape.total_nnz;
+
+#pragma omp parallel sections num_threads(4)
```
nguidotti, line 1624: NITPICK: Do the parallel allocation has a significant performance impact? If not then, I would use the serial version for simplicity

cpp/src/io/experimental_mps_fast/fast_parser.cpp
```diff
@@ -0,0 +1,3221 @@
...
+                                         const column_merge_shape_t<i_t>& shape,
+                                         int num_threads)
+{
+  scoped_timer_t timer("scatter_into_csr");
+  {
+    scoped_timer_t matrix_timer("scatter_matrix_entries");
+#ifdef MPS_FAST_PERF_COUNTERS
+    std::vector<perf_counter_snapshot_t> perf_snapshots((size_t)shape.num_chunks);
+#endif
+#pragma omp parallel for num_threads(num_threads)
```
nguidotti, line 1669: Ideally, you should a single parallel section for the entire parser (to avoid the creation/destruction overhead of the thread pool).

cpp/src/io/experimental_mps_fast/fast_parser.cpp
```diff
@@ -0,0 +1,3221 @@
...
+    }
+  }
+
+  {
+    scoped_timer_t timer("materialize_problem_var_names");
+    const bool col_dense_ordered = state.col_index_mode == index_mode_t::dense_ordered;
+    size_t n = col_dense_ordered ? (size_t)state.problem.n_vars_ : state.var_names_sv.size();
+    state.problem.var_names_.resize(n);
+    if (col_dense_ordered && n >= 1'000'000 && num_threads > 1) {
+#pragma omp parallel for schedule(static) num_threads(num_threads)
```
nguidotti, line 2868: NITPICK: `schedule(static)` is the default scheduling policy

cpp/src/io/experimental_mps_fast/fast_parser.cpp
```diff
@@ -0,0 +1,3221 @@
...
+#pragma omp task depend(in : quadratic_ready, columns_done) depend(out : quadratic_done)
+      {
+        run_parser_task([&] {
+          MPS_NVTX_RANGE("task_quadratic", nvtx::colors::generic);
+          parse_quadratic_range(state, input.registry->range(mps_phase_kind::quadratic));
+          phase_end("quadratic");
+        });
+      }
+
+#pragma omp taskwait
```
nguidotti: You do not need the `taskwait` here. There is an implicit barrier at the end of the single and the parallel section that waits for all tasks to be completed before proceeding

cpp/src/io/experimental_mps_fast/fast_parser.cpp
```diff
@@ -0,0 +1,3221 @@
...
+  problem.row_names_.reserve(reserve_dim);
+  problem.var_names_.reserve(reserve_dim);
+  problem.constraint_lower_bounds_.reserve(reserve_dim);
+  problem.constraint_upper_bounds_.reserve(reserve_dim);
+  return reserve_dim;
+}
+
+// Contract every input stream fed to parse_mps_fast_stream must satisfy.
+template <typename Stream>
+concept InputStream = requires(Stream stream)
```
nguidotti, line 2934: I think the `InputStream` concept is not needed here since it is only used once and the templated method is enclosed in this file.

cpp/src/io/experimental_mps_fast/file_reader.cpp
```diff
@@ -0,0 +1,371 @@
...
+std::size_t system_page_size()
+{
+  static std::size_t page_size = [] {
+    long value = ::sysconf(_SC_PAGESIZE);
+    return value > 0 ? (std::size_t)value : (std::size_t)4096;
+  }();
+  return page_size;
+}
+
+bool pread_full(int fd, char* dst, std::size_t bytes, std::size_t offset)
```
nguidotti, line 128: Is there a function in the STL that does this? Seems like a common method

cpp/src/io/experimental_mps_fast/file_reader.hpp
```diff
@@ -0,0 +1,319 @@
...
+};
+
+// Work-stealing parallel loop over [0, count). Each of thread_count workers pulls
+// the next index from a shared counter and invokes body(index). An exception
+// escaping body is captured into the latch and stops the loop; the caller is
+// responsible for calling latch.rethrow_if_error() after this returns. Workers
+// are named "<thread_name_prefix><worker-id>" when a prefix is supplied.
+// OMP just doesn't really play well with blocking pread()
+template <typename Body>
+void parallel_for_indexed(std::size_t count,
```
nguidotti, line 156: In OpenMP, you already have `dynamic` scheduling policies that dynamically balance the load across the threads

cpp/src/utilities/perf_counters.hpp
```diff
@@ -0,0 +1,194 @@
+// SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES. All rights
```
nguidotti, line 1: How this differ from running `perf` from outside?

nguidotti (commented):
Thanks for the hard work, Alice! A fast parser is a great improvement.
I have a few general comments:
- Is there a reason for not using AVX512? Nowadays, hardware support is quite good in newer processors
- I never work with ARM SIMD, but NEOS seems to be the most common choice. Due to Grace/Vera, I think is important for us to support this architecture.
- Unless you see a large performance improvement for your custom implementation, I would use the STL methods when possible for simplicity (e.g., for `malloc` or reading/parsing outside the hot path).

cpp/src/io/experimental_mps_fast/fast_fp64_parser.hpp
aliceb-nv: This is a 256bit integer, not a 256bit vector :) Necessary to build the eisel-lemire tables at constexpr time.

cpp/src/io/experimental_mps_fast/fast_fp64_parser.hpp
aliceb-nv: This is a constexpr-only path

cpp/src/io/experimental_mps_fast/fast_fp64_parser.hpp
aliceb-nv: Template metaprogramming tricks are awesome but tend to tank compile times a lot more than what modern constexpr expressivity allows :)

cpp/src/io/experimental_mps_fast/fast_fp64_parser.hpp
aliceb-nv: This is based on the observation that a fair amount of digit runs in MPS files fit in this range, and SWAR is more convenient for these smaller ranges when swizzling is required. On SIMD especially AVX2 this could require cross-lane communications which tend to have an increased latency especially compared to in-register ops. NB: you highlighted "blend" operations, but they're strictly in-lane compared to true swizzles, hence the very low latency/throughput numbers :)

cpp/src/io/experimental_mps_fast/fast_parse_primitives.hpp
aliceb-nv: In my defense, this is a common pattern in low-ish level C programming

cpp/src/io/experimental_mps_fast/fast_parser.cpp
aliceb-nv: Diminishing returns, the overhead trumps the parallelization benefits

cpp/src/io/experimental_mps_fast/fast_parser.cpp
aliceb-nv: On large instances it did. Ideally, I'd just change the problem representation directly, but that's a much bigger change and probably API breaking, in other words a headache for another time

cpp/src/io/experimental_mps_fast/file_reader.cpp
aliceb-nv: Not really for this scenario :( The goal here is to submit parallel chunk reads to a single file descriptor. The C stdlib only really provides serial stateful options (FILE*, or std::ifstream in C++)

cpp/src/io/experimental_mps_fast/file_reader.hpp
aliceb-nv: In this scenario (blocking i/o pread()s, the simplest option was just to stick to std::thread. These threads strictly do I/O, so it won't really lead to contention either way

cpp/src/utilities/perf_counters.hpp
aliceb-nv: Convenience (directly logged, not having to tweak my existing scripts)
If need be this can be dropped, naturally

aliceb-nv:
Regarding AVX512 - it is still not consistently present on modern consumer CPUs (either no support at all, or only a subset), and the AMD implementations are essentially still using 256bit ALUs in a trench coat; so I did not feel comfortable making this a requirement. I also don't expect the improvements to be huge without a significant refactor - most lines in MPS files are >16B and <32B wide, which fit AVX2 perfectly. AVX2 is also basically guaranteed to exist on any modern x86 CPU released in the last 10 years, consumer or server

SVE2 support could be awesome :) Would be lovely to benchmark. As it stands, NEON is guaranteed to be present on all aarch64 CPUs and that's what SIMDe uses under the hood to translate the intel-style intrinsics. 

All custom implementations used here were motivated by benchmarking. std::from_chars came close to the fast fp64 parser used here on modern GCC libstdc++, but I did not feel too comfortable using it directly since if it falls back to a locale-based parser (or if compiled with any other libstdc+++), performance would tank significantly and seemingly unpredictably to the customer

cpp/src/utilities/perf_counters.hpp
nguidotti: It is fine to keep it. I just curious if there is a difference.

nguidotti:
> Regarding AVX512 - it is still not consistently present on modern consumer CPUs (either no support at all, or only a subset), and the AMD implementations are essentially still using 256bit ALUs in a trench coat; so I did not feel comfortable making this a requirement. I also don't expect the improvements to be huge without a significant refactor - most lines in MPS files are >16B and <32B wide, which fit AVX2 perfectly. AVX2 is also basically guaranteed to exist on any modern x86 CPU released in the last 10 years, consumer or server
> 
> SVE2 support could be awesome :) Would be lovely to benchmark. As it stands, NEON is guaranteed to be present on all aarch64 CPUs and that's what SIMDe uses under the hood to translate the intel-style intrinsics.
> 
> All custom implementations used here were motivated by benchmarking. std::from_chars came close to the fast fp64 parser used here on modern GCC libstdc++, but I did not feel too comfortable using it directly since if it falls back to a locale-based parser (or if compiled with any other libstdc+++), performance would tank significantly and seemingly unpredictably to the customer

Fair enough. Thanks for explaining it to me!

cpp/tests/linear_programming/parser_test.cpp
```diff
@@ -56,6 +56,21 @@ bool file_exists(const std::string& file)
...
 namespace {
 
+struct mps_reader_param_t {
+  const char* name;
+  mps_reader_type_t reader;
+};
+
+constexpr mps_reader_param_t default_mps_reader_param{"default_reader",
+                                                      mps_reader_type_t::default_reader};
+constexpr mps_reader_param_t fast_mps_reader_param{"fast_experimental",
```
mlubin, line 66: When the fast reader becomes mature, will we want to rename the command-line argument? Maybe just call it fast?

mlubin (commented):
This would be a great material for a technical blog post.

cpp/src/io/experimental_mps_fast/fast_parser.cpp
aliceb-nv: I'm just keeping it as self-documentation since I'm still getting ramped up on OpenMP :)

cpp/src/io/experimental_mps_fast/fast_parser.cpp
aliceb-nv: I'm keeping it for self-documentation purposes (makes it clear what an input stream is supposed to be and supposed to do)

cpp/tests/linear_programming/parser_test.cpp
aliceb-nv: Yeah :) I just wanted to make it clear that this isn't supposed to be the reference implementation for now, and that it may potentially fail on some technically noncompliant MPS files. I'd like to run some rounds of fuzzy testing before promoting it to "fast"

nguidotti:
When you moved from GCC to LLVM OpenMP, did you check if this makes a performance difference? Also there are subtle differences between them that may trigger some bugs.

aliceb-nv:
@nguidotti I didn't, but my understanding is that LLVM's libomp is more involved and extensive than libgomp. Will run benches

nguidotti:
Yeah, I consider the LLVM implementation better than GOMP one (like proper support for taskyield and other features). Mostly to make sure it does not have incompatibles with GCC

Bubullzz (commented):
I don't think I have the knowledge required to give an in depth review of all these optimizations but overall structure LGTM. Thank you for this crazy work !

nguidotti:
Hey @aliceb-nv,
Is there any blockers for merging your new parser?

aliceb-nv:
I'd like to do a bit of fuzzy testing over it, but otherwise it only requires infra review.

nguidotti (approved):
I will approve it, so you can merge it when you feels it is ready.

ramakrishnap-nv (approved):

PR #1430, opened 2026-06-12
add a Python interface to the C++ grpc asyc client

tmckayus:
This lets Python use an async grpc client with Problem type in additon to existing blocking Problem.solve() with remote execution env vars set (zero-code-change remote solve)

For some scenarios,  an async client with polling, or results fetched by a different process/thread, or job cancellation, etc is desirable. It also helps with agents that formulate problems and then want to cancel the job and modify the model.


tmckayus:
Check test_grpc_client.py to see usage. Adding user-facing documentation/example as a follow up on the PR

tmckayus:
Also, some of the name-spacing around linear_programming should change in a future PR. The grpc client is not necessarily specific to linear programming, although it started there. We should move it later. For the moment, left it alone to reduce churn.

python/cuopt/cuopt/grpc/linear_programming/grpc_client.pxd
```diff
@@ -0,0 +1,87 @@
...
+        bint job_complete
+
+    ctypedef int (*grpc_log_line_callback_t)(
+        const char* line, size_t line_len, int job_complete, void* user_data
+    ) noexcept nogil
+
+    cdef cppclass grpc_python_client_t:
+        grpc_python_client_t(const string& host, int port) except +
+        bint connect(string& error_out)
+        grpc_submit_result_t submit(
```
ramakrishnap-nv, line 67: Should we create assuming we will be adding routing support in future ? should there be a routing and numerical optimization set ?

python/cuopt/cuopt/tests/fixtures/grpc_server_fixtures.py
```diff
@@ -0,0 +1,203 @@
...
+import os
+import shutil
+import signal
+import socket
+import subprocess
+import time
+
+import pytest
+
+# Port offsets (added to CUOPT_TEST_PORT_BASE). Keep unique per test class.
```
ramakrishnap-nv, line 32: Is this meant just for test? May be we should move it under testing.

python/cuopt/conftest.py
```diff
@@ -0,0 +1,5 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
ramakrishnap-nv:
```suggestion
# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```

ramakrishnap-nv (changes requested):

python/cuopt/cuopt/tests/fixtures/grpc_server_fixtures.py
tmckayus: moved

python/cuopt/cuopt/grpc/linear_programming/grpc_client.pxd
tmckayus: created submodule cuopt.grpc.numerical so we can add more in the future

cpp/include/cuopt/grpc/cython_grpc_client.hpp
```diff
@@ -0,0 +1,152 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2025-2026, NVIDIA CORPORATION & AFFILIATES. All rights
```
ramakrishnap-nv:
```suggestion
 * SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES. All rights
```

cpp/include/cuopt/grpc/grpc_client_env.hpp
```diff
@@ -0,0 +1,22 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2025-2026, NVIDIA CORPORATION & AFFILIATES. All rights
```
ramakrishnap-nv:
```suggestion
 * SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES. All rights
```

Same for any other new files

python/cuopt/pyproject.toml
```diff
@@ -80,6 +80,9 @@ max_allowed_size_compressed = '55Mi'
 
 [tool.pytest.ini_options]
 testpaths = ["cuopt/tests"]
+# Avoid prepending the project root onto sys.path, which would shadow the
+# installed cuopt wheel (Cython extensions live in site-packages, not source).
+addopts = "--import-mode=importlib"
```
ramakrishnap-nv: Did this cause any issue ?

ramakrishnap-nv (approved):
Rest looks good, minor suggestions. And also lets create an issue for the follow-up PR for docs.

python/cuopt/pyproject.toml
tmckayus: It depends where you run the tests from. I ran into a problem where I was running tests from the project root, and it was not finding the correct Cython pieces which had just been built because it was looking at the local path.  This keeps pytest from importing source from the local path.

tmckayus: It depends where you run the tests from. I ran into a problem where I was running tests from the project root, and it was not finding the correct Cython pieces which had just been built because it was looking at the local path.  This keeps pytest from importing source from the local path.

python/cuopt/cuopt/linear_programming/solver/solver_wrapper.pyx
```diff
@@ -129,6 +129,21 @@ cdef object _vector_to_numpy(const vector[double]& vec):
     return np.asarray(<double[:size]> data_ptr, dtype=np.float64).copy()
 
 
+def _vars_dict(variable_names, primal_solution):
+    if len(primal_solution) == 0:
+        # Unbounded/infeasible solves may return no primal vector; match legacy
```
mlubin: What makes this "legacy" behavior, is there a plan to change it?

python/cuopt/cuopt/grpc/mathematical_optimization/__init__.py
```diff
@@ -0,0 +1,11 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+from cuopt.grpc.mathematical_optimization.grpc_client import (
```
mlubin: I'm not sure if introducing `cuopt.grpc.mathematical_optimization` is the right thing to do while `cuopt.linear_programming` is still around, the naming inconsistency is likely to confuse users. I would just stick with `linear_programming` until we're ready to rename everything on the python side.

Also is there a technical reason to do `cuopt.grpc.linear_programming` over `cuopt.linear_programming.grpc`?

python/cuopt/cuopt/tests/linear_programming/test_grpc_client.py
```diff
@@ -0,0 +1,266 @@
...
+    problem.addConstraint(x <= 1, name="c2")
+    problem.setObjective(x, sense=MAXIMIZE)
+    return problem
+
+
+def test_grpc_package_is_namespace_only():
+    """cuopt.grpc is a namespace; domain clients live in subpackages."""
+    assert "Client" not in grpc_pkg.__dict__
+    assert (
+        importlib.util.find_spec("cuopt.grpc.mathematical_optimization")
```
mlubin: This seems strange to test. We're already importing from `cuopt.grpc.mathematical_optimization` above.

python/cuopt/pyproject.toml
mlubin: Feels like this change is more appropriate for a standalone PR so it gets proper visibility? I'm not sure what the implications are.

mlubin (approved):
Approving with comments.

python/cuopt/pyproject.toml
tmckayus: that's a good point @mlubin. I can back it out and submit separately

python/cuopt/cuopt/grpc/mathematical_optimization/__init__.py
tmckayus: @mlubin good note on the rename.

For the module hierarchy, cuopt.grpc is at the top because we were anticipating in the future a cuopt.grpc.routing (and potentially others) which would take different problem types. Common stuff (like job status, job cancel, and job delete (remove results from the server)) can live in cupot.grpc and be inherited.

I suppose another way to do it is have a grpc common, and change the order, and just have the leaves import and inherit from grpc common.  Not sure which is better

mlubin: I'm ok with `cuopt.grpc.linear_programming`

python/cuopt/cuopt/tests/linear_programming/test_grpc_client.py
tmckayus: The first part is about ensuring that a "Client" is not exported from the namespace package (because what would that client be? lp, routing, one combined client, just common job admin stuff?). We can remove that and just add a comment. The second part is largely redundant, yes.

tmckayus: The first part is about ensuring that a "Client" is not exported from the namespace package (because what would that client be? lp, routing, one combined client, just common job admin stuff?). We can remove that and just add a comment. The second part is largely redundant, yes.

mlubin: The first part seems ok to me, the second part is redundant.

python/cuopt/cuopt/linear_programming/solver/solver_wrapper.pyx
tmckayus: Bad wording here, "legacy” here means pre-existing local Solve behavior. Before this PR we built vars with dict(zip(names, primal)), which returns {} when the primal is empty. We extracted _vars_dict for the gRPC client and added explicit name/length checks (based on coderabbit feedback iirc); the early return prevents those checks from turning “no primal” into a ValueError when the caller still passes variable names so that existing behavior is not broken. I'll fix the comment.

python/cuopt/cuopt/tests/linear_programming/test_grpc_client.py
tmckayus: I just removed the whole thing and added an explicit comment to the __init__.py, thanks

PR #1438, opened 2026-06-15
Fix the bug detecting nonconvex quadratic constraints in the general …

### LINKED ISSUE #1434: [BUG] incorrect "optimal" solution on nonconvex qcqp
**Describe the bug**

Claude-generated LP file is below. The comments describe how cuOpt returns an answer that is incorrectly declared optimal.

```
\     maximize  x0 + x1
\     subject to  x0 + x1     <= 100            (trivial slack row)
\                 2 x0 * x1   <= 0.5            (binding indefinite quadratic)
\                -1 <= x0, x1 <= 1
\
\ True optimum: 1.25  at (x0, x1) = (1, 0.25) (or (0.25, 1)).
\   max x0 + x1 s.t. x0*x1 <= 0.25, bounds [-1,1]
\   --> on x0*x1 = 0.25 with x0 in [0.25, 1], obj = x0 + 0.25/x0; convex U-shape
\       minimized at x0 = 0.5 (obj 1.0), maximized at the endpoints x0 = 1
\       and x0 = 0.25 (obj 1.25). The bound x0 <= 1 is the binding side.
\
\ With cuOpt's general-path LDLT silently dropping indefinite blocks, the
\ quadratic constraint is lost and the solver reports obj = 2 at (1, 1) --
\ violating x0*x1 = 1 > 0.25.
\
\ Path:
\   LP parser splits "2 x0 * x1" into the symmetric COO pair (0,1,1)+(1,0,1)
\   via lp_parser.cpp build_symmetric_q_coo.
\   Nonzero rhs (0.5) sets has_nonzero_rhs in the SOC dispatcher
\   (translate_soc.hpp:264, 277), forcing the general convex-quadratic path.
\   That builds symmetric H = [[0,2],[2,0]] -- indefinite with zero diagonals.
\   right_looking_ldlt's diagonal-only Markowitz cannot pivot, breaks at
\   rank=0 with no INDEFINITE_MATRIX_RETURN sentinel, and the rotated/SOC
\   construction emits a trivially-feasible 2D cone -- the quadratic
\   constraint is silently dropped.
\
\ Choosing rhs = 0.5 (not -0.5) keeps the dropped cone's downstream
\ s_0 = alpha + 0.5 = 1, s_{r+1} = alpha - 0.5 = 0 feasible (||0|| <= 1),
\ so the barrier converges cleanly to the wrong answer instead of bailing
\ with a numerical error.
Maximize
 obj: x0 + x1
Subject To
 c0: x0 + x1 <= 100
 q0: [ 2 x0 * x1 ] <= 0.5
Bounds
 -1 <= x0 <= 1
 -1 <= x1 <= 1
End
```

Run `cuopt_cli` with the above LP file as input.

**Expected behavior**
An explicit error rejecting the nonconvex QCQP, or a generic numerical error.

Local build of cuOpt 26.06

yuwenchen95:
## Description

Cross-only indefinite quadratic constraints (e.g. `2·x₀·x₁ ≤ 0.5` with `H = [[0,2],[2,0]]`) previously passed convexity checks on the general SOC conversion path. Diagonal LDLT returned `rank = 0` without `INDEFINITE_MATRIX_RETURN`, and a degenerate `r = 0` SOC lift silently dropped the quadratic term, allowing wrong solutions to be reported as Optimal (e.g. obj=2 at (1,1) instead of rejecting the problem).

Fixes:
- `right_looking_ldlt`: return `INDEFINITE_MATRIX_RETURN` when factorization stalls at `rank = 0` on a nonzero matrix.
- `translate_soc.hpp` (general path): `cuopt_expects(rank >= 1, …)` before building the SOC lift.

Adds LDLT and SOC-conversion unit tests for the cross-only indefinite case.

## Issue

closes #1434

## Checklist

- Testing
   - [x] Added tests
   - [ ] NA
- Documentation
   - [ ] Added new documentation
   - [x] NA

## Test plan

- [ ] `./build.sh libcuopt` with conda env active, then run built/installed `cuopt_cli` against `$CONDA_PREFIX/lib/libcuopt.so`
- [ ] C++ GTest:
  - `./cpp/build/tests/dual_simplex/DUAL_SIMPLEX_TEST --gtest_filter='right_looking_ldlt.indefinite_cross_only_2x2'`
  - `./cpp/build/tests/socp/SOCP_TEST --gtest_filter='general_quadratic.rejects_cross_only_indefinite'`
  - `./cpp/build/tests/socp/SOCP_TEST --gtest_filter='general_quadratic.rejects_non_convex'`
- [ ] Manual: `datasets/qcqp/issue_1434_nonconvex.lp` → `ValidationError` (not Optimal obj=2)


cpp/tests/socp/general_quadratic_test.cu
```diff
@@ -207,6 +207,59 @@ TEST(general_quadratic, rejects_non_convex)
     cuopt::logic_error);
 }
 
+// Test: cross-only indefinite Q (issue #1434). H = [[0, 2]; [2, 0]] has zero diagonals so LDLT
+// returns rank 0 without a negative pivot; must still be rejected as non-convex.
+TEST(general_quadratic, rejects_cross_only_indefinite)
```
mlubin, line 223: Instead of testing an internal converter, what about the end-to-end behavior on a non-convex instance? What status does the solver return?

yuwenchen95: I'm following the style of previous `rejects_non_convex` test, just to check if we can capture the error correctly.

yuwenchen95: Here is the full info after the change:
```
(cuopt_dev)~:/cuopt_workspace/cuopt$ ./cpp/build/cuopt_cli ./datasets/qcqp/issue_1434_nonconvex.lp 
Setting CUDA_MODULE_LOADING to EAGER
Reading file issue_1434_nonconvex.lp
Read file issue_1434_nonconvex.lp in 0.01 seconds
cuOpt version: 26.8.0, git hash: 07f7b3f7, host arch: x86_64, device archs: 90a-real
CPU: AMD EPYC 7413 24-Core Processor, threads (physical/logical): 24/48, RAM: 211.39 GiB
CUDA 12.9, device: NVIDIA H200 (ID 0), VRAM: 139.80 GiB
CUDA device UUID: 9f8075aa-e02a-0953-1955-32f638fb2993

Problem has 1 quadratic constraints. Converting to second-order cones and solving with barrier.
Error in solve_qcqp: {"CUOPT_ERROR_TYPE": "ValidationError", "msg": "Quadratic constraint 'q0' is non-convex (Q matrix is indefinite)"}
```

cpp/src/barrier/translate_soc.hpp
rg20, line 137: How long does the conversion take? I think this loop can be parallelized to speed this up. 

You can do it as another PR though.

rg20 (approved):

cpp/tests/socp/general_quadratic_test.cu
mlubin: I still think we should have an end to end test checking the solver status. This is more important than testing internal functions.

cpp/src/barrier/translate_soc.hpp
yuwenchen95: It really depends on the problem class. Usually it takes less than 1%, but there is a problem class that may take up to 8% of total solve time.

I would do it later in another PR aiming efficiency improvement of setup phase for QCQPs.

cpp/tests/socp/general_quadratic_test.cu
yuwenchen95: Added.

yuwenchen95:
/ok to test [e863795](https://github.com/NVIDIA/cuopt/pull/1438/commits/e863795e9f16fafe05acf383f937f1a301786b57)

cpp/tests/socp/general_quadratic_test.cu
```diff
@@ -214,6 +221,44 @@ TEST(general_quadratic, rejects_non_convex)
...
+// End-to-end: cross-only indefinite Q (issue #1434). H = [[0, 2]; [2, 0]] has zero diagonals so
+// LDLT returns rank 0 without a negative pivot; must still be rejected as non-convex via
+// solve_qcqp.
+TEST(general_quadratic, rejects_cross_only_indefinite)
+{
+  raft::handle_t handle{};
+  init_handler(&handle);
+
+  // H = [[0, 2]; [2, 0]] from [ 4 x * y ] in LP bracket notation.
+  auto problem = cuopt::mathematical_optimization::io::read_lp_from_string<i_t, f_t>(R"LP(
```
mlubin: `io::read_lp_from_string`

cpp/tests/socp/general_quadratic_test.cu
```diff
@@ -214,6 +221,44 @@ TEST(general_quadratic, rejects_non_convex)
...
+Bounds
+  -1 <= x <= 1
+  -1 <= y <= 1
+End
+)LP");
+
+  ASSERT_TRUE(problem.has_quadratic_constraints());
+  ASSERT_EQ(problem.get_quadratic_constraints().size(), 1u);
+
+  cuopt::mathematical_optimization::optimization_problem_t<i_t, f_t> op_problem(&handle);
```
mlubin:
```suggestion
  optimization_problem_t<i_t, f_t> op_problem(&handle);
```

cpp/tests/socp/general_quadratic_test.cu
```diff
@@ -214,6 +221,44 @@ TEST(general_quadratic, rejects_non_convex)
...
+  cuopt::mathematical_optimization::populate_from_mps_data_model(&op_problem, problem);
```
mlubin:
```suggestion
  populate_from_mps_data_model(&op_problem, problem);
```

cpp/tests/socp/general_quadratic_test.cu
```diff
@@ -214,6 +221,44 @@ TEST(general_quadratic, rejects_non_convex)
...
+  ASSERT_TRUE(op_problem.has_quadratic_constraints())
+    << "populate_from_mps_data_model dropped quadratic constraints";
+  ASSERT_EQ(op_problem.get_quadratic_constraints().size(), 1u);
+
+  cuopt::mathematical_optimization::pdlp_solver_settings_t<i_t, f_t> settings;
```
mlubin:
```suggestion
  pdlp_solver_settings_t<i_t, f_t> settings;
```

cpp/tests/socp/general_quadratic_test.cu
```diff
@@ -214,6 +221,44 @@ TEST(general_quadratic, rejects_non_convex)
...
+  auto solution = cuopt::mathematical_optimization::solve_lp(op_problem, settings);
```
mlubin:
```suggestion
  auto solution = solve_lp(op_problem, settings);
```

cpp/tests/socp/general_quadratic_test.cu
```diff
@@ -5,13 +5,20 @@
...
+#include <gmock/gmock.h>
 #include <gtest/gtest.h>
 
 #include <barrier/translate_soc.hpp>
+#include <cuopt/mathematical_optimization/io/parser.hpp>
+#include <cuopt/mathematical_optimization/optimization_problem.hpp>
 #include <cuopt/mathematical_optimization/optimization_problem_interface.hpp>
+#include <cuopt/mathematical_optimization/optimization_problem_utils.hpp>
+#include <cuopt/mathematical_optimization/pdlp/solver_settings.hpp>
+#include <cuopt/mathematical_optimization/solve.hpp>
```
mlubin: Is this the right file for an end-to-end solve test if `cuopt/mathematical_optimization/solve.hpp` was not previously included?

mlubin (changes requested):
We should very rarely have code that fully qualifies nanes with `cuopt::mathematical_optimization`

cpp/tests/socp/general_quadratic_test.cu
yuwenchen95: I think it's ok since the file is solely for tests related to general quadratic convex constraints.

mlubin: We should follow the pattern in the surrounding context and write the test with `solve_linear_program_with_barrier` instead of `solve_lp`.

ramakrishnap-nv (approved):

cpp/tests/socp/general_quadratic_test.cu
```diff
@@ -230,7 +226,7 @@ TEST(general_quadratic, rejects_cross_only_indefinite)
   init_handler(&handle);
 
   // H = [[0, 2]; [2, 0]] from [ 4 x * y ] in LP bracket notation.
-  auto problem = cuopt::mathematical_optimization::io::read_lp_from_string<i_t, f_t>(R"LP(
+  auto mps = io::read_lp_from_string<i_t, f_t>(R"LP(
```
mlubin: The `mps` name is confusing because the problem is provided in LP format.

cpp/tests/socp/general_quadratic_test.cu
```diff
@@ -241,22 +237,67 @@ Bounds
...
+  EXPECT_EQ(m, 0);
 
-  cuopt::mathematical_optimization::pdlp_solver_settings_t<i_t, f_t> settings;
-  auto solution = cuopt::mathematical_optimization::solve_lp(op_problem, settings);
+  user_problem_t<i_t, f_t> user_problem(&handle);
+  user_problem.num_rows  = m;
+  user_problem.num_cols  = n;
+  user_problem.objective = mps.get_objective_coefficients();
+
+  user_problem.A.m      = m;
```
mlubin, line 254: What happens if someone copies this code that ignores the constraint matrix into another test? We need a comment here.

cpp/tests/socp/general_quadratic_test.cu
```diff
@@ -241,22 +237,67 @@ Bounds
...
-  EXPECT_THAT(error_status.what(), testing::HasSubstr("non-convex"));
-  EXPECT_THAT(error_status.what(), testing::HasSubstr("q0"));
+  simplex_solver_settings_t<i_t, f_t> settings;
+  settings.barrier          = true;
+  settings.barrier_presolve = true;
+  settings.dualize          = 0;
+
+  try {
+    convert_quadratic_constraints_to_second_order_cones<i_t, f_t>(n, qcs, csr_A, user_problem);
+    csr_A.to_compressed_col(user_problem.A);
```
mlubin, line 293: Why is user_problem.A initialized above if it's set here?

mlubin (approved):

cpp/tests/socp/general_quadratic_test.cu
yuwenchen95: We need `csr_A` for the use in `convert_quadratic_constraints_to_second_order_cones`. The current input pipeline has several transformation between csc and csr format for the efficiency of each block.

Issue #1354, opened 2026-06-01
[BUG] Rotated SOC with single (asymmetric) cross term returns nan instead of raising the logged ValidationError


ramakrishnap-nv:
**Describe the bug**

A rotated second-order cone constraint written the intuitive way — with a single
cross term `- x3*x4` — is rejected internally by the SOC conversion (a
`ValidationError` is logged), but the error is **not propagated to the caller**.
`solve()` returns `Status = NoTermination` with `ObjValue = nan` instead of either
solving the cone or raising the validation error.

The SOC converter (`cpp/src/barrier/translate_soc.hpp`) requires the rotated-cone
cross term to be supplied as a **symmetric** off-diagonal pair `(-d, -d)`. A single
off-diagonal entry trips:

```
Quadratic constraint 'Q0' rotated SOC Q must contain exactly one symmetric
off-diagonal pair (-d,-d); found 1 off-diagonal entries
```

The quadratic-*objective* path symmetrizes `Q`, so users reasonably expect the
quadratic-*constraint* path to behave the same. At minimum, the validation error
should surface as a Python exception / non-OK termination status rather than a
silent `nan`.

**Steps/Code to reproduce bug**

```python
import numpy as np
from cuopt.linear_programming.problem import Problem, MINIMIZE

def solve(symmetric):
    p = Problem()
    x1 = p.addVariable(lb=-np.inf, name="x1")
    x2 = p.addVariable(lb=-np.inf, name="x2")
    x3 = p.addVariable(lb=0, name="x3")
    x4 = p.addVariable(lb=0, name="x4")
    if symmetric:
        # works: symmetric Q (Q[x3,x4] = Q[x4,x3] = -0.5)
        p.addConstraint(x1*x1 + x2*x2 - 0.5*x3*x4 - 0.5*x4*x3 <= 0)
    else:
        # fails: single off-diagonal entry Q[x3,x4] = -1
        p.addConstraint(x1*x1 + x2*x2 - x3*x4 <= 0)
    p.addConstraint(x1 + x2 >= 2)
    p.setObjective(x3 + x4, sense=MINIMIZE)
    p.solve()
    return p.Status, p.ObjValue

print("intuitive (- x3*x4):   ", solve(False))   # -> (0, nan)  (error only logged)
print("symmetric (-0.5 each): ", solve(True))    # -> (1, 2.8284271257194744)
```

Output:

```
[...]

PR #1439, opened 2026-06-15
fix(socp): canonicalize QC Q COO for rotated SOC detection

### LINKED ISSUE #1435: [BUG] Rotated second order cone detection assumes symmetrized inputs
This is a bug in the C++-level detection of rotated second order cones, but it's easy to demonstrate in python:

```python
from cuopt.linear_programming.problem import Problem, MAXIMIZE, CONTINUOUS

INF = float("inf")

def solve(label, build_constraint):
    p = Problem("rsoc")
    t  = p.addVariable(lb=0.5, ub=0.5, vtype=CONTINUOUS, name="t")
    u  = p.addVariable(lb=1.0, ub=1.0, vtype=CONTINUOUS, name="u")
    v0 = p.addVariable(lb=-INF, ub=INF, vtype=CONTINUOUS, name="v0")
    v1 = p.addVariable(lb=-INF, ub=INF, vtype=CONTINUOUS, name="v1")
    p.addConstraint(t + u + v0 + v1 <= 100.0, name="slack")
    p.addConstraint(build_constraint(t, u, v0, v1), name="rsoc")
    p.setObjective(v0 + v1, sense=MAXIMIZE)
    sol = p.solve()
    lhs = (v0.getValue()**2 + v1.getValue()**2
           - 2 * t.getValue() * u.getValue())
    print(f"\n[{label}]")
    print(f"  Status:    {sol.get_termination_reason()}")
    print(f"  Objective: {sol.get_primal_objective():.10g}   "
          f"(expected sqrt(2) ~ 1.41421)")
    print(f"  v0 = {v0.getValue():.6g},  v1 = {v1.getValue():.6g}")
    print(f"  v0^2 + v1^2 - 2*t*u = {lhs:.6g}   (should be <= 0)")


solve("A: v0*v0 + v1*v1 - 2*t*u <= 0  (natural idiom)",
      lambda t, u, v0, v1: v0*v0 + v1*v1 - 2*t*u <= 0.0)

solve("B: v0*v0 + v1*v1 - t*u - u*t <= 0  (manually symmetrized)",
      lambda t, u, v0, v1: v0*v0 + v1*v1 - t*u - u*t <= 0.0)
```

Yields:
```
[A: v0*v0 + v1*v1 - 2*t*u <= 0  (natural idiom)]
  Status:    Optimal
  Objective: 0.0001276033875   (expected sqrt(2) ~ 1.41421)
  v0 = 6.38017e-05,  v1 = 6.38017e-05
  v0^2 + v1^2 - 2*t*u = -1   (should be <= 0)
[B: v0*v0 + v1*v1 - t*u - u*t <= 0  (manually symmetrized)]
  Status:    Optimal
  Objective: 1.414213562   (expected sqrt(2) ~ 1.41421)
  v0 = 0.707107,  v1 = 0.707107
  v0^2 + v1^2 - 2*t*u = -4.58143e-10   (should be <= 0)
```

The C API is similarly affected (confirmed) and potentially the MPS reader too (not confirmed).

yuwenchen95:
## Description

Quadratic-constraint `Q` is now stored in **canonical COO** internally: one coefficient per variable pair (e.g. a single `-2` for `2·x₀·x₁`), with symmetric MPS halves merged at ingest. Canonicalization runs at ingest boundaries (MPS/LP parser, C API, Python→solver, gRPC, PDLP CPU/GPU problem setup). MPS `QCMATRIX` still accepts symmetric halves; the MPS writer expands canonical cross terms back to symmetric form on export.

The RSOC fast path now accepts a single eligible cross term (e.g. `-2·t·u` for `||tail||² ≤ 2·t·u`) instead of requiring two symmetric off-diagonal entries. Previously, natural Python/API forms were misrouted to the general path or failed pattern matching, producing wrong optima (e.g. ~0 instead of √2).

## Issue

closes #1435, #1354

## Checklist

- Testing
   - [x] Added tests
   - [ ] NA
- Documentation
   - [x] Added new documentation
   - [ ] NA

## Test plan

- [ ] `./build.sh libcuopt` with conda env active
- [ ] C++: `ctest --test-dir cpp/build -R 'SOCP_TEST|MPS_PARSER_TEST|C_API_TEST'`
- [ ] Parser: `qc_cross_term_stored_canonical` and related LP/MPS QC COO tests
- [ ] C API: rotated SOC / canonical QC COO tests in `C_API_TEST`
- [ ] Python: `pytest -v cuopt/cuopt/tests/socp/test_socp.py -k rotated_soc`
- [ ] Manual: rotated SOC with single `-2*t*u` cross term → objective ≈ √2 (not ~0)


cpp/include/cuopt/linear_programming/io/data_model_view.hpp
```diff
@@ -418,9 +418,15 @@ class data_model_view_t {
 
   /**
    * @brief Quadratic constraints (MPS QCMATRIX); owned copy for writers when not using spans.
+   *
+   * Q COO is stored as supplied (not canonicalized here). Canonicalization runs when
```
mlubin: The comment should describe the interface and API contracts, not the implementation.

cpp/include/cuopt/linear_programming/io/mps_data_model.hpp
```diff
@@ -262,7 +265,9 @@ class mps_data_model_t {
    * @note All span inputs are host memory; the model copies this data.
    * @param linear_values, linear_indices Same nnz; can be empty for a purely quadratic row (rare).
    * @param vals, rows, cols COO triplets for Q; same length; may all be empty if Q is empty.
-   *        Stored sorted by (row, col).
+   *        Stored sorted by (row, col) in canonical form (one entry per variable pair).
+   * @param require_symmetric_q_offdiagonal When true (MPS QCMATRIX), each off-diagonal pair must
```
mlubin: What's the motivation for this extra argument? Is it an assertion or does the change the behavior of the function?

cpp/include/cuopt/linear_programming/io/mps_data_model.hpp
```diff
@@ -385,4 +391,13 @@ class mps_data_model_t {
...
 };  // class mps_data_model_t
 
+/**
+ * @brief Canonicalize Q COO on each quadratic constraint in place (merge duplicates, collapse
+ * symmetric MPS halves when requested). Implemented in mps_data_model.cpp.
+ */
+template <typename i_t, typename f_t>
+void canonicalize_quadratic_constraints(
+  std::vector<typename mps_data_model_t<i_t, f_t>::quadratic_constraint_t>& constraints,
+  bool require_symmetric_q_offdiagonal = false);
```
mlubin: Same comment on `require_symmetric_q_offdiagonal`. Is it an assertion or does the change the behavior of the function?

cpp/include/cuopt/linear_programming/io/mps_data_model.hpp
```diff
@@ -385,4 +391,13 @@ class mps_data_model_t {
 
 };  // class mps_data_model_t
 
+/**
+ * @brief Canonicalize Q COO on each quadratic constraint in place (merge duplicates, collapse
+ * symmetric MPS halves when requested). Implemented in mps_data_model.cpp.
```
mlubin: > Implemented in mps_data_model.cpp.

This is obvious for everything in the header file, no?

cpp/include/cuopt/linear_programming/optimization_problem_interface.hpp
```diff
@@ -75,6 +75,10 @@ class optimization_problem_interface_t {
 
   /**
    * @brief Store quadratic constraints for MPS round-trip (linear + Q parts per QC row).
+   *
+   * Q COO must already be in canonical form (one entry per variable pair). Callers that ingest
```
mlubin: The documentation on quadratic_constraint_t already says the Q must be in canonical format. Is this comment redundant?

cpp/src/io/lp_parser.cpp
```diff
@@ -257,9 +220,8 @@ class LpParseEngine {
   //   Objective:  must be followed by '/ 2'; the inner-coefficient convention
   //               is QUADOBJ-style 0.5 x^T Q x, so off-diagonals are halved
   //               and linear-inside-bracket terms also get /2.
-  //   Constraint: must NOT be followed by '/ 2'; coefficients are taken at
-  //               face value (x^T Q x); bracket must contain at least one
-  //               quadratic term.
+  //   Constraint: must NOT be followed by '/ 2'; coefficients are face value
+  //               for x^T Q x (canonicalized in append_quadratic_constraint).
```
mlubin: This comment is describing the accepted syntax. canonicalization isn't relevant here, plus the edit dropped important information.

cpp/src/io/quadratic_constraint_coo.hpp
```diff
@@ -0,0 +1,67 @@
...
+namespace cuopt::linear_programming::io {
+
+/**
+ * @brief Options for quadratic-constraint Q COO canonicalization.
+ *
+ * Internal storage uses one triplet per variable pair: off-diagonal cross terms store the full
+ * coefficient on x_i*x_j (e.g. -2*d for rotated SOC), not symmetric halves.
+ *
+ * Canonicalization runs at ingestion boundaries only:
+ * - MPS/LP: append_quadratic_constraint()
```
mlubin: This is a utility function. It shouldn't list all of its users; the list will quickly get stale.

cpp/src/io/quadratic_constraint_coo.hpp
```diff
@@ -0,0 +1,67 @@
...
+ * - C API: optimization_problem add_quadratic_constraint()
+ * - Python view -> solver: populate_from_data_model_view()
+ * - gRPC decode: before set_quadratic_constraints()
+ * - MPS export: mps_writer QCMATRIX emission
+ */
+template <typename f_t>
+struct qc_coo_canonicalize_options_t {
+  /** When true (MPS QCMATRIX), every off-diagonal pair must appear symmetrically in the input. */
+  bool require_symmetric_offdiagonal_pairs{false};
```
mlubin: Again is this an assertion or behavior change?

cpp/src/io/quadratic_constraint_coo.hpp
```diff
@@ -0,0 +1,67 @@
...
+#include <tuple>
+#include <type_traits>
+#include <unordered_map>
+#include <utility>
+#include <vector>
+
+namespace cuopt::linear_programming::io {
+
+/**
+ * @brief Options for quadratic-constraint Q COO canonicalization.
```
mlubin: Instead of calling this a "quadratic constraint Q" can we describe the properties of the matrix more generically? I think we can just talk about canonicalizing into lower or upper triangular matrices that are implicitly symmetric. The idea is to reduce the context needed to understand the code; I don't think there's anything specific to optimization here.

cpp/src/io/quadratic_constraint_coo.hpp
```diff
@@ -0,0 +1,67 @@
...
+ * - C API: optimization_problem add_quadratic_constraint()
+ * - Python view -> solver: populate_from_data_model_view()
+ * - gRPC decode: before set_quadratic_constraints()
+ * - MPS export: mps_writer QCMATRIX emission
+ */
+template <typename f_t>
+struct qc_coo_canonicalize_options_t {
+  /** When true (MPS QCMATRIX), every off-diagonal pair must appear symmetrically in the input. */
+  bool require_symmetric_offdiagonal_pairs{false};
+  f_t tol{1e-12};
```
mlubin: Why do we need a numerical tolerance for the canonicalization? I'd remove this and, if we need one, add a documented zero tolerance elsewhere.

cpp/src/io/quadratic_constraint_coo.hpp
```diff
@@ -0,0 +1,67 @@
...
+ * - Sorts output by (row, col).
+ */
+template <typename i_t, typename f_t>
+void canonicalize_qc_coo(std::vector<i_t>& rows,
+                         std::vector<i_t>& cols,
+                         std::vector<f_t>& vals,
+                         const qc_coo_canonicalize_options_t<f_t>& opts = {});
+
+/** Canonicalize Q on a constraint entry with rows/cols/vals and constraint_row_name. */
+template <typename QC>
```
mlubin: Why does this need to be templated on QC? Looks like we could take a matrix instead (rows, cols, vals).

cpp/tests/linear_programming/c_api_tests/c_api_test.c
```diff
@@ -1741,6 +1742,236 @@ cuopt_int_t test_general_quadratic_constraint_problem(cuopt_int_t* termination_s
...
+
+  cuopt_float_t constraint_bounds[] = {2.0};
+  char constraint_sense[]           = {CUOPT_GREATER_THAN};
+
+  cuopt_float_t var_lower_bounds[] = {-CUOPT_INFINITY, -CUOPT_INFINITY, 0.0, 0.0};
+  cuopt_float_t var_upper_bounds[] = {
+    CUOPT_INFINITY, CUOPT_INFINITY, CUOPT_INFINITY, CUOPT_INFINITY};
+  char variable_types[] = {CUOPT_CONTINUOUS, CUOPT_CONTINUOUS, CUOPT_CONTINUOUS, CUOPT_CONTINUOUS};
+
+  // Canonical cross term Q[x3, x4] = -1 (not MPS symmetric halves).
```
mlubin: We should remove most of not all references to "MPS halves". This isn't a special property of MPS and there are many ways to input a problem into cuOpt. Just say what it is: a single, non-symmetric term.

docs/cuopt/source/cuopt-c/convex/convex-c-api.rst
```diff
@@ -68,9 +68,10 @@ For problems with quadratic constraints, first create a problem, and then use
    supports three types of quadratic constraints:
    - **Convex quadratic constraints** of the form ``x^T Q x + d^T x <= alpha`` where ``H=(Q+Q^T)/2`` is a symmetric positive semidefinite matrix. `Q` need not be symmetric.
    - **Second-order cone constraints** of the form ``sum_{i=1}^n x_i^2 <= x_0``, ``x_0 >= 0``
-   - **Rotated second-order cone constraints** of the form ``sum_{i=2}^n x_i^2  - 0. 5 * x_0 * x_1 - 0. 5 * x_1 * x_0 <= 0``, ``x_0 >= 0``, ``x_1 >= 0``
+   - **Rotated second-order cone constraints** of the form ``sum_{i=2}^n x_i^2 - 2 * x_0 * x_1 <= 0``, ``x_0 >= 0``, ``x_1 >= 0``
+     (supply ``Q[x_0, x_1] = -2`` in COO, or ``-x_0*x_1`` for the ``<= x_0*x_1`` cone form; see :doc:`/convex-features`)
 
-   For the rotated second-order cone constraints, cuOpt expects the quadratic matrix to be symmetric.
+   MPS QCMATRIX requires symmetric off-diagonal halves; the C API accepts a single cross coefficient per variable pair (canonicalized on ingest).
```
mlubin: MPS QCMATRIX isn't relevant for the documentation of the C API.

mlubin: Be clear what's accepted. If the inputs are canonicalized then it sounds like the C API accepts more than just a single cross coefficient per variable pair.

docs/cuopt/source/cuopt-c/convex/convex-examples.rst
```diff
@@ -269,9 +269,9 @@ Rotated Second-Order Cone Example
 ---------------------------------
 
 This example adds a **rotated** cone with :c:func:`cuOptAddQuadraticConstraint`. The
-quadratic matrix ``Q`` must be supplied **symmetrically**, so the cross term of
-``x1^2 + x2^2 <= x3*x4`` is given as the two equal off-diagonal entries
-``Q[x3, x4] = Q[x4, x3] = -0.5`` (triplets ``(2, 3, -0.5)`` and ``(3, 2, -0.5)``).
+cone ``x1^2 + x2^2 <= x3*x4`` uses the canonical cross term ``Q[x3, x4] = -1``
+(i.e. ``-x3*x4`` in ``x^T Q x``). MPS QCMATRIX uses symmetric halves instead; see
```
mlubin: MPS QCMATRIX isn't relevant for the documentation of the C API.

docs/cuopt/source/convex-features.rst
```diff
@@ -199,16 +210,16 @@ When any quadratic constraint is present, cuOpt automatically selects the barrie
...
+    # x2^2 + x3^2 <= 2*x0*x1
     problem.addConstraint(
-        x2*x2 + x3*x3 - 0.5*x0*x1 - 0.5*x1*x0 <= 0, name="rotated_soc"
+        x2*x2 + x3*x3 - 2*x0*x1 <= 0, name="rotated_soc"
     )
 
 .. note::
-   cuOpt expects the quadratic matrix ``Q`` to be symmetric. Supply each cross
-   term as two equal halves, as in ``-0.5*x0*x1 - 0.5*x1*x0`` above.
+   MPS QCMATRIX files require symmetric cross-term halves; the Python and C APIs accept a single
```
mlubin: MPS QCMATRIX isn't relevant for the documentation of the Python API.

cpp/tests/linear_programming/c_api_tests/c_api_test.c
yuwenchen95: I agree

cpp/include/cuopt/linear_programming/io/mps_data_model.hpp
yuwenchen95: This is the flag for additional check with MPS file input that requires symmetric halved inputs for cross terms.

docs/cuopt/source/cuopt-c/convex/convex-c-api.rst
yuwenchen95: We allow users to give duplicate cross coefficients per variable pair and then merge them in the canonicalization step.

cpp/src/io/quadratic_constraint_coo.hpp
yuwenchen95: I should use the epsilon value of`f_t` rather than a separate parameter.

cpp/src/io/quadratic_constraint_coo.hpp
yuwenchen95: Removed

cpp/src/io/quadratic_constraint_coo.hpp
yuwenchen95: Clarified to be lower-triangular form of a matrix to store.

yuwenchen95: Clarified to be lower-triangular form of a matrix to store.

cpp/include/cuopt/linear_programming/io/mps_data_model.hpp
yuwenchen95: This is the flag for additional check with MPS file input that requires symmetric halved inputs for cross terms.

cpp/tests/linear_programming/parser_test.cpp
```diff
@@ -2176,14 +2176,12 @@ End
 )LP");
   ASSERT_EQ(m.get_quadratic_constraints().size(), 1u);
   const auto& qc = nth_qc(m, 0);
-  // Q has 4 entries (all of [[1,2],[2,1]]) stored as COO triplets.
-  EXPECT_EQ(qc.rows, (std::vector<int>{0, 0, 1, 1}));
-  EXPECT_EQ(qc.cols, (std::vector<int>{0, 1, 0, 1}));
-  ASSERT_EQ(qc.vals.size(), 4u);
+  EXPECT_EQ(qc.rows, (std::vector<int>{0, 0, 1}));
```
mlubin: Nit: pre-existing but should be fixed. The idiomatic way to write this line (and similar lines) is:
```
EXPECT_THAT(qc.rows, ElementsAre(0, 0, 1))
```
with
```
#include <gmock/gmock.h>

using ::testing::ElementsAre;
```
up top.

cpp/include/cuopt/linear_programming/io/mps_data_model.hpp
```diff
@@ -262,7 +264,11 @@ class mps_data_model_t {
    * @note All span inputs are host memory; the model copies this data.
    * @param linear_values, linear_indices Same nnz; can be empty for a purely quadratic row (rare).
    * @param vals, rows, cols COO triplets for Q; same length; may all be empty if Q is empty.
-   *        Stored sorted by (row, col).
+   *        Canonicalized on ingest to one triplet per variable pair (full x^T Q x coefficient).
```
mlubin: Is it always the lower triangle, upper triangle, or could it be canonicalized in either form?

cpp/include/cuopt/linear_programming/io/mps_data_model.hpp
```diff
@@ -262,7 +264,11 @@ class mps_data_model_t {
...
+   * @param require_symmetric_q_offdiagonal Input validation only (default false). When false,
```
mlubin: I still find this `require` argument a bit confusing. Can we more directly write a separate helper function that checks the desired property rather than adding additional extraneous functionality to `append_quadratic_constraint`?

cpp/src/io/quadratic_constraint_coo.hpp
```diff
@@ -0,0 +1,37 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2025-2026, NVIDIA CORPORATION & AFFILIATES. All rights
+ * reserved. SPDX-License-Identifier: Apache-2.0
+ */
+#pragma once
```
mlubin: Consider renaming this file; the contents don't mention anything about quadratic constraints.

docs/cuopt/source/convex-features.rst
```diff
@@ -157,20 +157,32 @@ Second-order cone constraints must be specified as a quadratic constraint with a
...
 
 .. code-block:: text
 
-   x_2^2 + x_2^2 + ... + x_n^2 - 0.5 * x_0 * x_1 - 0.5 * x_1 * x_0 <= 0,  x_0 >= 0,  x_1 >= 0.
+   x_2^2 + x_3^2 + ... + x_n^2 - 2 * x_0 * x_1 <= 0,  x_0 >= 0,  x_1 >= 0.
 
-Note that for rotated second-order cone constraints, cuOpt expects the quadratic matrix to be symmetric.
-For any cross term, provide both ``Q_i,j`` and ``Q_j,i``. cuOpt expects only symmetric ``Q``:
+In the Python and C APIs, supply one cross coefficient per variable pair (e.g. ``-2 * x_0 * x_1``);
+cuOpt stores one canonical COO entry per pair. The MPS QCMATRIX interchange format requires each
```
mlubin: Defining the MPS QCMATRIX format seems out of scope here; it's standardized across solvers.

mlubin (commented):
Big improvement!

We have test coverage of the C interface, what about also adding tests covering the python example in the original issue?

cpp/include/cuopt/linear_programming/io/mps_data_model.hpp
yuwenchen95: The input can be either form but the upper triangle form is used after canonicalization.

cpp/include/cuopt/linear_programming/io/mps_data_model.hpp
yuwenchen95: I moved it into `mps_parser_internal.hpp` since it is only applied to MPS files.

cpp/src/io/quadratic_constraint_coo.hpp
yuwenchen95: I moved the canonicalization function into `mps_parser_*` files though it is also used for lp file input.

ramakrishnap-nv (approved):

cpp/include/cuopt/linear_programming/io/mps_data_model.hpp
```diff
@@ -385,4 +388,14 @@ class mps_data_model_t {
 
 };  // class mps_data_model_t
 
+/**
+ * @brief Canonicalize Q COO on each quadratic constraint in place.
+ *
+ * Used for raw API / data_model_view input (single cross term per pair). MPS QCMATRIX
```
mlubin: "Used for" is a red flag in documentation. First, documentation should use the active voice (https://developers.google.com/style/voice). Second it's unclear if it's describing the API contract or giving examples of callers. Describe what the function does and what assumptions it makes on the input, no more no less. "MPS QCMATRIX symmetric-half validation is done in the MPS parser before append." does not fit here.

cpp/src/io/lp_parser.cpp
```diff
@@ -964,9 +927,8 @@ void LpParseEngine<i_t, f_t>::parse_quadratic_bracket(int outer_sign,
                        "constraint must contain at least one quadratic term",
                        peek().line);
 
-    // Coefficients are at face value — the post-pass that flushes the
-    // quadratic_constraint_block_t to the data model handles the symmetric
-    // expansion and the /2 split for off-diagonals.
+    // Coefficients are face value for x^T Q x; canonicalization runs in
```
mlubin: Pre-existing, but it's not really clear to me what "face value" means, can we be more precise? Also "x^T Q x" does not have coefficients, Q does.

cpp/src/io/lp_parser.hpp
```diff
@@ -73,9 +73,9 @@ class lp_parser_t {
     std::vector<i_t> linear_indices{};
     std::vector<f_t> linear_values{};
     f_t rhs_value{};
-    // Upper-triangular (i <= j) raw triples directly from the LP source
-    // (face value, no /2). The post-pass mirrors and halves off-diagonals
-    // to build the symmetric Q in CSR.
+    // Upper-triangular (i <= j) face-value triples from the LP source; each
```
mlubin: Pre-existing but, try to avoid "face value"

cpp/src/io/mps_parser.cpp
```diff
@@ -176,6 +197,126 @@ std::string_view trim(std::string_view str)
...
+                     error_type_t::ValidationError,
+                     "COO rows/cols/vals length mismatch");
+
+  std::unordered_map<std::pair<i_t, i_t>, f_t, qcmatrix_pair_hash<i_t>> agg;
+  agg.reserve(n);
+  for (size_t t = 0; t < n; ++t) {
+    const i_t r = rows[t];
+    const i_t c = cols[t];
+    const f_t v = vals[t];
+    if (std::abs(v) <= std::numeric_limits<f_t>::epsilon()) { continue; }
```
mlubin: Why does this process need a numerical nonzero check at all? What are we trying to catch by zeroing out some of the structurally nonzero inputs?

cpp/src/io/mps_parser.cpp
```diff
@@ -176,6 +197,126 @@ std::string_view trim(std::string_view str)
...
+    agg[{r, c}] += v;
+  }
+
+  const f_t eps = std::numeric_limits<f_t>::epsilon();
+  std::unordered_map<std::pair<i_t, i_t>, char, qcmatrix_pair_hash<i_t>> visited;
+  for (const auto& [rc, coeff] : agg) {
+    (void)coeff;
```
mlubin: What does `(void)coeff` do?

cpp/src/io/mps_parser.cpp
```diff
@@ -176,6 +197,126 @@ std::string_view trim(std::string_view str)
   return str.substr(start, end - start + 1);
 }
 
+template <typename i_t, typename f_t>
+void check_symmetric_offdiagonal_pairs(const std::vector<i_t>& rows,
```
mlubin, line 182: `std::unordered_map<std::pair<i_t, i_t>>` is a very unusual data structure for storing a sparse matrix. I'm suspicious of the custom hash as well. We have existing code in the codebase to convert COO to CSR/CSC. If you convert the COO and its transpose to two CSR matrices then check those two for equality, you'll have a relatively efficient symmetry check.

python/cuopt/cuopt/tests/socp/test_socp.py
```diff
@@ -28,6 +28,11 @@
 EXPECTED_SOCP_3_OBJECTIVE = -1.932105
 EXPECTED_SOCP_3_X = (0.83666003, -0.54772256)
 
+# Rotated SOC with a single canonical cross term (-2*t*u); optimum max v0+v1 is sqrt(2).
+_SQRT2 = np.sqrt(2.0)
+EXPECTED_ROTATED_SOC_NATURAL_CROSS_OBJECTIVE = _SQRT2
```
mlubin: These values are used in one function, why define file-level constants?

yuwenchen95: Moved them into the corresponding tests.

cpp/src/io/mps_parser.cpp
yuwenchen95: Aha, it doesn't matter when we are trying to merge duplicate terms.

cpp/src/io/mps_parser.cpp
yuwenchen95: It is deleted.

cpp/src/io/mps_parser.cpp
yuwenchen95: This was used for merging the same quadratic terms with different coefficients, but we have put the merging into the canonicalization step and it is ok we restrict the user to provide non-repeting terms in MPS files. Now, I have moved the symmetric check into `flush_qcmatrix_block`, where the coo format of a `QCMATRIX` first appears, and used two vectors for local sort. It should be more reasonable and efficient now.

mlubin: The sorts of the triplets are still potentially expensive. @rg20 do you have pointers to conversion utils, I think you looked at this recently?

yuwenchen95: I would support current sorting over count-sort conversion here. Count-sort conversion is of complexity `O(nnz_soc+n)` for a quadratic constraint. However, we usually have multiple quadratic constraints of small dimensionalities `nnz_soc<<n` so the dependency on `n` for each quadratic constraint is less favorable compared to current sorting of complexity `O(nnz_soc*log(nnz_soc))`. The only worst case is for a general quadratic constraint that is dense and of dimensionality close to `n`, but it rarely happens in practice.

That's also why we keep triplet format for quadratic constraints across the data pipeline, and only use count-sort conversion for quadratic objective only.

mlubin (approved):

cpp/include/cuopt/linear_programming/io/mps_data_model.hpp
```diff
@@ -239,7 +239,9 @@ class mps_data_model_t {
    * - row identity and type (from ROWS),
    * - sparse linear coefficients (from COLUMNS),
    * - RHS value (from RHS),
-   * - quadratic matrix Q in COO (SoA: row, col, value) from QCMATRIX — one triplet per nonzero.
+   * - quadratic matrix Q in canonical COO (SoA: row, col, value): one triplet per variable pair,
+   *   sorted by (row, col). Off-diagonal cross terms store the full coefficient on x_i*x_j
```
chris-maes: nit: do these need to be sorted?

cpp/include/cuopt/linear_programming/io/mps_data_model.hpp
```diff
@@ -239,7 +239,9 @@ class mps_data_model_t {
    * - row identity and type (from ROWS),
    * - sparse linear coefficients (from COLUMNS),
    * - RHS value (from RHS),
-   * - quadratic matrix Q in COO (SoA: row, col, value) from QCMATRIX — one triplet per nonzero.
+   * - quadratic matrix Q in canonical COO (SoA: row, col, value): one triplet per variable pair,
```
chris-maes: Is SoA important? This seems like something an agent added.

cpp/src/io/lp_parser.cpp
```diff
@@ -257,9 +220,9 @@ class LpParseEngine {
   //   Objective:  must be followed by '/ 2'; the inner-coefficient convention
   //               is QUADOBJ-style 0.5 x^T Q x, so off-diagonals are halved
   //               and linear-inside-bracket terms also get /2.
-  //   Constraint: must NOT be followed by '/ 2'; coefficients are taken at
-  //               face value (x^T Q x); bracket must contain at least one
-  //               quadratic term.
+  //   Constraint: must NOT be followed by '/ 2'; Q coefficients are stored as upper triangular
```
chris-maes: Nit: `Q coefficients are stored as upper triangular from` -> `Only the upper triangle of Q is stored`

cpp/src/io/lp_parser.cpp
```diff
@@ -964,9 +927,8 @@ void LpParseEngine<i_t, f_t>::parse_quadratic_bracket(int outer_sign,
                        "constraint must contain at least one quadratic term",
                        peek().line);
 
-    // Coefficients are at face value — the post-pass that flushes the
-    // quadratic_constraint_block_t to the data model handles the symmetric
-    // expansion and the /2 split for off-diagonals.
+    // Q coefficients are stored as upper triangular form; canonicalization runs in
```
chris-maes: Same comment as above

cpp/src/io/mps_parser.cpp
```diff
@@ -176,6 +178,123 @@ std::string_view trim(std::string_view str)
...
+    cols.clear();
+    vals.clear();
+    return;
+  }
+
+  // Map every entry to its upper-triangular key (min(r,c), max(r,c)). Both
+  // orientations of an off-diagonal pair and any duplicate entries collapse to
+  // the same key, so summing the values per key yields the full x^T Q x
+  // coefficient for that variable pair.
+  std::vector<std::tuple<i_t, i_t, f_t>> triples;
```
chris-maes: I think it likely better to store this as three separate arrays: std::vector<i_t> i, std::vector<i_t> j, std::vector<f_t> x.

cpp/src/io/mps_parser.cpp
```diff
@@ -176,6 +178,123 @@ std::string_view trim(std::string_view str)
...
+  triples.reserve(n);
+  for (size_t t = 0; t < n; ++t) {
+    triples.emplace_back(std::min(rows[t], cols[t]), std::max(rows[t], cols[t]), vals[t]);
+  }
+
+  // Sort triples by (row, col) key.
+  std::sort(triples.begin(), triples.end(), [](const auto& a, const auto& b) {
```
chris-maes: Is this sort necessary?

cpp/src/io/mps_parser.cpp
```diff
@@ -176,6 +178,123 @@ std::string_view trim(std::string_view str)
...
+      cols[a],
+      cols[a],
+      rows[a],
+      vals[a],
+      vals[b]);
+  }
+}
+
+template <typename i_t, typename f_t>
+void canonicalize_coo_matrix(std::vector<i_t>& rows, std::vector<i_t>& cols, std::vector<f_t>& vals)
```
chris-maes, line 246: I'm concerned that the asympototics for this function are not good. Sorting introduces O(n * log n). I'm not sure it is necessary to sort. And if sorting must be done an efficient bucket sort that is O(m + n) should be used.

chris-maes (changes requested):
Let's get the asymptotics right on deduping COO before merging.

cpp/src/io/mps_parser.cpp
rg20: @yuwenchen95 I think you can use the code from: 

https://github.com/NVIDIA/cuopt/blob/08cdd8292908649ab06c1fe6a4f4e6073b820cf9/cpp/src/pdlp/cuopt_c.cpp#L88

cpp/include/cuopt/linear_programming/io/mps_data_model.hpp
yuwenchen95: We conduct sorting for efficient merging duplicates later.

cpp/src/io/mps_parser.cpp
yuwenchen95: We conduct sorting for efficient merging duplicates later.

chris-maes (approved):
Approving so this can be merged while I am out of office. Please try to get correct asymptotics and fast run time before merging.

yuwenchen95:
/ok to test [5fcbcc6](https://github.com/NVIDIA/cuopt/pull/1439/commits/5fcbcc63f7fd897cd94f1fc712b55cc69ecafce7)

PR #1449, opened 2026-06-22
Assign gRPC workers to distinct GPUs via cudaSetDevice

tmckayus:
Round-robin workers across visible CUDA devices at worker startup and initialize a per-process RMM pool (CUOPT_GIGABYTES_PER_PROC) so multiple workers can share a GPU without exhausting memory.

The logic to actually map a worker process to a specific GPU was missing previously (oversight).

### DIFF
--- a/cpp/src/grpc/server/grpc_worker.cpp
+++ b/cpp/src/grpc/server/grpc_worker.cpp
@@ -9,6 +9,60 @@
 #include "grpc_pipe_serialization.hpp"
 #include "grpc_server_types.hpp"
 
+#include <rmm/mr/cuda_memory_resource.hpp>
+#include <rmm/mr/pool_memory_resource.hpp>
+
+#include <memory>
+
+namespace {
+
+void init_worker_rmm_pool()
+{
+  int pool_gigs = 1;
+  if (const char* env = std::getenv("CUOPT_GIGABYTES_PER_PROC")) {
+    const int parsed = std::atoi(env);
+    if (parsed > 0) { pool_gigs = parsed; }
+  }
+
+  // Keep the pool alive for the lifetime of this worker process.
+  static std::unique_ptr<rmm::mr::pool_memory_resource> pool_mr;
+  static bool initialized = false;
+  if (initialized) { return; }
+
+  pool_mr = std::make_unique<rmm::mr::pool_memory_resource>(
+    rmm::mr::cuda_memory_resource(), static_cast<std::size_t>(pool_gigs) * (1ULL << 30));
+  rmm::mr::set_current_device_resource(*pool_mr);
+  initialized = true;
+
+  SERVER_LOG_INFO("[Worker] RMM pool size: %d GiB", pool_gigs);
+}
+
+}  // namespace
+
+void init_worker_cuda_environment(int worker_id)
+{
+  int device_count            = 0;
+  const cudaError_t count_err = cudaGetDeviceCount(&device_count);
+  if (count_err != cudaSuccess || device_count <= 0) {
+    SERVER_LOG_WARN("[Worker %d] cudaGetDeviceCount failed (%s); using default device",
+                    worker_id,
+                    cudaGetErrorString(count_err));
+    return;
+  }
+
+  const int device          = worker_id % device_count;
+  const cudaError_t set_err = cudaSetDevice(device);
+  if (set_err != cudaSuccess) {
+    SERVER_LOG_ERROR(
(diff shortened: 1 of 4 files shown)


cpp/src/grpc/server/grpc_worker_infra.cpp
```diff
@@ -8,6 +8,67 @@
 #include "grpc_pipe_serialization.hpp"
 #include "grpc_server_types.hpp"
 
+#include <cctype>
```
tmckayus, line 11: Note that these routines  below are simply for logging in the main server, so that we can log information about the number of workers and gpus being used without actually initializing CUDA before forking the worker processes (which causes issues). 

The workers themselves simply ask CUDA for available devices and apply a mod on worker id

ramakrishnap-nv (approved):

cpp/src/grpc/server/grpc_worker.cpp
```diff
@@ -9,6 +9,60 @@
...
+#include <rmm/mr/cuda_memory_resource.hpp>
+#include <rmm/mr/pool_memory_resource.hpp>
+
+#include <memory>
+
+namespace {
+
+void init_worker_rmm_pool()
+{
+  int pool_gigs = 1;
```
ramakrishnap-nv, line 24: Are we using 1GB by default as rmm pool memory?

tmckayus: I think so, let me double check that against the Python server

tmckayus: @ramakrishnap-nv yes the default in the Python server is 1GB, can be overridden with environment variable

cpp/src/grpc/server/grpc_worker.cpp
```diff
@@ -9,6 +9,76 @@
...
+  }
+  return pool_gigs;
+}
+
+void init_worker_rmm_pool()
+{
+  const int pool_gigs = parse_pool_gigs_env();
+
+  // Keep the pool alive for the lifetime of this worker process.
+  static std::unique_ptr<rmm::mr::pool_memory_resource> pool_mr;
```
akifcorduk, line 44: Do you think we will ever do multi-threading or is it always going to be multi-processing? If we plan to do multi threading, static variables might cause issues.

cpp/src/grpc/server/grpc_worker.cpp
```diff
@@ -9,6 +9,76 @@
...
+
+  const int device          = worker_id % device_count;
+  const cudaError_t set_err = cudaSetDevice(device);
+  if (set_err != cudaSuccess) {
+    SERVER_LOG_ERROR(
+      "[Worker %d] cudaSetDevice(%d) failed: %s", worker_id, device, cudaGetErrorString(set_err));
+    return false;
+  }
+
+  init_worker_rmm_pool();
```
akifcorduk, line 76: Shouldn't this be done once per device instead of once per worker, otherwise we are allocating a new memory resource and assign it to the device again which might cause issues when there are multiple workers using the same device?

cpp/src/grpc/server/grpc_worker.cpp
tmckayus: I think we'll always do multi-process, because it is the easiest way to cancel a job (we don't have cancel hooks in the solver) or restart a worker if it has CUDA errors or something.  And it makes it easy to map a GPU to an isolated entity

cpp/src/grpc/server/grpc_worker.cpp
tmckayus: good feedback, I'll take a look at that

tmckayus: @akifcorduk 
This is the way we did it in the Python server -- each process gets it own pool on a device, and that only happens once per process.  Not sure how we would share a pool across processes -- is there a simple way to do that?

If it might be a problem, the simple fix is to cap the number of workers to the number of GPUs available. This would guarantee only 1 pool per device (held by 1 process) unless you ran multiple servers. It would just prevent oversubscribing.

akifcorduk: Oh worker is a process. Sorry for the confusion, in cpp OMP worker keyword is per thread. Yes, one pool per process is needed, it won't work otherwise, a different process cannot access cuda memory.

akifcorduk (approved):
LGTM, thanks Trevor!

PR #1462, opened 2026-06-24
ci: fix unbound CONDA_PREFIX in test_doc_examples.sh

### LINKED ISSUE #1459: ci: test_doc_examples.sh fails silently in wheel-tests-cuopt-server due to unbound CONDA_PREFIX
`./ci/test_doc_examples.sh` is called from `ci/test_wheel_cuopt_server.sh` but fails silently at line 460 with:

```
./ci/test_doc_examples.sh: line 460: CONDA_PREFIX: unbound variable
```

The script references `CONDA_PREFIX`, which is only set in conda environments. In the wheel test path (`wheel-tests-cuopt-server`), there is no conda environment, so the variable is unbound. Because `test_wheel_cuopt_server.sh` uses `set +e` around the doc example step, the failure is swallowed and the job still reports success — but doc example tests are never actually run in the wheel-tests-cuopt-server CI job.

**Fix:** The script should either check whether `CONDA_PREFIX` is set before using it, or the doc example step should be skipped gracefully when running outside a conda environment.

Observed in PR #1457 run: job `wheel-tests-cuopt-server / 12.9.2, 3.14, amd64, ubuntu24.04, h100` (job ID 83257859948).

ramakrishnap-nv:
- Use `${CONDA_PREFIX:-}` so the variable defaults to empty when unset (wheel environments). The search loop skips empty entries, so conda behavior is unchanged. Fixes #1459.
- Add missing `Makefile` to `docs/cuopt/source/cuopt-c/mip/examples/`; without it, `test_doc_examples.sh` marked all MIP C examples as failed.


docs/cuopt/source/cuopt-c/mip/examples/Makefile
```diff
@@ -0,0 +1,148 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
jameslamb: What's the goal with this file?

It seems totally unrelated to the title / description / linked issue.

jameslamb (changes requested):
It seems you might have included something you didn't mean to, will re-review once you clarify.

docs/cuopt/source/cuopt-c/mip/examples/Makefile
ramakrishnap-nv: Apologies, so it started with doc testing failing silently and once I fixed it, there was another error where one of the test missed a Cmake file, so I fixed that, but missed to update the description.

jameslamb:
```suggestion
# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```

Ok got it, thanks. It's 2026, so let's please fix the copyright year here.

docs/cuopt/source/cuopt-c/mip/examples/Makefile
```diff
@@ -0,0 +1,148 @@
...
+CC = gcc
+
+# Set include and library paths
+# REQUIRED: Set these paths either:
+#   1. As environment variables
+#   2. On command line: make INCLUDE_PATH=... LIBCUOPT_LIBRARY_PATH=...
+#   3. Uncomment and edit below for your system:
+#
+# INCLUDE_PATH = /path/to/cuopt/include
+# LIBCUOPT_LIBRARY_PATH = /path/to/cuopt/lib
```
jameslamb: IMO it'd be simpler to just pick 1 way and document that that's how this works.

All this extra flexibility is just maintenance cost for no benefit unless you have a specific reason to want to support 3 different patterns.

The `make INCLUDE_PATH=...` form would be my preference.

jameslamb (approved):
Approving so you have what you need to merge, do what you want with my small suggestions.

docs/cuopt/source/cuopt-c/mip/examples/Makefile
ramakrishnap-nv: Done — copyright updated to 2026.

docs/cuopt/source/cuopt-c/mip/examples/Makefile
ramakrishnap-nv: Done — simplified to command-line only (`make INCLUDE_PATH=... LIBCUOPT_LIBRARY_PATH=...`), removed env-var and hardcoded options. Also removed the verbose help/list targets.

Iroy30 (approved):

PR #1465, opened 2026-06-25
chore(ci): exclude NVSkills-generated files from trailing-whitespace hook

ramakrishnap-nv:
NVSkills-Eval regenerates `skill-card.md` and `BENCHMARK.md` with trailing whitespace after `SKILL.md` changes; exclude them alongside `skill.oms.sig`.\n\nRelated: #1464

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -8,7 +8,13 @@ repos:
       - id: end-of-file-fixer
         exclude: ^(datasets|helmchart)/.*\.(mps|json|yaml|yml|txt)$|^skills/.*/skill\.oms\.sig$
       - id: trailing-whitespace
-        exclude: ^datasets/.*\.(mps|json|yaml|yml|txt)$|^skills/.*/skill\.oms\.sig$|^skills/.*/skill-card\.md$|^skills/.*/BENCHMARK\.md$
+        exclude: |
+          (?x)^(
+            ^datasets/.*\.(mps|json|yaml|yml|txt)$|
+            ^skills/.*/skill\.oms\.sig$|
+            ^skills/.*/skill-card\.md$|
+            ^skills/.*/BENCHMARK\.md$
+          )
       - id: check-builtin-literals
       - id: check-executables-have-shebangs
       - id: check-json
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,20 +1,21 @@
 # SPDX-FileCopyrightText: Copyright (c) 2019-2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 
+exclude: |
+  (?x)^(
+    skills/.*/skill\.oms\.sig|
+    skills/.*/skill-card\.md|
+    skills/.*/BENCHMARK\.md
+  )$
+
 repos:
   - repo: https://github.com/pre-commit/pre-commit-hooks
     rev: 'v6.0.0'
     hooks:
       - id: end-of-file-fixer
-        exclude: ^(datasets|helmchart)/.*\.(mps|json|yaml|yml|txt)$|^skills/.*/skill\.oms\.sig$
+        exclude: ^(datasets|helmchart)/.*\.(mps|json|yaml|yml|txt)$
       - id: trailing-whitespace
-        exclude: |
-          (?x)^(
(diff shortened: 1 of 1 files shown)


.pre-commit-config.yaml
```diff
@@ -8,7 +8,7 @@ repos:
       - id: end-of-file-fixer
         exclude: ^(datasets|helmchart)/.*\.(mps|json|yaml|yml|txt)$|^skills/.*/skill\.oms\.sig$
       - id: trailing-whitespace
-        exclude: ^datasets/.*\.(mps|json|yaml|yml|txt)$|^skills/.*/skill\.oms\.sig$
+        exclude: ^datasets/.*\.(mps|json|yaml|yml|txt)$|^skills/.*/skill\.oms\.sig$|^skills/.*/skill-card\.md$|^skills/.*/BENCHMARK\.md$
```
jameslamb: Why do this? Can't the process that modifies these files be modified to also not have any trailing whitespace?

jameslamb (commented):
Left a question, this PR is confusing to me.

.pre-commit-config.yaml
ramakrishnap-nv: This is generated or added by internal CI projects and it is controlled by nvcarps-team. So, it would take time to push them to update this.

bdice: Agree with James's question. Also I recommend breaking this up for readability. See example: https://github.com/rapidsai/cudf/blob/69dc0799906273230297f18c5f5aa980bee23935/.pre-commit-config.yaml#L66-L71
```suggestion
        exclude: |
          (?x)^(
            ^datasets/.*\.(mps|json|yaml|yml|txt)$|
            ^skills/.*/skill\.oms\.sig$|
            ^skills/.*/skill-card\.md$|
            ^skills/.*/BENCHMARK\.md$
          )
```

jameslamb: In our offline discussion, @ramakrishnap-nv said this has to be excluded because some internal process produces it alongside a signature (content-based checksum), and we don't have much control over that process.

If that's the case, that these files should never be modified, then they should be added to the top-level `exclude:` in this file, not just this one specific hook.

ramakrishnap-nv: Pushed an update

ramakrishnap-nv: Update: Agree regarding the exclusion, I have updated it, @jameslamb for viz

Ignore:
Top-level `exclude:` would skip these files from all hooks including `validate-skills` and `verify-copyright`, which we do want to run on `SKILL.md` changes (and the generated files alongside it). Hook-level exclusion keeps the other checks active. Applied @bdice's formatting suggestion.

ramakrishnap-nv: Updated — moved all three NVSkills-generated files (`skill.oms.sig`, `skill-card.md`, `BENCHMARK.md`) to the top-level `exclude:` and removed the per-hook patterns.

.pre-commit-config.yaml
```diff
@@ -1,14 +1,21 @@
 # SPDX-FileCopyrightText: Copyright (c) 2019-2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 
+exclude: |
+  (?x)^(
+    skills/.*/skill\.oms\.sig|
+    skills/.*/skill-card\.md|
+    skills/.*/BENCHMARK\.md
+  )$
```
jameslamb, line 9: Should this be `skills/.*`? Wouldn't `SKILL.md` (for example) changing also change the signature?

ramakrishnap-nv: Yes, in that case we need to run CI which would generate new files and signature associated with it and push as commit, for reference https://github.com/NVIDIA/cuopt/pull/1460#issuecomment-4800927129

ramakrishnap-nv: https://github.com/NVIDIA/cuopt/pull/1460/changes/ee81be4be29ac3d9095d2a57c861a3c15ab39654

ramakrishnap-nv: But we want to follow guidelines on all other things that we write and control like SKILL.md or any other underlying references, only want to skip which are generated.

jameslamb: ok

jameslamb (approved):
Going to say unhelpfully that this seems annoying to have to manage, but the changes are fine based on my understanding of how the process is working. Thanks for answering my questions.

PR #1468, opened 2026-06-26
Add support for sparsity exploitation for SOCs of large dimensionality

yuwenchen95:
## Description

Adds sparsity exploitation for second-order cone programs (SOCP) that contain
large-dimensional cones, so the barrier solver no longer materializes a dense
Hessian block for every cone.

- **Hybrid dense/sparse cone handling.** Cones with dimension smaller than a certain threshold
  keep the existing explicit dense Hessian block; larger cones now use a sparse
  rank-2 (Nesterov-Todd) representation in the augmented KKT system, with an
  expansion block whose RHS rows are zero-filled.
- **New SOCP device kernels** for sparse NT-scaling, sparse Hessian diagonal
  (`Hs`), and the sparse augmented matvec, plus assembly of the sparse expansion
  into the augmented system.
- **Segmented-sum reduction** over cone vectors that distinguishes small/medium/large
  cones, with an ILP path (multiple accumulators per thread) for large "medium" cones.
- **Initial-point bug fix** removes an incorrect sign flip of the augmented-system diagonal during initial-point setup. With this fix, the barrier method solves `ns1688926` in Han’s LP dataset and `QGROW15` in Maros–Mészáros set.
- **Kernel fusion.** Sparse SOC augmented-value assembly (packed `Hs` diagonal,
  Hessian scatter, and rank-2 expansion couplings) is fused into a single
  entry-parallel kernel launch per refactor. Cone step-length computation is
  likewise fused so each cone accumulates its three tail scalars and writes
  `alpha` in one size-aware pass instead of separate segmented sums.
- **GPU CSR KKT buildup.** Augmented-system CSR construction moves from host
  loops onto the device (`csr_kkt_build.cuh`): metadata (dense/sparse routing,
  block offsets) and row fill/count run as CUDA kernels, so structure and
  values are built without staging the full pattern through the CPU.
 
  

### Improvements

Achieve significant speedup on large-dimensional SOCP (especially large-cone instances such as chainsing-* and several sched_* / 2013_firL2* models).

| Instance | Solve time (s) | Notes |
| -------- | -------------- | ----- |
| chainsing-10000-3 | 180.08 → 0.70 | 257× faster |
| chainsing-10000-2 | 143.47 → 0.63 | 228× faster |
| chainsing-50000-3 | 500 → 2.81 | out of memory ->solved |
| chainsing-50000-2 | 500 → 3.07 | out of memory ->solved |
| sched_200_100_scaled | 18.92 → 0.61 | 31× faster |
| 2013_firL2L1alph | 82.25 → 3.79 | 22× faster |
| sched_200_100_orig | 21.71 → 1.01 | 22× faster |
| sched_100_100_scaled | 3.28 → 0.38 | 8.6× faster |
| sched_100_100_orig | 3.94 → 0.57 | 6.9× faster |
| 2013_firL2Linfeps | 43.10 → 7.49 | 5.8× faster |
| chainsing-1000-3 | 1.19 → 0.21 | 5.7× faster |
| sched_100_50_scaled | 1.34 → 0.29 | 4.6× faster |
| sched_100_50_orig | 1.71 → 0.45 | 3.8× faster |
| chainsing-1000-2 | 0.74 → 0.21 | 3.5× faster |
| sched_50_50_scaled | 0.52 → 0.22 | 2.4× faster |
| sched_50_50_orig | 0.72 → 0.36 | 2.0× faster |
| 2013_firL2a | 18.90 → 13.25 | 1.4× faster |
| 2013_wbNRL | 13.98 → 12.00 | 1.2× faster |
| nb_L2 | 0.29 → 0.22 | 1.3× faster |
### Regressions
| Instance | Solve time (s) | Notes |
| -------- | -------------- | ----- |
| 2013_firL1Linfalph | 31.46 → 35.06 | 1.11× slower; only loss vs `main` |


cpp/include/cuopt/mathematical_optimization/pdlp/solver_settings.hpp
```diff
@@ -283,6 +283,8 @@ class pdlp_solver_settings_t {
   bool eliminate_dense_columns{true};
   pdlp_precision_t pdlp_precision{pdlp_precision_t::DefaultPrecision};
   bool barrier_iterative_refinement{true};
+  bool barrier_csr_ir_matvec{false};
```
rg20: Lets remove this if its faster always faster to avoid the operator based version.

cpp/src/barrier/cusparse_view.hpp
```diff
@@ -73,4 +75,31 @@ class cusparse_view_t {
   rmm::device_scalar<f_t> d_minus_one_;
   rmm::device_scalar<f_t> d_zero_;
 };
+
+// cuSparse CSR SpMV view for the augmented KKT matrix stored in device_csr_matrix_t.
+// The sparsity pattern is fixed; numerical values are read from matrix.x in place.
+template <typename i_t, typename f_t>
+class augmented_cusparse_view_t {
```
rg20: Why do you need a separate class for this?  Is n't this same as cusparse_view_t ?

cpp/src/barrier/iterative_refinement.hpp
```diff
@@ -89,7 +88,8 @@ f_t vector_norm2(const rmm::device_uvector<f_t>& x)
 template <typename i_t, typename f_t, typename T>
 f_t iterative_refinement_simple(T& op,
```
rg20, line 58: this is not being used anywhere in the code right?

cpp/include/cuopt/mathematical_optimization/constants.h
```diff
@@ -48,6 +48,8 @@
 #define CUOPT_ORDERING                             "ordering"
 #define CUOPT_BARRIER_DUAL_INITIAL_POINT           "barrier_dual_initial_point"
 #define CUOPT_BARRIER_ITERATIVE_REFINEMENT         "barrier_iterative_refinement"
+#define CUOPT_BARRIER_CSR_IR_MATVEC                "barrier_csr_ir_matvec"
+#define CUOPT_BARRIER_SOC_THRESHOLD                "barrier_soc_threshold"
```
rg20: is this parameter needed?

yuwenchen95: These are parameters I want to tune, so I want to expose them during the developing phase. They would be removed when I'm going to merge PR.

cpp/src/barrier/iterative_refinement.hpp
yuwenchen95: It's not been used but easier for test purpose when I try to track the error in IR. I would prefer to keep it.

cpp/include/cuopt/mathematical_optimization/pdlp/solver_settings.hpp
yuwenchen95: So far, I haven't see significant speedup for others but fail optimality for several examples in cblib. It's better to keep it as an option but not default until I can improve the numerical stability of it.

rg20: which problem are you seeing the issue with?

cpp/src/barrier/iterative_refinement.hpp
rg20: Sure!

cpp/include/cuopt/mathematical_optimization/pdlp/solver_settings.hpp
rg20: We want to avoid code divergence as much as possible.   This definitely should not be a user facing option.

yuwenchen95: > We want to avoid code divergence as much as possible. This definitely should not be a user facing option.

I agree not to expose it when the PR is merged.

yuwenchen95: > which problem are you seeing the issue with?

`sched_` problems in cblib suffer numerical degradation on it.

cpp/src/barrier/cusparse_view.hpp
yuwenchen95: Good point. The previous `cusparse_view_t ` was limited to CPU matrix initialization only, but you're right we don't create a separate class for the augmented system. I have unified them into a new `cusparse_view_t` class.

ramakrishnap-nv (approved):
Approving CMAKE changes

cpp/include/cuopt/mathematical_optimization/pdlp/solver_settings.hpp
rg20: For QP and QCQP, there is no performance impact if I change this to true.

yuwenchen95: Mine failed on`sched_100_100_orig`, it is probably affected by non-determinism in cudss if the same problem is solved on your machine.

The performance improvement is not significant in absolute time `O(ms)`, so the performance is nearly the same on the benchmarks, but will make slightly difference for the solve time of a portfolio optimization problem.

cpp/src/barrier/barrier.cu
```diff
@@ -40,6 +40,8 @@
 #include <optional>
 #include <span>
 
+#include <cstdio>
```
chris-maes: Nit: includes go from closest to the code to farthest away. So cstdio should be at the end. 

Also why do you need this include?

cpp/src/barrier/barrier.cu
```diff
@@ -512,9 +544,20 @@ class iteration_data_t {
...
         d_augmented_rhs_.resize(augmented_size, stream_view_);
         d_augmented_soln_.resize(augmented_size, stream_view_);
+        d_aug_x1_.resize(lp.num_cols, stream_view_);
+        d_aug_x2_.resize(lp.num_rows, stream_view_);
+        d_aug_y1_.resize(lp.num_cols, stream_view_);
+        d_aug_y2_.resize(lp.num_rows, stream_view_);
+        d_aug_y_exp_.resize(augmented_expansion_count(), stream_view_);
+        d_aug_y_exp_orig_.resize(augmented_expansion_count(), stream_view_);
+        if (settings.barrier_csr_ir_matvec) {
+          settings.log.printf("IR matvec                     : csr\n");
```
chris-maes: I'm not sure we should put this in the log. And if we do we should write out IR as "Iterative Refinement".

cpp/src/barrier/barrier.cu
```diff
@@ -684,52 +743,95 @@ class iteration_data_t {
...
+        // TODO: Once form_augmented() is moved to GPU, we can avoid D2H round-trip.
         cone_offsets_host.resize(n_cones + 1);
-        cone_block_offsets_host.resize(n_cones + 1);
         raft::copy(
           cone_offsets_host.data(), cones().cone_offsets.data(), n_cones + 1, stream_view_);
         handle_ptr->sync_stream();
+
+        cone_sparse_idx.assign(n_cones, i_t(-1));
+        if (n_sparse_soc > 0) {
+          // TODO: Once form_augmented() is moved to GPU, we can avoid D2H round-trip for
```
chris-maes: Can we do this TODO now? 
Search the code for: 
`cub::DeviceScan::ExclusiveSum(
    nullptr, scan_bytes, row_counts.data(), Arow.row_start.data(), m, stream);`
to see how to do prefix sum on the GPU.

cpp/src/barrier/barrier.cu
```diff
@@ -684,52 +743,95 @@ class iteration_data_t {
...
+        dense_cone_block_offsets_host = {0};
+        dense_cone_block_offsets_host.reserve(n_cones + 1);
+        i_t dense_idx = 0;
         for (i_t k = 0; k < n_cones; ++k) {
-          const auto q_k                 = cone_offsets_host[k + 1] - cone_offsets_host[k];
-          cone_block_offsets_host[k + 1] = cone_block_offsets_host[k] + q_k * q_k;
+          if (cone_sparse_idx[k] < 0) {
+            dense_cone_idx_by_cone[k] = dense_idx;
+            dense_cone_ids_host.push_back(k);
+            const auto q_k = cone_offsets_host[k + 1] - cone_offsets_host[k];
```
chris-maes: Avoid `auto` for base types. I think this should be `const i_t q_k = ...`

cpp/src/barrier/barrier.cu
```diff
@@ -822,27 +976,77 @@ class iteration_data_t {
...
         for (i_t k = n; k < n + m; ++k) {
-          // A block, we can use AT in csc directly
           augmented_CSR.row_start[k] = q;
           const i_t l                = k - n;
           const i_t col_beg          = AT.col_start[l];
           const i_t col_end          = AT.col_start[l + 1];
-          for (i_t p = col_beg; p < col_end; ++p) {
-            augmented_CSR.j[q]   = AT.i[p];
-            augmented_CSR.x[q++] = AT.x[p];
+          for (i_t idx = col_beg; idx < col_end; ++idx) {
```
chris-maes: The code uses `p` and `q` for loop indexs into the CSC and CSR arrays. Avoid `idx` here for consistency.

cpp/src/barrier/barrier.cu
```diff
@@ -822,27 +976,77 @@ class iteration_data_t {
...
+        for (i_t s = 0; s < n_sparse_soc; ++s) {
+          const i_t k   = sparse_cone_ids_host[s];
+          const i_t q_k = static_cast<i_t>(cone_offsets_host[k + 1] - cone_offsets_host[k]);
+          const i_t cone_col_start = cone_start() + static_cast<i_t>(cone_offsets_host[k]);
+          const size_t flat_base   = sparse_entry_offsets[s];
+          const i_t prow_v         = n + m + 2 * s;
+          const i_t prow_u         = n + m + 2 * s + 1;
+
+          augmented_CSR.row_start[prow_v] = q;
+          for (i_t j = 0; j < q_k; ++j) {
```
chris-maes: We use `j` for column indices. You might want to use a different variable here, unless this corresponds to a column index.

cpp/src/barrier/barrier.cu
```diff
@@ -822,27 +976,77 @@ class iteration_data_t {
...
+        augmented_CSR.row_start[factorization_size] = q;
+        augmented_CSR.nz_max                        = q;
         augmented_CSR.j.resize(q);
         augmented_CSR.x.resize(q);
-        i_t expected_nnz = 2 * nnzA + (n - m_c) + total_block_nnz + m + off_diag_Qnz;
+        i_t expected_nnz =
+          2 * nnzA + (n - m_c) + dense_soc_kkt_nnz + m + off_diag_Qnz + sparse_soc_kkt_nnz;
         settings_.log.debug("augmented nz %d predicted %d\n", q, expected_nnz);
-        cuopt_assert(q == expected_nnz, "augmented nnz != predicted");
+        cuopt_assert(q == expected_nnz, "augmented nz != predicted");
```
chris-maes: nnz is `number of nonzeros`. Not sure there is a need to change this comment in the assert.

cpp/src/barrier/barrier.cu
```diff
@@ -2256,7 +2658,8 @@ int barrier_solver_t<i_t, f_t>::initial_point(iteration_data_t<i_t, f_t>& data)
     } op(data);
 
     if (settings.barrier_iterative_refinement) {
-      iterative_refinement<i_t, f_t, op_t>(op, rhs, soln);
+      const f_t ir_tol = data.has_sparse_cones() ? f_t(1e-12) : f_t(1e-8);
```
chris-maes, line 2261: Asking for 1e-12 seems pretty tight here.

cpp/src/barrier/barrier.cu
```diff
@@ -2583,21 +2984,30 @@ void barrier_solver_t<i_t, f_t>::gpu_compute_residual_norms(const rmm::device_uv
 }
 
 template <typename i_t, typename f_t>
-f_t barrier_solver_t<i_t, f_t>::compute_nonnegative_step_length(iteration_data_t<i_t, f_t>& data,
-                                                                const rmm::device_uvector<f_t>& x,
-                                                                const rmm::device_uvector<f_t>& dx)
+std::pair<f_t, f_t> barrier_solver_t<i_t, f_t>::compute_nonnegative_step_length_pair(
```
chris-maes, line 2587: Does this result in a performance increase? If not, consider reverting back to the original code.

cpp/src/barrier/barrier.cu
```diff
@@ -2834,18 +3249,6 @@ i_t barrier_solver_t<i_t, f_t>::gpu_compute_search_direction(iteration_data_t<i_
     raft::copy(data.d_r1_prime_.data(), data.d_tmp3_.data(), data.d_tmp3_.size(), stream_view_);
   }
 
-  if (!use_augmented) {
```
chris-maes, line 2837: Why is this no longer needed?

cpp/src/barrier/barrier.cu
```diff
@@ -3454,46 +3886,52 @@ void barrier_solver_t<i_t, f_t>::compute_target_mu(
...
-    step_dual_aff   = std::min(step_dual_aff, cone_d);
+    step_primal_aff = std::min(step_primal_aff, cone_combined);
+    step_dual_aff   = std::min(step_dual_aff, cone_combined);
   }
 
   if (data.Q.n > 0 || has_soc) {
     step_primal_aff = step_dual_aff = std::min(step_primal_aff, step_dual_aff);
   }
 
-  // Compute complementarity_xz_aff_sum = sum(x_aff * z_aff),
```
chris-maes, line 3481: Please restore this lost comment.

cpp/src/barrier/barrier.cu
```diff
@@ -3503,20 +3941,26 @@ void barrier_solver_t<i_t, f_t>::compute_target_mu(
     data.d_x_.size(),
     stream_view_);
 
-  // Here the update of w_aff and v_aff are done temporarily and sum of their products is
```
chris-maes, line 3506: Please restore this lost comment.

chris-maes (changes requested):
Thanks for this nice performance improvement. 

I haven't made it through the whole PR. But I don't think we should merge as is with two different iterative refinement modes. We should try to converge on a single way to do iterative refinement. 

Also, please run all the barrier benchmarks and report geomean results in the commit message to ensure there are no regressions. Ideally, these runs should be on the lego machines.

cpp/src/barrier/barrier.cu
yuwenchen95: Yes, we could. I will try to add it.

cpp/src/barrier/barrier.cu
yuwenchen95: This is for debugging purpose and I should remove it.

cpp/src/barrier/barrier.cu
yuwenchen95: It's related to I/O in `cholesky_debug_check`, which only executes in debug mode. Missing it would crash the debug mode.

cpp/src/barrier/barrier.cu
yuwenchen95: This is necessary when we have sparse socs, where KKT matrix has worser conditioning number after sparsification. I only enable it for sparse socs so the performance for dense soc blocks won't have performance regression.

cpp/src/barrier/barrier.cu
yuwenchen95: Yes, since we are always addressing step size for primal and dual pair of same dimensionality, we can reduce kernel launches from `2` to `1`. It's a small performance trick on GPU.

cpp/src/barrier/barrier.cu
yuwenchen95: I'm polishing code to remove unnecessary branching. The code block has been moved to a branching later with the same nvtx range `"Barrier: GPU compute H"`.

yuwenchen95:
> Thanks for this nice performance improvement.
> 
> I haven't made it through the whole PR. But I don't think we should merge as is with two different iterative refinement modes. We should try to converge on a single way to do iterative refinement.
> 
> Also, please run all the barrier benchmarks and report geomean results in the commit message to ensure there are no regressions. Ideally, these runs should be on the lego machines.

I have cleaned up the iterative refinement path and also supported CSR KKT build-up on GPU.

cpp/src/barrier/barrier.cu
chris-maes: How much does this impact the geomean of the total solve time?

cpp/src/dual_simplex/scaling.cpp
```diff
@@ -8,6 +8,8 @@
 #include <dual_simplex/scaling.hpp>
```
chris-maes, line 8: This looks like debug / profiling code. Can you revert changes to this file?

cpp/src/dual_simplex/presolve.cpp
```diff
@@ -13,6 +13,8 @@
 #include <dual_simplex/solve.hpp>
```
chris-maes, line 13: This looks like debug / profiling code. Can you revert changes to this file?

cpp/src/barrier/iterative_refinement.hpp
```diff
@@ -23,9 +23,10 @@
```
chris-maes, line 23: Let's revert changes to this file that add the `nvtx::range`

cpp/src/barrier/barrier.cu
```diff
@@ -684,201 +749,92 @@ class iteration_data_t {
...
-        }
-        augmented_CSR.row_start[n + m] = q;
-        augmented_CSR.nz_max           = q;
-        augmented_CSR.j.resize(q);
-        augmented_CSR.x.resize(q);
-        i_t expected_nnz = 2 * nnzA + (n - m_c) + total_block_nnz + m + off_diag_Qnz;
-        settings_.log.debug("augmented nz %d predicted %d\n", q, expected_nnz);
-        cuopt_assert(q == expected_nnz, "augmented nnz != predicted");
-        cuopt_assert(A.col_start[n] == AT.col_start[m], "A nz != AT nz");
+      augmented_csr_side_buffers_t<i_t, f_t> side{d_augmented_diagonal_indices_,
```
chris-maes: What is `side_buffers`? Should this be `size_buffers`? That is are you trying to correctly size some vectors?

cpp/src/barrier/barrier.cu
```diff
@@ -268,6 +280,15 @@ class iteration_data_t {
       d_augmented_diagonal_indices_(0, lp.handle_ptr->get_stream()),
       d_cone_csr_indices_(0, lp.handle_ptr->get_stream()),
       d_cone_Q_values_(0, lp.handle_ptr->get_stream()),
+      d_dense_cone_diag_csr_indices_(0, lp.handle_ptr->get_stream()),
```
chris-maes: Maybe it would make sense to store all these in a specific data structure?

cpp/src/barrier/barrier.cu
```diff
@@ -684,201 +749,92 @@ class iteration_data_t {
...
+                                                  d_sparse_exp_u_row_,
+                                                  d_sparse_expansion_D_};
+
+      raft::device_span<const i_t> element_cone_ids_span{};
+      raft::device_span<const size_t> cone_offsets_span{};
+      raft::device_span<const i_t> sparse_cone_ids_span{};
+      raft::device_span<const i_t> sparse_entry_offsets_span{};
+      size_t n_sparse_cone_entries = 0;
+      if (has_soc) {
+        element_cone_ids_span     = cuopt::make_span(cones().element_cone_ids);
```
chris-maes: Maybe all these should be stored in a separate data structure?

cpp/src/barrier/barrier.cu
```diff
@@ -1921,6 +1946,10 @@ class iteration_data_t {
   csc_matrix_t<i_t, f_t> ADAT;
   // csc_matrix_t<i_t, f_t> augmented;
   device_csr_matrix_t<i_t, f_t> device_augmented;
+  std::optional<device_csc_matrix_t<i_t, f_t>> device_A_csc_;
```
chris-maes: Do we need `std::optional` on all of these? Can't we just have them be empty matrices if they aren't used?

cpp/src/barrier/barrier.cu
```diff
@@ -1950,6 +1979,15 @@ class iteration_data_t {
   rmm::device_uvector<i_t> d_augmented_diagonal_indices_;
   rmm::device_uvector<i_t> d_cone_csr_indices_;
   rmm::device_uvector<f_t> d_cone_Q_values_;
+  rmm::device_uvector<i_t> d_dense_cone_diag_csr_indices_;
```
chris-maes: I'd. group these into a data structure.

cpp/src/barrier/barrier.cu
chris-maes: If it is less than 1%, please revert the code.

cpp/src/barrier/barrier.cu
```diff
@@ -4194,23 +4283,16 @@ lp_status_t barrier_solver_t<i_t, f_t>::solve(f_t start_time, lp_solution_t<i_t,
...
+      {
+        raft::common::nvtx::range fun_scope("Barrier: search_direction (affine)");
+        status =
+          gpu_compute_search_direction(data, dual_perturb, primal_perturb, max_affine_residual);
+      }
       if (settings.concurrent_halt != nullptr && *settings.concurrent_halt == 1) {
         settings.log.printf("Barrier solver halted\n");
         return lp_status_t::CONCURRENT_LIMIT;
       }
-      // D2D: save affine directions on device (same stream — no sync needed)
```
chris-maes, line 4203: Why is this no longer needed?

cpp/src/barrier/barrier.cu
```diff
@@ -3962,11 +4055,7 @@ lp_status_t barrier_solver_t<i_t, f_t>::check_for_suboptimal_solution(
       data.relative_dual_residual_save < settings.barrier_relaxed_optimality_tol &&
       data.relative_complementarity_residual_save < settings.barrier_relaxed_complementarity_tol) {
     settings.log.printf("Restoring previous solution\n");
-    raft::copy(data.x.data(), data.d_x_.data(), data.d_x_.size(), stream_view_);
```
chris-maes, line 3965: Why is this no longer needed?

chris-maes:
Please try to minimize unnecessary changes in the PR. For instance many of the changes are additions on nvtx::ranges here. This creates a lot of churn in the code and makes the PR longer and more difficult to review.

cpp/src/barrier/barrier.cu
yuwenchen95: The copies were moved into `compute_target_mu` that we fused computing complementarity and saving affine direction together. This can reduce memory access times of `dx, dz, dw, dv` by one.

cpp/src/barrier/barrier.cu
yuwenchen95: They are not removed but wrapped into the function `restore_saved_iterate`.

cpp/src/barrier/barrier.cu
yuwenchen95: We can save 4 kernel launches and 4 D2H synchronizations per iteration in total (`x,z` and `w,v` pairs in both affine and corrector steps). It is worth for fast portfolio optimization problem, where per iteration time is `< 4ms` and the this can save `>0.1ms`.

Besides, I don't think it would harm in other cases since we always have to address nonnegative variables in primal-dual pairs of the same dimensionality, e.g. `x,z` and `w,v`.

yuwenchen95:
> Please try to minimize unnecessary changes in the PR. For instance many of the changes are additions on nvtx::ranges here. This creates a lot of churn in the code and makes the PR longer and more difficult to review.

Profiling `nvtx::ranges` have been removed as requested. Vectors and views related to cones have been grouped into new structs `cone_kkt_data_t`, `cone_kkt_views_t` and `sparse_cone_views_t`.

cpp/src/barrier/barrier.cu
```diff
@@ -446,6 +467,9 @@ class iteration_data_t {
       if (n_upper_bounds > 0) {
         settings.log.printf("Upper bounds                : %d\n", n_upper_bounds);
       }
+
+      d_dw_aff_.resize(n_upper_bounds, stream_view_);
```
chris-maes: Can this move into the constructor? Or i s n_upper_bounds not known until later?

cpp/src/barrier/barrier.cu
```diff
@@ -684,201 +737,71 @@ class iteration_data_t {
...
+      raft::common::nvtx::range scope("Barrier: augmented: device CSR build");
 
-      i_t new_nnz = 2 * nnzA + n + m + nnzQ + total_block_nnz;  // conservative estimate of nnz
-      csr_matrix_t<i_t, f_t> augmented_CSR(n + m, n + m, new_nnz);
-      std::vector<i_t> augmented_diagonal_indices(n + m, -1);
-      std::vector<i_t> cone_csr_indices_host(total_block_nnz, -1);
-      std::vector<f_t> cone_Q_values_host(total_block_nnz, f_t(0));
-      i_t q            = 0;
-      i_t off_diag_Qnz = 0;
+      const size_t n_sparse_entries = has_soc && p > 0 ? cones().n_sparse_cone_entries : size_t{0};
```
chris-maes: Nit: probably better to call this `n_sparse_cone_entries`

chris-maes:
The code looks good to me. Please run final benchmarks on the LP, QP, QCP and SOCP datasets, and report the performance improvements before merging.

chris-maes (approved):
LGTM! Approved! Thanks @yuwenchen95

PR #1486, opened 2026-06-26
ci(test): enable pytest-xdist for cuopt Python tests

ramakrishnap-nv:
## Summary

- Add `pytest-xdist` to `test_python_common` dependencies (propagated to conda envs and pyproject.toml files via dependency file generator)
- Pass `-n auto` to the non-nightly cuopt pytest invocation in `ci/run_cuopt_pytests.sh`
- Server tests (`run_cuopt_server_pytests.sh`) are unchanged — they bind a fixed port (`localhost:18900`) that would conflict across parallel workers

The script already `cd`s into `python/cuopt/cuopt` where xdist + coverage is known to work (noted in the existing comment on line 7).

## Test plan

- [ ] CI `conda-python-tests` job shows parallel test workers in the log
- [ ] Test results and coverage report are complete (no missing tests)
- [ ] `conda-python-tests` wall time is reduced compared to baseline
- [ ] Server tests are unaffected


tmckayus (approved):

gforsyth (approved):
This looks ready to go.  My only comment is that you could possibly use `-n auto` and take advantage of more cores/processes when running on a beefier runner, but that shouldn't block this.

ramakrishnap-nv:
> This looks ready to go. My only comment is that you could possibly use `-n auto` and take advantage of more cores/processes when running on a beefier runner, but that shouldn't block this.

I had tried auto, but due to GPU contention and few other issues, it was failing. So I reduced it just to 2 to be safe.

bdice:
Generally we use `-n 6` or `-n 8` in most of RAPIDS. The safe maximum depends on the GPU memory usage of your tests, not the number of CPU cores! Try to design it to run safely on a 16GB GPU.

ramakrishnap-nv:
I will try to experiment and see which one would work the best

PR #1489, opened 2026-06-29
skills: consolidate three numerical-optimization API skills into one

ramakrishnap-nv:
Merges `cuopt-numerical-optimization-api-python`, `-api-c`, and `-api-cli` into a single `cuopt-numerical-optimization-api` skill.

- Shared content (LP/MILP/QP selection, QP rules, dual values, common issues) in `SKILL.md`
- Interface-specific syntax and examples in `references/python_api.md`, `references/c_api.md`, `references/cli_api.md`
- Assets reorganized under `assets/python/`, `assets/c/`, `assets/cli/`
- Evals merged (9 total)
- Updated `AGENTS.md`, `marketplace.json`, and cross-references


ramakrishnap-nv:
## Skill evaluation — PR #1489 (merged `cuopt-numerical-optimization-api`) vs latest `main`

Ran an agent-based eval harness over **99 numeric LP/MILP benchmark problems** (3 attempts each), comparing this PR's merged skill against main's separate skills and a no-skills baseline. Agent = Opus 4.8 via the NVIDIA inference endpoint with a `run_python` tool executing cuOpt in `nvidia/cuopt:26.8.0a-cu13`; a separate LLM judge verifies the **objective value only**. The PR config was run **twice** (unchanged) to gauge run-to-run variance.

| Config | pass@1 | pass@3 | avg steps | avg tokens | avg wall |
|---|---:|---:|---:|---:|---:|
| PR #1489 — run 1 | 83.8% | 91.9% | 4.4 | 33,938 | 68.7 s |
| PR #1489 — run 2 | 87.5% | 94.9% | 4.5 | 31,102 | 50.8 s |
| latest `main` | 84.8% | 94.9% | 3.9 | 30,459 | 51.8 s |
| no-skills | 81.1% | 88.9% | 4.9 | 31,602 | 85.8 s |

### Findings

- **No regression from the consolidation.** Across two full runs the merged skill matches main (pass@3 94.9% = 94.9%; pass@1 within noise). Both skill configs beat no-skills on accuracy, steps, and time.
- **The earlier PR-vs-main gap was sampling noise** — run-to-run swing is ~±3–4 pp. The 3 evals that flipped between the two identical PR runs (`lpmilp-020`, `lpmilp-047`, `lpmilp-060`) are exactly the borderline ones.
- **The skill is used fully**: the agent loads the merged skill at the same rate as main's separate skill and reliably opens `references/python_api.md` — progressive disclosure works.
- **Minor structural token cost**: the merged skill's routing needs a second tool call (`SKILL.md` → `references/python_api.md`) to reach the Python details that main's self-contained `SKILL.md` delivers in one load — i.e. ≥1 extra tool-call round when the skill is used. The effect is small (net ~+11% tokens, and it washed out in run 2). *Optional:* inline the common Python path into `SKILL.md` to remove it.
- **Benchmark note (not a skill issue)**: eval `lpmilp-047` is a goal-programming problem — pass/fail is a coin-flip on whether the agent reports the goal-deviation objective (0, legitimately) vs the profit (11250, the ground truth).

_Eval harness and raw per-attempt results available on request._

skills/cuopt-numerical-optimization-api/SKILL.md
```diff
@@ -0,0 +1,95 @@
...
+---
+
+
+# cuOpt Numerical Optimization API
+
+Model and solve LP, MILP, and QP problems using NVIDIA cuOpt's GPU-accelerated solver.
+
+## Interface Selection
+
+Choose the reference for the user's interface:
```
mlubin, line 25: External modeling tools are also user interfaces to cuOpt, do we mention them here?

ramakrishnap-nv: We cover them as thirdparty in docs, but not here. It would be a good idea.  should we add a section underneath this or a new skill would be better ?

mlubin: I don't think we need a new skill, just worth mentioning that the third party interfaces exist.

ramakrishnap-nv: Done — added a short note under **Interface Selection** that cuOpt works as a solver backend for AMPL, GAMS/GAMSPy, PuLP, and JuMP (near-zero code changes), pointing to the Third-Party Modeling Languages docs. No new skill, per your suggestion. Commit 6217a923.

skills/cuopt-numerical-optimization-api/SKILL.md
```diff
@@ -0,0 +1,101 @@
...
+| Interface | When to use | Reference |
+|-----------|-------------|-----------|
+| **Python** | User is writing Python code | [references/python_api.md](references/python_api.md) |
+| **C / C++** | User is embedding in a C/C++ application | [references/c_api.md](references/c_api.md) |
+| **CLI** | User is solving from MPS files on the command line | [references/cli_api.md](references/cli_api.md) |
+
+If the interface is not yet clear, ask before writing any code.
+
+**Already using a modeling language?** cuOpt also works as a solver backend for third-party
+modeling tools — **AMPL, GAMS / GAMSPy, PuLP, and JuMP** — with near-zero code changes (point the
```
Iroy30: if it's an exhaustive list maybe we should add pyomo and cvxpy

ramakrishnap-nv: I was thinking same whether we want to maintain multiple lists, one in doc, readme and one here. May be for now I will add and we can revisit it

Iroy30 (approved):
minor comment, looks good

mlubin (approved):

skills/cuopt-numerical-optimization-api/SKILL.md
ramakrishnap-nv: Added Pyomo and CVXPY to the list under Interface Selection (commit 5403868c). Verified both are real cuOpt interfaces — CVXPY is documented (LP/QP, plus beta QCQP/SOCP), and Pyomo has a CI-tested `cuopt_direct` integration. Agreed on the multiple-lists concern; noted to revisit consolidating the README/docs/skill lists later.

ramakrishnap-nv: Also propagated Pyomo and CVXPY to the docs so the skill and docs stay consistent (commit 71e2c370): added Pyomo + CVXPY sections on the Third-Party Modeling Languages page and to the introduction list, added Pyomo to the LP/MILP list in milp-features, and a Pyomo row to the convex-features table. CVXPY is intentionally left out of milp-features since it's convex-only (no MILP). Agree we should eventually maintain a single source for this list rather than three.

PR #1494, opened 2026-06-30
Improve cuopt-numerical-optimization-api-cli

mlubin:
cuopt_cli accepts LP and QPS files in addition to MPS (dispatched by extension), and the LP/MPS parsers both support quadratic objectives. Update the CLI skill to reflect this, and replace the partial MPS/LP format specs with guidance to define models through the cuOpt Python API or a preferred modeling interface rather than hand-authoring files. This is industry standard practice.

### DIFF
--- a/skills/cuopt-numerical-optimization-api-cli/SKILL.md
+++ b/skills/cuopt-numerical-optimization-api-cli/SKILL.md
@@ -23,24 +23,13 @@ Confirm problem type and formulation (variables, objective, constraints, variabl
 
 This skill is **CLI only** (MPS or LP file input).
 
-## Input formats
-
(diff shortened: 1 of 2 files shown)


ramakrishnap-nv:
> cuopt_cli accepts LP and QPS files in addition to MPS (dispatched by extension), and the LP/MPS parsers both support quadratic objectives. Update the CLI skill to reflect this, and replace the partial MPS/LP format specs with guidance to define models through the cuOpt Python API or a preferred modeling interface rather than hand-authoring files. This is industry standard practice.

@mlubin if you don't mind me asking, may I ask you to create this PR using branch in the main repo, this would enable use to run evaluation and scan for the skill changes. And don't bother to update skill card, it will get updated by skill signing.

mlubin:
Done: #1495

PR #1495, opened 2026-06-30
Improve cuopt-numerical-optimization-api-cli

mlubin:
cuopt_cli accepts LP and QPS files in addition to MPS (dispatched by extension), and the LP/MPS parsers both support quadratic objectives. Update the CLI skill to reflect this, and replace the partial MPS/LP format specs with guidance to define models through the cuOpt Python API or a preferred modeling interface rather than hand-authoring files. This is industry standard practice.


ramakrishnap-nv (approved):

ramakrishnap-nv:
Removed my approval so that we can wait for the next commit from CI with signature and skill card

ramakrishnap-nv (approved):

PR #1498, opened 2026-07-01
Use pytest importlib mode to avoid shadowing installed cuopt.

tmckayus:
When pytest is run from the repo root, prepending the source tree to sys.path can hide Cython extensions built into the installed wheel.

Pretty minor, but it bit me when testing Python gRPC bindings in the development environment when I ran tests from the repo root.

### DIFF
--- a/python/cuopt/pyproject.toml
+++ b/python/cuopt/pyproject.toml
@@ -80,6 +80,9 @@ max_allowed_size_compressed = '55Mi'
 
 [tool.pytest.ini_options]
 testpaths = ["cuopt/tests"]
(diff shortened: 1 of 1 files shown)


tmckayus:
This is an incomplete change, and also ultimately not needed (it compensates for a root-level conftest.py which is missing here, but that root level conftest.py can be moved anyway)

tmckayus:
This is an incomplete change, and also ultimately not needed (it compensates for a root-level conftest.py which is missing here, but that root level conftest.py can be moved anyway)

PR #1513, opened 2026-07-02
fix(ci): restrict UBI10 build, test, and manifest to CUDA 13.x only

ramakrishnap-nv:
## Summary

- UBI10 CUDA base images (`nvidia/cuda:*-ubi10`) are not published for any CUDA 12.x version on Docker Hub or NGC
- CI was failing when attempting to build the UBI10 image for CUDA 12.9.0 with `not found`
- Skip UBI10 build/push steps in `build_images.yaml` when `CUDA_VER` starts with `12.`
- Skip UBI10 test job in `test_images.yaml` for CUDA 12.x
- Skip UBI10 manifest creation in `create_multiarch_manifest.sh` when `CUDA_MAJOR != 13`


ramakrishnap-nv:
Superseded — switching from UBI10 to UBI9 (which has CUDA images for both 12.x and 13.x), so CU13-only restriction is no longer needed.

PR #1516, opened 2026-07-03
Improve QC COO canonicalization performance

yuwenchen95:
## Description

Improve QC COO canonicalization performance by reusing a scratch workspace across constraints, instead of allocating fresh lookup buffers on every call. Canonicalization output is unchanged; this is a performance optimization for batch ingest paths (MPS/LP parse, gRPC, incremental `add_quadratic_constraint`).

## Checklist

- Testing
   - [ ] Added tests
   - [ ] NA
- Documentation
   - [ ] Added new documentation
   - [x] NA


yuwenchen95:
/ok to test [fc12d4a](https://github.com/NVIDIA/cuopt/pull/1516/commits/fc12d4ac27192f83ad68c41cd1ebd99329ae812b)

cpp/include/cuopt/mathematical_optimization/io/mps_data_model.hpp
```diff
@@ -15,6 +15,72 @@
...
+    std::vector<value_t>& vals,
+    coo_canonicalization_workspace_t<index_t, value_t>& workspace);
+};
+
+/**
+ * @brief Canonicalize a symmetric matrix in COO form to upper-triangular storage
+ *        using reusable workspace storage. Defined in mps_parser.cpp.
+ */
+template <typename i_t, typename f_t>
+void canonicalize_coo_matrix(std::vector<i_t>& rows,
```
mlubin, line 26: There's a mismatch between the header where the function is declared (mps_data_model.hpp) and the cpp file where it's defined (mps_parser.cpp).

cpp/src/io/mps_parser.cpp
```diff
@@ -256,78 +261,124 @@ void canonicalize_coo_matrix(std::vector<i_t>& rows, std::vector<i_t>& cols, std
...
+    output_cols.reserve(entries.size());
+    output_vals.reserve(entries.size());
+    for (const auto& entry : entries) {
+      if (std::abs(entry.val) > eps) {
+        output_rows.push_back(entry.row);
+        output_cols.push_back(entry.col);
+        output_vals.push_back(entry.val);
+      }
     }
+  } catch (...) {
```
mlubin: What's the purpose of the try-catch? what are we catching?

cpp/src/io/mps_parser.cpp
```diff
@@ -256,78 +261,124 @@ void canonicalize_coo_matrix(std::vector<i_t>& rows, std::vector<i_t>& cols, std
...
-    max_row     = std::max(max_row, r);
-    max_col     = std::max(max_col, c);
+    mps_parser_expects(rows[k] >= 0 && cols[k] >= 0,
+                       error_type_t::ValidationError,
+                       "COO indices must be non-negative");
+    max_index = std::max(max_index, std::max(rows[k], cols[k]));
   }
+  workspace.ensure_index_capacity(max_index);
+
+  auto& active_rows = workspace.active_rows_;
```
mlubin: don't use auto when the type is easy to write down: https://google.github.io/styleguide/cppguide.html#Type_deduction

cpp/src/io/mps_parser.cpp
```diff
@@ -243,8 +243,13 @@ void check_symmetric_offdiagonal_pairs(const std::vector<i_t>& rows,
...
 template <typename i_t, typename f_t>
-void canonicalize_coo_matrix(std::vector<i_t>& rows, std::vector<i_t>& cols, std::vector<f_t>& vals)
+void canonicalize_coo_matrix(std::vector<i_t>& rows,
+                             std::vector<i_t>& cols,
+                             std::vector<f_t>& vals,
+                             coo_canonicalization_workspace_t<i_t, f_t>& workspace)
 {
+  using workspace_t = coo_canonicalization_workspace_t<i_t, f_t>;
+
   const size_t n = vals.size();
```
mlubin: `n` is a very confusing name for the number of nonzeros

cpp/src/io/mps_parser.cpp
```diff
@@ -256,78 +261,124 @@ void canonicalize_coo_matrix(std::vector<i_t>& rows, std::vector<i_t>& cols, std
...
+        } else {
+          entries[workspace.col_position_[col]].val += vals[k];
+        }
       }
+
+      const size_t row_output_end = entries.size();
+      for (size_t p = row_output_start; p < row_output_end; ++p) {
+        workspace.col_position_[entries[p].col] = workspace_t::invalid_index;
+      }
+      std::sort(entries.begin() + row_output_start,
```
mlubin: Do we actually need columns to be sorted within rows?

cpp/src/io/mps_parser.cpp
```diff
@@ -256,78 +261,124 @@ void canonicalize_coo_matrix(std::vector<i_t>& rows, std::vector<i_t>& cols, std
...
-      const i_t row = std::min(rows[static_cast<size_t>(k)], cols[static_cast<size_t>(k)]);
-      const i_t col = std::max(rows[static_cast<size_t>(k)], cols[static_cast<size_t>(k)]);
-      const i_t c   = col - min_idx;
-      if (marker[static_cast<size_t>(c)] <= row_start) {
-        entries.push_back(entry_t{row, col, vals[static_cast<size_t>(k)]});
-        marker[static_cast<size_t>(c)] = static_cast<i_t>(entries.size());
-      } else {
-        entries[static_cast<size_t>(marker[static_cast<size_t>(c)] - 1)].val +=
-          vals[static_cast<size_t>(k)];
+    std::sort(active_rows.begin(), active_rows.end(), [](const auto& lhs, const auto& rhs) {
```
mlubin: Do we actually need rows to be sorted?

cpp/src/io/mps_parser.cpp
```diff
@@ -256,78 +261,124 @@ void canonicalize_coo_matrix(std::vector<i_t>& rows, std::vector<i_t>& cols, std
...
-  std::vector<i_t> cursor(offsets.begin(), offsets.end() - 1);
-  for (size_t k = 0; k < n; ++k) {
-    const i_t r                          = std::min(rows[k], cols[k]) - min_idx;
-    perm[static_cast<size_t>(cursor[r])] = static_cast<i_t>(k);
-    cursor[static_cast<size_t>(r)]++;
-  }
+  try {
+    // Discover active canonical rows and count entries in first-appearance order.
+    for (size_t k = 0; k < n; ++k) {
+      const i_t row = std::min(rows[k], cols[k]);
```
mlubin: Consider doing a pass at the beginning to put all the entries in the correct triangle. This would reduce the context needed to understand the code and reduce the incidence of `max` and `min` in various spots.

cpp/src/io/mps_parser.cpp
```diff
@@ -256,78 +261,124 @@ void canonicalize_coo_matrix(std::vector<i_t>& rows, std::vector<i_t>& cols, std
...
-      rows.push_back(e.row);
-      cols.push_back(e.col);
-      vals.push_back(e.val);
+    const f_t eps = std::numeric_limits<f_t>::epsilon();
+    output_rows.reserve(entries.size());
+    output_cols.reserve(entries.size());
+    output_vals.reserve(entries.size());
+    for (const auto& entry : entries) {
+      if (std::abs(entry.val) > eps) {
+        output_rows.push_back(entry.row);
```
mlubin: Why do we need `output_rows`, `output_cols`, and `output_vals`, versus writing directly to `rows`, `cols`, and `vals`?

cpp/include/cuopt/mathematical_optimization/io/mps_data_model.hpp
```diff
@@ -15,6 +15,72 @@
...
 namespace cuopt::mathematical_optimization::io {
 
+/**
+ * @brief Reusable workspace for COO canonicalization.
+ *
+ * Lookup vectors grow to fit the largest variable index seen so far. Each call resets only
+ * entries that were touched. A workspace must not be shared across concurrent calls.
+ */
+template <typename i_t, typename f_t>
+class coo_canonicalization_workspace_t {
```
mlubin: This class is an implementation detail and doesn't really belong in a header file. The use of "friend" below hints at this. Most of the call sites are using canonicalize_quadratic_constraints so they don't need to see coo_canonicalization_workspace_t. Perhaps non-performance-sensitive sites can call a version that allocates a new coo_canonicalization_workspace_t?

Alternatively find a way to make this helper opaque so the body can be moved to a cpp file.

cpp/src/io/mps_parser.cpp
yuwenchen95: Agreed. This is unnecessary.

cpp/src/io/mps_parser.cpp
yuwenchen95: This is a tradeoff between a cleaner code design and avoiding an extra for loop.

cpp/src/io/mps_parser.cpp
yuwenchen95: I'm afraid we can't avoid it because the original bucket sorting relies on the assumption of sorted indices that we can do prefix sum and nnz entries are implicitly row-ordered by putting them into the ordered buckets 1 to n. Otherwise, we have to use unordered_map for counting.

cpp/src/io/mps_parser.cpp
yuwenchen95: Same reason as for the sorting on rows.

cpp/include/cuopt/mathematical_optimization/io/mps_data_model.hpp
yuwenchen95: Moved implementation of this function to `mps_data_model.cpp`

cpp/include/cuopt/mathematical_optimization/io/mps_data_model.hpp
yuwenchen95: The class is exposed due to the existence of `append_quadratic_constraint` where we add quadratic constraint incrementally while want to keep reusing workspace. Dropping reusing workspace will slow the speed a lot when using `add_quadratic_constraint`.

I made a new commit trying to keep reusing workspace without exposing the class details in `mps_data_model.hpp`.

yuwenchen95:
The PR is unsolved. Multi-threading is possible since each quadratic constraint is independent.

mlubin:
Why didn't we want to merge this?

yuwenchen95:
> Why didn't we want to merge this?

The current change relies on reusing the existing memory for processing quadratic constraints sequentially. I'm considering if it is worth to explore multi-threading since each quadratic constraint is isolated.

PR #1518, opened 2026-07-06
[DRAFT] Solver persistence pipeline

Iroy30:
## Checklist

- Testing
   - [ ] Added tests
   - [ ] NA
- Documentation
   - [ ] Added new documentation
   - [ ] NA


python/cuopt/cuopt/linear_programming/problem.py
```diff
@@ -1629,6 +1635,92 @@ def _to_data_model(self):
...
+            self.model is None
+            or self.constraint_csr_matrix is None
+            or self.objective is None
+            or len(self.objective) != n
+            or self.rhs is None
+        ):
+            self._to_data_model()
+            return
+
+        for j in range(n):
```
yuwenchen95: Vectorized update will be faster than looping over `n`.

python/cuopt/cuopt/linear_programming/problem.py
```diff
@@ -1629,6 +1635,92 @@ def _to_data_model(self):
...
+        self._index_to_var_cache = None
+        self._constraint_csr_scipy = None
+
+    def _constraint_csr_scipy_matrix(self):
+        csr_dict = self.constraint_csr_matrix
+        if csr_dict is None or self.rhs is None:
+            return None
+        if self._constraint_csr_scipy is not None:
+            return self._constraint_csr_scipy
+        n_rows = (
```
yuwenchen95: Redundant branching here. Can set `n_rows = len(self.rhs)`

python/cuopt/cuopt/linear_programming/problem.py
```diff
@@ -1629,6 +1635,92 @@ def _to_data_model(self):
...
+            or len(self.objective) != n
+            or self.rhs is None
+        ):
+            self._to_data_model()
+            return
+
+        for j in range(n):
+            self.objective[j] = self.vars[j].getObjectiveCoefficient()
+
+        rhs_arr = (
```
yuwenchen95: Redundant branching here.

python/cuopt/cuopt/linear_programming/problem.py
```diff
@@ -1629,6 +1635,92 @@ def _to_data_model(self):
...
+            (
+                np.asarray(csr_dict["values"], dtype=np.float64),
+                np.asarray(csr_dict["column_indices"], dtype=np.int32),
+                np.asarray(csr_dict["row_pointers"], dtype=np.int32),
+            ),
+            shape=(n_rows, len(self.vars)),
+        )
+        return self._constraint_csr_scipy
+
+    def _refresh_data_model_values(self):
```
yuwenchen95: Only linear cost and rhs of linear constraints are updated. Missing update of quadratic cost, lower and upper bounds of variables, constraint matrix, row sense and objective constraint.

python/cuopt/cuopt/linear_programming/problem.py
```diff
@@ -1629,6 +1635,92 @@ def _to_data_model(self):
...
+        self.model.set_objective_coefficients(self.objective)
+        self.model.set_constraint_bounds(rhs_arr)
+
+    def _populate_slacks_vectorized(self, primal_sol):
+        """Assign constraint slacks via CSR matvec (structure-stable problems)."""
+        A = self._constraint_csr_scipy_matrix()
+        if A is None:
+            return False
+
+        rhs_arr = (
```
yuwenchen95: Redundant branching here.

python/cuopt/cuopt/linear_programming/problem.py
```diff
@@ -1629,6 +1635,92 @@ def _to_data_model(self):
...
+            if constr.is_quadratic:
+                continue
+            rhs_arr[linear_row] = constr.RHS
+            linear_row += 1
+        self.rhs = rhs_arr
+
+        self.model.set_objective_coefficients(self.objective)
+        self.model.set_constraint_bounds(rhs_arr)
+
+    def _populate_slacks_vectorized(self, primal_sol):
```
yuwenchen95: Why do we need to recompute slack values rather than extract them from existing values in the solver?

python/cuopt/cuopt/linear_programming/problem.py
```diff
@@ -2186,21 +2288,44 @@ def populate_solution(self, solution):
...
             dual_sol = solution.get_dual_solution()
-        linear_row = 0
-        for constr in self.constrs:
-            if constr.is_quadratic:
-                continue
-            if dual_sol is not None and len(dual_sol) > linear_row:
-                constr.DualValue = dual_sol[linear_row]
-            constr.Slack = constr.compute_slack()
-            linear_row += 1
+        if not (
```
yuwenchen95: What are these new lines `2291-2312` for?`

python/cuopt/cuopt/linear_programming/problem.py
```diff
@@ -2212,10 +2337,13 @@ def solve(self, settings=solver_settings.SolverSettings()):
         >>> problem.setObjective(x + y, sense=MAXIMIZE)
         >>> problem.solve()
         """
-        if self.model is None:
+        if self.model is None or self.constraint_csr_matrix is None:
```
yuwenchen95: Why we need additional `self.constraint_csr_matrix is None` here?

python/cuopt/cuopt/linear_programming/problem.py
```diff
@@ -2212,10 +2337,13 @@ def solve(self, settings=solver_settings.SolverSettings()):
...
             self._to_data_model()
-        # Call Solver
-        solution = solver.Solve(self.model, settings)
-        # Post Solve
+        else:
+            self._refresh_data_model_values()
+        active_session = session if session is not None else self._session
+        solution = solver.Solve(self.model, settings, session=active_session)
+        if getattr(settings, "session_enabled", False) and solution.lp_solve_session is not None:
```
yuwenchen95, line 2230: If `session_enabled=False`. Why do we still set `self._session = solution.lp_solve_session`?

cpp/src/pdlp/utilities/lp_solve_session.cu
```diff
@@ -0,0 +1,83 @@
+/* clang-format off */
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+/* clang-format on */
+
+#include <cuopt/linear_programming/utilities/lp_solve_session.hpp>
```
yuwenchen95, line 8: It's better to move `lp_solve_session` files under `barrier` folder. `barrier_resolve_session` may be a better name.

cpp/src/pdlp/utilities/lp_solve_session.cu
```diff
@@ -0,0 +1,83 @@
...
+
+#include <barrier/barrier_symbolic_cache.hpp>
+
+#include <optional>
+#include <utility>
+
+namespace cuopt::cython {
```
yuwenchen95, line 15: solver_session is not limited to the use in python. Move it out of `namespace cuopt::cython`.

cpp/src/pdlp/utilities/cython_solve.cu
```diff
@@ -71,59 +86,102 @@ cuopt::mathematical_optimization::lp_solution_interface_t<int, double>* call_sol
...
   // Determine memory backend based on execution mode
-  auto memory_backend = cuopt::mathematical_optimization::get_memory_backend_type();
+  auto memory_backend = cuopt::linear_programming::get_memory_backend_type();
 
   solver_ret_t response;
 
+  auto& pdlp_settings = solver_settings->get_pdlp_settings();
+  const bool session_enabled = pdlp_settings.session_enabled;
+  const bool barrier_path    = uses_barrier_session_path(*solver_settings, *data_model);
+  const bool want_session    = (session_in != nullptr || session_enabled) && barrier_path &&
```
yuwenchen95, line 136: Do we also have to exclude concurrent mode?

cpp/src/pdlp/utilities/cython_solve.cu
```diff
@@ -71,59 +86,102 @@ cuopt::mathematical_optimization::lp_solution_interface_t<int, double>* call_sol
...
+  raft::handle_t* solve_handle = &ephemeral_handle;
+
   // Create problem instance and CUDA resources based on memory backend
-  if (memory_backend == cuopt::mathematical_optimization::memory_backend_t::GPU) {
-    // GPU memory backend: Create CUDA resources and GPU problem
-    rmm::cuda_stream stream(static_cast<rmm::cuda_stream::flags>(flags));
-    const raft::handle_t handle_{stream};
+  if (memory_backend == cuopt::linear_programming::memory_backend_t::GPU) {
+    if (want_session) {
+      if (active_session == nullptr) {
```
yuwenchen95, line 151: Moving timing one layer up instead of repetition in different branch.

cpp/src/pdlp/utilities/cython_solve.cu
```diff
@@ -71,59 +86,102 @@ cuopt::mathematical_optimization::lp_solution_interface_t<int, double>* call_sol
...
+  const bool barrier_path    = uses_barrier_session_path(*solver_settings, *data_model);
+  const bool want_session    = (session_in != nullptr || session_enabled) && barrier_path &&
+                            memory_backend == cuopt::linear_programming::memory_backend_t::GPU &&
+                            !is_batch_mode;
+
+  std::unique_ptr<lp_solve_session_t> owned_session;
+  lp_solve_session_t* active_session = session_in;
+  pdlp_settings.lp_solve_session     = nullptr;
+
+  rmm::cuda_stream ephemeral_stream(static_cast<rmm::cuda_stream::flags>(flags));
```
yuwenchen95, line 144: Why do we create a new stream rather than using the existing one in `lp_solve_session_t`?

cpp/include/cuopt/linear_programming/pdlp/solver_settings.hpp
```diff
@@ -326,6 +330,10 @@ class pdlp_solver_settings_t {
   // Used to force batch PDLP to solve a subbatch of the problems at a time
   // The 0 default value will make the solver use its heuristic to determine the subbatch size
   i_t fixed_batch_size{0};
+  /** When true, first GPU barrier/QCQP solve returns an ``lp_solve_session_t`` capsule. */
+  bool session_enabled{false};
+  /** Non-owning session pointer set by ``call_solve`` for barrier symbolic reuse. */
+  cuopt::cython::lp_solve_session_t* lp_solve_session{nullptr};
```
yuwenchen95, line 336: Can we separate the session pointer from `solver_settings_t` struct? It is set to `nullptr` in every call of `call_solve` in `cython_solve.cu`. Parameters in `solver_settings_t` shouldn't be changed within the algorithm call.

cpp/src/pdlp/solve.cu
```diff
@@ -61,14 +62,61 @@
...
 #include <set>
 #include <tuple>
 
 #define CUOPT_LOG_CONDITIONAL_INFO(condition, ...) \
   if ((condition)) { CUOPT_LOG_INFO(__VA_ARGS__); }
 
-namespace cuopt::mathematical_optimization {
+namespace cuopt::linear_programming {
+
+namespace {
```
yuwenchen95, line 75: What are these lines of code for?

cpp/src/pdlp/solve.cu
```diff
@@ -1821,10 +1884,14 @@ optimization_problem_solution_t<i_t, f_t> solve_qcqp(
       CUOPT_LOG_INFO("Writing user problem to file: %s", settings.user_problem_file.c_str());
       op_problem.write_to_mps(settings.user_problem_file);
     }
+    {
```
yuwenchen95, line 1887: What is this step for?

cpp/src/barrier/barrier_symbolic_cache.hpp
```diff
@@ -0,0 +1,82 @@
...
+      d_augmented_diagonal_indices_(0, stream),
+      device_ADAT(stream),
+      device_AD(stream),
+      device_A(stream),
+      d_original_A_values(0, stream),
+      device_A_x_values(0, stream)
+  {
+  }
+
+  void clear()
```
yuwenchen95, line 63: We also need to clear vector and matrix memory at the same time.

cpp/src/barrier/barrier_symbolic_cache.hpp
```diff
@@ -0,0 +1,82 @@
...
+    valid         = false;
+    cusparse_info.reset();
+  }
+
+  [[nodiscard]] bool matches_reuse(barrier_sparsity_hash_t hash,
+                                   bool augmented,
+                                   raft::handle_t const* handle) const
+  {
+    return valid && handle != nullptr && handle_ptr == handle && use_augmented == augmented &&
+           sparsity_hash == hash;
```
yuwenchen95, line 78: The hash comparison is a simplified but not theoretically equivalent one. Can we have an exact comparison of row, col vectors in csr format, e.g. the same size of vectors and the same index and shift values in them? This should also be efficient on GPU.

cpp/src/barrier/sparse_cholesky.cuh
```diff
@@ -509,33 +516,43 @@ class sparse_cholesky_cudss_t : public sparse_cholesky_base_t<i_t, f_t> {
...
       "cudssDataGet for LU_NNZ");
-    settings_.log.printf("Symbolic nonzeros in factor : %.2e\n", static_cast<f_t>(lu_nz) / 2.0);
+    settings_->log.printf("Symbolic nonzeros in factor : %.2e\n", static_cast<f_t>(lu_nz) / 2.0);
     // TODO: Is there any way to get nonzeros in the factors?
     // TODO: Is there any way to get flops for the factorization?
     RAFT_CUDA_TRY(cudaStreamSynchronize(stream));
     handle_ptr_->get_stream().synchronize();
 
+    symbolic_done_        = true;
+    numeric_factor_valid_ = false;
```
yuwenchen95, line 534: This seems to be never used.

cpp/src/barrier/sparse_cholesky.cuh
```diff
@@ -865,6 +882,63 @@ class sparse_cholesky_cudss_t : public sparse_cholesky_base_t<i_t, f_t> {
     this->positive_definite = positive_definite;
   }
 
+  void rebind_settings(const simplex_solver_settings_t<i_t, f_t>& settings) override
+  {
+    settings_ = &settings;
+  }
+
+  void invalidate_numeric_factor() override { numeric_factor_valid_ = false; }
```
yuwenchen95, line 890: The function can be removed since `numeric_factor_valid_` is not used.

cpp/src/barrier/sparse_cholesky.cuh
```diff
@@ -865,6 +882,63 @@ class sparse_cholesky_cudss_t : public sparse_cholesky_base_t<i_t, f_t> {
     this->positive_definite = positive_definite;
   }
 
+  void rebind_settings(const simplex_solver_settings_t<i_t, f_t>& settings) override
```
yuwenchen95, line 885: Do we want additional parameter settings update beyond value update in objective and constraints?

cpp/src/barrier/barrier.cu
```diff
@@ -4037,18 +4486,57 @@ lp_status_t barrier_solver_t<i_t, f_t>::solve(f_t start_time, lp_solution_t<i_t,
     }
 
     csc_matrix_t<i_t, f_t> Q(lp.num_cols, 0, 0);
+    std::unique_ptr<iteration_data_t<i_t, f_t>> owned_data;
+
+    auto finish_session = [&](lp_status_t status) -> lp_status_t {
```
yuwenchen95, line 4491: Define `finish_session` explicitly outside of `barrier_solver_t<i_t, f_t>::solve`

cpp/src/barrier/barrier.cu
```diff
@@ -604,36 +616,420 @@ class iteration_data_t {
...
-      raft::common::nvtx::range scope("Barrier: LP Data: symbolic analysis");
-      // Perform symbolic analysis
-      symbolic_status = 0;
-      if (use_augmented) {
-        {
-          raft::common::nvtx::range form_scope("Barrier: LP Data: form augmented");
-          // Build the sparsity pattern of the augmented system
-          form_augmented(true);
+
+      auto adopt_augmented_symbolic = [&]() -> bool {
```
yuwenchen95, line 623: Define a function explicitly outside.

cpp/src/barrier/barrier.cu
```diff
@@ -1950,6 +2349,53 @@ class iteration_data_t {
   rmm::device_uvector<i_t> d_augmented_diagonal_indices_;
   rmm::device_uvector<i_t> d_cone_csr_indices_;
   rmm::device_uvector<f_t> d_cone_Q_values_;
+  device_csr_matrix_t<i_t, f_t>* pinned_device_augmented_{nullptr};
```
yuwenchen95, line 2352: It's a bit strange we have pinned memory defined here. Why we need this additional pointer if we already have `device_` vectors in the existing `iteration_data_t`. 

If it is related to the issue that reusable space outlive `iteration_data_t`. We'd better group all reusable memory into a struct and shared it to `iteration_data_t`.

cpp/src/barrier/barrier.cu
```diff
@@ -4001,7 +4448,9 @@ lp_status_t barrier_solver_t<i_t, f_t>::check_for_suboptimal_solution(
 }
 
 template <typename i_t, typename f_t>
-lp_status_t barrier_solver_t<i_t, f_t>::solve(f_t start_time, lp_solution_t<i_t, f_t>& solution)
+lp_status_t barrier_solver_t<i_t, f_t>::solve(f_t start_time,
```
yuwenchen95, line 4451: Some thoughts on the new solve function:

Current `solve` function of `barrier_solver_t` can be divided into two parts: the first phase is the initialization of all data structures and vectors we needed, and the second phase is the true `solve` phase (running a barrier method). We can split it into a `setup` for the first phase and a `solve` function for the second phase. Then, all changes relevant to reoptimization reside in `setup` function only.


Additionally, we can verify if reusing previous factorization is possible or not once the `barrier_presolve` is finished, by verifying objective and constraints (vectors and matrices) are the same as the previous problem. That means we can possibly reuse `iteration_data_t` by updating only the reusing workspace.

cpp/src/barrier/barrier.cu
```diff
@@ -604,36 +616,420 @@ class iteration_data_t {
...
+      cache.cusparse_info       = std::move(cusparse_info_);
+    }
+
+    cache.valid = true;
+    settings_.log.printf(
+      "Barrier: stored ADAT symbolic cache hash=0x%016llx\n",
+      static_cast<unsigned long long>(cache.sparsity_hash));
+  }
+
+  bool refresh_augmented_values()
```
yuwenchen95, line 800: It's better to define refresh functions for each blocks (e.g. `Q`, `A`, ) respectively. Only `Q` and `A` entries in the augmented KKT needs to be refreshed in the setup of reoptimization, since other entries will be recomputed for every barrier iteration.

In addition, updating blocks seperately also help in speed, since users usually update a subset of problem coefficients and we may not need to update all numerical values.

cpp/src/barrier/barrier_symbolic_cache.hpp
```diff
@@ -0,0 +1,82 @@
...
+class sparse_cholesky_cudss_t;
+
+/**
+ * @brief Cached cuDSS symbolic state and GPU buffers for hash-gated barrier reuse.
+ *
+ * Holds reordering + symbolic factorization in @p chol, the sparsity hash of the matrix
+ * that was analyzed, and path-specific GPU workspace (augmented KKT or ADAT + cuSPARSE).
+ */
+template <typename i_t, typename f_t>
+struct barrier_symbolic_cache_t {
```
yuwenchen95, line 33: We need to save previous problem coefficients, e.g. `Q`, `A` and bounds, etc, for checking if reoptimization is possible.

python/cuopt/cuopt/tests/linear_programming/test_lp_solve_session.py
```diff
@@ -0,0 +1,221 @@
...
+
+def _resolve_after_cache_clear(prob, settings, session):
+    """If adopt hits a stale hash and clears cache, one retry must succeed."""
+    sol, log_text, profile = solve_with_log(prob, settings, session=session)
+    if count_log_matches(log_text, _CLEAR_CACHE_LINE) >= 1:
+        sol, log_text, profile = solve_with_log(prob, settings, session=session)
+    return sol, log_text, profile
+
+
+def test_adat_session_warm_reuse():
```
yuwenchen95, line 53: These tests seems to be generated by agent. It's best we clearly define a deterministic small examples rather than relying on random generation in unit tests.

Also, it would be clear to define a problem in `lp` format for better visualization.

python/cuopt/cuopt/tests/linear_programming/session_cache_helpers.py
```diff
@@ -0,0 +1,350 @@
...
+    cold_log: str,
+    warm_log: str,
+    cold_profile: dict[str, float],
+    warm_profile: dict[str, float],
+    *,
+    expect_adat: bool = False,
+    expect_augmented: bool = False,
+) -> None:
+    """Cold run stores cache; warm run reuses symbolic factorization."""
+    if expect_adat:
```
yuwenchen95, line 298: Capturing lines in log is fragile since we may not be able to include so much info in logging and the logging will vary over time. The best way is to clearly define a `data_update` function where we update new problem coefficients and return the flag that if previous factorization could be reused or not.

script_session_cache_tests.py
```diff
@@ -0,0 +1,150 @@
+#!/usr/bin/env python3
```
yuwenchen95, line 1: It's a test file that have to be under `test` folder.

script_perf_eval.py
```diff
@@ -0,0 +1,927 @@
...
+    c07_w = bench.get("warm_cache_ms_avg", {}).get("C07", 0.0)
+
+    if mode == "baseline":
+        if bench.get("cold_reuse_log_count", 0) != 0 or any(c != 0 for c in warm_reuse):
+            raise AssertionError(
+                f"baseline: unexpected sparsity reuse log "
+                f"(cold={bench.get('cold_reuse_log_count')}, warm={warm_reuse})"
+            )
+        if c07_w < 50.0:
+            raise AssertionError(
```
yuwenchen95, line 720: The check based on time heuristic is very fragile, the result may not be consistent across different hardwares and also vary depending on the dimension of a problem. We need clear flag information.

script_perf_eval.py
```diff
@@ -0,0 +1,927 @@
+#!/usr/bin/env python3
```
yuwenchen95, line 1: It's a test file that have to be under `test` folder. 

Also try to rename the file as `portfolio_optimization_test.py`.

cpp/include/cuopt/linear_programming/utilities/solver_cache_profiler.hpp
```diff
@@ -0,0 +1,152 @@
...
+#include <cstdarg>
+#include <cstdint>
+#include <cstdio>
+#include <cstdlib>
+#include <cstring>
+#include <string>
+
+namespace cuopt::linear_programming::cache_profile {
+
+enum class cache_id : int {
```
yuwenchen95, line 23: Is it only for temporary code verification?

script_perf_eval.py
```diff
@@ -0,0 +1,927 @@
...
+    "C03": "Problem fingerprint",
+    "C04": "Augmented vs ADAT choice",
+    "C05": "ADAT / augmented sparsity pattern",
+    "C06": "cuDSS handle + config",
+    "C07": "cuDSS symbolic factorization",
+    "C08": "Dense-column / SOC layout metadata",
+    "C09": "Device buffer allocation sizes",
+}
+
+_CACHE_PROFILE_LINE = re.compile(
```
yuwenchen95, line 68: The same concern as in `session_cache_helpers.py`: We need clear flag information returned by a check function rather than capturing log or time heuristics.

script_perf_eval.py
Iroy30: this will be removed after testing, not for the PR.

cpp/include/cuopt/linear_programming/utilities/solver_cache_profiler.hpp
Iroy30: Yes, this is temporary profiling code for testing, I will remove it once the PR is out of Draft and thoroughly evaluated.

script_session_cache_tests.py
Iroy30: To be Removed. Only for temporary testing.

script_perf_eval.py
Iroy30: To be Removed. Only for temporary testing.

script_perf_eval.py
Iroy30: To be Removed. Only for temporary testing.

cpp/src/barrier/barrier.cu
Iroy30: Yes it is because it is not owned here. Python owns this data in session.

PR #1522, opened 2026-07-06
make selection of CPU or GPU memory for problem automatic in C api

tmckayus:
This change makes the selection of CPU vs GPU memory automatic based on the presence of a GPU. Previously, CPU backed problems were only created when remote execution is enabled. Now, they can be created automatically on a CPU host or on a GPU host with CUDA_VISIBLE_DEVICES="".  This lays the foundation for more flexibility in the future for use of the parser on CPU hosts, etc.

### DIFF
--- a/cpp/include/cuopt/mathematical_optimization/backend_selection.hpp
+++ b/cpp/include/cuopt/mathematical_optimization/backend_selection.hpp
@@ -40,29 +40,10 @@ bool is_remote_execution_enabled();
 execution_mode_t get_execution_mode();
 
 /**
- * @brief Check if CPU memory should be used for local execution (test mode)
- *
- * Undocumented. Intended for testing CPU problem/solution structures without remote execution.
- * When enabled, local solve will convert CPU problems to GPU, solve, and convert back.
- *
- * @return true if CUOPT_USE_CPU_MEM_FOR_LOCAL is set to "true" or "1" (case-insensitive)
- */
-bool use_cpu_memory_for_local();
-
-/**
- * @brief Whether a CPU-memory-backed problem has a configured solve path without re-probing CUDA.
- *
- * True when remote execution is enabled or undocumented test mode CUOPT_USE_CPU_MEM_FOR_LOCAL is
- * set (local CPU-to-GPU bridge on a GPU host).
- */
-bool cpu_memory_backend_solve_allowed();
-
-/**
  * @brief Determine which memory backend to use based on execution mode and hardware
  *
  * Logic:
  *   - REMOTE execution -> always CPU memory
- *   - LOCAL + CUOPT_USE_CPU_MEM_FOR_LOCAL -> CPU memory (undocumented test mode)
  *   - LOCAL + no visible CUDA devices -> CPU memory (auto-detect CPU-only host)
  *   - LOCAL otherwise -> GPU memory
  *
--- a/cpp/include/cuopt/mathematical_optimization/backend_selection.hpp
+++ b/cpp/include/cuopt/mathematical_optimization/backend_selection.hpp
@@ -42,7 +42,7 @@ execution_mode_t get_execution_mode();
 /**
  * @brief Check if CPU memory should be used for local execution (test mode)
  *
- * This is intended for testing CPU problem/solution structures without remote execution.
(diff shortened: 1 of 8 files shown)


cpp/include/cuopt/mathematical_optimization/backend_selection.hpp
```diff
@@ -42,20 +42,29 @@ execution_mode_t get_execution_mode();
 /**
  * @brief Check if CPU memory should be used for local execution (test mode)
  *
- * This is intended for testing CPU problem/solution structures without remote execution.
+ * Undocumented. Intended for testing CPU problem/solution structures without remote execution.
```
mlubin: Looks documented to me, heh. What's the plan for CUOPT_USE_CPU_MEM_FOR_LOCAL? Can we deprecate/remove it?

cpp/src/pdlp/backend_selection.cpp
```diff
@@ -40,11 +42,39 @@ bool use_cpu_memory_for_local()
...
+  // Local run with no usable device: fall back to CPU memory, which requires remote
+  // execution at solve time. Log at INFO (the shipped level) so this surprising
+  // fallback is visible without a recompile -- the CUDA error code distinguishes
+  // "no device" from a driver/OS fault, and CUDA_VISIBLE_DEVICES reveals an
+  // intentionally hidden GPU. This query is invoked at both problem creation and
+  // solve time; the decision is fixed for the life of the process, so log only once
+  // to avoid duplicate lines.
+  static std::atomic<bool> already_logged{false};
+  if (!already_logged.exchange(true)) {
+    CUOPT_LOG_INFO(
```
mlubin, line 48: Nit: it would be nice to have a version of the logging macros that logs only once. absl has `LOG_ONCE`.

cpp/include/cuopt/mathematical_optimization/backend_selection.hpp
tmckayus: @mlubin hahaha, I thought the same thing. I could change it to "not officially supported"

So, the point of this env var is to do a round trip through the solver but beginning and ending with CPU data structures. In effect, it lets us test the conversion routines CPU -> GPU problem and GPU -> CPU solution on a real solve without using a gRPC server. Perhaps overkill, but it seems like a good smoke test.

cpp/src/pdlp/backend_selection.cpp
tmckayus: that is a great idea, I'll add an issue for it

cpp/include/cuopt/mathematical_optimization/backend_selection.hpp
mlubin: I'd prefer to delete the extra code if it's just for testing a code path we can test via public APIs. "Test via Public APIs" (https://abseil.io/resources/swe-book/html/ch12.html) has been beaten into me

tmckayus: Done, extra test mode removed (which shortens the C api testing as well)

cpp/tests/linear_programming/c_api_tests/c_api_test.c
```diff
@@ -2847,3 +2617,324 @@ cuopt_int_t test_pdlp_precision_single(const char* filename,
...
+/**
+ * Build a small MIP via cuOptCreateProblem on a CPU-backed host and verify getters (no solve).
+ * Caller must set CUDA_VISIBLE_DEVICES="" before invoking.
+ */
+cuopt_int_t test_cpu_host_create_problem_api()
+{
+  cuOptOptimizationProblem problem = NULL;
+  cuopt_int_t status;
+
+#define CPU_HOST_NUM_ITEMS       8
```
mlubin: Why are defines needed? can we scope the constants to this test?

mlubin (approved):

cpp/tests/linear_programming/c_api_tests/c_api_test.c
tmckayus: replaced with enum

ramakrishnap-nv (approved):

PR #1525, opened 2026-07-07
Python grpc async api allow arguments for TLS

tmckayus:
Allow TLS settings to be settable in the Python/C++ gRPC client init (rather than read only from env vars)

Env var mode still works, but if the client explicitly disables TLS or sets TLS values, the environment is ignored.


python/cuopt/cuopt/grpc/linear_programming/grpc_client.pyx
```diff
@@ -156,10 +229,14 @@ cdef class Client:
         self._incumbent_thread_errors = {}
         self._host = host
         self._port = port
-        cdef string error_out
+        self._tls = tls
         if not self._client.get().connect(error_out):
             raise GrpcError(error_out.decode("utf-8"))
 
+    def _spawn_client(self):
```
tmckayus, line 236: note this is for the background thread used to retrieve incumbents, for example. It needs to clone the setup of the main thread's client

python/cuopt/cuopt/tests/fixtures/grpc_server_fixtures.py
```diff
@@ -168,6 +171,177 @@ def kill_server(proc):
 cpu_only_env = client_remote_env
 
 
+def generate_test_certs(cert_dir):
```
ramakrishnap-nv: There are few certs in the repo which are being used to test legacy server, should we reuse that ?

tmckayus: hmm, yes we might be able to

ramakrishnap-nv (approved):

rg20 (approved):

PR #1215, opened 2026-05-14
Add symbol visibility checks

bdice:
## Summary
- Adds a `ci/check_symbols.sh` exported-symbol guard modeled on cuDF, with cuOpt-specific forbidden internal namespace patterns.
- Wires the symbol check into the `libcuopt` conda recipe after installing `libmps_parser` and `libcuopt`.
- Starts hidden-by-default symbol visibility treatment by adding export macros for `libcuopt` C API and `libmps_parser` parser/writer APIs.

## Status
This is intentionally opened as a draft. The new guard is useful, but `libcuopt.so` still exports CUB/Thrust/RAFT/RMM and cuOpt detail symbols after the initial hidden-visibility pass.

## Validation
- `cmake --build /home/coder/cuopt/cpp/build/latest --target mps_parser cuopt -j 8` passed.
- `./ci/check_symbols.sh /home/coder/cuopt/cpp/build/latest/libmps_parser/libmps_parser.so` passed.
- `./ci/check_symbols.sh /home/coder/cuopt/cpp/build/latest/libcuopt.so` currently fails as expected; first failing pattern is `cub::` with 1910 matches in the rebuilt library.
- C API exports remain visible in the rebuilt `libcuopt.so` (`cuOptCreateProblem`, `cuOptGetFloatSize`, `cuOptSolve`, `cuOptDestroySolution`).

## References
- Closes NVIDIA/cuopt#1213 when complete.
- Related: rapidsai/build-infra#53.
- Examples: rapidsai/cuml#8037, rapidsai/cudf#15982, rapidsai/raft#3006, rapidsai/rmm#1654.

PR #1526, opened 2026-07-07
Fix symbol visibility and add checks

### LINKED ISSUE #1213: Hide internal C++ symbols exported from libcuopt
## Summary
`libcuopt.so` currently exports a large internal C++/CUDA symbol surface, including CUB/Thrust/RAFT/RMM symbols and cuOpt `detail` implementation symbols. This matches the RAPIDS-wide symbol visibility problem tracked in rapidsai/build-infra#53.

## Evidence
Local audit of `cpp/build/latest/libcuopt.so` with `readelf --dyn-syms --wide | c++filt` found approximately:

- 10,383 defined dynamic symbols
- 2,020 `cub::` symbols
- 2,318 `thrust::` symbols
- 1,028 `raft::` symbols
- 2,691 `rmm::` symbols
- 2,628 `cuopt::linear_programming::detail` symbols
- 1,574 `cuopt::routing::detail` symbols
- 376 `grpc::` symbols
- 361 `google::protobuf` symbols
- 579 `tbb::` symbols

`detect-weak-linking` (https://github.com/rapidsai/detect-weak-linking) did not report default weak CUDA kernel duplicates for the sampled test executable, but `compute-sanitizer` reports duplicate CUB kernel registration in test runs and the broad exported-symbol surface increases collision risk.

## Desired direction
Follow the treatment used in RAPIDS libraries such as cuML, cuDF, RAFT, and RMM:

- Set hidden-by-default visibility for cuOpt shared library targets.
- Add a `CUOPT_EXPORT` macro for public ABI symbols.
- Mark public C API declarations and any public C++ API/types/template instantiations that must remain visible.
- Add a CI/package symbol check similar to cuDF's `ci/check_symbols.sh`, with cuOpt-specific forbidden export patterns.

## References
- rapidsai/build-infra#53
- rapidsai/cuml#8037
- rapidsai/cudf#15982
- rapidsai/raft#3006
- rapidsai/rmm#1654

arhag23:
## Description

Updates CMake code to build with hidden visibility. Adds explicit exports for symbols that should be visible in the shared library. Adds a separate static library for use with the tests which require access to the hidden symbols. Adds a CI check for ensuring no symbols from dependencies and implementation details are leaking.

## Issue

Closes #1213
Builds off #1215 which can be closed after this is merged 

## Checklist

- Testing
   - [x] Added tests
   - [ ] NA
- Documentation
   - [ ] Added new documentation
   - [x] NA


ci/check_symbols.sh
```diff
@@ -0,0 +1,79 @@
...
+    | c++filt --no-params \
+    | awk '$0 !~ /_error/' \
+    > "${symbol_file}"
+
+patterns=(
+    'cub::'
+    'thrust::'
+    'raft::'
+    'rmm::'
+    'cuopt::linear_programming::detail'
```
aliceb-nv: We've recently renamed this namespace (linear_programming) to mathematical_optimization

ci/check_symbols.sh
```diff
@@ -0,0 +1,87 @@
...
+    'cuopt::routing::detail'
+    'grpc::'
+    'google::protobuf'
+    'tbb::'
+)
```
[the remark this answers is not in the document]
aliceb-nv: We should also probably catch dejavu::, papilo::, boost:: symbols.

cpp/src/pdlp/solve.cu
aliceb-nv, line 2364: We should probably export all of these as well. @Kh4ster to check

cpp/CMakeLists.txt
```diff
@@ -587,13 +576,99 @@ configure_file(
...
+        "$<$<COMPILE_LANGUAGE:CUDA>:${CUOPT_CUDA_FLAGS}>"
+)
+
+target_compile_definitions(cuopt PUBLIC
+        CUOPT_CUDA_ARCHITECTURES="${JOINED_CUDA_ARCHITECTURES}"
+        CUOPT_CPU_ARCHITECTURE="${CMAKE_SYSTEM_PROCESSOR}"
+        "CUOPT_LOG_ACTIVE_LEVEL=RAPIDS_LOGGER_LOG_LEVEL_${LIBCUOPT_LOGGING_LEVEL}"
+        CUSPARSE_ENABLE_EXPERIMENTAL_API
+)
+
```
aliceb-nv, line 703: @ramakrishnap-nv should we add a `target_link_options(cuopt PRIVATE "LINKER:--exclude-libs,ALL")`? Otherwise PSLP symbols are embedded with default visibility afaik

aliceb-nv (approved):
Approving engine side

cpp/tests/linear_programming/grpc/CMakeLists.txt
```diff
@@ -24,7 +24,7 @@ target_include_directories(GRPC_CLIENT_TEST
 
 target_link_libraries(GRPC_CLIENT_TEST
     PRIVATE
-    cuopt
+    cuopt_static
```
ramakrishnap-nv, line 27: Any reason to change this to static?

cpp/CMakeLists.txt
ramakrishnap-nv: Agree

cpp/src/pdlp/solve.cu
arhag23: I checked these against the public include file, `batch_pdlp_solve` and `mps_data_model_to_optimization_problem` are present in `<cuopt/mathematical_optimization/solve.hpp>` so I will export them also. The rest aren't in any public header file.

cpp/tests/linear_programming/grpc/CMakeLists.txt
arhag23: Yes, nearly all tests now link against the static library since almost all directly use internal functions that are hidden when linking against the shared library.

cpp/CMakeLists.txt
arhag23: I went with this, but is there any reason for using `ALL` rather than explicitly specifying `PSLP` (and if any other PRIVATE static libraries)?

arhag23:
@ramakrishnap-nv We now build a static version of cuopt to link the tests against since they directly reference internal symbols not in the public header API. This grealy increases the size of the test binaries (~100 MB -> ~10 GB) since this embeds the kernel device code in the tests. Is this acceptable or do we want to look at alternatives? We can try exporting every internal symbol used for the tests, but this may lead to too much of the internals being visible.

ramakrishnap-nv:
> @ramakrishnap-nv We now build a static version of cuopt to link the tests against since they directly reference internal symbols not in the public header API. This grealy increases the size of the test binaries (~100 MB -> ~10 GB) since this embeds the kernel device code in the tests. Is this acceptable or do we want to look at alternatives? We can try exporting every internal symbol used for the tests, but this may lead to too much of the internals being visible.

Yeah, that was my main concern that the size of the tests would increase significantly. Are we doing same across all other RAPIDS projects? And these are normally uploaded to nightly registries, so that is another concern.

arhag23:
> Are we doing same across all other RAPIDS projects?

AFAIK other RAPIDS projects didn't need such a change for symbol visibility controls.

ramakrishnap-nv:
> > Are we doing same across all other RAPIDS projects?
> 
> AFAIK other RAPIDS projects didn't need such a change for symbol visibility controls.

Hmm, in the issue @bdice mentioned that this has been done for cudf, cuml and others, so what is the extent. If possible, I would like to avoid static linking with test.

arhag23:
> > > Are we doing same across all other RAPIDS projects?
> > 
> > 
> > AFAIK other RAPIDS projects didn't need such a change for symbol visibility controls.
> 
> Hmm, in the issue @bdice mentioned that this has been done for cudf, cuml and others, so what is the extent. If possible, I would like to avoid static linking with test.

I think those libraries didn't have tests against the library internals so they never ran into this issue. To avoid static linking with the tests, we have to either export all the internals tested against or disable visibility controls when building for the tests. Unfortunately none of these options seem like the "correct" thing to do.

ramakrishnap-nv:
@arhag23 I have created a PR to minimize the number of tests that require static linking to reduce overall size. Still testing, may be if you think that is good, you can merge it to your PR and close that one #1566

arhag23:
I like where those changes are heading and I can incorporate them into this PR, but are you content with the results? Looking through the CI, I still see that the tests use 4 or 5 GB. This is a huge reduction, but we are still a lot above the original footprint. I'll try and get a list of what is being tested internally and maybe some of these things should be exported anyway.

Also, if we have internal vs external tests, we can then work towards drop allowing the use of internal headers for the external tests?

ramakrishnap-nv:
> I like where those changes are heading and I can incorporate them into this PR, but are you content with the results? Looking through the CI, I still see that the tests use 4 or 5 GB. This is a huge reduction, but we are still a lot above the original footprint. I'll try and get a list of what is being tested internally and maybe some of these things should be exported anyway.
> 
> Also, if we have internal vs external tests, we can then work towards drop allowing the use of internal headers for the external tests?

Yeah, completely agree, I am currently evaluating what all tests are needed and add-ons. And also we would need a way so we can may be expose parts of internal APIs so we can use those with shared library instead of adding static.

arhag23:
Closed in favor of #1625

PR #1550, opened 2026-07-09
Add front-completion workflow to cuopt-multi-objective-exploration skill

cafzal:
## Description

The skill teaches weighted-sum vs ε-constraint scalarization and stops at "ε-constraint reaches non-supported points." This adds the operational sequel as a new Step 5 — complete the frontier: measure what a sweep missed (the boxes between adjacent swept points), fill the largest gaps first with targeted ε solves, keep time-limited feasible incumbents with their gap as the suboptimality bound (heuristics-only mode covers the no-certificates-needed case), and tag every point exact vs approximate so a mixed frontier is never over-claimed. Adds no solver features; orchestrates shipped APIs (MIP starts, time limits, heuristics-only mode). The interpret section is renumbered to Step 6.

Motivation, measured against exactly enumerated ground-truth fronts on a fixed-cost selection MILP family (median over 10 seeds × 3 instances, n = 60–240): the supported points any weighted-sum sweep can reach cap at 0.82–0.90 of front hypervolume — more weights cannot close that — while the same budget on targeted ε solves reaches ~0.99. On a near-convex variant the miss persists (weight grids cluster at corners), so measure-then-fill pays regardless of the front's convexity.

Verified end to end with cuOpt 26.06.00 on a T4, driving the section exactly as written on models adapted from cuopt-examples:

| Model | Sweep (before) | Completed (after) | Fill cost |
|---|---|---|---|
| [Workforce multi-objective](https://github.com/NVIDIA/cuopt-examples/blob/main/workforce_optimization/workforce_optimization_multiobjective.ipynb) | 3 distinct plans from 15 all-`Optimal` solves — hypervolume 0.489 | 9 plans, all exact and on the 52-point reference front — 0.681 vs the enumeration's 0.735 (93% of coverage) | 6 solves vs 52 |
| [Capacitated p-median](https://github.com/NVIDIA/cuopt-examples/blob/main/emergency_management/capacitated_emergency_shelter.ipynb) | 7 distinct plans | 9 plans (6 exact, 3 approximate) — 99.9% of the best-known front's hypervolume | 3 solves |
| [Portfolio QP](https://github.com/NVIDIA/cuopt-examples/blob/main/portfolio_optimization/QP_portfolio_frontier_duals.ipynb) (control) | all 15 points distinct; largest gap box 1.2% of the frame | measure bounds any possible miss at 4.5% — verdict: stop | 0 solves |

The p-median run also exercised the degrade path for real: time-capped solves returned `FeasibleFound` incumbents, kept and tagged approximate, and heuristics-only mode populated the same subproblem.

The verification loop also caught and fixed one defect in the measure: a sweep that returns only a handful of points is all gaps, so no box stands out from the median — the text now also flags boxes covering a large share of the frontier's span.

Evals: +2 positive (gap detection and fill; provenance reporting), +1 negative (single-objective decoy), mirroring the existing schema. Setting and status names verified in source and docs on this branch; `validate_skills.sh` passes; `BENCHMARK.md` / skill card / signature regenerate via NVSkills CI. Benchmark harness and per-seed tables available on request.

## Checklist

- Testing
   - [x] Added tests
   - [ ] NA
- Documentation
   - [x] Added new documentation
   - [ ] NA

### DIFF
--- a/skills/cuopt-multi-objective-exploration/SKILL.md
+++ b/skills/cuopt-multi-objective-exploration/SKILL.md
@@ -1,7 +1,7 @@
 ---
 name: cuopt-multi-objective-exploration
 version: "26.08.00"
-description: Trace and interpret the Pareto frontier across competing objectives using repeated single-objective cuOpt solves (weighted-sum and ε-constraint).
+description: Trace, complete, and interpret the Pareto frontier across competing objectives using repeated single-objective cuOpt solves (weighted-sum and ε-constraint).
 license: Apache-2.0
 origin: cuopt-skill-evolution
 metadata:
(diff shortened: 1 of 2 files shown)


cafzal:
@ramakrishnap-nv I realized this might be a helpful follow-up to the multi-objective skill. Small addition so I opened a PR directly but happy to convert to an issue instead if it warrants further discussion.

mlubin (approved):
I'm a bit concerned that we're asking agents to do too much mechanical work in text form and code/harness guardrails would be more appropriate, but this is fine as an addition to the existing skill.

ramakrishnap-nv (approved):

PR #1552, opened 2026-07-10
fix(deps): add explicit cufile dependency

gforsyth:
## Description

Testing a smaller base CI image for wheel tests in #1548 -- that image does not have cuda toolkit installed and we rely on installing all dependencies from wheels.
That test bubbled up a missing `cufile` dependency (which was previously satisfied by the `libcufile.so` that was present on the CI image already).

I've also added in a workaround for a bug with the `cufile` wheels on ARM64, where they fail to load required dynamic libraries.

We had to make the same changes in rapidsai/kvikio#1000

xref rapidsai/build-planning#143

CI run before this change: https://github.com/NVIDIA/cuopt/actions/runs/29044331506/job/86218256034?pr=1548

And after this change: https://github.com/NVIDIA/cuopt/actions/runs/29054246782/job/86246625279?pr=1548

## Checklist

- Testing
   - [ ] Added tests
   - [ ] NA
- Documentation
   - [ ] Added new documentation
   - [x] NA

### DIFF
--- a/dependencies.yaml
+++ b/dependencies.yaml
@@ -154,6 +154,7 @@ files:
       - test_python_common
       - test_python_cuopt
       - depends_on_rapids_logger
+      - test_cufile
   py_build_cuopt_server:
     output: pyproject
     pyproject_dir: python/cuopt_server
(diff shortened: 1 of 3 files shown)


gforsyth:
Ah, this is not necessary -- I just clocked that the `cufile` loading error is bubbling up from `libkvikio`, so I'll go put in additional fixes there

PR #1553, opened 2026-07-10
fix(deps): add cufile to test dependencies for cuda12

gforsyth:
## Description

Testing a smaller base CI image for wheel tests in #1548 -- that image does not
have cuda toolkit installed and we rely on installing all dependencies from
wheels.

In rapidsai/kvikio#1000 and rapidsai/kvikio#1003 we added handling for loading
`cufile` from an installed wheel -- but we add that dependency to `libkvikio`
wheels as `cuda-toolkit[cufile]` and there was no `cufile` extra available
before `cuda-toolkit==12.6.3`, so for the testing setup here, we explicitly
install `nvidia-cufile-cu12` for earlier versions of `cuda-toolkit` so that
`libcufile` is available for tests.

(cuOpt doesn't have a direct dependency on `cufile`, but instead is running into this in `libkvikio` via `pylibcudf`.)

xref rapidsai/build-planning#143

## Checklist

- Testing
   - [ ] Added tests
   - [ ] NA
- Documentation
   - [ ] Added new documentation
   - [x] NA

### DIFF
--- a/dependencies.yaml
+++ b/dependencies.yaml
@@ -154,6 +154,7 @@ files:
       - test_python_common
       - test_python_cuopt
       - depends_on_rapids_logger
+      - test_cufile
   py_build_cuopt_server:
     output: pyproject
     pyproject_dir: python/cuopt_server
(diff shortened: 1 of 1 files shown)


ramakrishnap-nv (approved):

jakirkham (approved):
Thanks Gil! 🙏

PR #1555, opened 2026-07-10
fix(barrier): destroy cuDSS descriptors before freeing their backing buffers

### LINKED ISSUE #1554: [BUG] QP solver segfaults in test_qp.py::test_solver during full test suite run
## Description

`tests/quadratic_programming/test_qp.py::test_solver` crashes the pytest process with SIGSEGV (exit 139) when run as part of the full test suite. The segfault is **flaky** — it does not reproduce when the test is run in isolation, but consistently appears when the full suite runs together.

The crash occurs in two independent CI jobs from the same nightly run, across different Python versions, OSes, GPUs, and dep vintages — pointing to a bug in the QP solver's C++ layer, not an ABI or dependency issue.

## Affected CI jobs (run [29074991116](https://github.com/NVIDIA/cuopt/actions/runs/29074991116))

| Job | Config | Job ID |
|-----|--------|--------|
| [86304536002](https://github.com/NVIDIA/cuopt/actions/runs/29074991116/job/86304536002) | CUDA 12.9.2, Python 3.11, Rocky 8, L4, oldest-deps | wheel-tests-cuopt |
| [86304536010](https://github.com/NVIDIA/cuopt/actions/runs/29074991116/job/86304536010) | CUDA 12.9.2, Python 3.12, Ubuntu 24.04, H100, latest-deps | wheel-tests-cuopt |

## Stack trace (both jobs identical)

```
Primal infeasibility (abs/rel): 2.50e-09/4.17e-10
Dual infeasibility   (abs/rel): 2.00e-08/1.18e-09
Complementarity gap  (abs/rel): 2.29e-07/6.95e-09

Fatal Python error: Segmentation fault

Thread (most recent call first):
  File "cuopt/linear_programming/solver/solver.py", line 102 in Solve   ← solver_wrapper.Solve() call
  File "cuopt/utilities/exception_handler.py", line 24 in func
  File "python/cuopt/cuopt/tests/quadratic_programming/test_qp.py", line 65 in test_solver
```

The QP barrier solver completes successfully (infeasibilities are at near-optimal tolerance), then the process segfaults — likely during result copy-back or post-solve cleanup in the C++ extension.

## Crash isolation result

CI retried all 128 tests individually after the crash. Zero consistent crashes, zero flaky crashes — the segfault only manifests when tests run together, suggesting a **use-after-free or shared GPU/CUDA state issue** triggered by a preceding test.

## Test that crashes

`python/cuopt/cuopt/tests/quadratic_programming/test_qp.py::test_solver` — a simple QP with `CUOPT_ITERATION_LIMIT=10` set; the barrier converges in 1 iteration.

## Steps to reproduce

Run the full cuopt Python test suite (`pytest tests/`); the crash does not reproduce when `test_qp.py::test_solver` is run alone.

ramakrishnap-nv:
## Summary

- In \`sparse_cholesky_cudss_t\` destructor, \`cudaFreeAsync\` was called on device buffers before the cuDSS descriptors that reference them were destroyed, causing a use-after-free when cuDSS teardown triggered an internal stream sync.
- Fix: destroy cuDSS descriptors first, then free the backing buffers, then sync the stream.

Closes #1554

### DIFF
--- a/cpp/src/barrier/sparse_cholesky.cuh
+++ b/cpp/src/barrier/sparse_cholesky.cuh
@@ -352,16 +352,14 @@ class sparse_cholesky_cudss_t : public sparse_cholesky_base_t<i_t, f_t> {
 
   ~sparse_cholesky_cudss_t() override
   {
-    cudaFreeAsync(csr_values_d, stream);
-    cudaFreeAsync(csr_columns_d, stream);
-    cudaFreeAsync(csr_offset_d, stream);
-
-    cudaFreeAsync(x_values_d, stream);
-    cudaFreeAsync(b_values_d, stream);
+    // Destroy cuDSS descriptors before freeing the device buffers they reference.
+    // cudssDataDestroy/cudssDestroy may internally enqueue work on the stream via
+    // the custom memory handler (cudss_device_dealloc -> cudaFreeAsync); freeing
+    // the user-provided backing buffers first would create a use-after-free if
+    // cuDSS triggers a stream sync during teardown.
     if (A_created) {
       CUDSS_CALL_AND_CHECK_EXIT(cudssMatrixDestroy(A), status, "cudssMatrixDestroy for A");
     }
-
     CUDSS_CALL_AND_CHECK_EXIT(
       cudssMatrixDestroy(cudss_x), status, "cudssMatrixDestroy for cudss_x");
     CUDSS_CALL_AND_CHECK_EXIT(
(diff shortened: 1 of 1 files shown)


hlinsen (approved):

cpp/src/barrier/sparse_cholesky.cuh
```diff
@@ -352,23 +352,29 @@ class sparse_cholesky_cudss_t : public sparse_cholesky_base_t<i_t, f_t> {
 
   ~sparse_cholesky_cudss_t() override
   {
-    cudaFreeAsync(csr_values_d, stream);
-    cudaFreeAsync(csr_columns_d, stream);
-    cudaFreeAsync(csr_offset_d, stream);
-
-    cudaFreeAsync(x_values_d, stream);
-    cudaFreeAsync(b_values_d, stream);
+    // Destroy cuDSS descriptors before freeing the device buffers they reference.
```
yuwenchen95: NIT: comment should be more concise here.

chris-maes (approved):
LGTM

PR #1556, opened 2026-07-10
Routing DataModel: store-then-build (defer device construction to solve)

ramakrishnap-nv:
## Summary
Routing `DataModel` now records its setter calls and builds the device model lazily at solve, instead of pushing each input to the GPU as it is set. This keeps inputs host-resident while a problem is assembled and is the foundation for host-only (remote / gRPC) problem construction.

Behavior-preserving, except that C++ validation and dtype-cast warnings now surface at solve time rather than at the setter call. Full routing suite passes.


ramakrishnap-nv:
@coderabbitai review

ramakrishnap-nv:
@coderabbit review

python/cuopt/cuopt/tests/routing/test_deferred.py
```diff
@@ -0,0 +1,27 @@
...
+from cuopt.routing._lazy import _SKIP_GETTERS
+
+
+def test_lazy_covers_wrapper_surface():
+    """Every public wrapper DataModel method must be handled by the recording
+    layer (installed as a recorder/getter or an explicit override). This fails
+    loudly if a new wrapper method -- e.g. a mutator not named set_*/add_* --
+    is added without being recorded, instead of silently dropping its data.
+    """
+    dm = routing.DataModel(1, 1)  # triggers _install_methods
```
Iroy30, line 17: What happens if user calls dm.random_func_call(x, y, z) ?
Should this be a test or a validation within cuopt

Iroy30:
Just a few thoughts:
1) How would I a developer add a new function to data model , what needs to be done?
2) The naming of the Lazy Data Model and file could be better

python/cuopt/cuopt/tests/routing/test_deferred.py
ramakrishnap-nv: Good question. `dm.random_func_call(x, y, z)` raises `AttributeError` — the recording layer installs only the declared setter/getter surface and has no `__getattr__` catch-all, so an unknown or typo'd call fails loudly and never enters the IR (`_calls` stays empty). So it's standard Python behavior rather than something cuOpt needs to validate. I've added `test_unknown_method_is_not_recorded` to lock this behavior.

ramakrishnap-nv:
Thanks @Iroy30 — both addressed:

**1) Adding a new function to the DataModel.** Three steps, now documented in the `_deferred.py` module docstring ("Adding a new setter/getter"): (a) implement it on the C++ `data_model_view_t` + Cython wrapper, (b) add the public method in `vehicle_routing.py` (validation + docstring, forwarding to `super()`), (c) add its name to `_SETTERS`/`_GETTERS`. `test_deferred` cross-checks the declared surface against the wrapper, so a missing spec entry fails CI rather than silently dropping the call's data.

**2) Naming.** Renamed `_LazyDataModel`/`_lazy.py` → `_DeferredDataModel`/`_deferred.py`. "Lazy" was too generic, and "host" would be misleading since cuDF inputs pass through and stay on device — "deferred" names the actual mechanism (the device build is deferred to `Solve`, for host and device inputs alike) and avoids colliding with the Cython "wrapper".

Also from the earlier thread: an unknown/typo'd call like `dm.random_func_call(...)` raises `AttributeError` and never enters the IR (no `__getattr__` catch-all) — locked by `test_unknown_method_is_not_recorded`.

Iroy30 (approved):

PR #1559, opened 2026-07-10
change semantics of problem construction in cuOptReadProblem

tmckayus:
Since the mps data model is constructed internally, use move semantics when building a CPU problem to avoid an  unnecessary copy.  This is a small optimization but on a large problem with 10M variables and 200k constraints, it can save half a second in problem creation.

### DIFF
--- a/cpp/cuopt_cli.cpp
+++ b/cpp/cuopt_cli.cpp
@@ -138,8 +138,8 @@ int run_single_file(const std::string& file_path,
       std::make_unique<cuopt::mathematical_optimization::cpu_optimization_problem_t<int, double>>();
   }
 
-  cuopt::mathematical_optimization::populate_from_mps_data_model(problem_interface.get(),
-                                                                 mps_data_model);
+  cuopt::mathematical_optimization::adopt_from_mps_data_model(problem_interface.get(),
+                                                              std::move(mps_data_model));
 
   const bool is_mip = (problem_interface->get_problem_category() ==
                          cuopt::mathematical_optimization::problem_category_t::MIP ||
@@ -152,7 +152,7 @@ int run_single_file(const std::string& file_path,
       initial_solution_file.empty()
         ? std::vector<double>()
         : cuopt::mathematical_optimization::solution_reader_t::get_variable_values_from_sol_file(
-            initial_solution_file, mps_data_model.get_variable_names());
+            initial_solution_file, problem_interface->get_variable_names());
 
     if (is_mip) {
       auto& mip_settings = settings.get_mip_settings();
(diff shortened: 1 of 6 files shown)


akifcorduk (approved):

cpp/include/cuopt/mathematical_optimization/cpu_optimization_problem.hpp
```diff
@@ -18,6 +18,11 @@
 #include <string>
 #include <vector>
 
+namespace cuopt::mathematical_optimization::io {
```
mlubin: Nit: move this inside the `cuopt::mathematical_optimization` block below with a `namespace io`.

cpp/cuopt_cli.cpp
```diff
@@ -138,8 +138,11 @@ int run_single_file(const std::string& file_path,
       std::make_unique<cuopt::mathematical_optimization::cpu_optimization_problem_t<int, double>>();
   }
 
-  cuopt::mathematical_optimization::populate_from_mps_data_model(problem_interface.get(),
-                                                                 mps_data_model);
+  // adopt_from_mps_data_model moves the parsed storage into a HOST-backed problem and falls back to
```
mlubin: Why is this comment needed? It looks like it just says what the function does; any relevant details should be on the function docstring.

cpp/tests/linear_programming/unit_tests/solution_interface_test.cu
```diff
@@ -424,6 +424,97 @@ End
...
+
+  // The moved-from model must be left empty (adopt consumed it).
+  EXPECT_EQ(model_for_move.get_n_variables(), 0);
+  EXPECT_EQ(model_for_move.get_n_constraints(), 0);
+}
+
+// adopt_from_mps_data_model must produce a problem identical to populate_from_mps_data_model.
+TEST_F(SolutionInterfaceTest, adopt_matches_populate_lp_file)
+{
+  const auto model = cuopt::mathematical_optimization::io::read_mps<int, double>(lp_file_);
```
mlubin:
```suggestion
  const auto model = io::read_mps<int, double>(lp_file_);
```

cpp/tests/linear_programming/unit_tests/solution_interface_test.cu
```diff
@@ -424,6 +424,97 @@ End
...
+  expect_adopt_matches_populate(model);
+}
+
+TEST_F(SolutionInterfaceTest, adopt_matches_populate_quadratic_constraints)
+{
+  const auto model = cuopt::mathematical_optimization::io::read_lp_from_string<int, double>(R"LP(
```
mlubin:
```suggestion
  const auto model = io::read_lp_from_string<int, double>(R"LP(
```
We should check why code is generated in this style and if there's anything we can do to prevent it.

mlubin (approved):

PR #1564, opened 2026-07-13
Fix SOC index collision and simplify QC-to-SOC conversion by routing linear QCs through general LDLT path

yuwenchen95:
## Description

Simplify SOC conversion in `translate_soc.hpp` and fix a variable-index collision when mixing general convex QCs with affine-style QCs (e.g. `hub1.mps.gz`).

**Bug fix:** When a general LDLT QC was converted before an affine QC with a linear part, both paths could allocate overlapping auxiliary column indices. This produced `SOC variable index ... collision` at runtime.

**Refactor — merge affine fast path into general LDLT path:** QCs with a nonzero linear part (`COLUMNS`) no longer use a separate affine auxiliary-variable + linking-row fast path. They always route through the general LDLT lift, which handles the linear term directly in the `s_0` / `s_{r+1}` rows. This removes:
- `qc_affine_heads` / `n_affine_linear_aux` pre-allocation
- post-loop affine linking rows
- the rotated-SOC-with-implicit-head affine branch

**Refactor — clearer path routing:** Classification is inverted to match the header comment: a QC is eligible for standard or rotated SOC fast paths when it matches those patterns (`rhs = 0`, no linear part); otherwise it uses the general path. 

**Tests:** Add `mixed_general_and_affine_soc_conversion` regression test that reproduces the pre-fix failure mode (general QC first, affine QC second).

## Checklist

- Testing
   - [x] Added tests
   - [ ] NA
- Documentation
   - [ ] Added new documentation
   - [x] NA


cpp/src/barrier/translate_soc.hpp
```diff
@@ -87,21 +87,23 @@ void convert_quadratic_constraints_to_second_order_cones(
...
-  // then apply a single column permutation to the linear model.
+  // SOC conversion routes each quadratic constraint as follows:
+  //
+  // Fast path (pattern-matched, rhs = 0, no linear part in COLUMNS):
+  //   1) standard Lorentz SOC — diagonal QCMATRIX with one head -s and tail diagonals +s
+  //        (common s > 0; dividing the row by s normalizes without changing feasibility)
+  //   2) rotated SOC — one off-diagonal cross -2*d on two heads plus tail diagonals +s
+  //        (d > 0, s > 0; stored as Q cross (head0, head1, -2*d); lift uses sqrt(d/s))
+  //
+  // General path (LDLT lift on H = Q + Q^T, must be PSD):
```
chris-maes: What is a LDLT lift?

cpp/src/barrier/translate_soc.hpp
```diff
@@ -258,9 +245,9 @@ void convert_quadratic_constraints_to_second_order_cones(
       tail_vars.push_back(pr.first);
     }
 
-    // Determine whether to use the general convex quadratic path.
-    // The general path is needed when Q does not fit any special SOC pattern,
-    // or when the RHS is nonzero (special cases require rhs = 0).
+    // Route to a fast SOC path when Q matches a known pattern (rhs = 0, no linear part);
+    // otherwise use the general LDLT lift. Eligibility must be exact: valid SOCs have an
```
chris-maes: use the general LDLT check?

cpp/src/barrier/translate_soc.hpp
yuwenchen95: It's the general path for quadratic constraints where we need to factorize the quadratic cost. We are using `LDL` factorization for factorization and positive semidefinite check.

chris-maes: I wrote that code. But I'm not sure how the word "lift" relates to that code.

yuwenchen95: I think we considered any introduction of new variables as a `lift`.

cpp/src/barrier/translate_soc.hpp
```diff
@@ -258,9 +242,9 @@ void convert_quadratic_constraints_to_second_order_cones(
       tail_vars.push_back(pr.first);
     }
 
-    // Determine whether to use the general convex quadratic path.
-    // The general path is needed when Q does not fit any special SOC pattern,
-    // or when the RHS is nonzero (special cases require rhs = 0).
+    // Route to a fast SOC path when Q matches a known pattern (rhs = 0, no linear part);
```
chris-maes: Nit: Q isn't matching a known pattern of rhs = 0 and no linear part--- the quadratic constraint is matching the pattern. Suggestion replace `when Q` with `when the quadratic constraint`

cpp/src/barrier/translate_soc.hpp
```diff
@@ -328,167 +289,89 @@ void convert_quadratic_constraints_to_second_order_cones(
...
-              qc.constraint_row_name.c_str());
-          }
-
-          cone.reserve(tail_vars.size() + 1);
-          cone.push_back(affine_head);
-          cone.insert(cone.end(), tail_vars.begin(), tail_vars.end());
-          cone_dim   = static_cast<i_t>(tail_vars.size() + 2);
-          is_rotated = 1;
-          rotated_cones.push_back(rotated_soc_t{affine_head, -1, tail_vars, true, 1});
+            "Quadratic constraint '%s' SOC Q: all positive diagonal coefficients must match; got "
```
chris-maes: Nit: Avoid using abbreviation `SOC` write out `second-order cone transformation: Q matrix all positive diagonal coefficients must match.

chris-maes (approved):
LGTM. Please test on the SOCP test set and QPLIB QCP test set to ensure that there are no regressions before merging. 

Thanks!

yuwenchen95:
> LGTM. Please test on the SOCP test set and QPLIB QCP test set to ensure that there are no regressions before merging.
> 
> Thanks!

I have retested SOCPs and the performance is consistent.

PR #1566, opened 2026-07-13
[DRAFT] Minimize cuopt_static test linkage to avoid 5 GB package size explosion

ramakrishnap-nv:
## Description

Follow-on to #1526 (includes all its commits). That PR correctly introduces hidden symbol visibility and `cuopt_static` for tests that need internal symbols — but switches **all** tests via the `ConfigureTest` macro, causing a ~67× size explosion in `libcuopt-tests`.

**Measured impact of #1526 as-is (from its own CI artifacts):**
| Package | Before | After #1526 |
|---|---|---|
| `libcuopt-tests` cuda13 linux-64 | ~75 MB | **4.99 GiB** |

This PR adds one commit on top of #1526 that limits `cuopt_static` linkage to only the tests that actually need it.

## Analysis

**Tests that genuinely need `cuopt_static`** — routing tests only. They all include `routing_test.cuh` which pulls in `cuopt::routing::detail` types (`solution_t`, `problem_t`, `fleet_order_constraints_t`, `fill_data_model_matrices`, etc.), and `local_search_cand_test.cu` / `top_k.cu` directly construct internal solver state.

**Tests that do NOT need `cuopt_static`:**
- MIP, LP, QP, SOCP, dual_simplex, distance engine, CLI, utilities tests
- Several of these call `raft::linalg::detail::cublassetpointermode` / `cusparsesetpointermode` — these are **raft inline headers** compiled into the TU, not cuopt hidden symbols, and require no special linkage
- `cuopttestutils` keeps `cuopt_static` PRIVATE (unchanged from #1526); non-routing tests never call `check_constraints.cu` functions so the linker never pulls in that object file

**GRPC tests** are left unchanged — they have explicit `target_link_libraries` and `grpc_client.cpp` lacks `CUOPT_EXPORT`, so `cuopt_static` remains correct.

## Changes (one commit on top of #1526)

- `cpp/tests/CMakeLists.txt`: Keep `ConfigureTest` → `cuopt` shared; add `ConfigureInternalTest` → `cuopt_static`
- `cpp/tests/routing/CMakeLists.txt`: Switch all 8 routing targets to `ConfigureInternalTest`

## Local build verification

```
cmake --build cpp/build --target LP_UNIT_TEST MIP_TEST ROUTING_UNIT_TEST

# Binary sizes confirm correct linkage:
# LP_UNIT_TEST        6.2 MB  (shared)
# MIP_TEST            2.2 MB  (shared)
# ROUTING_UNIT_TEST  44.0 MB  (static — expected, accesses detail:: types)
```

## Checklist

- Testing
  - [x] Existing tests cover these changes (no test logic changed, only linkage)
- Documentation
  - [x] NA


ramakrishnap-nv:
closing for #1581

PR #1570, opened 2026-07-14
refactor(grpc): expose public blocking client, move internal-only tests to mock suite

ramakrishnap-nv:
Add `grpc_client_config_t` as a public header and a slim blocking client `grpc_cpp_client_t` (pimpl over the internal `grpc_client_t`) so that `GRPC_INTEGRATION_TEST` no longer needs internal headers. Tests exercising the async job lifecycle (submit/check_status/cancel/delete) are removed from the integration test — they are already covered by `GRPC_CLIENT_TEST` mock tests. `StreamLogs_NotFound` is added to `GRPC_CLIENT_TEST`. Fixes the pre-existing bad relative include in `grpc_client_env.hpp`.

### DIFF
--- a/cpp/tests/linear_programming/grpc/CMakeLists.txt
+++ b/cpp/tests/linear_programming/grpc/CMakeLists.txt
@@ -99,7 +99,6 @@ target_include_directories(GRPC_INTEGRATION_TEST
     PRIVATE
     "${CUOPT_SOURCE_DIR}/include"
     "${CUOPT_SOURCE_DIR}/src/grpc"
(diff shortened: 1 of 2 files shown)


mlubin:
Do we have any code coverage utilities to verify that we're not losing coverage from this change?

ramakrishnap-nv:
> Do we have any code coverage utilities to verify that we're not losing coverage from this change?

We do run static analysis, but it catches mostly security threats and not anything else, may be we need to add one.

PR #1574, opened 2026-07-14
[DRAFT] refactor(tests): replace cuopt_static with CUOPT_INTERNAL_EXPORT + version script

ramakrishnap-nv:
Builds on #1566. Eliminates `cuopt_static` from all test linkage by introducing `CUOPT_INTERNAL_EXPORT` for the small set of internal symbols that test TUs reference directly. A GNU ld version script (`cmake/cuopt.map`) assigns public symbols to `CUOPT_1.0` and test-only internals to `CUOPT_PRIVATE_1.0`. Only `generator::dataset_t` and `generate_dataset` needed marking — all other routing internal types are template/inline and instantiate in-place. Removes `ConfigureInternalTest` entirely; all tests now use `ConfigureTest` (shared linkage).


cpp/cmake/cuopt.map
```diff
@@ -0,0 +1,30 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+#
+# GNU ld version script for libcuopt.so.
+#
+# CUOPT_1.0       — stable public API (CUOPT_EXPORT symbols)
```
mlubin: None of the C++ API is part of the stable ABI from what I understand. Only the C and Python interfaces are.

ramakrishnap-nv: Agree, I am currently trying a way to work on a way which would solve this problem of internal API usage as a whole. So just testing as of now.

ramakrishnap-nv:
Closing to rethink approach. The mix of CUOPT_INTERNAL_EXPORT + static-linked tests is not clean. Plan: assess which tests depend on internal APIs, determine if they're valuable, and consolidate into single per-module internal test binaries (one for routing, one for numerical optimization) that link against cuopt_static.

PR #1575, opened 2026-07-14
ci: enable codecov for Python test coverage

ramakrishnap-nv:
Removes the explicit `run_codecov: false` override, adds `codecov.yml` (patch checks informational, never blocks), suppresses spurious coverage.py warnings, and excludes `codecov.yml` from changed-files triggers.

Requires `CODECOV_TOKEN` to be set as a repo/org secret.

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -416,9 +416,7 @@ jobs:
     with:
       build_type: pull-request
       script: ci/test_python.sh
-    secrets:
-      script-env-secret-4-key: CODECOV_TOKEN
-      script-env-secret-4-value: ${{ secrets.CODECOV_TOKEN }}
+    secrets: inherit # zizmor: ignore[secrets-inherit]
   docs-build:
     needs: [conda-python-build, changed-files]
     permissions:
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -416,13 +416,7 @@ jobs:
     with:
       build_type: pull-request
       script: ci/test_python.sh
-    secrets:
-      script-env-secret-1-key: CUOPT_S3_URI
-      script-env-secret-1-value: ${{ secrets.CUOPT_S3_URI }}
-      script-env-secret-2-key: CUOPT_AWS_ACCESS_KEY_ID
-      script-env-secret-2-value: ${{ secrets.CUOPT_AWS_ACCESS_KEY_ID }}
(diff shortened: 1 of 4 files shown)


.github/workflows/pr.yaml
```diff
@@ -410,16 +414,9 @@ jobs:
...
       build_type: pull-request
       script: ci/test_python.sh
-    secrets:
-      script-env-secret-1-key: CUOPT_S3_URI
-      script-env-secret-1-value: ${{ secrets.CUOPT_S3_URI }}
-      script-env-secret-2-key: CUOPT_AWS_ACCESS_KEY_ID
-      script-env-secret-2-value: ${{ secrets.CUOPT_AWS_ACCESS_KEY_ID }}
-      script-env-secret-3-key: CUOPT_AWS_SECRET_ACCESS_KEY
-      script-env-secret-3-value: ${{ secrets.CUOPT_AWS_SECRET_ACCESS_KEY }}
+    secrets: inherit # zizmor: ignore[secrets-inherit]
```
KyleFromNVIDIA: Why has this been changed to `inherit`? Why not add the secret that you need?

ramakrishnap-nv: Hmm, I went with that since it was simpler, but I agree, better to use the detail

ramakrishnap-nv: Updated — since S3/AWS secrets are not needed for PR builds, slot 1 is free and we now pass `CODECOV_TOKEN` explicitly via `script-env-secret-1`.

ramakrishnap-nv: Also I created this PR for future in case I need to pass extra secrets, may I get review on this as well https://github.com/rapidsai/shared-workflows/issues/593

KyleFromNVIDIA (approved):

.github/workflows/pr.yaml
ramakrishnap-nv: Reverted to `secrets: inherit` for now — the explicit `script-env-secret-1` approach injects `CODECOV_TOKEN` into `GITHUB_ENV` but the current shared workflow overrides it with `env: CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}` (empty since it's not a declared input), causing the upload to fail. Will switch to explicit once rapidsai/shared-workflows#593 merges.

Iroy30 (approved):

PR #1576, opened 2026-07-15
Increase cuSparse version requierment for SpMVOp from 12.7 to 12.8

Bubullzz:
# Problem 

When running (and compiling) cuOpt with an env that has cuSparse 12.7, solving a problem creates a memory leak.

# Cause

The experimental API cuSparse_SpMVOp causes memory leaks in version 12.7 on the SpMVOpCreateDescr call.
This problem is solved in cuSparse 12.8.

# Solution

This PR increases the cuSparse version requierment for SpMVOp from 12.7 to 12.8. The memory leak is gone, and it results in simpler code because there is no need to do dlsym dispatching depending on the version. 

Disclaimer : cuSparse version **!=** cuda version. 
- cuSparse 12.7 is associated with cuda 13.2
- cuSparse 12.8 is associated with cuda 13.3

### DIFF
--- a/cpp/src/pdlp/cusparse_view.cu
+++ b/cpp/src/pdlp/cusparse_view.cu
@@ -270,51 +270,30 @@ void my_cusparsespmm_preprocess(cusparseHandle_t handle,
 }
 #endif
 
-#if CUOPT_CUSPARSE_VER_12_7_UP
-// SpMVOp symbols. Resolved at runtime via dlsym, because the runtime cuSPARSE version
-// might not match the headers used at compile time. cuSPARSE 12.7 corresponds to CUDA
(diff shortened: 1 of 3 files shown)


Bubullzz:
/ok to test [28ec571](https://github.com/NVIDIA/cuopt/pull/1576/commits/28ec571ba1c1ebf6361f41fc67b40e37b2c45895)

Bubullzz:
/ok to test [1566871](https://github.com/NVIDIA/cuopt/pull/1576/commits/1566871bce2fa7ac06127799c002d33b747f856c)

nguidotti (approved):
Looks good to me! Thanks for the quick fix @Bubullzz!

Kh4ster (approved):
All good!

PR #1577, opened 2026-07-15
Add more strategies of init point computation in barrier

yuwenchen95:
## Description

Extend barrier initial-point handling for SOCP and fix augmented-path KKT assembly at initialization.

- Add `ensure_initial_point_interior` to push linear variables into the nonnegative orthant and SOC blocks into strict Lorentz cone interior; apply to `x` and `z` at the end of `initial_point`.
- Fix augmented init sign: remove erroneous `diag.multiply_scalar(-1.0)` so the first-call augmented build produces `-(Q+D)` on the primal block, matching `augmented_multiply` and the refactor path.
- Size augmented primal/dual RHS and solution vectors to `max(n+m, device_augmented.n)` for sparse SOC expansion, zero-padding extra entries.
- Enable initial-point iterative refinement for LP/QP augmented solves by seeding `d_diag_` from host `diag`; skip IR for cone problems and direct-free linear variables where operator/factorization would disagree.
- Move primal, bound, and dual residual checks to after interior enforcement so diagnostics reflect the final starting point.
- Remove unused `inv_sqrt_diag` member and initialization.

## Checklist

- Testing
   - [ ] Added tests
   - [x] NA
- Documentation
   - [ ] Added new documentation
   - [x] NA


cpp/src/barrier/barrier.cu
```diff
@@ -2155,41 +2183,49 @@ int barrier_solver_t<i_t, f_t>::initial_point(iteration_data_t<i_t, f_t>& data)
...
-    const f_t norm_b = vector_norm_inf<i_t, f_t>(lp.rhs);
-    const f_t norm_c = vector_norm_inf<i_t, f_t>(lp.objective);
-    const f_t mu     = std::sqrt((1.0 + norm_b) * (1.0 + norm_c));
-    const f_t sqrt2  = std::sqrt(2.0);
-    const f_t x_soc  = mu * sqrt2;
-    const f_t z_soc  = mu * sqrt2;
-    // Linear orthant
-    for (i_t j = 0; j < cs; ++j) {
+  // Full primal+dual point; no factorization/solve (main loop factorizes later).
+  if (init_strategy == 2) {
```
rg20: Use enums instead of constants for initial point strategy type.

rg20:
@yuwenchen95 did you run QCQP benchmark? are there any regressions?

yuwenchen95:
> @yuwenchen95 did you run QCQP benchmark? are there any regressions?

Current change won't have any impact on the performance of QCQP since I only add the options but force the initialization of QCQP as before.

cpp/src/barrier/barrier.cu
yuwenchen95: Do you think we'd better rename `barrier_dual_initial_point` to `barrier_initial_point` now?

cpp/src/barrier/barrier.cu
```diff
@@ -94,6 +95,39 @@ bool validate_barrier_cone_layout(const lp_problem_t<i_t, f_t>& problem,
...
+// Push entries into interior of nonnegative orthant and SOC.
+template <typename i_t, typename f_t>
+static void ensure_initial_point_interior(dense_vector_t<i_t, f_t>& values,
+                                          f_t epsilon_adjust,
+                                          const std::vector<i_t>& linear_mask,
+                                          i_t linear_end,
+                                          const std::vector<i_t>& cone_dims)
+{
+  f_t min_linear = inf;
+  for (i_t j = 0; j < linear_end; ++j) {
```
mlubin, line 109: Can we call `values.ensure_positive(epsilon_adjust, linear_mask);` instead of this loop?

yuwenchen95: Good point. Now we push the entry when `min_linear <= epsilon_adjust` compared to `min_x <= 0.0` in the existing `ensure_positive`. We should redo the LP/QP test to see if we can use a consistent `ensure_positive` @rg20.

cpp/include/cuopt/mathematical_optimization/utilities/internals.hpp
```diff
@@ -142,5 +142,20 @@ enum presolver_t : int {
   PSLP    = CUOPT_PRESOLVE_PSLP
 };
 
+/**
```
chris-maes: I think it would be better if you put this in `barrier.hpp`.

cpp/src/barrier/barrier.cu
```diff
@@ -2460,6 +2487,43 @@ int barrier_solver_t<i_t, f_t>::initial_point(iteration_data_t<i_t, f_t>& data)
   settings.log.printf("min v %e min z %e\n", data.v.minimum(), data.z.minimum());
 #endif
 
+  // Residual checks below reflect the final initial point, after positivity shifts.
```
chris-maes: We did the residual checks before the positivity shift to see how well we are solving the least squares problem. We already print out the residuals in the log after the positivity shift. So I think these are not necessary. I would restore the previous code.

cpp/src/barrier/barrier.cu
chris-maes: I agree it would be better to move this code into `ensure_positive`. But we should test if this code actually yields an improvement.

ramakrishnap-nv (approved):
Lets add any documentation if we are missing

cpp/src/barrier/barrier.cu
```diff
@@ -2229,15 +2271,18 @@ int barrier_solver_t<i_t, f_t>::initial_point(iteration_data_t<i_t, f_t>& data)
   dense_vector_t<i_t, f_t> DinvFu(lp.num_cols);  // DinvFu = Dinv * Fu
   data.inv_diag.pairwise_product(Fu, DinvFu);
   dense_vector_t<i_t, f_t> q(lp.num_rows);
+  const i_t aug_base = lp.num_cols + lp.num_rows;
+  const i_t aug_size = use_augmented ? std::max<i_t>(aug_base, data.device_augmented.n) : aug_base;
```
hlinsen: What path makes `device_augmented.n` exceed `lp.num_cols + lp.num_rows`? The SOC layout keeps cone vars inside `num_cols` (`cone_var_start_ + total_cone_dim == num_cols`) and direct-free vars are linear columns `< cone_start`, so I can't construct a divergence — which would make this `max()` and the `rhs.set_scalar(0.0)` defensive no-ops rather than the actual fix. If there *is* a case where they diverge, the LMS-path buffers sized `lp.num_rows`/`lp.num_cols` later in this function would be under-sized too. Could you point at the instance that motivated the resize?

cpp/include/cuopt/mathematical_optimization/constants.h
```diff
@@ -145,7 +145,8 @@
 #define CUOPT_MIP_HYPER_SUBMIP_ENABLE_CPUFJ          "mip_hyper_submip_enable_cpufj"
 
 /* @brief QCQP (barrier) scaling hyper-parameters */
-#define CUOPT_QCQP_HYPER_RUIZ_EQUILIBRATION "qcqp_hyper_ruiz_equilibration"
+#define CUOPT_QCQP_HYPER_RUIZ_EQUILIBRATION         "qcqp_hyper_ruiz_equilibration"
```
rg20: is this a hyper parameter?  I think it should be a regular parameter

cpp/src/barrier/barrier.cu
```diff
@@ -2175,41 +2203,55 @@ int barrier_solver_t<i_t, f_t>::initial_point(iteration_data_t<i_t, f_t>& data)
   const bool use_augmented          = data.use_augmented;
   const bool has_direct_free_linear = data.n_direct_free_linear > 0;
 
-  // SOCP: data-dependent initial point following SeDuMi (Sturm, 1999).
-  //   mu = sqrt((1 + ||b||_inf) * (1 + ||c||_inf))
-  //   primal and dual: x = mu * e_K,  z = mu * e_K
+  const barrier_dual_initial_point_t input_strategy =
+    static_cast<barrier_dual_initial_point_t>(settings.barrier_dual_initial_point);
```
rg20: make the member variable itself an enum

cpp/include/cuopt/mathematical_optimization/constants.h
```diff
@@ -200,6 +201,11 @@
 #define CUOPT_METHOD_BARRIER      3
 #define CUOPT_METHOD_UNSET        4
 
+#define CUOPT_BARRIER_DUAL_INITIAL_POINT_AUTOMATIC             -1
+#define CUOPT_BARRIER_DUAL_INITIAL_POINT_LUSTIG_MARSTEN_SHANNO 0
+#define CUOPT_BARRIER_DUAL_INITIAL_POINT_DUAL_LEAST_SQUARES    1
```
chris-maes: Dual initial point dual least squares is a bit strange. There are two duals.

cpp/src/math_optimization/solver_settings.cu
```diff
@@ -124,6 +124,8 @@ solver_settings_t<i_t, f_t>::solver_settings_t() : pdlp_settings(), mip_settings
     {CUOPT_MIP_HYPER_SUBMIP_MIN_FIXRATE_CAP, &mip_settings.submip_params.min_fixrate_cap, f_t(0.0), f_t(1.0), f_t(0.1), "hard cap on the minimum fix rate for solving a sub-MIP"},
     {CUOPT_MIP_HYPER_SUBMIP_TARGET_MIP_GAP, &mip_settings.submip_params.target_mip_gap, f_t(0.0), f_t(1.0), f_t(0.01), "MIP gap target for the sub-MIP"},
     {CUOPT_MIP_HYPER_SUBMIP_ITERATION_LIMIT_RATIO, &mip_settings.submip_params.iteration_limit_ratio, f_t(0.0), f_t(1.0), f_t(0.8), "sub-MIP simplex-iteration limit as a factor of parent B&B iterations"},
+    // QCQP (barrier) hyper-parameter (hidden from default --help: name contains "hyper_")
+    {CUOPT_BARRIER_HYPER_INITIAL_POINT_SAFEGUARD, &pdlp_settings.barrier_hyper_initial_point_safeguard, f_t(0.0), std::numeric_limits<f_t>::infinity(), f_t(10.0), "margin pushing the barrier initial iterate into the interior of the nonnegative orthant / SOC"},
```
chris-maes: I don't think this should be a hyper parameter.

chris-maes: Let's make it a regular parameter

cpp/include/cuopt/mathematical_optimization/constants.h
yuwenchen95: Since we are including more initial point strategies, it's better to rename `dual initial point` as 'initial point' for all existing parameters. The only concern is we have already made '--barrier-dual-initial-point' visible to users and renaming will also affect it.

cpp/include/cuopt/mathematical_optimization/pdlp/solver_settings.hpp
```diff
@@ -294,13 +294,16 @@ class pdlp_solver_settings_t {
...
   i_t ordering{-1};
-  i_t barrier_dual_initial_point{-1};
+  barrier_dual_initial_point_t barrier_dual_initial_point{barrier_dual_initial_point_t::Automatic};
   i_t postsolve_info{-1};
   i_t barrier_presolve_bound_free_variables{-1};  // -1 automatic, 0 disabled, 1 enabled
   // Ruiz equilibration for QCQP (barrier) scaling: -1 automatic (row/column
   // imbalance heuristic), 0 disabled, 1 enabled. Distinct from PDLP's own Ruiz
   // scaling in pdlp_hyper_params_t.
   i_t qcqp_ruiz_equilibration{-1};
+  // Margin used to push the barrier method's initial iterate into the interior of the
```
chris-maes, line 304: I think this comment applies to another parameter.

chris-maes: Oh whoops. This is for the safeguard not the dual initial point. My mistake. Please ignore.

cpp/src/barrier/barrier.cu
```diff
@@ -97,6 +97,34 @@ bool validate_barrier_cone_layout(const lp_problem_t<i_t, f_t>& problem,
...
+                                          i_t linear_end,
+                                          const std::vector<i_t>& cone_dims)
+{
+  // Linear shift
+  std::vector<i_t> linear_only_mask(values.size(), 0);
+  std::copy(linear_mask.begin(), linear_mask.begin() + linear_end, linear_only_mask.begin());
+  values.ensure_positive(epsilon_adjust, linear_only_mask);
+
+  // Cone shift
+  i_t off = 0;
```
chris-maes: off -> offset

cpp/src/barrier/barrier.cu
```diff
@@ -97,6 +97,34 @@ bool validate_barrier_cone_layout(const lp_problem_t<i_t, f_t>& problem,
   return true;
 }
 
+// Push entries into interior of nonnegative orthant and SOC.
```
chris-maes: SOC -> second-order cone

cpp/src/barrier/barrier.cu
```diff
@@ -2182,41 +2210,54 @@ int barrier_solver_t<i_t, f_t>::initial_point(iteration_data_t<i_t, f_t>& data)
...
-    i_t off = 0;
-    for (size_t k = 0; k < lp.second_order_cone_dims.size(); k++) {
-      i_t q_k          = lp.second_order_cone_dims[k];
-      data.x[cs + off] = x_soc;
-      data.z[cs + off] = z_soc;
-      for (i_t j = 1; j < q_k; ++j) {
-        data.x[cs + off + j] = 0.0;
-        data.z[cs + off + j] = 0.0;
+
+    // SOC blocks: x = z = mu * e, with e = (sqrt(2), 0, ..., 0)
```
chris-maes: SOC -> second-order cone

cpp/src/barrier/barrier.cu
```diff
@@ -2182,41 +2210,54 @@ int barrier_solver_t<i_t, f_t>::initial_point(iteration_data_t<i_t, f_t>& data)
...
+    if (data.has_cones()) {
+      const i_t cs    = data.cone_start();
+      const f_t x_soc = mu * sqrt2;
+      const f_t z_soc = mu * sqrt2;
+      i_t off         = 0;
```
chris-maes: off -> offset

cpp/src/barrier/barrier.cu
```diff
@@ -2182,41 +2210,54 @@ int barrier_solver_t<i_t, f_t>::initial_point(iteration_data_t<i_t, f_t>& data)
...
-      i_t q_k          = lp.second_order_cone_dims[k];
-      data.x[cs + off] = x_soc;
-      data.z[cs + off] = z_soc;
-      for (i_t j = 1; j < q_k; ++j) {
-        data.x[cs + off + j] = 0.0;
-        data.z[cs + off + j] = 0.0;
+
+    // SOC blocks: x = z = mu * e, with e = (sqrt(2), 0, ..., 0)
+    if (data.has_cones()) {
+      const i_t cs    = data.cone_start();
```
chris-maes: cs -> cone_start

cpp/src/barrier/barrier.cu
```diff
@@ -2359,9 +2400,18 @@ int barrier_solver_t<i_t, f_t>::initial_point(iteration_data_t<i_t, f_t>& data)
 #endif
   }
 
-  float64_t epsilon_adjust = 10.0;
+  const f_t epsilon_adjust = settings.barrier_initial_point_safeguard;
+  // Push entries into interior of nonnegative orthant and SOC.
```
chris-maes: SOC -> second-order cone

cpp/src/barrier/barrier.cu
```diff
@@ -2359,9 +2400,18 @@ int barrier_solver_t<i_t, f_t>::initial_point(iteration_data_t<i_t, f_t>& data)
...
+  const bool has_soc   = data.has_cones();
+  const i_t linear_end = has_soc ? data.cone_start() : lp.num_cols;
+  auto ensure_interior = [&](dense_vector_t<i_t, f_t>& values,
```
chris-maes: Please don't create a lambda for this. Just call `ensure_initial_point_interior` directly.

chris-maes (approved):
Please make the following changes before merging. Thanks!

rg20 (approved):

PR #1580, opened 2026-07-16
[BUG] Fix variable fixing for initial solutions

nguidotti:
When feeding initial solutions to the population, the diversity manager was fixing the variables on a default constructed problem. If the problem has free variables, it means that it was fixed to an infinity bound.

## Checklist

- Testing
   - [ ] Added tests
   - [ ] NA
- Documentation
   - [ ] Added new documentation
   - [ ] NA

### DIFF
--- a/cpp/src/mip_heuristics/diversity/diversity_manager.cu
+++ b/cpp/src/mip_heuristics/diversity/diversity_manager.cu
@@ -240,12 +240,6 @@ void diversity_manager_t<i_t, f_t>::add_user_given_solutions(
     }
 
     if (problem_ptr->pre_process_assignment(init_sol_assignment)) {
-      // Seed the solution with the provided assignment BEFORE fixing integers and running the
-      // completion LP. Otherwise the fix uses the solution_t constructor's default assignment
-      // (the variable lower bounds), which is -inf for free variables; fixing such a variable
-      // makes fix_given_variables compute a NaN constraint bound (inf - inf) and trips the
-      // problem-representation validation. Seeding first also lets us keep the LP's repaired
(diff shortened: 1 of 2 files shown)


cpp/src/mip_heuristics/diversity/diversity_manager.cu
```diff
@@ -240,6 +240,16 @@ void diversity_manager_t<i_t, f_t>::add_user_given_solutions(
     }
 
     if (problem_ptr->pre_process_assignment(init_sol_assignment)) {
+      // Seed the solution with the provided assignment BEFORE fixing integers and running the
```
akifcorduk: I would remove both comments. They are just describing the bug

cpp/src/mip_heuristics/diversity/diversity_manager.cu
```diff
@@ -250,10 +260,6 @@ void diversity_manager_t<i_t, f_t>::add_user_given_solutions(
                              problem_ptr->integer_indices,
                              lp_settings,
                              static_cast<bound_presolve_t<i_t, f_t>*>(nullptr));
-      raft::copy(sol.assignment.data(),
-                 init_sol_assignment.data(),
-                 init_sol_assignment.size(),
-                 sol.handle_ptr->get_stream());
       bool is_feasible = sol.compute_feasibility();
```
akifcorduk, line 257: I would do one addition: check both the solution before lp and after lp and take whichever is feasible. PDLP (`run_lp_with_vars_fixed`) might make some solutions infeasible and pdlp returns the final solution.

akifcorduk (approved):

PR #1581, opened 2026-07-16
refactor(tests): link internal tests against cuopt_static

ramakrishnap-nv:
## Summary

Prerequisite structural work for enabling `-fvisibility=hidden` on `libcuopt.so` ([issue #1213](https://github.com/NVIDIA/cuopt/issues/1213)). An earlier attempt using `CUOPT_INTERNAL_EXPORT` annotations ([PR #1574](https://github.com/NVIDIA/cuopt/pull/1574)) was closed after the scope of symbols to annotate proved too large (including static data members that cannot carry visibility attributes).

This PR puts the test infrastructure in place so that hidden visibility can be enabled in a follow-up.

- **`cuopt_objs` OBJECT library** — compile sources once; both `cuopt` (shared) and `cuopt_static` (static, `BUILD_TESTS` only) consume the same object files
- **`STATIC_LIB` option in `ConfigureTest`** — links `cuopt_static` and omits `cuopttestutils` (which transitively pulls in the shared lib) for tests that need compiled internal symbols
- **`ROUTING_INTERNAL_TEST`** — new target consolidating GES solver, cross-route insertion, local-search candidate, and top-k tests; also registers four previously unbuilt files (`l0_scross_test.cu`, `local_search_cand_test.cu`, `load_balancing_test.cu`, `feasibility_jump_tests.cu`)
- **gRPC tests** — `GRPC_CLIENT_TEST` and `GRPC_INTEGRATION_TEST` now compile the gRPC client/mapper sources directly into the test binary, following the pattern already used by `GRPC_PIPE_SERIALIZATION_TEST`; no internal symbols need exporting from the shared lib
- **`cuOptIsCusparseRuntimeMixedPrecisionSupported()`** — promoted from an internal function to the public C API; `c_api_tests.cpp` updated to call the public version


cpp/tests/utilities/base_fixture.hpp
```diff
@@ -111,7 +111,12 @@ inline auto parse_test_options(int argc, char** argv)
  * the original functionality. In addition, this custom `main` function parses
  * the command line to customize test behavior, like the allocation mode used
  * for creating the default memory resource.
+ *
+ * When CUOPT_DISABLE_TEST_MAIN is defined (e.g. when compiling multiple test
```
mlubin: Do we need this macro? What about deleting `CUOPT_TEST_PROGRAM_MAIN` in the test files that no longer need a main?

akifcorduk:
`load_balancing_test.cu, feasibility_jump_tests.cu` the description says these are now part of `ROUTING_INTERNAL_TEST` but they are MIP tests.

cpp/include/cuopt/mathematical_optimization/cuopt_c.h
```diff
@@ -1055,6 +1055,14 @@ cuopt_int_t cuOptGetDualObjectiveValue(cuOptSolution solution,
  */
 cuopt_int_t cuOptGetReducedCosts(cuOptSolution solution, cuopt_float_t* reduced_cost_ptr);
 
+/**
+ * @brief Query whether the runtime cuSPARSE version supports mixed-precision SpMV
+ *        (FP32 matrix × FP64 vector). Requires cuSPARSE >= 12.5 at runtime.
+ *
+ * @return 1 if supported, 0 otherwise.
+ */
+cuopt_int_t cuOptIsCusparseRuntimeMixedPrecisionSupported();
```
akifcorduk: Why do we make this a public function?

ramakrishnap-nv: The test needs to know whether the cuSPARSE runtime supports mixed-precision SpMV so it can set the right expectation (success vs. error/non-optimal). Originally I added it as a public C API so the test binary (which links only against the shared `cuopt`) could call it — but that unnecessarily grows our public surface with an implementation detail.

Replaced with a compile-time `#if CUSPARSE_VERSION >= 12500` check in the test. For cuOpt's deployment model the compile-time and runtime toolkit versions match, so this is equivalent and avoids adding anything to the public API.

ramakrishnap-nv:
Good catch — those files were tried during development but removed because their constructor signatures no longer match the current API (`load_balancing_test.cu` and `feasibility_jump_tests.cu` are MIP tests, not routing, and both had stale APIs). The description was wrong; updated to accurately describe what's in each target.

mlubin (approved):
Wait for @akifcorduk's approval as well

cpp/tests/internal/CMakeLists.txt
```diff
@@ -0,0 +1,37 @@
+# cmake-format: off
+# SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+# cmake-format: on
+
+# Single binary for all numopt tests that require internal symbols (cuopt_static).
+# Add new internal numopt tests here instead of creating separate STATIC_LIB targets.
+ConfigureTest(NUMOPT_INTERNAL_TEST
```
akifcorduk, line 8: Is there a reason we are creating one giant binary? It kind of removes the isolation of test that belong to different functional parts of the solver. This increases link time as well. It might also create name collusions on which you have encountered some of them and fixed with static functions. Also now when we want to run a test, it is more cumbersome with `--gtest-filter`.

cpp/tests/linear_programming/c_api_tests/c_api_tests.cpp
```diff
@@ -441,23 +438,35 @@ TEST(c_api, pdlp_precision_single)
...
-    auto status = test_pdlp_precision_mixed(filename.c_str(), &termination_status, &objective);
-    bool solve_returned_error = (status != CUOPT_SUCCESS);
-    bool solve_returned_non_optimal =
-      (status == CUOPT_SUCCESS && termination_status != CUOPT_TERMINATION_STATUS_OPTIMAL);
-    EXPECT_TRUE(solve_returned_error || solve_returned_non_optimal);
-    return;
+  // Mixed-precision SpMV (FP32 matrix × FP64 vector) requires cuSPARSE >= 12.5 at BOTH
+  // compile time (header) and runtime (loaded .so). The header version (#if) guards symbol
+  // availability; the runtime check below mirrors is_cusparse_runtime_mixed_precision_supported().
+#if CUSPARSE_VERSION >= 12500
```
akifcorduk, line 452: What's the reason for the code duplication? Is this because this test belongs to the public API?

cpp/CMakeLists.txt
```diff
@@ -634,8 +619,6 @@ target_link_libraries(cuopt
         ${CUDSS_LIB_FILE}
         PRIVATE
         ${CUOPT_PRIVATE_CUDA_LIBS}
-        simde::simde
```
akifcorduk, line 637: I think this is needed for the compilation and internal objects(Your compilation already fails because of this). OpenMP is also needed IMO.

akifcorduk (commented):
Thanks Ram! My biggest concern is having a single binary for all tests. If the binary sizes are not exploding, I would keep the separate binaries.

ramakrishnap-nv:
> Thanks Ram! My biggest concern is having a single binary for all tests. If the binary sizes are not exploding, I would keep the separate binaries.

If we have multiple internal Api testing binaries, each would need libcuopt statically linked, which adds 500mb to each such test binary, increasing size drastically.

cpp/tests/utilities/base_fixture.hpp
ramakrishnap-nv: The macro is still used in 8 standalone test binaries (routing level0/level1, distance_engine, local_search_cand). Those are built without `internal/main.cu` — each source file provides its own `main()` via this macro. Deleting it would break them. Tests consolidated into `NUMOPT_INTERNAL_TEST` and `ROUTING_INTERNAL_TEST` already had their `CUOPT_TEST_PROGRAM_MAIN()` calls removed in commit `0c2f0797`.

cpp/tests/linear_programming/c_api_tests/c_api_tests.cpp
ramakrishnap-nv: Yes — `C_API_TEST` links against the shared `cuopt` (not `cuopt_static`) because it tests the public C ABI. The runtime cuSPARSE check is inlined here because `C_API_TEST` only has access to the public API — there's no internal helper it can call. The `EXPECT_NE` appearing twice (once in the `!runtimeSupported` branch and once in the `#else` block) is intentional: both code paths produce an error return, just from different root causes (runtime mismatch vs. compile-time unsupported headers). The alternative would be a helper function, but given how small the check is and that it only appears twice, inlining felt clearer.

cpp/CMakeLists.txt
ramakrishnap-nv: Fixed in commits `8f23153c` (OpenMP::OpenMP_CUDA) and `c1598967` (simde::simde). Both were PRIVATE deps of the old `cuopt` SHARED target that were providing compile-time flags/includes; after the refactor to `cuopt_objs`, they need to be on the OBJECT library where sources are actually compiled.

ramakrishnap-nv:
@akifcorduk may I get another round of the review

akifcorduk (approved):

cpp/tests/utilities/base_fixture.hpp
```diff
@@ -111,6 +111,7 @@ inline auto parse_test_options(int argc, char** argv)
  * the original functionality. In addition, this custom `main` function parses
  * the command line to customize test behavior, like the allocation mode used
  * for creating the default memory resource.
+ *
  */
```
tmckayus, line 115: was this a style thing? seems odd to insert a blank comment line.

tmckayus (approved):

cpp/tests/utilities/base_fixture.hpp
ramakrishnap-nv: Mostly, but I will merge this PR and may be create another PR to fix this.

ramakrishnap-nv: Good catch — it's a stray blank comment line with no purpose (not a style requirement). Removed it in #1625 (commit 969511b7), which merges this area.

PR #1589, opened 2026-07-20
Fix clique size and numerical issue bugs

akifcorduk:
Small cliques are reduced to pairwise conflicts in adj list format and the limit was 512. For some problems, this is too agressive and creates large amount of adj list entries that exceed int32 size. This PR fixes it by reducing the adj list storage threshold to 64.

This PR also fixes a roundoff error that is revealed after extraction of cliques from mixed rows were instroduced. We check with tolerance now.


aliceb-nv (approved):
Good forward thinking for the int64_t for offsets :) We should probably go over the whole codebase someday to ensure all CSRs use 64bit for the offsets. Perhaps a `off_t` typedef or such and agentic checks

PR #1590, opened 2026-07-20
Print logs when exception is caught

akifcorduk:
Previously, exceptions inside the MIP OpenMP region were caught silently and rethrown after the region ended. Now their message is logged when caught while preserving the deferred rethrow.


aliceb-nv (approved):
Let's also check the lp/ side of the codebase for similar patterns

mlubin (approved):

PR #1482, opened 2026-06-26
Recursive RINS

nguidotti:
This PR implements the recursive RINS heuristics. More specifically, it applies a dive-like procedure to create a RINS sub-MIP, presolved it with Papilo and then passes to the cut-and-branch procedure. It also launch a CPU FJ worker over the sub-MIP that runs in parallel to cut-and-branch.

Since the current implementation, does not have an unified problem representation, this PR also adds:
- A routine for converting a `lp_problem_t` (standard form) to `user_problem_t` (range form).
- Allow Papilo to be applied to an `user_problem_t` instead of `optimization_problem_t`.
- Adds an upper bound callback for retrieving the upper bound of the main solver (since it may change if another worker find a better incumbent).

Additionally, the number of worker per diving heuristic is no longer tied to the total number of workers, which allows the solver to use all diving heuristics in low thread counts (we rotate between the different heuristics for each diving worker launch).

## Results
MIPLIB2017, 10min, GH200

```
================================================================================
 main-2026-07-03 (1) vs recursive-submip (2)
================================================================================

------------------------------------------------------------------------------------------------------------------------------
|                                        |       Run 1        |       Run 2        |     Abs. Diff.     |   Rel. Diff. (%)   |
------------------------------------------------------------------------------------------------------------------------------
| Imported                                                 240                  240                   +0                 --- |
| Feasible                                                 227                  228                   +1                 --- |
| Optimal                                                   86                   91                   +5                 --- |
| Solutions with <0.1% primal gap                          138                  151                  +13                 --- |
| Nodes explored (mean)                              1.326e+07            1.296e+07           -3.057e+05               -2.31 |
| Nodes explored (shifted geomean)                   1.262e+04            1.239e+04                 -230               -1.82 |
| Relative MIP gap (mean)                               0.2777               0.2726             -0.00512               -1.84 |
| Relative MIP gap (shifted geomean)                   0.09089              0.08901            -0.001883               -2.07 |
| Solve time (mean)                                      422.6                435.4               +12.79               +3.03 |
| Solve time (shifted geomean)                           196.9                226.6               +29.66               +15.1 |
| Primal gap (mean)                                      9.969                 9.29               -0.679               -6.81 |
| Primal gap (shifted geomean)                          0.4584               0.3491              -0.1093               -23.8 |
| Primal integral (mean)                                 25.02                22.97               -2.051                -8.2 |
| Primal integral (shifted geomean)                      2.849                2.421              -0.4275                 -15 |
[...]

PR #1591, opened 2026-07-20
Make cuopt.routing importable without a GPU

ramakrishnap-nv:
## Summary
Importing `cuopt.routing` required a GPU: it pulled in `cuopt.routing.utils`/`utils_wrapper`, whose `generate_dataset` used `cudf.Series()` as **default argument values**. Default args are evaluated at import time, and constructing even an empty `cudf.Series` needs a GPU — so `import cuopt.routing` failed with `cudaErrorNoDevice` on a GPU-less host (e.g. a CPU-only client building/serializing a routing problem to submit to a remote solver).

Fix at the source: move the empty-series construction out of the signatures into the function bodies via `None` sentinels. `import cudf` itself doesn't touch the GPU, so the modules now import GPU-free and the helpers behave identically when called. Adds `__all__` for explicit re-exports and a subprocess test that imports the package with no visible GPU.

Validated: `import cuopt.routing` + helpers + `DataModel` build all work under `CUDA_VISIBLE_DEVICES=""`; the routing suite passes on GPU.


tmckayus (approved):

ramakrishnap-nv:
Thanks for the review. Addressed in ca9e016:

- **Cover the complete set of removed top-level names** (test_gpu_free_import.py) — done. The test now enumerates all five moved names (`generate_dataset`, `DatasetDistribution`, `add_vehicle_constraints`, `create_pickup_delivery_data`, `update_routes_and_vehicles`) and fails if any leaks back onto `cuopt.routing`.

The remaining CodeRabbit comments were against earlier revisions of this PR and are now stale:
- `__init__.py` `__getattr__`/`__dir__` — the lazy-import approach was dropped in favour of removing the helpers from the namespace entirely; that code no longer exists.
- `utils.py` "document the `None` defaults" — the `cudf.Series()`→`None` sentinel change was reverted; `utils.py` matches `main` again.
- `test_routing_utils_imports_without_a_gpu` suggestions — that test function was removed along with the sentinel change.

ramakrishnap-nv:
### `wheel-tests-cuopt` failures are unrelated to this PR

The failing jobs crash in three **MILP / linear-programming** tests with `Fatal Python error: Aborted` (SIGABRT), taking down the xdist workers (`--max-worker-restart=0`, so the job goes red):

- `tests/linear_programming/test_incumbent_callbacks.py::test_incumbent_get_callback[/mip/swath1.mps]` (gw3)
- `tests/linear_programming/test_incumbent_callbacks.py::test_incumbent_get_set_callback[/mip/neos5-free-bound.mps]` (gw1)
- `tests/linear_programming/test_lp_solver.py::test_heuristics_only` (gw2)

This PR changes **only** routing Python namespace (`git diff main...HEAD` touches zero LP/MILP/solver files), so it cannot cause a native abort in the MILP solver. Notably **`conda-python-tests` pass** — the crash is wheel-build-specific and reproduces on the MILP code that came in via the merge of `main` (recent MILP/MPS work: #1429, #1589, #1482).

I've re-triggered the failed jobs to rule out flakiness under `-n 4`. If they crash again, this is a `main`-level wheel-build regression in the MILP solver that needs a separate fix — it isn't introduced or fixable here.

ramakrishnap-nv:
The CI failures are not caused by this PR's changes.

All three failing tests (`test_incumbent_get_callback[swath1.mps]`, `test_incumbent_get_set_callback[swath1.mps]` / `neos5-free-bound.mps`, `test_heuristics_only`) use `swath1.mps` or another MPS file and crash with `std::bad_array_new_length` — a C++ exception thrown when an invalid size is passed to `new[]`.

This regression was introduced in main and is not present in the PR 1549 run from today (same runner configs, all green). The two PRs have different main bases: PR 1549 ran on `24bf137f` while PR 1591's base is `3e8acd7f` (18 commits newer). The most likely culprit is **#1429 "Fast MPS parser for free-format MPS files"** (`3e8acd7f`), which is the top commit in main and directly touches MPS file parsing. A parser bug that produces a negative or overflowed count would explain `bad_array_new_length` on these exact MPS test files.

This PR can be unblocked once the main regression is identified and fixed.

PR #1593, opened 2026-07-21
Fix libomp ABI bug

### LINKED ISSUE #1592: wheel: swath1.mps aborts with std::bad_array_new_length after fast MPS parser (#1429)
## Summary

Three wheel tests crash with `std::bad_array_new_length` on all GPU runners since the fast MPS parser (#1429) merged into main on 2026-07-20:

- `tests/linear_programming/test_incumbent_callbacks.py::test_incumbent_get_callback[/mip/swath1.mps]`
- `tests/linear_programming/test_incumbent_callbacks.py::test_incumbent_get_set_callback[/mip/swath1.mps]`
- `tests/linear_programming/test_lp_solver.py::test_heuristics_only` (also uses swath1.mps)

## Symptom

```
terminate called after throwing an instance of 'std::bad_array_new_length'
  what():  std::bad_array_new_length
Fatal Python error: Aborted
```

The crash happens inside `solver.Solve()` before any callback fires.

## Key observations

- Fails on all 6 wheel-test runners (rtxpro6000 and l4, amd64 and arm64)
- `neos5-free-bound.mps` variants of the same tests pass
- conda-python-tests pass on identical GPU configs — the conda binary either parses swath1.mps correctly or solves it at root (0 callbacks → pytest.skip), suggesting a build-configuration difference between conda and wheel binaries
- Tests were previously disabled (May 2026, #1208) for a different swath1 crash (PDLP CUDA graph capture race), then re-enabled in July 2026

## Root cause hypothesis

The fast MPS parser (#1429) mis-parses `swath1.mps` in the wheel build, producing an invalid array size that causes `new[]` to throw. The conda build is unaffected, likely due to different compiler flags or linked library versions.

## Workaround

swath1.mps tests temporarily skipped in #1591 while this is investigated.

aliceb-nv:
This PR adds a workaround resolving a buggy interaction between GCC's generated OMP helpers and LLVM's OpenMP runtime.
This was introduced as a latent issue in the Recursive RINS PR, and then exposed by the fast MPS PR which moved the wheel's runtime to LLVM OpenMP.

In more details:
> The failure was caused by an LLVM libomp bug in its GCC GOMP_task compatibility layer: for an included OpenMP task (if(false)), libomp skipped GCC’s generated
  firstprivate copy function, leaving a captured non-trivial C++ object unconstructed; the RINS sub-MIP path then read this invalid object and eventually threw
  std::bad_array_new_length. The unrelated merge merely exposed the latent runtime-dependent bug in libomp-linked wheels; explicitly executing the included-task
  path synchronously avoids the faulty compatibility path.

I will file a min repro and bug to LLVM later.

closes #1592 

## Checklist

- Testing
   - [ ] Added tests
   - [ ] NA
- Documentation
   - [ ] Added new documentation
   - [ ] NA

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -53,7 +53,7 @@ jobs:
       packages: read
       pull-requests: read
     secrets: inherit # zizmor: ignore[secrets-inherit]
-    uses: rapidsai/shared-workflows/.github/workflows/conda-cpp-build.yaml@main
+    uses: rapidsai/shared-workflows/.github/workflows/conda-cpp-build.yaml@release/26.08
     with:
       build_type: ${{ inputs.build_type || 'branch' }}
       branch: ${{ inputs.branch }}
@@ -69,7 +69,7 @@ jobs:
       packages: read
       pull-requests: read
     secrets: inherit # zizmor: ignore[secrets-inherit]
-    uses: rapidsai/shared-workflows/.github/workflows/conda-python-build.yaml@main
+    uses: rapidsai/shared-workflows/.github/workflows/conda-python-build.yaml@release/26.08
     with:
       build_type: ${{ inputs.build_type || 'branch' }}
       branch: ${{ inputs.branch }}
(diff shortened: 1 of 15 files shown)


akifcorduk (approved):

ramakrishnap-nv (approved):

aliceb-nv:
test failures unrelated. addressed in https://github.com/NVIDIA/cuopt/pull/1594

PR #1596, opened 2026-07-21
test(routing): add null-value rejection test for cost matrix

ramakrishnap-nv:
Closes #1220.

`add_cost_matrix` rejects NULL values via `validate_matrix`; adds a negative test to lock down that contract.


ramakrishnap-nv:
Closing in favour of the original PR #1220, which will be retargeted to main.

PR #1597, opened 2026-07-21
Routing over gRPC: VRP server + compiled C++/Cython client

ramakrishnap-nv:
End-to-end VRP (vehicle routing) over the existing gRPC server, using a single architecture consistent with LP/MILP (compiled C++/Cython client, no pure-Python protobuf path). Rebased on `main` (which now includes #1591).

- **Server** (from @tmckayus's `feature/grpc-server-vrp`, rebased onto main): `cuopt_routing.proto`, `vrp_request` arm, server dispatch + `grpc_routing_problem_mapper` + `cpu_routing_problem_t`.
- **Client**: `cpu_routing_solution_t` + client-side parse; `submit_vrp/get_vrp_result/solve_vrp` on `grpc_client_t` (unary-only); a `grpc_python_client_t` shim; and `cuopt.grpc.routing.RoutingClient` (Cython) that walks the DataModel store-then-build IR into `cpu_routing_problem_t` — the single mapping site that replaces `to_host_problem` (removed).

### Testing
- **Coverage**: `test_grpc_serialization` asserts every `cuopt.routing._deferred._SETTERS` name is mapped (fail-loud on a new setter), exercises each field category via a no-server `problem_summary` probe, and covers **numpy / pandas / cuDF** inputs.
- **End-to-end** against `cuopt_grpc_server`: remote solve matches local (objective, route, vehicle count) for cost-matrix, capacity, and time-window problems.
- **GPU-less client**: validated in a completely GPU-disconnected container (`docker run` without `--gpus`) — a *local* solve fails (`CUDARuntimeError`), while remote solves with **numpy and pandas** inputs succeed against a host GPU server.

VRP is unary-only (no >2 GiB chunking yet), matching the server POC.


ramakrishnap-nv:
Thanks for the review — all findings addressed.

**Client (002c971):**
- **Falsy-zero `time_limit=0`** (`grpc_client.pyx`): `_apply_settings` now uses `is not None`, so an explicit `0` is applied instead of silently dropped.
- **`result_vrp` `not_ready`** (`cython_grpc_client.cpp`): added the same in-flight `status()` pre-check as `result()`, so polling `result()` on a running job returns a structured not-ready signal (`RoutingClient.result()` returns `None`) instead of a generic `GetResult` failure. Re-added `not_ready` to the `.pxd`.

**Server (7a61e65):**
- **Error-message leak** (`grpc_routing_problem_mapper.cpp`): `map_routing_solution_to_proto` now runs `get_error_status()` through a sanitizer that mirrors `format_cuopt_error` (parses the structured `{"error_type","msg"}` payload to `type: msg`), so raw internal detail isn't sent to clients. Inlined rather than cross-including the heavy `server/grpc_server_types.hpp` into the client-linked mapper.
- **`to_device` host-buffer lifetime** (`cpu_routing_problem.cu`): the transient host buffers in `copy_u8_as_bool()` and the initial-solutions `types` conversion feed async H2D copies; the stream is now drained before they go out of scope, so the copies can't read freed memory.
- **`to_device` capacity validation** (`cpu_routing_problem.cu`): each capacity dimension is now checked for `num_orders` demand and `fleet_size` capacity values before building the device view, preventing an OOB read from an undersized (e.g. raw-proto) request.

**Stale:** the `solve_vrp` log-streaming comment is moot — `solve_vrp` was removed earlier (no caller).

Verified: clean rebuild; valid capacity/bool-array/time-window solves match locally; undersized capacity is rejected; 8 serialization tests + GPU-less container path still pass. Full async parity (`status`/`cancel`/live log streaming) remains a planned follow-up.

cpp/src/grpc/codegen/generated/cuopt_remote_data.proto
```diff
@@ -51,6 +51,7 @@ enum VariableType {
 enum ProblemCategory {
   LP = 0;
   MIP = 1;
+  VRP = 2;
```
rg20: may be use a different number here,  we will add QP, SOCP,  and many more. So for VRP, does it make sense to use something like 100?

cpp/src/grpc/codegen/generated/cuopt_remote_data.proto
```diff
@@ -343,4 +344,6 @@ message ChunkedResultHeader {
   double ws_last_restart_kkt_score = 3005;
   double ws_sum_solution_weight = 3006;
   int32 ws_iterations_since_last_restart = 3007;
+  bool is_vrp = 4000;
```
rg20, line 349: is_vrp a  boolean, it should be true or false

cpp/src/grpc/codegen/generated/generated_enum_converters_problem.inc
```diff
@@ -27,6 +27,7 @@ cuopt::remote::ProblemCategory to_proto_problem_category(problem_category_t v)
   switch (v) {
     case problem_category_t::LP: return cuopt::remote::LP;
     case problem_category_t::MIP: return cuopt::remote::MIP;
+    case problem_category_t::IP: return cuopt::remote::MIP;
```
rg20: I don't think this should be done here,   the conversion should be done in the solve calls

cpp/src/grpc/codegen/field_registry.yaml
```diff
@@ -165,9 +165,17 @@ enums:
...
+    # (LP, MIP, VRP) intentionally diverge: VRP is a transport-only routing
+    # category with no math-opt enumerator, and IP is a C++-only value that
+    # serializes as MIP on the wire.
     values:
     - LP
     - MIP
+    - VRP:
+        proto_only: true   # proto value with no problem_category_t; from_proto throws
+    cpp_aliases:
+      IP: MIP              # C++-only enumerator; to_proto serializes it as MIP
```
rg20: don't think this decision should happen here, we dont want any logic related to problem representation in the grpc code

cpp/src/grpc/cuopt_routing.proto
```diff
@@ -0,0 +1,168 @@
+// SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+syntax = "proto3";
```
rg20, line 4: is this generated code?

cpp/src/grpc/codegen/generated/cuopt_remote_data.proto
ramakrishnap-nv: Done — moved VRP to 1000 (`VRP = 1000`) so LP/MIP and future math-opt categories (QP, SOCP, ...) keep the low sequential numbers. Set in field_registry.yaml and regenerated.

cpp/src/grpc/codegen/generated/cuopt_remote_data.proto
ramakrishnap-nv: Yes, `is_vrp` is a plain proto3 bool (defaults to false; set true only for routing results). It stays as the result-header signal the client reads to know how to decode the payload — it does not carry any problem-representation mapping, so it's fine as a boolean.

cpp/src/grpc/codegen/generated/generated_enum_converters_problem.inc
ramakrishnap-nv: Agreed. That converter was actually unused — the job category is assigned directly from the request oneof in the service (`has_lp_request`/`has_mip_request`/`has_vrp_request`), which is the 'solve call' you mean. I marked `problem_category` as `proto_enum_only`, so codegen now emits only the wire enum and no C++ <-> proto converter. The IP->MIP / VRP->throw decisions are gone from the generated .inc.

cpp/src/grpc/codegen/field_registry.yaml
ramakrishnap-nv: Removed. Dropped the `cpp_aliases: {IP: MIP}` and `VRP: {proto_only}` decisions from the registry; `problem_category` is now just `proto_enum_only` with the wire values (LP, MIP, VRP=1000). No problem-representation mapping is generated into the grpc code anymore.

cpp/src/grpc/cuopt_routing.proto
ramakrishnap-nv: No — `cpp/src/grpc/cuopt_routing.proto` is hand-authored. It's not in `codegen/generated/` and isn't produced by `generate_conversions.py`; it imports `ProblemCategory` from the generated `cuopt_remote.proto`.

ramakrishnap-nv: Follow-up plan: bringing `cuopt_routing.proto` under the field_registry codegen (so it's generated like the math-opt protos) is worthwhile but sizable — the routing messages fit the existing `repeated_messages` mechanism, but `RoutingSolution.objective_values` is a `map<int32,double>` (no map support in the generator yet) and the 377-line mapper has bespoke logic (cost-matrix grouping by vehicle_type, capacity dims, per-vehicle service times, PDP/breaks/matching/precedence, initial solutions). To keep this PR focused and landable, I'll do the full VRP registry + codegen (proto **and** mapper) as a dedicated follow-up PR rather than expand this one.

ramakrishnap-nv: Tracked as #1616 for the follow-up.

cpp/src/grpc/codegen/generated/cuopt_remote_data.proto
rg20: here the value is 4000 for a boolean, is this intentional?

cpp/src/grpc/codegen/field_registry.yaml
```diff
@@ -165,9 +165,17 @@ enums:
...
     domain: problem
+    # Transport-only wire enum. Deliberately generates NO C++ <-> proto converter
+    # (proto_enum_only): the job's category is assigned directly from the request
+    # oneof in the service (has_lp_request / has_mip_request / has_vrp_request), so
+    # no enum conversion — and no problem-representation decision — lives in the
+    # grpc codegen. VRP sits at 1000 to leave the low numbers for future math-opt
+    # categories (QP, SOCP, ...).
+    proto_enum_only: true
     values:
     - LP
```
rg20: may be explicitly state the values for LP and MIP as well?

cpp/src/grpc/grpc_routing_problem_mapper.cpp
```diff
@@ -0,0 +1,377 @@
...
+    p.cost_matrices.push_back(std::move(out));
+  }
+  for (auto const& tm : pb.transit_time_matrices()) {
+    cuopt::routing::cpu_cost_matrix_t out;
+    out.vehicle_type = static_cast<uint8_t>(tm.vehicle_type());
+    copy_repeated_to_vector(tm.values(), out.matrix);
+    p.transit_time_matrices.push_back(std::move(out));
+  }
+
+  copy_repeated_to_vector(pb.vehicle_start_locations(), p.vehicle_start_locations);
```
rg20, line 109: this is kind of error prone, how do we ensure we don't miss out anything when we add new vectors?  

ideally this should be auto-generated as well

cpp/src/grpc/grpc_routing_problem_mapper.cpp
```diff
@@ -0,0 +1,377 @@
...
+    case solution_status_t::TIMEOUT: return cuopt::remote::ROUTING_TIMEOUT;
+    case solution_status_t::EMPTY: return cuopt::remote::ROUTING_EMPTY;
+    case solution_status_t::ERROR: return cuopt::remote::ROUTING_ERROR;
+  }
+  return cuopt::remote::ROUTING_ERROR;
+}
+
+}  // namespace
+
+void map_proto_to_routing_problem(const cuopt::remote::RoutingProblem& pb,
```
rg20, line 88: Does having unit tests for round trip make sense?

cpp/src/routing/cpu_routing_problem.cu
```diff
@@ -0,0 +1,322 @@
...
+  // stream before it goes out of scope so the copy does not read freed host
+  // memory.
+  stream.synchronize();
+  return d;
+}
+
+}  // namespace
+
+std::pair<data_model_view_t<int, float>, cpu_routing_problem_t::device_data_ptr>
+cpu_routing_problem_t::to_device(raft::handle_t* handle) const
```
rg20, line 97: This looks like a lot of boiler plate code. Error prone?  can we avoid?

python/cuopt/cuopt/grpc/routing/grpc_client.pyx
```diff
@@ -0,0 +1,400 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
rg20, line 1: Why do we need any code on python side?  

we do not have any python code for MIP/LP right?

ramakrishnap-nv: LP/MIP does have Python code — `cuopt.grpc.linear_programming.Client` (added in #1430), with the same submit/result surface. This `cuopt.grpc.routing.RoutingClient` is its symmetric counterpart. Both are thin Cython wrappers over the *same* C++ shim `grpc_python_client_t` — no grpcio/protobuf in the wheel, no `_pb2` stubs. The Python layer's only job is to take the native Python problem object (`DataModel` / `data_model_view`), populate the C++ host struct, call the typed shim method, and decode the result — it's the GPU-less Python entry point. Unifying/generating these per-domain clients is folded into #1616.

cpp/src/routing/cpu_routing_problem.cu
ramakrishnap-nv: Agreed — the host-struct population is boilerplate, the same class of hand-written mapping as the proto mapper. Both are slated to be auto-generated from field_registry in #1616, which removes this by construction. Added to that issue's scope.

cpp/src/grpc/grpc_routing_problem_mapper.cpp
ramakrishnap-nv: Yes. Rather than hand-write a large round-trip test against a mapper that #1616 will regenerate, I've folded a round-trip + coverage test into #1616 (written against the generated mapper). In the meantime the path is covered end-to-end by `GRPC_VRP_TEST_DRIVER` (problem→proto→server→solve) and `test_remote_solve_matches_local` (remote result matches a local solve), plus the Python `test_routing_grpc_serialization` setter-coverage guard.

cpp/src/grpc/grpc_routing_problem_mapper.cpp
ramakrishnap-nv: Agreed, and that's exactly the goal of #1616 — generate the mapper (and proto) from field_registry so a new array can't be added in one place and missed in the mapper. The Python side already has this guard (`test_routing_grpc_serialization` asserts the client handles every recorded setter); #1616 brings the equivalent structural guarantee to the C++ mapper.

cpp/src/grpc/codegen/field_registry.yaml
ramakrishnap-nv: Done in ae8d5d8a — `LP: 0`, `MIP: 1`, `VRP: 1000`. No wire change (they were already the auto-numbered values).

cpp/src/grpc/codegen/generated/cuopt_remote_data.proto
ramakrishnap-nv: Yes, intentional. `4000` is the proto field number (tag), not the value — and it follows the ChunkedResultHeader block-numbering scheme documented at the top of field_registry.yaml (LP 1000-1999, MIP 2000-2999, WS 3000-3999). VRP's header fields take the next block: `is_vrp = 4000`, `routing_solution = 4001`. Keeping VRP in its own block avoids colliding with future LP/MIP/WS header scalars.

ramakrishnap-nv:
Filed the review follow-ups as tracked issues so nothing is lost when this lands:

- #1629 — chunked upload/download for large VRP problems & solutions (+ preflight size check, review #7)
- #1630 — log streaming & incumbent parity with LP/MILP (async/log/incumbent)
- #1631 — server-side validation of VRP problem inputs (review #6)
- #1632 — client completeness: settings surface, e2e coverage, docs, perf nits (review #5, #8, #9)
- #1616 — generate `cuopt_routing.proto` + mapper via field_registry codegen (mapper boilerplate, round-trip tests, client unification, namespace)

ramakrishnap-nv:
Answering the two design questions:

**1. Why `bytes routing_solution` instead of a structured message (circular import)**

You're right that it's a workaround, and the cleaner fix is exactly what you suggested. The cycle is:

- `ChunkedResultHeader`, `LPSolution`, `MIPSolution` all live in the generated `cuopt_remote_data.proto` — the base data file that everything imports.
- `RoutingSolution` lives in the hand-authored `cuopt_routing.proto`, which `import`s `cuopt_remote.proto` → `cuopt_remote_data.proto`.
- So making `ChunkedResultHeader` embed `RoutingSolution` would force `cuopt_remote_data.proto` to import `cuopt_routing.proto`, which imports back → **cycle**. LP/MIP dodge it only because their solutions are defined *in the same file* as the header.

The `bytes` field sidesteps the cycle by not referencing the routing type at the proto level (the worker serializes `RoutingSolution` and stuffs the blob in). Agreed it's hacky and it's a POC shortcut.

The clean fix is to put `RoutingSolution` in the same file as the other solutions — which is exactly what the codegen work in #1616 does: once the routing messages are generated into `cuopt_remote_data.proto` alongside `LPSolution`/`MIPSolution`, `ChunkedResultHeader` can carry a structured `RoutingSolution routing_solution` field and the `bytes` workaround goes away. I've added that explicitly to #1616's scope (embed the structured type, drop the blob). Your "common file + per-problem files" instinct is the same idea; the registry codegen lands the routing messages in the right file for free.

**2. Transparent remote execution (env-var) parity**

Not done yet — not in another PR. #1597 adds only the *explicit* `cuopt.grpc.routing.RoutingClient`. LP/MIP additionally have **transparent** remote execution: `is_remote_execution_enabled()` (checks `CUOPT_REMOTE_HOST`/`CUOPT_REMOTE_PORT`) makes the normal `solve_lp()`/`solve_mip()` route to `solve_lp_remote()`/`solve_mip_remote()`, non-invasively from any interface. Routing's `Solve()` has no such hook. Agreed it's the natural parity — filed as #1633 (add `solve_vrp_remote()` and wire `is_remote_execution_enabled()` into routing `Solve()`). (Note: the `CUOPT_GRPC_SERVER` var the RoutingClient tests use is a different knob from the transparent `CUOPT_REMOTE_HOST`/`CUOPT_REMOTE_PORT` path.)

ramakrishnap-nv:
Done — `routing_solution` is now a **structured `RoutingSolution`** in both `ResultResponse` and `ChunkedResultHeader`, no more `bytes` blob (e4ccc751).

Approach (avoids hand-written code in the generated file):
- Moved `RoutingSolution`/`RoutingSolutionStatus` into a **leaf proto** `cuopt_routing_solution.proto` (no imports) that both result messages import — breaks the cycle without relocating anything into the generated file.
- The codegen learned a registry-driven `proto_imports`, so the generated `cuopt_remote_data.proto` emits `import "cuopt_routing_solution.proto";` and `RoutingSolution routing_solution = 4001;` — all generated, nothing hand-edited.
- Worker/service/client read & write the structured field directly (no `SerializeToString`/`ParseFromString`).

Wire-compatible as discussed (same field numbers; a `bytes` field and an embedded message encode identically), so no deployed client breaks.

Verified: protoc validates the full graph (no cycle), codegen verify passes, clean build, C++ grpc ctests 3/3, Python routing serialization + VRP client e2e 10 passed, structured round-trip returns the correct solution.

rg20 (approved):

cpp/tests/routing/grpc/CMakeLists.txt
```diff
@@ -0,0 +1,56 @@
+# cmake-format: off
+# SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+# cmake-format: on
+
+# Standalone VRP gRPC test driver (not a gtest). Requires a running cuopt_grpc_server.
+
+find_package(nlohmann_json QUIET)
```
jameslamb: Instead of a raw `find_package()` / `find_path()`, would you consider adding a `cmake/thirdparty/get_nlohmann_json.cmake`?

Like cuVS does: https://github.com/NVIDIA/cuvs/blob/da897db2346a22a0911541e39b917d9b26fcf8da/cpp/cmake/thirdparty/get_nlohmann_json.cmake

Doing that and using the RAPIDS CPM stuff like the cuVS example has some nice properties like:

* pin to the version that RAPIDS dependencies might use
* populate the CPM cache so other uses of nlohmann_json in the project don't have to re-do the search / download

cpp/CMakeLists.txt
```diff
@@ -395,6 +395,18 @@ if (NOT SKIP_GRPC_BUILD)
     add_compile_definitions(CUOPT_ENABLE_GRPC)
     message(STATUS "gRPC enabled (target gRPC::grpc++ is available)")
 
+    # The routing arm of the gRPC layer needs the routing engine. The wire
+    # protocol (including the routing protos) is always generated so LP-only
+    # builds still speak the same protocol; only the mappers and the VRP
+    # client/worker paths are dropped.
+    if (NOT SKIP_ROUTING_BUILD)
+        set(CUOPT_ENABLE_GRPC_ROUTING ON)
+        add_compile_definitions(CUOPT_ENABLE_GRPC_ROUTING)
```
jameslamb: `add_compile_definitions` will add `-DCUOPT_ENABLE_GRPC_ROUTING` to all targets. Is that intentional? If not, it'd be better to use `target_compile_definitions()` (https://cmake.org/cmake/help/latest/command/target_compile_definitions.html) on only the specific targets that need this, to avoid leaking flags into other targets.

jameslamb (approved):
Reviewed from a packaging / CI perspective. Left some small suggestions on the CMake code but none need to be blocking, do what you want there.

Code-generating bindings from the protobuf schemas (I think that's what's happening here, right?) makes sense to me.

I don't need to review again but please do `@` me if there are questions you want my opinion on.

ramakrishnap-nv:
> Reviewed from a packaging / CI perspective. Left some small suggestions on the CMake code but none need to be blocking, do what you want there.
> 
> Code-generating bindings from the protobuf schemas (I think that's what's happening here, right?) makes sense to me.
> 
> I don't need to review again but please do `@` me if there are questions you want my opinion on.

Yes, yaml is used as reference to build proto

cpp/CMakeLists.txt
ramakrishnap-nv: Not intentional — fixed in 4183bd53.

`grpc_worker.cpp` turned out to be the only source that tests the macro, and it belongs only to `cuopt_grpc_server`, so it's now `target_compile_definitions(cuopt_grpc_server PRIVATE CUOPT_ENABLE_GRPC_ROUTING)`. The CMake variable still gates which sources get compiled.

Verified in the generated `build.ninja`: the define now appears on `cuopt_grpc_server`'s 8 sources and nowhere in `cuopt_objs`. Previously it was on every target in the tree.

cpp/tests/routing/grpc/CMakeLists.txt
ramakrishnap-nv: Good call — done in 4183bd53. Added `cpp/cmake/thirdparty/get_nlohmann_json.cmake` using `rapids_cpm_find`, following the cuVS module you linked, and the test now just links `nlohmann_json::nlohmann_json`.

Besides the CPM-cache benefit, this removed a worse problem: the old code did an early `return()` when neither `find_package()` nor `find_path()` found anything, so `GRPC_VRP_TEST_DRIVER` would silently disappear from the build rather than fail.

Two notes on the pin:
- rapids-cmake has no `rapids_cpm_nlohmann_json` and no `nlohmann_json` entry in its versions file, so the version is pinned locally rather than by rapids-cmake.
- Pinned to 3.12.0 to match cuVS and the `nlohmann_json` conda package, so an existing install is reused rather than fetched. Configure now reports `CPM: Using local package nlohmann_json@3.12.0`.

FWIW cuVS looks to be the only RAPIDS repo carrying this module — cuml/cugraph/cudf/raft/kvikio/cuspatial all 404 on that path.

cpp/src/grpc/routing/grpc_routing_problem_mapper.cpp
```diff
@@ -0,0 +1,377 @@
...
+    p.cost_matrices.push_back(std::move(out));
+  }
+  for (auto const& tm : pb.transit_time_matrices()) {
+    cuopt::routing::cpu_cost_matrix_t out;
+    out.vehicle_type = static_cast<uint8_t>(tm.vehicle_type());
+    copy_repeated_to_vector(tm.values(), out.matrix);
+    p.transit_time_matrices.push_back(std::move(out));
+  }
+
+  copy_repeated_to_vector(pb.vehicle_start_locations(), p.vehicle_start_locations);
```
ramakrishnap-nv: Agreed, and tracked as #1730 rather than expanding this PR.

The two comments share one root cause: the field list is written out five times (struct, proto, host→proto, proto→host, `to_device`), with nothing enforcing that they stay in sync. Worth noting the failure mode is silent — a missed field doesn't crash, it just never crosses the wire, so the remote solve answers a different problem than the one submitted.

Current scale is 41 `std::vector` members, 6 `std::map`, and 5 nested struct types, so ~80% is one pattern repeated. The issue proposes a single source of truth (registry entry or X-macro) expanded into all five sites plus the round-trip test you suggested [here](https://github.com/NVIDIA/cuopt/pull/1597#discussion_r3648191407) — that test is the cheap first step and is worth keeping even after codegen lands.

cpp/src/routing/cpu_routing_problem.cu
ramakrishnap-nv: Agreed, and tracked as #1730 rather than expanding this PR.

The two comments share one root cause: the field list is written out five times (struct, proto, host→proto, proto→host, `to_device`), with nothing enforcing that they stay in sync. Worth noting the failure mode is silent — a missed field doesn't crash, it just never crosses the wire, so the remote solve answers a different problem than the one submitted.

Current scale is 41 `std::vector` members, 6 `std::map`, and 5 nested struct types, so ~80% is one pattern repeated. The issue proposes a single source of truth (registry entry or X-macro) expanded into all five sites plus the round-trip test you suggested [here](https://github.com/NVIDIA/cuopt/pull/1597#discussion_r3648191407) — that test is the cheap first step and is worth keeping even after codegen lands.

tmckayus (approved):

PR #1603, opened 2026-07-22
fix(grpc): terminate workers on server shutdown

tmckayus:
Fix an oversight on grpc server shutdown. The server would not exit if a worker was running a solve. The original Python REST server killed any running jobs if the server was shutdown and set the job status accordingly. This matches that behavior.

Without this behavior, the process has to be killed from the OS or every running job id has to be explicitly canceled.


ramakrishnap-nv (approved):

tmckayus:
thanks to @akifcorduk for suggesting watchdog time expansion

akifcorduk (approved):

PR #1606, opened 2026-07-23
fix: build against rmm 26.10 (use rmm's device_uvector declaration)

### LINKED ISSUE #1605: [BUG] Source build fails against rmm 26.10: hand-rolled rmm::device_uvector forward declaration is ambiguous
### Describe the bug

Building `libcuopt` from source against **rmm 26.10** fails to compile. rmm 26.10 places `device_uvector` inside an ABI inline namespace (`rmm::RMM_ABI_NAMESPACE`, see `rmm/detail/export.hpp`), while cuOpt hand-rolls a forward declaration of it in the *plain* `rmm` namespace in `cpp/src/pdlp/utilities/problem_checking.cuh`:

```cpp
namespace rmm {
template <typename T>
class device_uvector;
}  // namespace rmm
```

This injects a second, distinct `rmm::device_uvector`, so every use becomes ambiguous.

### Steps/Code to reproduce

Configure and build `libcuopt` from source with rmm 26.10 on the include path (CUDA 13.3 / GCC 14 / `sm_120a` here, but the ambiguity is compiler/arch independent).

### Error

```
cpp/src/pdlp/utilities/problem_checking.cuh(36): error: "rmm::device_uvector" is ambiguous
cpp/src/utilities/copy_helpers.hpp(82): error: "rmm::device_uvector" is ambiguous
...
```

The failures then cascade into follow-on `namespace "cuopt::raft" has no member "add_op"` and thrust `"this pragma must immediately precede a declaration"` errors in the same translation units; all of them trace back to this one forward declaration (fixing it alone clears them).

### Root cause

`problem_checking.cuh` is the only place in the tree that manually forward-declares an rmm type. cuOpt already opts out of thrust's and cub's ABI namespaces (`-DTHRUST_DISABLE_ABI_NAMESPACE` / `-DCUB_DISABLE_NAMESPACE_MAGIC`); rmm's ABI inline namespace is newer and has no such opt-out, so the plain forward declaration no longer matches rmm's real declaration.

### Fix

Replace the hand-rolled forward declaration with `#include <rmm/device_uvector.hpp>` (rmm's own, correctly ABI-namespaced declaration). PR to follow.

### Environment

- cuOpt: `main`
- rmm / raft: 26.10
- CCCL: 3.4.0, CUDA 13.3 toolkit, GCC 14
- GPU: RTX 5090 (`sm_120a`)

maxwbuckley:
## Description

Building `libcuopt` from source against **rmm 26.10** fails to compile because `cpp/src/pdlp/utilities/problem_checking.cuh` hand-rolls a forward declaration of `rmm::device_uvector` in the plain `rmm` namespace:

```cpp
namespace rmm {
template <typename T>
class device_uvector;
}  // namespace rmm
```

rmm 26.10 places `device_uvector` in an ABI inline namespace (`rmm::RMM_ABI_NAMESPACE`, see `rmm/detail/export.hpp`), so this hand-rolled declaration creates a *second*, distinct `rmm::device_uvector`. Every use then becomes ambiguous:

```
cpp/src/pdlp/utilities/problem_checking.cuh(36): error: "rmm::device_uvector" is ambiguous
cpp/src/utilities/copy_helpers.hpp(82): error: "rmm::device_uvector" is ambiguous
```

with cascading `namespace "cuopt::raft" has no member "add_op"` and thrust `"this pragma must immediately precede a declaration"` errors in the same translation units.

This is the only manual rmm forward declaration in the tree. cuOpt already opts out of thrust's and cub's ABI namespaces (`-DTHRUST_DISABLE_ABI_NAMESPACE` / `-DCUB_DISABLE_NAMESPACE_MAGIC`); rmm's ABI inline namespace is newer and has no such opt-out, so the plain forward declaration no longer matches rmm's real declaration.

## Change

Replace the hand-rolled forward declaration with `#include <rmm/device_uvector.hpp>`, so rmm's own (correctly ABI-namespaced) declaration is used.

## Testing

Built `libcuopt` (LP + MIP) from source against rmm/raft 26.10 + CCCL 3.4.0 (CUDA 13.3, GCC 14, `sm_120a`):
- **Before:** the `device_uvector is ambiguous` errors above.
- **After:** clean compile + link. `solve_LP` returns correct optima — `afiro` → `-464.7531` (known optimum), and a set of small LPs match HiGHS to ≤ 1e-6 relative error — and the MIP path runs.

Fixes #1605.

### DIFF
--- a/cpp/src/pdlp/utilities/problem_checking.cuh
+++ b/cpp/src/pdlp/utilities/problem_checking.cuh
@@ -10,10 +10,7 @@
 #include <cuopt/mathematical_optimization/optimization_problem.hpp>
 #include <cuopt/mathematical_optimization/pdlp/solver_settings.hpp>
 
-namespace rmm {
-template <typename T>
-class device_uvector;
-}  // namespace rmm
+#include <rmm/device_uvector.hpp>
 
 namespace cuopt::mathematical_optimization {
 


ramakrishnap-nv:
Thank you for the PR, but this has been resolved with this PR https://github.com/NVIDIA/cuopt/pull/1599

PR #1608, opened 2026-07-23
fix(security): resolve SonarQube critical/high/blocker security findings

ramakrishnap-nv:
## Summary

Resolves SonarQube critical, high, and blocker security findings across the server, C++ solver, and CI utilities.

**Python server (CWE-22, CWE-20):**
- **NVCF endpoint**: validate asset ID stays within `nvcf-asset-dir` via realpath+commonpath; require absolute path for `nvcf-large-output-dir`
- **`get_output_name`**: replace blocklist with realpath+commonpath containment check
- **Log endpoints**: validate `id` is a well-formed UUID before constructing file paths
- **`getsolverlogs`**: reject negative `frombyte` with HTTP 422

**C++ / CI (genuine bugs):**
- `fast_parser.cpp`: wrap `scoped_timer_t` destructor body in try-catch (S1048)
- `fast_fp64_parser.hpp`, `phase2.cpp`, `c_api_test.c`: replace `v==v`/`x!=x` NaN idiom with `std::isnan()`/`isnan()` (S1764)
- `pcgenerator.hpp`: replace unary minus on unsigned `(-rot)` with `(32u-rot)&31u` (S876)
- `aggregate_nightly.py`: canonicalize `output_dir` with `Path.resolve()` before use (S2083)

**False positives suppressed with `// NOSONAR`:**
- `fast_parser.cpp`: `find_line_start` return stays within bounds (S3519)
- `lz4_file_reader.cpp`: `window_count` initialized before dependent fields per C++ declaration order (S836)
- `file_to_string.cpp` (14 sites): `mps_parser_expects` throws before any null dereference (S2259)


tmckayus (approved):

cpp/src/io/experimental_mps_fast/fast_parser.cpp
```diff
@@ -222,13 +222,16 @@ class scoped_timer_t {
...
   {
 #ifdef MPS_FAST_TIMERS
-    auto end          = std::chrono::high_resolution_clock::now();
-    double elapsed_ms = std::chrono::duration<double, std::milli>(end - start_).count();
-    nvtx_.end();
-    if (accumulator_) { *accumulator_ += elapsed_ms; }
-    auto [rss_kb, hwm_kb] = current_process_rss_kb();
-    std::lock_guard<std::mutex> lock(get_timer_mutex());
-    get_timer_buffer().push_back({name_, elapsed_ms, rss_kb, hwm_kb});
+    try {
```
mlubin: What can throw here? cc @aliceb-nv

cpp/src/io/experimental_mps_fast/lz4_file_reader.cpp
```diff
@@ -1,4 +1,4 @@
-// SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES. All rights
+// SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 // reserved. SPDX-License-Identifier: Apache-2.0
```
mlubin: Incorrect header

cpp/src/io/experimental_mps_fast/fast_parser.cpp
aliceb-nv: Only things I can think of are the mutex (pthreads/futex error?) and the std::vector.push_back(), and that's a huge stretch
In any case I don't think silently catching these is a good idea

cpp/src/io/experimental_mps_fast/lz4_file_reader.cpp
```diff
@@ -452,9 +452,11 @@ struct lz4_pipeline_t {
     : input(input_),
       window_count(cuda::ceil_div(input.compressed_size_, window_bytes)),
       windows(window_count),
-      window_state_(std::make_unique<window_state_t[]>(window_count)),
+      window_state_(std::make_unique<window_state_t[]>(
+        window_count)),  // NOSONAR: window_count declared before window_state_ (line 869 vs 891)
```
mlubin: We shouldn't be hardcoding line numbers in comments, this will go stale very quickly.

mlubin (approved):

cpp/src/io/experimental_mps_fast/lz4_file_reader.cpp
ramakrishnap-nv: Addressed

PR #1615, opened 2026-07-23
Python api performance improvement

Iroy30:
## Description
- improved populate_solution slack computation. Drops ~68 ms to ~6 ms 
- Selective datamodel update depending on the data updated (as long as structure remains the same) instead of rebuild CSR and model each time. Drops ~40 ms to ~2 ms
- fixed a pre-existing attribute update bug

The 100ms shave off helps in consecutive solves of portfolio problems which solve in 300-500ms.

## Checklist

- Testing
   - [ ] Added tests
   - [x] NA
- Documentation
   - [ ] Added new documentation
   - [x] NA


Iroy30:
@coderabiitai review

python/cuopt/cuopt/linear_programming/problem.py
```diff
@@ -1466,6 +1468,34 @@ def __init__(self, model_name=""):
         self.lower_bound = None
         self.upper_bound = None
         self.var_type = None
+        self._index_to_var_cache = None
+        self._constraint_csr_scipy = None
+        self._constraint_index_to_csr_row = None
```
mlubin, line 1517: Consider refactoring so that the constraint index is always the same as the csr row index?

python/cuopt/cuopt/linear_programming/problem.py
```diff
@@ -1526,44 +1556,64 @@ def _from_data_model(self, dm):
...
+        self._constraint_index_to_csr_row = {
+            constr.index: row for row, constr in enumerate(linear_constrs)
+        }
+        self.rhs = np.fromiter(
+            (constr.RHS for constr in linear_constrs),
+            dtype=np.float64,
+            count=m,
+        )
+        self.row_sense = np.asarray(
+            [constr.Sense for constr in linear_constrs], dtype="S1"
```
mlubin, line 1619: Why `np.fromiter` above and `np.asarray` here?

python/cuopt/cuopt/linear_programming/problem.py
```diff
@@ -1526,44 +1556,64 @@ def _from_data_model(self, dm):
...
+        )
+        self.row_names = [
+            constr.ConstraintName or "R" + str(constr.index)
```
mlubin, line 1622: Do we need to generate row names if they're not provided?

python/cuopt/cuopt/linear_programming/problem.py
```diff
@@ -2186,13 +2411,23 @@ def populate_solution(self, solution):
         dual_sol = None
         if not IsMIP:
             dual_sol = solution.get_dual_solution()
+
+        # Vectorized slacks when a full primal and CSR/`rhs` are available;
+        # otherwise leave Slack as NaN (same outcome as unset Values).
+        slacks = None
+        if len(primal_sol) == len(self.vars):
+            A = self._constraint_csr_scipy_matrix()
```
mlubin, line 2489: This looks like the only place in the code where `_constraint_csr_scipy_matrix` is used. Are we keeping a duplicate copy of the matrix in memory just to compute slack values? We should be sensitive to changes in peak memory usage. An extra copy of the constraint matrix isn't cheap.

python/cuopt/cuopt/linear_programming/problem.py
Iroy30: CSR only stores the linear constraint matrix A. Constraint indices are assigned in add order across all constraints (linear and quadratic) for inspection/accessors. Quadratic constraints are excluded from CSR, so constr.index is not always the CSR row index; _constraint_index_to_csr_row maps between them.

python/cuopt/cuopt/linear_programming/problem.py
Iroy30: This was already existing code, the change just removed a if conditional-check hence the indentation showing up as new lines of code.
As to why we name it - it is to pass the user-given names to MPS ensuring that if a row is unnamed in python it is generated so that set_row_names can be called without empty strings

python/cuopt/cuopt/linear_programming/problem.py
Iroy30: was updated as a part of python overhead reduction strategy for consecutive solves. It saves ~10ms for an engine solve of 300ms for 275k nnz problem. Can handle numeric values. 
I agree that there needs to be consistency but these small wins can add up as problem grows.

python/cuopt/cuopt/linear_programming/problem.py
Iroy30: That's a good point, yes it is mainly for slack and it cuts down the computation quite a bit. We could delete it after compute instead of storing and maintaining.

python/cuopt/cuopt/linear_programming/problem.py
Iroy30: Updated

python/cuopt/cuopt/linear_programming/problem.py
```diff
@@ -129,6 +141,18 @@ def __init__(
         self.VariableName = vname
         self.MIPStart = float("nan")
 
+    def __setattr__(self, name, value):
+        object.__setattr__(self, name, value)
+        # Solution fields are written in hot loops; skip tracking lookups.
+        if name in self._OUTPUT_ATTRIBUTES:
+            return
+        problem = self.__dict__.get("_problem")
```
chris-maes, line 149: Could this be expensive if we are setting the objective or lower / upper bounds in a loop?

python/cuopt/cuopt/linear_programming/problem.py
```diff
@@ -111,6 +112,17 @@ class Variable:
         (unset). Only used for a MIP problem.
     """
 
+    # Input attrs: direct writes mark the matching Problem._stale key.
+    _STALE_ATTRIBUTE = {
```
chris-maes, line 116: Would `_INPUT_ATTRIBUTE` be a better name?

python/cuopt/cuopt/linear_programming/problem.py
```diff
@@ -111,6 +112,17 @@ class Variable:
...
+        "LB": "variable",
+        "UB": "variable",
+        "Obj": "objective",
+        "VariableType": "variable",
+        "VariableName": "variable",
```
chris-maes, line 121: Why does changing the variable name mark the variable as stale?

python/cuopt/cuopt/linear_programming/problem.py
```diff
@@ -111,6 +112,17 @@ class Variable:
...
+        "MIPStart": "variable",
```
chris-maes, line 122: Why does changing the MIP start mark the variable as stale?

python/cuopt/cuopt/linear_programming/problem.py
chris-maes: Maybe we should keep the quadratic constraints separate from the linear constraints when we call `addConstraint` to avoid this issue.

python/cuopt/cuopt/linear_programming/problem.py
chris-maes: Maybe you should have another value "variable_names". And only update the variable names in this case.

python/cuopt/cuopt/linear_programming/problem.py
chris-maes: Maybe you should have another value "start" and only update the start in this case?

python/cuopt/cuopt/linear_programming/problem.py
```diff
@@ -5,9 +5,10 @@
 import copy
 import os
 from enum import Enum
+from itertools import chain
 
 import numpy as np
-from scipy.sparse import coo_matrix
+from scipy.sparse import coo_matrix, csr_matrix
```
chris-maes, line 11: Not for this PR; but we should think if we want to make `scipy` a dependency for our Python interface. In general we should try to minimize the number of dependencies we have.

python/cuopt/cuopt/linear_programming/problem.py
```diff
@@ -1628,6 +1716,110 @@ def _to_data_model(self):
...
+                self.lower_bound[j] = var.getLowerBound()
+                self.upper_bound[j] = var.getUpperBound()
+                var_name = var.VariableName
+                if var_name == "":
+                    var_name = "C" + str(var.index)
+                self.var_names.append(var_name)
+                self.mip_start[j] = var.MIPStart
+            self.model.set_variable_lower_bounds(self.lower_bound)
+            self.model.set_variable_upper_bounds(self.upper_bound)
+            self.model.set_variable_types(self.var_type)
```
chris-maes, line 1791: Maybe add separate `self._stale["variable_types"]` and `self._stale['variable_names']` to avoid resetting the bounds if just the types or names change. Same for mip start

python/cuopt/cuopt/linear_programming/problem.py
```diff
@@ -1636,23 +1828,38 @@ def update(self):
...
+        # cannot be mapped back to cached A_values reliably.
+        self.constraint_csr_matrix = None
+        self._mark_stale("variable", "objective", "rhs", "A_values")
 
-    def reset_solved_values(self):
-        # Resets all post solve values
-        for var in self.vars:
-            var.Value = float("nan")
-            var.ReducedCost = float("nan")
+    def reset_solved_values(
```
chris-maes, line 1837: Should this be `_reset_solved_values`? Since we don't expect the user to call this function right?

python/cuopt/cuopt/linear_programming/problem.py
```diff
@@ -1715,12 +1932,19 @@ def addConstraint(self, constr, name=""):
         >>> problem.addConstraint(-x*x + y*y <= 0, name="soc")
         """
         if self.solved:
-            self.reset_solved_values()  # Reset all solved values
+            self.reset_solved_values()
+        self.constraint_csr_matrix = None
```
chris-maes, line 1936: Maybe group these lines into a single function call `_invalidate_problem_cache()`

python/cuopt/cuopt/linear_programming/problem.py
```diff
@@ -1715,12 +1932,19 @@ def addConstraint(self, constr, name=""):
...
+        self._mark_stale(
+            "structure", "variable", "objective", "rhs", "A_values"
+        )
         n = len(self.constrs)
         match constr:
             case Constraint():
                 constr.index = n
                 constr.ConstraintName = name
+                constr._problem = self
                 self.constrs.append(constr)
```
chris-maes, line 1948: We could have `self.linear_constraints` and `self.quadratic_constraints` and append to one or another depending on the the type of expression (LinearExpression or QuadraticExpression) in the constraint.

python/cuopt/cuopt/linear_programming/problem.py
```diff
@@ -1841,6 +2116,12 @@ def setObjective(self, expr, sense=MINIMIZE):
                 raise ValueError(
                     "Objective must be a Variable, Expression or a constant"
                 )
+        if not is_quadratic:
+            self.objective_qmatrix = None
+            if had_qmatrix:
+                # DataModel has no API to clear Q from its persistent C++
```
chris-maes, line 2122: Should we add an API to DataModel to allow for this?

python/cuopt/cuopt/linear_programming/problem.py
chris-maes: I think if we are only forming `_constraint_csr_scipy_matrix` to compute the slacks, it would be better for us to compute the slack values in the Cython code and include them in the solution. That would allow us to avoid forming `_constraint_csr_scipy_matrix` and minimize the dependency on scipy.

python/cuopt/cuopt/linear_programming/problem.py
Iroy30: Renaming / MIPStart / type currently all share _stale["variable"], so refresh resyncs bounds, types, names, and MIPStart all together if either changes and causes a _stale["variable"]. 

That path does not rebuild structure/CSR; it only rewrites the existing DataModel variable arrays setters- dm.set_* . The cost is one O(n) sync at solve() (n being num variables)which is a loop we need to anyway make if either of these attribute changes. For the warm LP path this PR mostly targets (bounds / obj / RHS) - name and MIPStart edits aren’t hot - so the extra setters are cheap vs. the solve and not worth finer flags here. Happy to split bounds / names / types / start later if a name- or start-heavy workload shows up in profiles.

python/cuopt/cuopt/linear_programming/problem.py
Iroy30: Renaming / MIPStart / type currently all share _stale["variable"], so refresh resyncs bounds, types, names, and MIPStart all together if either changes and causes a _stale["variable"]. 

That path does not rebuild structure/CSR; it only rewrites the existing DataModel variable arrays setters- dm.set_* . The cost is one O(n) sync at solve() (n being num variables)which is a loop we need to anyway make if either of these attribute changes. For the warm LP path this PR mostly targets (bounds / obj / RHS) - name and MIPStart edits aren’t hot - so the extra setters are cheap vs. the solve and not worth finer flags here. Happy to split bounds / names / types / start later if a name- or start-heavy workload shows up in profiles.

python/cuopt/cuopt/linear_programming/problem.py
Iroy30: Noted. Scipy is intertwined in more places, so maybe we can discuss and address this in a follow up

python/cuopt/cuopt/linear_programming/problem.py
Iroy30: Renaming / MIPStart / type currently all share _stale["variable"], so refresh resyncs bounds, types, names, and MIPStart all together if either changes and causes a _stale["variable"].

That path does not rebuild structure/CSR; it only rewrites the existing DataModel variable arrays setters- dm.set_* . The cost is one O(n) sync at solve() (n being num variables)which is a loop we need to anyway make if either of these attribute changes. For the warm LP path this PR mostly targets (bounds / obj / RHS) - name and MIPStart edits aren’t hot - so the extra setters are cheap vs. the solve and not worth finer flags here. Happy to split bounds / names / types / start later if a name- or start-heavy workload shows up in profiles.

python/cuopt/cuopt/linear_programming/problem.py

python/cuopt/cuopt/linear_programming/problem.py
Iroy30: Updated cython to make sure we clear it using a new object. It introduced a bug since the current C++ data model once created doesn't have a clear/unset API, so a QP once solved does not clear the datamodel qmatrix even if we update objective to purely linear.

python/cuopt/cuopt/linear_programming/problem.py
Iroy30: fixed

python/cuopt/cuopt/linear_programming/problem.py
Iroy30: It is ~6ms more expensive for a 10k loop just because it now triggers a function call now. But the setter is important to 1) Mark stale if attributes are directly changed thus avoiding silent discrepancies 2) But make sure that output attributes (that we set after solve for user output) don't mark model as stale

python/cuopt/cuopt/linear_programming/problem.py
Iroy30: Updated. Cython now computes slack and populates the solution object. No regression in comparison to scipy usage.

python/cuopt/cuopt/linear_programming/solution/solution.py
```diff
@@ -74,6 +74,10 @@ class Solution:
         Note: Applicable to only LP
         The reduced cost.
         It contains the dual multipliers for the linear constraints.
+    slack : numpy.array
+        Classical LP slack/surplus per linear constraint in CSR row order:
```
mlubin: Drop "classical LP". It's simply the slack/surplus per linear constraint in row order.

mlubin (approved):
Sorry I had one comment queued and forgot to submit.

Iroy30:
@tmckayus Thank you for catching edge cases in the caching system. Updated _from_data_model to be more complete with handling quadratic constraints and objectives. Those should be fixed now

python/cuopt/cuopt/linear_programming/solution/solution.py
Iroy30: Done

tmckayus (approved):

PR #1628, opened 2026-07-28
test(barrier): cap qplib_8515_ruiz_forced_off at the iteration limit

mlubin:
The test was flaky because sometimes it never converged. Enforce an iteration limit instead.


Bubullzz (approved):
clean fix. thanks Miles

rg20 (approved):

PR #1636, opened 2026-07-28
feat(c-api): add cuOptSetLogCallback

### LINKED ISSUE #184: [FEA] Add support for logging callback in the C interface
We received a request from a user to support a log callback in the C interface. The user would provide a function that would receive a string. cuOpt would call this function each time it wrote to the log. This would allow cuOpt's logs to be captured, without having to capture stdout.

### LINKED ISSUE #1752: cuOptSetLogCallback registration is process-global; concurrent solves can drop or leak a callback
Follow-up to review feedback on #1636 (raised by CodeRabbit, [endorsed by @mlubin](https://github.com/NVIDIA/cuopt/pull/1636#discussion_r3617260455): "this seems like a real issue").

## Problem

`cuOptSetLogCallback` stores a callback per settings object, but it reaches the
logger through a single process-global slot in `cpp/src/utilities/logger.cpp`:

```cpp
static log_callback_with_data_t g_pending_callback = nullptr;
static void* g_pending_callback_data               = nullptr;
```

Flow: `cuOptSolve` drops the settings' callback into these globals via
`set_pending_log_callback()`, and `init_logger_t` picks it up, copies it into the
logger guard, and installs a `callback_sink_mt`.

Two failure modes follow, both only with concurrent or overlapping solves.

**1. Clobbering.** The mutex makes each write atomic, but not the
drop-then-consume pair. Thread A calls `set_pending_log_callback(cbA)`, thread B
calls `set_pending_log_callback(cbB)` before A's `init_logger_t` runs, and A
silently gets B's callback.

**2. Silent drop, with a stale pointer left behind.** `init_logger_t` returns
early when a guard is already alive:

```cpp
auto existing_guard = g_active_guard.lock();
if (existing_guard) { guard_ = existing_guard; return; }
```

So if solve A is running, solve B's `init_logger_t` never reads the mailbox: B's
callback is never installed, and B's registration stays in the globals. A later
solve can then pick it up and invoke B's function pointer with B's `user_data`,
which may already have been freed by then.

The second is the more serious one — it is not "some log lines go missing" but a
stale callback plus stale user data surviving the solve that registered them.

## Root cause

`default_logger()` is a process-wide singleton with one global sink list, so
there is no per-solve context for `user_log_bridge` to route on. Per-solve
callback delivery needs solve identity threaded through the logging path.

## Options

1. **Fail loudly.** If a solve begins while a guard is active with a different
   callback, return an error from `cuOptSolve` rather than silently dropping it.
   Small and honest; rejects a legitimate use case.
2. **Thread-local registration.** The bridge looks up a `thread_local` callback,
   so each solving thread gets its own. Covers lines emitted on the solve thread;
   lines from internal worker threads would reach no callback.
3. **Per-solve logger context.** Thread solve identity through the logging path
   so the bridge can route correctly. Correct, and the largest change.

Whichever is chosen, clearing the pending slot on the early-return path would
remove the stale-pointer hazard on its own, and is worth doing regardless.

## Scope

Single-solve use — the common case — is correct today, which is why #1636 ships
with this documented as a known limitation rather than blocked on it.

ramakrishnap-nv:
Closes #184.

Adds a C API function plus an environment variable for solver log control:

- `cuOptSetLogCallback(settings, callback, user_data)` — registers a callback invoked once per standard solver log line (in addition to any console/file sink). Pass `NULL` to clear.
- `CUOPT_LOG_LEVEL` environment variable — overrides log verbosity at runtime (`TRACE`, `DEBUG`, `INFO`, `WARN`, `ERROR`, `CRITICAL`, `OFF`, case-insensitive). Bounded by the compile-time `LIBCUOPT_LOGGING_LEVEL` (default `INFO`): lowering verbosity always works, raising it above the build level has no effect since those statements are compiled out.

**No log level is exposed through the API** (per @chris-maes's review). There is no `cuOptSetLogLevel`, no level enum, and the callback carries no severity:

```c
typedef void (*cuOptLogCallback)(const char* message, void* user_data);
```

Callers receive the lines the solver would print to the console and nothing else — debug and trace are filtered in the log bridge, so internal diagnostics cannot reach user code even in a build with a lower `LIBCUOPT_LOGGING_LEVEL`. Verbosity is controlled by the env var rather than a setter.

**Design**: the callback is stored in `solver_settings_handle_t` and installed as a `rapids_logger::callback_sink_mt` via a RAII scope guard in `cuOptSolve`, so `init_logger_t` picks it up without changes to the C++ internal solver-settings structs. Registration is thread-local, so the bridge reads state owned by the solving thread.

**Scope note**: cuOpt's standard solver output is emitted at `info` (`CUOPT_LOG_INFO` covers "Solve status", "PDLP finished" and the progress table, ~153 call sites), so the cut is at `info` — filtering it out would leave the callback with nothing to deliver. Separating user-facing output from internal `info` diagnostics would mean auditing those call sites and is not attempted here.

### Testing
- `C_API_TEST` covers callback delivery, clearing (`NULL`), user-data forwarding, and multiple settings objects.
- Built and tested locally in a repo-local conda env on an RTX 8000: `C_API_TEST` passes, `ci/check_symbols.sh` clean, `cuOptSetLogCallback` verified exported.
- `C_API_TEST` asserts the callback fires at least once during a solve and stops after clearing, so thread-local routing still delivers real solver output.
- `CpuOnlyWithServerTest.log_callback_remote` forks a real `cuopt_grpc_server` and asserts a line from the *server's* solver log reaches the callback — a bare call-count check would pass even with no forwarding, since the client emits its own lines.
- Not covered: concurrent solves with different callbacks, which need two threads solving at once.

### Docs
No user-guide changes. The header documents the callback contract, including that message formatting is not a stable API.

### Concurrency
Closes #1752. Registration is thread-local, held for the duration of `cuOptSolve`, and the bridge sink is always installed (it no-ops unless the logging thread has a registration). Delivery depends only on which thread is logging, so concurrent solves cannot capture each other's callback and a solve that reuses an existing logger guard still gets its own callback.

Log lines emitted on internal worker threads are not delivered, since those threads carry no registration.

### Remote solves
With `CUOPT_REMOTE_HOST`/`CUOPT_REMOTE_PORT` set, the solve runs on a `cuopt_grpc_server` and the server's log is streamed back to the callback, so registration behaves the same locally and remotely. Previously the streamed log went only to stdout/log file and the callback saw just the local lines.

Covered by `CpuOnlyWithServerTest.log_callback_remote`. Reverting the forwarding makes it fail with "delivered 1 lines but none from the server's solver log".

The callback runs on the calling thread for a local solve and on the log-streaming thread for a remote one; this is documented on `cuOptLogCallback`.


cpp/src/utilities/logger.cpp
```diff
@@ -146,6 +146,43 @@ struct logger_config_guard {
...
+// Pending user log callback set by the C API before a solve.
+// Accessed under g_guard_mutex.
+static log_callback_with_data_t g_pending_callback      = nullptr;
+static void* g_pending_callback_data                    = nullptr;
+static int g_pending_log_level                          = -1;  // -1 = use compiled default
+
+static void user_log_bridge(int lvl, const char* msg)
+{
+  if (g_pending_callback) { g_pending_callback(lvl, msg, g_pending_callback_data); }
+}
```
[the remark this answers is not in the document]
mlubin: @ramakrishnap-nv this seems like a real issue.

cpp/include/cuopt/mathematical_optimization/constants.h
```diff
@@ -146,6 +146,15 @@
 /* @brief QCQP (barrier) scaling hyper-parameters */
 #define CUOPT_QCQP_HYPER_RUIZ_EQUILIBRATION "qcqp_hyper_ruiz_equilibration"
 
+/* @brief Log level constants for cuOptSetLogLevel */
```
chris-maes: Are we sure we want to expose these log levels to the user?

chris-maes (dismissed):
I think we should discuss before merging. I'm supportive of adding an API like
`cuOptSetLogCallback`. But I don't think we should expose the different log levels to the user. These log levels are really more for developers and debugging. I'm fine with having an environmental variable that can change the log level.

cpp/include/cuopt/mathematical_optimization/constants.h
ramakrishnap-nv: We can always set this Error or warn by default and it is upto user what they want to check.

ramakrishnap-nv:
> I think we should discuss before merging. I'm supportive of adding an API like
> `cuOptSetLogCallback`. But I don't think we should expose the different log levels to the user. These log levels are really more for developers and debugging. I'm fine with having an environmental variable that can change the log level.

Updated, so removed set level and provided an option. in env to choose from existing levels.

ramakrishnap-nv:
May I get another round of review ?

ramakrishnap-nv:
@chris-maes May I get another round of review on this PR

cpp/include/cuopt/mathematical_optimization/cuopt_c.h
```diff
@@ -828,6 +828,43 @@ cuopt_int_t cuOptGetFloatParameter(cuOptSolverSettings settings,
...
+ * @param user_data Opaque pointer passed to cuOptSetLogCallback.
+ *
+ * @note The callback is invoked from the solver thread. Do not call back into
+ *  cuOpt from inside the callback.
+ * @warning Log message formatting is not part of the stable API and may change
+ *  between releases. The callback is intended for display purposes (GUI integration,
+ *  log forwarding, stdout capture) — do not parse message content for programmatic
+ *  control flow.
+ */
+typedef void (*cuOptLogCallback)(const char* message, void* user_data);
```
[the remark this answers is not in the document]
ramakrishnap-nv: Not an ABI break — `cuOptLogCallback` has never shipped. It is introduced by this PR, so there is no released signature to migrate from:

```
$ git show origin/main:cpp/include/cuopt/mathematical_optimization/cuopt_c.h | grep -c cuOptLogCallback
0
$ git show v26.10.00a:cpp/include/cuopt/mathematical_optimization/cuopt_c.h | grep -c cuOptLogCallback
0
```

The three-parameter form existed only in this branch's earlier commit (3756aaf7), and the two-parameter form (dc14e3bd) replaced it before merge. The diff you compared is intra-PR churn, not a change to a public API.

No release-note migration guidance is needed. The signature narrowing was made in response to @chris-maes's review — dropping the level parameter so no severity is exposed through the C API.

cpp/include/cuopt/mathematical_optimization/constants.h
ramakrishnap-nv: Addressed — no log level is exposed through the API at all now, per your Slack comments as well.

- The `constants.h` level enum and `cuOptSetLogLevel` are gone; this PR no longer touches `constants.h`.
- The callback itself no longer carries severity:

```c
-typedef void (*cuOptLogCallback)(int level, const char* message, void* user_data);
+typedef void (*cuOptLogCallback)(const char* message, void* user_data);
```

There is now no enum, no numeric mapping, and nothing for callers to branch on. The previous version tried to handle this with a doc comment ("not a stable programmatic API"), which does not stop code being written against a value it can see.

- The bridge also filters below `info`, so debug and trace never reach user code. That previously held only by accident of the build — `CUOPT_LOG_ACTIVE_LEVEL` defaults to INFO so those statements are compiled out, but a lower-level build, or `CUOPT_LOG_LEVEL=DEBUG` against one, would have routed internal diagnostics into a user callback.

**One judgement call worth your view.** You said the callback should fire only on a standard log line and that no code should see our debug *and info* messages. Those cannot both hold literally here: cuOpt's standard solver output *is* emitted at `info` — `CUOPT_LOG_INFO` covers "Solve status", "PDLP finished" and the progress table, roughly 153 call sites (level census: TRACE 109, DEBUG 295, INFO 153, WARN 13, ERROR 46). Filtering `info` out would leave the callback with essentially nothing to deliver.

So I read "debug and info" as internal diagnostics and cut at `info`: callers get what the console would show, and nothing below it. If you meant something stricter — a dedicated user-log channel separate from `info` diagnostics — that means auditing those ~153 sites and splitting them, which I would rather do as a follow-up than fold in here. Happy to go that way if you prefer.

ramakrishnap-nv:
@chris-maes May I get review on this PR

cpp/include/cuopt/mathematical_optimization/cuopt_c.h
```diff
@@ -828,6 +828,43 @@ cuopt_int_t cuOptGetFloatParameter(cuOptSolverSettings settings,
...
+ * file or console sink already enabled via ``log_to_console`` / ``log_file``
+ * parameters. Pass NULL to remove a previously registered callback.
+ *
+ * @param[in] settings  The solver settings object.
+ * @param[in] callback  Callback function, or NULL to clear.
+ * @param[in] user_data Opaque pointer forwarded to the callback unchanged.
+ *
+ * @return A status code indicating success or failure.
+ */
+cuopt_int_t cuOptSetLogCallback(cuOptSolverSettings settings,
```
rg20, line 865: How this handled with GRPC?

ramakrishnap-nv: Good catch — it wasn't, and it is now (30632cad).

With `CUOPT_REMOTE_HOST`/`CUOPT_REMOTE_PORT` set, `cuOptSolve` dispatches through `solve_lp_remote` / `solve_mip_remote`, and the callback saw almost nothing. Two reasons:

1. The streamed server log was written straight to `std::cout` and the log file, bypassing `default_logger()` and therefore the callback sink.
2. `config.stream_logs` was only enabled when `log_to_console` or `log_file` was set, so a registered callback alone did not even request streaming.

Both remote paths now forward each streamed line to the user callback, and request streaming when a callback is present. The registration is captured at call time rather than read inside the lambda, because `grpc_client_t` invokes `log_callback` from its own `log_thread_`, which carries no thread-local registration (registration is per-thread as of 6562dda6, so concurrent solves cannot capture each other's callback).

Verified against a live `cuopt_grpc_server` with a small C probe — the same LP delivers the same number of callback lines either way:

```
local :  solve rc=0 callback_calls=16
remote:  solve rc=0 callback_calls=16
```

and the remote lines are the server's own solver log (`Using PSLP presolver`, `PSLP Presolved problem: ...`, `Status: Optimal`), cross-checked against the server-side log.

One behavioural note now documented in the header: the callback runs on the calling thread for a local solve, and on the log-streaming thread for a remote one.

rg20: Can you please add a test for this?

ramakrishnap-nv: Follow-up: this is now covered by a test as well (31e6c106).

`CpuOnlyWithServerTest.log_callback_remote` reuses the existing fixture that forks a real `cuopt_grpc_server` and points `CUOPT_REMOTE_HOST`/`PORT` at it.

Worth noting why the assertion is what it is: the client emits its own `Using remote GPU backend` line locally, so a bare "callback fired at least once" check passes even when no server output is forwarded at all. The test therefore asserts that a line from the *server's* solver log arrives. Reverting the forwarding in `solve_remote.cpp` makes it fail with:

```
Remote solve delivered 1 lines but none from the server's solver log
[  FAILED  ] CpuOnlyWithServerTest.log_callback_remote
```

ramakrishnap-nv:
Merged current `main` (`e7eb4d41`) in. Two conflicts, one of which needed a real decision rather than a mechanical merge.

**`logger.hpp` — union.** Both sides appended right after `buffered_entry`: this PR added the callback registration API, `main` added the `log_buffer` infrastructure. They share no declarations, so it was textual adjacency rather than a real collision. Both kept.

**`logger.cpp` — delete/modify.** #1778 (`refactor(logger): give each component library its own logger`) deleted this file and made the logger header-only with hidden visibility. This PR was adding `thread_local` callback state, a `static` mutex and the `scoped_log_callback_t` ctor/dtor to it.

Restoring the file would undo that PR's central point, so the callback state moved inline into `logger.hpp` instead — which is already the pattern there (`main` holds `g_guard_mutex` and `g_active_guard` as `inline`). The `thread_local` now lives in a `detail` namespace, with a comment recording the consequence of hidden visibility: **each component library gets its own copy**, so a callback registered through one is not seen by another. That looks right given registration is scoped to the solve that installed it and every solver builds its own `init_logger_t` on entry — but it is a real behavioural property of the merge, so flagging it explicitly.

**One thing the merge would have broken silently.** `main`'s new `apply_logger_config()` calls `sinks().clear()` and rebuilds the sink list, and it does not know about `user_log_bridge`. Left as-is, a callback registered via `cuOptSetLogCallback` would have stopped receiving lines the moment any solver configured logging — with no error. The bridge is now re-attached inside `apply_logger_config()`.

**Verified:** full build with tests, 0 errors. All 5 log-callback tests pass, including `CpuOnlyWithServerTest.log_callback_remote`. The 24 other `C_API_TEST` failures in my local run are the `square41.mps` fixtures — that dataset is fetched by `ci/test_cpp.sh` and I did not download it; they are unrelated to this change.

rg20 (approved):
Thanks for the changes!

ramakrishnap-nv:
Merged `main` (`8fc01156`) in. One conflict, in `logger.hpp` again.

`main` picked up #1825 (*Route Java console logging through System.out instead of native stdout*), which touches many of the same files as this PR — `cuopt_c.h`, `logger.hpp`, `cuopt_c.cpp`, and the C API tests. Two different features landing in the same region:

| | |
|---|---|
| this PR | `log_callback_with_data_t` — **per-thread** user callback behind `cuOptSetLogCallback` |
| #1825 | `log_console_callback_t` — **process-wide** console-sink override for the Java bindings |

No shared names, so the declarations block resolved as a union.

The part worth checking was `apply_logger_config()`, since both features push sinks there. That auto-merged correctly: #1825's branch chooses between its console callback and `std::cout`, and this PR's `user_log_bridge` re-attach follows it. They do different jobs and both survive — a registered `cuOptSetLogCallback` still receives lines whether or not a console override is installed.

**Verified:**
- build with tests — 0 errors
- this PR's callback tests: `log_callback`, `log_callback_cleared`, `log_callback_not_leaked_across_solves`, `CpuOnlyWithServerTest.log_callback_remote` — all pass
- #1825's `CONSOLE_LOG_CALLBACK_TEST` — passes

The remaining `C_API_TEST` failures locally are the `square41.mps` / `50v-10.mps` / `cvs16r128-89.mps` fixtures; those datasets are fetched by `ci/test_cpp.sh` and I have not downloaded them. Unrelated to this change.

PR #1640, opened 2026-07-30
Reduce concurrent LP host memory footprint

hlinsen:
- Release parsed MPS/QPS host buffers after GPU population completes.
  - Remove redundant full-size problem, name, and solution copies from concurrent LP solves.
  - Share the CPU problem between Barrier and dual simplex while preserving Barrier’s independent RAFT handle and CUDA stream.
  - Convert only the winning solver result and release losing solver storage before copying the solution back to the GPU.
  - Report both total and available host RAM in version information.

On B200, `mcf_2500_100_500` peak memory dropped from `168.12 GiB` on main to `143.73 GiB` with this change, a `24.39 GiB (14.5%)` reduction


cpp/src/utilities/version_info.cpp
```diff
@@ -180,11 +193,14 @@ void print_version_info(int num_devices)
...
                  CUOPT_CPU_ARCHITECTURE,
                  CUOPT_CUDA_ARCHITECTURES);
-  CUOPT_LOG_INFO("CPU: %s, threads (physical/logical): %d/%d, RAM: %.2f GiB",
-                 get_cpu_model().c_str(),
-                 get_physical_cores(),
-                 std::thread::hardware_concurrency(),
-                 get_available_memory_gb());
+  const auto memory = get_host_memory_info();
+  CUOPT_LOG_INFO(
+    "CPU: %s, threads (physical/logical): %d/%d, RAM: %.2f GiB total, %.2f GiB available",
```
chris-maes: I think we should just say `RAM (available/total): %.2f / %.2f GiB`

cpp/src/dual_simplex/solve.cpp
```diff
@@ -351,10 +351,11 @@ template <typename i_t, typename f_t>
 lp_status_t solve_linear_program_with_barrier(const user_problem_t<i_t, f_t>& user_problem,
                                               const simplex_solver_settings_t<i_t, f_t>& settings,
                                               f_t start_time,
-                                              lp_solution_t<i_t, f_t>& solution)
+                                              lp_solution_t<i_t, f_t>& solution,
+                                              const raft::handle_t* handle_ptr)
```
chris-maes, line 355: Why does passing in the handle_ptr (rather than getting it from the user_problem) help reduce memory?

cpp/src/dual_simplex/solve.cpp
```diff
@@ -688,7 +699,6 @@ lp_status_t solve_linear_program(const user_problem_t<i_t, f_t>& user_problem,
   std::vector<i_t> new_slacks;
   dualize_info_t<i_t, f_t> dualize_info;
   convert_user_problem(user_problem, settings, original_lp, new_slacks, dualize_info);
-  solution.resize(user_problem.num_rows, user_problem.num_cols);
```
chris-maes, line 700: Why is this resize no longer needed/wanted?

cpp/src/pdlp/solve.cu
```diff
@@ -487,9 +488,10 @@ optimization_problem_solution_t<i_t, f_t> convert_dual_simplex_sol(
 
 template <typename i_t, typename f_t>
 std::tuple<simplex::lp_solution_t<i_t, f_t>, simplex::lp_status_t, f_t, f_t, f_t> run_barrier(
-  simplex::user_problem_t<i_t, f_t>& user_problem,
+  const simplex::user_problem_t<i_t, f_t>& user_problem,
```
chris-maes, line 491: Why does making this const help with memory?

cpp/src/pdlp/solve.cu
```diff
@@ -562,26 +572,28 @@ optimization_problem_solution_t<i_t, f_t> run_barrier(
...
 }
 
 template <typename i_t, typename f_t>
 std::tuple<simplex::lp_solution_t<i_t, f_t>, simplex::lp_status_t, f_t, f_t, f_t> run_dual_simplex(
-  simplex::user_problem_t<i_t, f_t>& user_problem,
+  const simplex::user_problem_t<i_t, f_t>& user_problem,
   pdlp_solver_settings_t<i_t, f_t> const& settings,
-  const timer_t& timer)
+  const timer_t& timer,
+  bool preserve_solution_shape = true)
```
chris-maes: Why would we ever not want to preserve the solution shape?

cpp/src/pdlp/solve.cu
```diff
@@ -595,9 +607,14 @@ std::tuple<simplex::lp_solution_t<i_t, f_t>, simplex::lp_status_t, f_t, f_t, f_t
     dual_simplex_settings.log.log = false;
   }
 
-  simplex::lp_solution_t<i_t, f_t> solution(user_problem.num_rows, user_problem.num_cols);
+  simplex::lp_solution_t<i_t, f_t> solution(0, 0);
```
chris-maes: How can we pass a solution with incorrect size into `solve_linear_program`?

cpp/include/cuopt/mathematical_optimization/optimization_problem_utils.hpp
```diff
@@ -155,6 +157,10 @@ void adopt_from_mps_data_model(optimization_problem_interface_t<i_t, f_t>* probl
     return;
   }
   populate_from_mps_data_model(problem, data_model);
+  if (auto* gpu_problem = dynamic_cast<optimization_problem_t<i_t, f_t>*>(problem)) {
```
chris-maes, line 160: I assume the `=` is intentional and that you didn't want `==` but you might want move that out of the if to be clear.

```
auto* gpu_problem = dynamic_cast<optimization_problem_t<i_t, f_t>*>(problem);
if (gpu_problem) {
 ...
}
```

cpp/src/pdlp/solve.cu
hlinsen: This is only for readability, there's one instance of that problem shared and we want to make sure it's not being modified in run_barrier.

cpp/src/dual_simplex/solve.cpp
hlinsen: This is refactoring that allows us to get rid of this copy:
```
  auto barrier_problem       = dual_simplex_problem;
  barrier_problem.handle_ptr = barrier_handle_ptr.get();
```

cpp/src/dual_simplex/solve.cpp
hlinsen: Uncrush primal/dual does the resising on valid solutions: 
- [https://github.com/NVIDIA/cuopt/blob/main/cpp/src/dual_simplex/solve.cpp#L701](https://github.com/NVIDIA/cuopt/blob/main/cpp/src/dual_simplex/solve.cpp#L701)
- [https://github.com/NVIDIA/cuopt/blob/main/cpp/src/dual_simplex/presolve.cpp#L1695-L1711](https://github.com/NVIDIA/cuopt/blob/main/cpp/src/dual_simplex/presolve.cpp#L1695-L1711).

cpp/src/pdlp/solve.cu
hlinsen: This is to allocate only when dual simplex wins in concurrent mode. We save ~1.8GB.
I agree that the code is complex for the return. I'll revert.

cpp/src/dual_simplex/solve.cpp
hlinsen: Reverted the solution changes

cpp/include/cuopt/mathematical_optimization/optimization_problem_utils.hpp
rg20: You can do this as well 

```
if (auto* gpu_problem = dynamic_cast<optimization_problem_t<i_t, f_t>*>(problem); gpu_problem) {
 ...
}
```

cpp/src/pdlp/solve.cu
```diff
@@ -521,11 +523,19 @@ std::tuple<simplex::lp_solution_t<i_t, f_t>, simplex::lp_status_t, f_t, f_t, f_t
...
   auto status = simplex::solve_linear_program_with_barrier<i_t, f_t>(
-    user_problem, barrier_settings, timer.get_tic_start(), solution);
+    user_problem,
+    barrier_settings,
+    timer.get_tic_start(),
+    solution,
+    handle_ptr == nullptr ? user_problem.handle_ptr : handle_ptr);
```
rg20: Why will this be a nullptr?

cpp/src/pdlp/translate.hpp
```diff
@@ -112,7 +112,7 @@ static simplex::user_problem_t<i_t, f_t> cuopt_problem_to_user_problem(
 
 template <typename i_t, typename f_t>
 static simplex::user_problem_t<i_t, f_t> cuopt_problem_to_user_problem(
-  raft::handle_t const* handle_ptr, mip::problem_t<i_t, f_t>& model)
+  raft::handle_t const* handle_ptr, mip::problem_t<i_t, f_t>& model, bool copy_names = true)
```
rg20, line 115: I think we should just remove names from `problem_t` class, its internal to the solver and not exposed to public APIs.

Kh4ster (approved):

cpp/src/dual_simplex/solve.cpp
chris-maes: where is that in the code? I see `barrier_lp` but that is after column scaling?

cpp/src/pdlp/solve.cu
```diff
@@ -521,11 +523,19 @@ std::tuple<simplex::lp_solution_t<i_t, f_t>, simplex::lp_status_t, f_t, f_t, f_t
     barrier_settings.log.log = false;
   }
 
-  simplex::lp_solution_t<i_t, f_t> solution(user_problem.num_rows, user_problem.num_cols);
+  simplex::lp_solution_t<i_t, f_t> solution(0, 0);
```
chris-maes: Do we still need this? This looks scary to me.

cpp/src/dual_simplex/solve.cpp
hlinsen: Here: https://github.com/NVIDIA/cuopt/blob/main/cpp/src/pdlp/solve.cu#L1594-L1599

rg20 (approved):

PR #1645, opened 2026-07-31
Pin PaPILO to flush empty-row fix from akifcorduk/papilo.

akifcorduk:
Change the papilo pin to my fork as Nicolas is on vacation. This fixes the bug by moving the empty-row redundancy check after removeFixedCols.

### DIFF
--- a/cpp/CMakeLists.txt
+++ b/cpp/CMakeLists.txt
@@ -270,14 +270,14 @@ endif ()
 
 FetchContent_Declare(
         papilo
-        GIT_REPOSITORY "https://github.com/nguidotti/papilo.git"
+        GIT_REPOSITORY "https://github.com/akifcorduk/papilo.git"
         # We would want to get the main branch. However, the main branch
         # does not have some of the presolvers and settings that we need
         # Mainly, probing and clique merging.
         # This is the reason we are using the development branch
         # from Oct 12, 2025. Once these changes are merged into the main branch,
         #we can switch to the main branch.
-        GIT_TAG "0cfea20c5655249174dbd04f0fb7bd6b1f6e9a0c"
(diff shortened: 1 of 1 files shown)


cpp/CMakeLists.txt
```diff
@@ -270,14 +270,14 @@ endif ()
 
 FetchContent_Declare(
         papilo
-        GIT_REPOSITORY "https://github.com/nguidotti/papilo.git"
+        GIT_REPOSITORY "https://github.com/akifcorduk/papilo.git"
```
rg20, line 273: I don't think its a good idea to use individual forks. Can you create another branch in Papilo repo itself?  

@ramakrishnap-nv for viz!

ramakrishnap-nv: Agree

ramakrishnap-nv: If you can't push branch in papilo, may be we can fallback to this.

akifcorduk: The individual fork(nicolas) is already merged to main. This just changes to fork to mine.

I am not a maintainer of papilo, how can I create a branch?

ramakrishnap-nv: Yeah, that's my next concern, if you don't have an option to push , then what you have is good enough for now.

ramakrishnap-nv: And you don't need to be a maintainer to push, if you have write access or the repo has option for you to create new branch, that should be good.

akifcorduk: I don't have write access and the repo doesn't seem to allow external developers to create a branch.

ramakrishnap-nv (approved):

PR #1646, opened 2026-07-31
add a smoke test to ensure server images start correctly

tmckayus:
This adds a simple smoke test after image builds to make sure the server instances start. The goal is to detect defects in the Dockerfile used to build the image (like faulty configuration of shared library paths, etc).


.github/workflows/test_images.yaml
```diff
@@ -87,3 +87,35 @@ jobs:
       - name: Test cuopt
         run: |
           bash ./ci/docker/test_image.sh
+
+  # Host-side startup smoke: exercises ENTRYPOINT/CMD (REST) and
+  # CUOPT_SERVER_TYPE=grpc. The jobs above run *inside* the image as a GHA
+  # container and never launch the servers, so they cannot catch packaging
+  # gaps such as UBI10's RHEL lib/lib64 NCCL path miss.
+  smoke:
```
ramakrishnap-nv, line 95: Can't we add the test script as part of test_image.sh ?

tmckayus: We can't put this in test_image.sh as wired today — that job runs inside the image as a GHA container, so it never exercises ENTRYPOINT/CMD, and it already patches LD_LIBRARY_PATH for the nvidia wheels (which is what masked the UBI10 NCCL packaging bug). The smoke test has to be host-side docker run.

The bug we hit was invoking the ubi10 image with just a docker command, as a user would.

ramakrishnap-nv (approved):

PR #1650, opened 2026-07-31
build: don't install libcuopt into conda env by default

ramakrishnap-nv:
## Summary

- `build.sh` previously ran `cmake --install` into `\$CONDA_PREFIX` on every build. Agents or developers that rebuilt without reinstalling (e.g. called `ninja` directly) would run the stale installed library rather than freshly compiled code — a silent mismatch with no easy diagnosis path.
- Default is now **build-only**. Pass `--install` to explicitly opt in to installing into the active conda environment.
- The old `-n` (no-install) flag is removed since build-only is now the default.
- The separate `make ninja-build` code path (only reachable via `-n`) is replaced by a single `cmake --build` call with an optional `--target install`.
- The conda recipe for `libcuopt` already passed `-n` explicitly; that flag is dropped since it is now redundant — the recipe's separate `cmake --install cpp/build` output step is unaffected.


KyleFromNVIDIA (approved):

mlubin (approved):
Maybe hold off on landing until after 26.08, we don't want broken build scripts at release time.

tmckayus (approved):

PR #1652, opened 2026-07-31
Scale the variables with variable bounds

rg20:
## Description

Pre-scale columns by variable bound magnitude before Ruiz equilibration to improve conditioning when variable bounds span many orders of magnitude. This is enabled only for QP/SOCP similar to Ruiz scaling.

**Changes:**

- Add a pre-scaling step that scales each column so variables become O(1) using the geometric mean of finite bounds
- Applied before Ruiz equilibration, which then balances coefficient magnitudes

**Motivation:**

Some QP problems (particularly network-flow formulations) have variable bounds ranging from O(1) to O(10^11). Without pre-scaling, the barrier solver sees variables at vastly different scales, leading to ill-conditioned KKT systems. By normalizing variables to O(1) before Ruiz equilibration, we improve numerical stability.

**Benchmark Results:**

- QP (Maros benchmark):  No regression
- QCQP:  QPLIB_9002 converges to optimality.  This model was highly unstable and exhibits severe non-deterministic because of that prior to this code change
- SOCP:  db-joint dataset fails to converge

**Acknowledgment:** 
This improvement was proposed by the [Hiverge](https://www.hiverge.ai/) AI discovery engine with experiments by [@kerry-hiverge](https://github.com/kerry-hiverge).

## Checklist

- Testing
   - [ ] Added tests
   - [ ] NA
- Documentation
   - [ ] Added new documentation
   - [x] NA

### DIFF
--- a/cpp/src/dual_simplex/scaling.cpp
+++ b/cpp/src/dual_simplex/scaling.cpp
@@ -115,6 +115,78 @@ i_t scaling(const lp_problem_t<i_t, f_t>& unscaled,
         q_ratio);
     }
 
+    // -----------------------------------------------------------------------
+    // Bound-magnitude column pre-scaling.
+    // -----------------------------------------------------------------------
+    // Ruiz equilibration only balances the *coefficient* magnitudes of A and Q.
+    // On problems whose variable BOUNDS span many orders of magnitude (e.g. a
+    // network-flow QP with |bound| ranging from ~1e4 to ~1e11), the constraint
+    // matrix can already be perfectly balanced (all +/-1) while the variables
+    // themselves live at wildly different scales. The interior-point diagonal
+    // D = z/x then spans those same many orders of magnitude, wrecking the
+    // conditioning of the KKT factorization and stalling convergence.
+    //
+    // We fix this by first scaling each column so that the variable it
+    // represents becomes O(1): c0[j] = (geometric mean of the finite bound
+    // magnitudes of x_j). Substituting x_j = c0[j] * x'_j leaves the feasible
+    // region shape unchanged but brings every variable to a common scale, which
+    // Ruiz then finishes off on the coefficient side. Columns with no finite,
+    // nonzero bound are left at scale 1.
+    {
+      std::vector<f_t> c0(n, 1.0);
+      const i_t cone_start0 =
+        unscaled.second_order_cone_dims.empty() ? n : unscaled.cone_var_start;
+      f_t geo_sum = 0.0;
+      i_t geo_count = 0;
+      for (i_t j = 0; j < cone_start0; ++j) {
+        f_t lo = std::abs(scaled.lower[j]);
+        f_t hi = std::abs(scaled.upper[j]);
+        f_t mag;
+        bool lo_fin = scaled.lower[j] > -1e20 && lo > 0;
+        bool hi_fin = scaled.upper[j] < 1e20 && hi > 0;
+        if (lo_fin && hi_fin) {
+          mag = std::sqrt(lo * hi);
+        } else if (lo_fin) {
+          mag = lo;
+        } else if (hi_fin) {
+          mag = hi;
+        } else {
+          continue;  // free / one-sided-zero: leave at scale 1
+        }
+        c0[j] = mag;
+        geo_sum += std::log(mag);
+        geo_count++;
+      }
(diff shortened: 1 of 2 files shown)


mlubin:
In the past I've seen some benchmark problems with very large, artificial bounds that don't correspond to the appropriate magnitude of the variables. I'm a bit concerned that such large bounds could have a noisy or negative impact on the solve if we rescale like this.

rg20:
> In the past I've seen some benchmark problems with very large, artificial bounds that don't correspond to the appropriate magnitude of the variables. I'm a bit concerned that such large bounds could have a noisy or negative impact on the solve if we rescale like this.

Benchmarks show regressions on QP though there are some good wins on LP benchmarks.  I think its not worth pushing this change.

mlubin:
Ok to close the PR?

cpp/src/dual_simplex/scaling.cpp
```diff
@@ -115,6 +115,78 @@ i_t scaling(const lp_problem_t<i_t, f_t>& unscaled,
...
+      if (geo_count > 0) {
+        // Normalize so the average column scale is 1, keeping the overall
+        // problem magnitude centered rather than uniformly shrinking it.
+        const f_t geo_mean = std::exp(geo_sum / static_cast<f_t>(geo_count));
+        for (i_t j = 0; j < cone_start0; ++j) {
+          c0[j] /= geo_mean;
+        }
+        // Apply x_j = c0[j] * x'_j : A(:,j) *= c0[j], obj[j] *= c0[j],
+        // bounds /= c0[j], Q(i,j) *= c0[i]*c0[j], accumulate into col_scale.
+        for (i_t j = 0; j < n; ++j) {
```
yuwenchen95, line 169: Could we apply the bound scaling only if it is above a threshold, like what we did for ruiz scaling? Would it help to fix QCQP regression?

rg20: Actually I don't see any regressions now.  Updated the PR description.

rg20:
> Ok to close the PR?

I no longer see the regressions.  Its actually helping stabilize QPLIB_9002 dataset as this has very large bounds.

cpp/src/dual_simplex/scaling.cpp
```diff
@@ -115,6 +115,79 @@ i_t scaling(const lp_problem_t<i_t, f_t>& unscaled,
...
 
+    // -----------------------------------------------------------------------
+    // Bound-magnitude column pre-scaling.
+    // -----------------------------------------------------------------------
+    // Ruiz equilibration only balances the *coefficient* magnitudes of A and Q.
+    // On problems whose variable BOUNDS span many orders of magnitude (e.g. a
+    // network-flow QP with |bound| ranging from ~1e4 to ~1e11), the constraint
+    // matrix can already be perfectly balanced (all +/-1) while the variables
+    // themselves live at wildly different scales. The interior-point diagonal
+    // D = z/x then spans those same many orders of magnitude, wrecking the
```
mlubin, line 126: "wrecking" sounds a bit too dramatic

cpp/src/dual_simplex/scaling.cpp
```diff
@@ -115,6 +115,79 @@ i_t scaling(const lp_problem_t<i_t, f_t>& unscaled,
...
+      std::vector<f_t> c0(n, 1.0);
+      std::vector<char> has_bound(n, 0);
+      const i_t cone_start0 = unscaled.second_order_cone_dims.empty() ? n : unscaled.cone_var_start;
+      f_t geo_sum           = 0.0;
+      i_t geo_count         = 0;
+      for (i_t j = 0; j < cone_start0; ++j) {
+        f_t lo = std::abs(scaled.lower[j]);
+        f_t hi = std::abs(scaled.upper[j]);
+        f_t mag;
+        bool lo_fin = scaled.lower[j] > -1e20 && lo > 0;
```
mlubin, line 145: Where is the hardcoded 1e20 coming from? Should this be a named constant instead?

mlubin, line 145: 1. Don't use nonstandard abbreviations like "fin"
2. "fin" I'm guessing means finite, but this naming is confusing because we're also checking for nonzero.

mlubin (approved):

PR #1653, opened 2026-08-01
docs: add Python async gRPC client guide under gRPC remote execution

tmckayus:
Document remote execution vs explicit gRPC clients, add quick-start and streaming examples, and clarify integrated-client environment variables. Also clarify some existing docs on API and server behavior.

PR #1655, opened 2026-08-03
normalize MILP to MIP in documentation

tmckayus:
We refer to both MILP and MIP in the documentation, so we decided to normalize around MIP everywhere.

Left a few references changed separately in #1653

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@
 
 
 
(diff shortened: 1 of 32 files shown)


ramakrishnap-nv (approved):

cwilkinson76 (commented):
Doc changes all look good, but do the files with "milp" in the filename need to be renamed to "mip"?

tmckayus:
> Doc changes all look good, but do the files with "milp" in the filename need to be renamed to "mip"?

Thanks @cwilkinson76 I'll take a look at that. They shouldn't affect labels, etc but it might be better to rename them

tmckayus:
> > Doc changes all look good, but do the files with "milp" in the filename need to be renamed to "mip"?
> 
> Thanks @cwilkinson76 I'll take a look at that. They shouldn't affect labels, etc but it might be better to rename them

@cwilkinson76 maybe better to leave renames to post-release to make sure we don't break links etc. "milp" isn't wrong, just slightly inconsistent in that case.

ramakrishnap-nv:
> > Doc changes all look good, but do the files with "milp" in the filename need to be renamed to "mip"?
> 
> Thanks @cwilkinson76 I'll take a look at that. They shouldn't affect labels, etc but it might be better to rename them

Agree, lets work on a follow-up PR in next release since this might break.

PR #1656, opened 2026-08-04
Presolve work and iteration limits

### LINKED ISSUE #1068: [FEA] cuOpt presolve should use work estimates instead of time limits
This keeps the solver deterministic. And would allow us to better limit certain presolve techniques.

akifcorduk:
This PR introduces determinsitic iteration and work limits to papilo and cuopt presolve (mainly probing). With this PR, we remove the time limits. This might impact very short runtime runs, so we might need more iterations depending on feedback. Slight regression on average error gap, possibly due to varying probing time. However, we have more wins vs loses in terms of number of inttances. Few instances are impacted more from the probing time. All other metrics are improving. Both papilo and cuopt probing execute under 60s cap that was there previously. 

On MIPLIB2017 benchmarks on H100, improvements are as follows:

main (avg of 2 runs):
Optimal: 74.5
Error gap: 12.25
MIP gap(geomean): 20.0
<0.1% error gap: 120.5

This branch(avg of 2 runs):
Optimal: 75.5
Error gap: 12.34
MIP gap(geomean): 19.67
<0.1% error gap: 122.5

closes #1068


cpp/src/mip_heuristics/presolve/presolve_budget_policy.hpp
```diff
@@ -0,0 +1,170 @@
...
+
+namespace cuopt::mathematical_optimization::mip {
+
+// Work the probing loop charges per unit of effort. These are exact counts rather than timings,
+// which is what makes the budget reproducible, and probing_work_scale below is expressed against
+// them -- changing either invalidates it.
+inline constexpr double probing_probe_work = 0.02;  // per probed variable, host overhead
+inline constexpr double probing_iter_work  = 0.01;  // per multi-probe propagation iteration
+
+// Probing work allowed per unit of the cost proxy below; dividing by that proxy is what turns it
```
akifcorduk, line 25: Note for reviewer: I think these commands are useful for future tuning and improvements.

cpp/src/mip_heuristics/presolve/presolve_budget_policy.hpp
```diff
@@ -0,0 +1,170 @@
...
+  // units per second, so the same budget was worth 360x more time on one instance than another.
+  //
+  // The ceiling is the only work bound; there is no coverage target alongside it. A coverage
+  // fraction cuts every instance by the same proportion whether or not it is the expensive one,
+  // which both obscured which of the two stopped a run and penalised the cheap instances. Dropping
+  // it left probing running over a second on 102 instances against 81 before, while runs over ten
+  // seconds fell from 34 to 11, since the ceiling truncates by cost rather than uniformly.
+  const double probing_cost_proxy = nnz + n_cand * acl;
+  b.probing_work_limit            = probing_work_scale / probing_cost_proxy;
+  b.probing_step_size             = probing_budget_step_size;
```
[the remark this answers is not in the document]
akifcorduk: The point of this is to get rid of timers completely. So it is okay if we hit that, we can improve it later.

cpp/src/mip_heuristics/presolve/probing_cache.cu
```diff
@@ -926,8 +946,15 @@ bool compute_probing_cache(bound_presolve_t<i_t, f_t>& bound_presolve,
...
     }  // implicit barrier that waits for all iterations to finish before proceeding
 
-    // TODO when we have determinism, check current threads work/time counter and filter queue
-    // items that are smaller or equal to that
+    // Single-threaded from here to the end of the step, so folding the per-task counts in a fixed
+    // order gives the same work_used for any thread count.
+    for (size_t t = 0; t < num_tasks; ++t) {
+      iters_done += iter_accum_pool[t];
+      iter_accum_pool[t] = 0.0;
+    }
```
aliceb-nv, line 954: Should this be a max-reduction? Since we want a deterministic proxy for wall-time (and the OMP block should ensure little contention)

aliceb-nv (approved):
Thanks a ton for this Akif!

ramakrishnap-nv (approved):

cpp/src/mip_heuristics/presolve/probing_cache.cu
akifcorduk: You are right. Let me fix and rerun benchmarks.

akifcorduk: Actually, there are a lot of contention because the actual probes are running on GPU on different streams. The point of multiple streams is to create contention so the GPU is fully utilized. So, summation might approximate the wall time better. What do you think?

aliceb-nv: Ah right :) Yup, it makes sense here then, I agree.

PR #1660, opened 2026-08-04
GF2 presolve rank-deficiency fix

aliceb-nv:
The GF(2) presolver treated a missing pivot column as proof of infeasibility, so rank-deficient but consistent GF2 blocks were wrongly reported infeasible. It now runs a full RREF that separates inconsistency from underdetermination and fixes only variables whose value is identical in every solution.

The GF2 presolver also now supports rectangular matrices, and test coverage has been expanded.

## Checklist

- Testing
   - [ ] Added tests
   - [ ] NA
- Documentation
   - [ ] Added new documentation
   - [ ] NA


cpp/src/mip_heuristics/presolve/third_party_presolve.cpp
```diff
@@ -657,46 +657,55 @@ void check_postsolve_status(const papilo::PostsolveStatus& status)
...
 
 template <typename f_t>
-void set_presolve_methods(papilo::Presolve<f_t>& presolver,
-                          problem_category_t category,
-                          bool dual_postsolve)
+void set_presolve_methods(
+  papilo::Presolve<f_t>& presolver,
+  problem_category_t category,
+  bool dual_postsolve,
+  std::optional<std::unordered_set<std::string>> const& method_allowlist = std::nullopt)
```
nguidotti, line 664: Are you using the `allowlist` for anything? Maybe a better approach is pass all the desired reductions as a list (`empty = all`) and set the presolve methods based on that using a loop + switch case

cpp/src/mip_heuristics/presolve/third_party_presolve.hpp
```diff
@@ -96,6 +98,12 @@ class third_party_presolve_t {
     double time_limit,
     i_t num_cpu_threads = 0);
 
+  // If set, only Papilo methods whose getName() is listed are registered
+  void set_reduction_allowlist(std::optional<std::unordered_set<std::string>> allowlist)
```
akifcorduk, line 102: It this just for the tests or are we planning to expose it to user or developers with some config?

akifcorduk (approved):
Thanks Alice!

cpp/src/mip_heuristics/presolve/third_party_presolve.hpp
aliceb-nv: Just for testing convenience, I wasn't too sure how to do this best to be fair :/ Seemed like the least bad option. But it will be useful for later presolver pass PRs

cpp/src/mip_heuristics/presolve/third_party_presolve.cpp
aliceb-nv: I went with this path since it allows expressing the whitelist as strings, without needing to pull any Papilo headers or machinery on the calling side

nguidotti: My suggestion is to use just a vector of string (or enums) and then translate from there, e.g.,
```cpp
std::vector<std::string> reductions = {"SingletonCols", "GF2Presolve"};
set_presolve_methods(reductions);
```
and 
```cpp
void set_presolve_methods(reductions) {
  for(auto r : reductions) {
    if (r == "SingletonCols") presolver.addPresolveMethod(uptr(new papilo::SingletonCols<f_t>()));
    else if ...
  }
}
```

aliceb-nv: The idea here being that we can just crosscheck against the presolver names reported by Papilo reductions,  so that we don't have to manually if-else over all possible options

ramakrishnap-nv (approved):

PR #1661, opened 2026-08-04
ensure nightly builds always produce new packages

jameslamb:
Contributes to https://github.com/rapidsai/build-planning/issues/218

* uses the new patterns for versioning nightly packages (see https://github.com/rapidsai/shared-workflows/pull/603), to ensure they're always uploaded on new builds. 

For wheels:

```text
# before
26.10.0a55 

# after
26.10.0a55.post260803200854
```

And for conda:

```text
# before
26.10.0a55[build=cuda12_260803_9da146ef]

# after
26.10.0a55[build=cuda12_260803200854_9da146ef]
```

### How I tested this

See https://github.com/rapidsai/cugraph-gnn/pull/508


.github/workflows/build.yaml
```diff
@@ -45,33 +45,38 @@ concurrency:
...
   cpp-build:
     permissions:
       actions: read
       contents: read
       id-token: write
       packages: read
       pull-requests: read
+    needs: [build-details]
     secrets: inherit # zizmor: ignore[secrets-inherit]
-    uses: rapidsai/shared-workflows/.github/workflows/conda-cpp-build.yaml@release/26.08
```
jameslamb: @ramakrishnap-nv why are all these references on `main` set to `@release/26.08`? Should they be set to `main`?

I'd only expect them to be version-pinned on a `release/YY.MM` branch here.

ramakrishnap-nv: @jameslamb With our new release process, main will stick to current release version  till release date so we will cut once we ready for release, since main moved to 26.10 in rapids, I had to switch.

jameslamb: Ok, got it. Until that happens, then, we won't be able to apply this migration to `cuopt` (because it depends on `shared-workflows` changes on `main`).

jameslamb: These have moved to `main`, updated this branch. This is ready for review.

ramakrishnap-nv (approved):

python/cuopt_server/cuopt_server/utils/job_queue.py
```diff
@@ -51,7 +51,7 @@ class PickleForbidden(Exception):
 def check_client_version(client_vers):
     logging.debug(f"client_vers is {client_vers} in check")
     if os.environ.get("CUOPT_CHECK_CLIENT", True) in ["True", True]:
-        major, minor, _ = __version__.split(".")
+        major, minor, *_ = __version__.split(".")
```
jameslamb, line 54: cuopt server tests were failing like this:

```text
  File "/pyenv/versions/3.12.13/lib/python3.12/site-packages/cuopt_server/utils/job_queue.py", line 54, in check_client_version
    major, minor, _ = __version__.split(".")
    ^^^^^^^^^^^^^^^
ValueError: too many values to unpack (expected 3)
```

([build link](https://github.com/NVIDIA/cuopt/actions/runs/31141567231/job/92889103705?pr=1661#step:9:444))

This PR now adds a 4th component to nightly versions (e.g. `26.10.0a14.post260807023309`).

Since this code only cares about the first 2 components (major and minor), just throwing away the others with `*` seems defensible.

I looked for other patterns like this in the codebase and didn't see any.

ramakrishnap-nv (approved):

PR #1668, opened 2026-08-05
docs: enforce heading title case with Vale

ramakrishnap-nv:
Heading capitalization has so far only been caught in human review. This adds a [Vale](https://vale.sh/) prose-lint hook so it is checked mechanically, and fixes the existing headings that do not comply.

The rule uses **Chicago** style, which keeps short prepositions and conjunctions lowercase (`Where to Find Examples`, `Working with Incumbent Solutions`) — matching how these docs already read. It also catches the opposite error: existing headings such as `LP Example With Data` that over-capitalize.

All 22 heading fixes are case-only, so RST underline lengths and HTML anchors are unchanged.

### Verification
- `vale docs/cuopt/source` — 0 errors in 66 files
- `pre-commit run --all-files` — all hooks pass
- `./build.sh docs` — exit 0, html + linkcheck, zero Sphinx warnings (confirms no `autosectionlabel` cross-reference broke)

### Reviewer note
Vale's `exceptions` match whole words, and a single match makes Vale skip the **entire** heading. A too-broad entry therefore silently disables the check for every heading that mentions it. The list must stay narrow, and `vale` should be re-run after editing it — documented in both `ci/vale/styles/cuOpt/Headings.yml` and `docs/cuopt/README.md`.

### Follow-up
A `substitution` rule for MILP → MIP is the natural next step, but it cannot be added until #1653 merges — `main` still has 21 MILP occurrences that #1653 removes, and CI runs `pre-commit` on all files.


cwilkinson76 (approved):
LGTM

tmckayus (approved):

PR #1683, opened 2026-08-06
Block Bounded Value Elimination presolving pass for binary problems

aliceb-nv:
This PR adds a binary reduction presolving pass inspired by SAT-related bounded variable elimination work such as "Effective Preprocessing in SAT through Variable and Clause Elimination" (SAT 2005).

Block BVE eliminates a small set of non-objective binary variables (the interior) by projecting the constraints they appear in onto the other binary variables of those same rows (the boundary),  which stay in the model.
For every assignment of the boundary it checks if the interior can be set to satisfy each row of the block.

The assignments admitting no such setting are everything the block still forces on the rest of the
model, so these can be turned into no-good rows while eliminating the interior columns altogether. Reductions requiring new no-good clauses are only committed if they would result in fewer total rows, to avoid excessive growth.

Candidates come from the probing cache. Each non-objective binary carrying at least one edge becomes a seed, tried in order of increasing row count, grown one variable at a time: among the eligible implication neighbors of the current interior, it absorbs whichever leaves the smallest boundary, and only while the boundary strictly shrinks.

Each grown interior is staged into a block and dropped if it breaks preset caps. Disjoint candidates are batched and enumerated on the GPU, yielding a feasibility bit and witness interior per boundary pattern, and the derived clauses are checked against that table and checked for growth.

Results are as follows on MIPLIB2017. Most reductions occur on mostly-binary combinatorial problems like the bnatt class, the piperout class, and the cryptanalysis classes.

We now find **+1 feasible/optimal**: bnatt400 The BKS is found in ~5s, and optimality proven after B&B in ~2min.

```
  Instance                     Relative row reduction    Relative column reduction
  ━━━━━━━━━━━━━━━━━━━━━━━━━━━  ━━━━━━━━━━━━━━━━━━━━━━━━  ━━━━━━━━━━━━━━━━━━━━━━━━━━━
   bnatt500                                    40.532%                      46.000%
  ───────────────────────────  ────────────────────────  ───────────────────────────
   bnatt400                                    40.720%                      45.167%
  ───────────────────────────  ────────────────────────  ───────────────────────────
   piperout-27                                 19.922%                      10.507%
  ───────────────────────────  ────────────────────────  ───────────────────────────
   piperout-08                                 14.991%                       7.020%
  ───────────────────────────  ────────────────────────  ───────────────────────────
   cryptanalysiskb128n5obj14                    7.480%                       9.175%
  ───────────────────────────  ────────────────────────  ───────────────────────────
   cryptanalysiskb128n5obj16                    7.480%                       9.169%
  ───────────────────────────  ────────────────────────  ───────────────────────────
   neos-950242                                  2.875%                       2.500%
  ───────────────────────────  ────────────────────────  ───────────────────────────
   physiciansched3-3                            2.128%                       3.874%
  ───────────────────────────  ────────────────────────  ───────────────────────────
   neos8                                        1.243%                       1.240%
  ───────────────────────────  ────────────────────────  ───────────────────────────
   roll3000                                     1.133%                       1.201%
```

Benchmark results:

```
        Seed    Feasible bnatt / main    Primal gap bnatt / main    Primal integral bnatt / main    SGM(0.001) bnatt / main
  ━━━━━━━━━━━  ━━━━━━━━━━━━━━━━━━━━━━━  ━━━━━━━━━━━━━━━━━━━━━━━━━  ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━  ━━━━━━━━━━━━━━━━━━━━━━━━━
           0                226 / 226        0.114377 / 0.114337             0.160337 / 0.165177        0.030480 / 0.031056
  ───────────  ───────────────────────  ─────────────────────────  ──────────────────────────────  ─────────────────────────
           1                226 / 225        0.113938 / 0.116821             0.159689 / 0.170257        0.030555 / 0.030913
  ───────────  ───────────────────────  ─────────────────────────  ──────────────────────────────  ─────────────────────────
           2                225 / 225        0.115021 / 0.117333             0.164569 / 0.165596        0.031414 / 0.029845
  ───────────  ───────────────────────  ─────────────────────────  ──────────────────────────────  ─────────────────────────
   Aggregate                677 / 676        0.114445 / 0.116164             0.161532 / 0.167010        0.030814 / 0.030600
   ```

## Checklist

- Testing
   - [ ] Added tests
   - [ ] NA
- Documentation
   - [ ] Added new documentation
   - [ ] NA

### CODE REVIEW

File: cpp/src/mip_heuristics/diversity/diversity_manager.cu [hunk 1]
```diff
@@ -304,19 +398,100 @@ bool diversity_manager_t<i_t, f_t>::run_presolve(f_t time_limit, timer_t global_
...
+       !problem_ptr->original_problem_ptr->get_problem_name().empty())
+        ? problem_ptr->original_problem_ptr->get_problem_name()
+        : std::string("cuopt");
+    const std::string mps_path = instance_name + "_gpupresolved.mps";
+    CUOPT_LOG_DEBUG("Exporting GPU-presolved problem to %s", mps_path.c_str());
+    auto model = problem_to_mps_data_model(*problem_ptr);
+    cuopt::mathematical_optimization::io::mps_writer_t<i_t, f_t> writer(model);
+    writer.write(mps_path);
+    exit(0);
+  }
```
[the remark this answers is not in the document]
aliceb-nv: This is a debug-only utility

cpp/src/mip_heuristics/presolve/block_bve.cu
```diff
@@ -0,0 +1,1372 @@
...
+  // Forcings mined from earlier projections. Pairs the cache never held become seed/absorb
+  // candidates, so a later round can grow blocks the first round could not see.
+  if (prior_original_id_findings != nullptr) {
+    for (const auto& forcing : prior_original_id_findings->forcings)
+      add_edge(forcing.var, forcing.forced_var);
+  }
+  std::vector<std::vector<i_t>> out(n_vars);
+  for (i_t v = 0; v < n_vars; ++v)
+    out[v].assign(adj[v].begin(), adj[v].end());
+  return out;
```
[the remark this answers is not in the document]
aliceb-nv: The determinism contract in the repo is run-to-run same env, so your finding is irrelevant.

skills/cuopt-developer/SKILL.md
```diff
@@ -171,6 +171,7 @@ cuopt/
 - Keep operations stream-ordered
 - Follow existing RAFT/RMM patterns
 - No raw `new`/`delete` - use RMM allocators
+- Prefer modern CCCL bit/math helpers in kernels (`cuda::bitfield_extract`, `cuda::bitmask`, pow2 utilities) over hand-rolled `%`/`/` by runtime powers of two — see [references/conventions.md](references/conventions.md)
```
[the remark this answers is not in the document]
aliceb-nv: this is out of my control

cpp/src/mip_heuristics/diversity/diversity_manager.cu
```diff
@@ -279,6 +337,42 @@ void diversity_manager_t<i_t, f_t>::add_user_given_solutions(
...
+    if (original_id < 0 || original_id >= (i_t)reverse_original_ids.size()) { continue; }
+    const i_t column = reverse_original_ids[original_id];
+    if (column < 0 || column >= problem.n_variables) { continue; }  // already eliminated
+    var_indices.push_back(column);
+    lb_values.push_back(value ? f_t(1) : f_t(0));
+    ub_values.push_back(value ? f_t(1) : f_t(0));
+  }
+  n_applied = (i_t)var_indices.size();
+  problem.update_variable_bounds(var_indices, lb_values, ub_values);
+  return true;
```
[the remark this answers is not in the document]
aliceb-nv: problem.cu:2387 opens with if (var_indices.size() == 0) { return; }

cpp/src/mip_heuristics/problem/problem.cu
```diff
@@ -2214,32 +2093,66 @@ void problem_t<i_t, f_t>::set_constraints_from_host_user_problem(
...
+  cuopt_assert(variables_in.size() == coefficients_in.size(), "csr index/value size mismatch");
+  cuopt_assert(static_cast<size_t>(offsets_in.back()) == variables_in.size(),
+               "CSR offsets/entries size mismatch");
+  cuopt_assert(names.empty() || names.size() == static_cast<size_t>(n_constraints),
+               "row names size mismatch");
+  for (i_t variable : variables_in) {
+    cuopt_assert(variable >= 0 && variable < n_variables, "CSR variable out of bounds");
   }
+  nnz   = static_cast<i_t>(variables_in.size());
+  empty = (nnz == 0 && n_constraints == 0 && n_variables == 0);
```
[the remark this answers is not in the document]
aliceb-nv: Block-BVE calls trivial_presolve immediately after installing the rewritten rows (block_bve.cu:1330), and nothing asserts in between: set_constraints_from_host_csr passes check_representation = false to recompute_auxilliary_data, and compute_transpose_of_problem early-returns on an empty matrix before reaching check_csr_representation. Inside trivial_presolve, update_from_csr treats every column with no matrix entry as unused, fixes it to its objective-best bound via assign_fixed_var_t, folds its objective contribution into objective_offset, and drops it. With zero rows that covers every column, so n_variables reaches 0 and trivial_presolve.cuh:370 sets empty = true before check_problem_representation runs.

Applying the change would also break an invariant the solver relies on. Post-presolve empty currently means "everything was eliminated", and solver.cu:240 acts on it by calling set_problem_fully_reduced() and returning sol without assigning anything. Setting empty = true while n_variables > 0 would return whatever sol.assignment happens to hold for the surviving columns. The bound-picking loop that handles a constraint-free model correctly lives at solve.cu:136, and that path only sees the constructor-set flag, never a presolve-set one.

aliceb-nv:
@coderabbitai full review

cpp/src/utilities/integer_scaling.hpp
```diff
@@ -0,0 +1,217 @@
...
+  }
+  return std::numeric_limits<double>::quiet_NaN();
+}
+
+}  // namespace detail
+
+// Continued-fractions approach: rationalize each coefficient, compute scm/gcd incrementally.
+// Returns the smallest positive multiplier s such that s * c is (near-)integer for every c, or NaN
+// if no such multiplier exists within the caps.
+inline double find_scaling_rational(const std::vector<double>& coefficients,
```
nguidotti, line 113: I think Chris also implemented a continued-fractional variable. Maybe it is better to use that.

nguidotti (commented):
Thanks for the hard work, Alice!

cpp/src/mip_heuristics/diversity/diversity_manager.cu
```diff
@@ -282,6 +284,42 @@ void diversity_manager_t<i_t, f_t>::add_user_given_solutions(
   }
 }
 
+// Pin variables that a BVE projection table showed to have a single admissible value. Ids arrive in
+// the original frame and may repeat across blocks and rounds. Returns false when two blocks
+// disagree on a variable, which proves infeasibility since each fixing is a consequence of its
+// block alone.
+template <typename i_t, typename f_t>
+static bool apply_bve_fixings(problem_t<i_t, f_t>& problem,
```
akifcorduk: I think this shouldn't reside on diversity_manager.cu.  Maybe as part of presolve?

cpp/src/mip_heuristics/diversity/diversity_manager.cu
```diff
@@ -282,6 +284,42 @@ void diversity_manager_t<i_t, f_t>::add_user_given_solutions(
...
+  std::vector<i_t> var_indices;
+  std::vector<f_t> lb_values;
+  std::vector<f_t> ub_values;
+  for (size_t k = 0; k < sorted.size(); ++k) {
+    const auto [original_id, value] = sorted[k];
+    if (k > 0 && original_id == sorted[k - 1].first) {
+      if (value != sorted[k - 1].second) { return false; }
+      continue;
+    }
+    if (original_id < 0 || original_id >= (i_t)reverse_original_ids.size()) { continue; }
```
akifcorduk: I think this should be an assert. This will miss issues.

cpp/src/mip_heuristics/diversity/diversity_manager.cu
```diff
@@ -324,9 +365,99 @@ bool diversity_manager_t<i_t, f_t>::run_presolve(f_t time_limit, timer_t global_
       std::chrono::duration<double>(std::chrono::steady_clock::now() - probing_t0).count());
     if (problem_is_infeasible) { return false; }
   }
-  const bool remap_cache_ids           = true;
-  problem_ptr->related_vars_time_limit = context.settings.heuristic_params.related_vars_time_limit;
+
   if (!global_timer.check_time_limit()) { trivial_presolve(*problem_ptr, remap_cache_ids); }
+
+  i_t max_bve_rounds          = 3;
```
akifcorduk: I think we should move all the bve logic in a new file in presolve folder. Including the apply_bve_fixings.

cpp/src/mip_heuristics/diversity/diversity_manager.cu
```diff
@@ -324,9 +365,99 @@ bool diversity_manager_t<i_t, f_t>::run_presolve(f_t time_limit, timer_t global_
...
+  const i_t n_vars_before_bve = problem_ptr->n_variables;
+  const i_t n_rows_before_bve = problem_ptr->n_constraints;
+
+  if (!run_probing_cache) max_bve_rounds = 0;
```
akifcorduk: I would check if the probing cache is empty. The probing cache might have run but it might be empty.

cpp/src/mip_heuristics/diversity/diversity_manager.cu
```diff
@@ -324,9 +365,99 @@ bool diversity_manager_t<i_t, f_t>::run_presolve(f_t time_limit, timer_t global_
...
+      break;
+    }
+    double bve_work_units = 0.0;
+    timer_t bve_timer(bve_stage_timer.clamp_remaining_time(
+      global_timer.clamp_remaining_time(presolve_timer.remaining_time())));
+    bool bve_proved_infeasible = false;
+    const bool reduced         = block_bve_presolve(
+      *problem_ptr, impl_adj, bve_timer, bve_work_units, &bve_findings, &bve_proved_infeasible);
+    if (bve_proved_infeasible) {
+      CUOPT_LOG_INFO("Block-BVE proved the problem infeasible");
```
akifcorduk: Have you checked if this was printed in any of the benchmarks?

cpp/src/mip_heuristics/presolve/probing_cache.cu
```diff
@@ -704,36 +713,52 @@ void apply_substitution_queue_to_problem(
...
   std::vector<f_t> coefficient_values;
 
-  // Get variable_mapping to convert current indices to original indices
+  // Get variable_mapping to convert current indices to post-Papilo frame
   auto h_variable_mapping =
     host_copy(problem.presolve_data.variable_mapping, problem.handle_ptr->get_stream());
   problem.handle_ptr->sync_stream();
 
+  // Collect AffineSub reconstructions, then append in deterministic order (by substituted_var).
+  std::vector<var_postsolve_t<i_t, f_t>> batch_recs;
```
akifcorduk: Can you explain a bit what is`batch_recs`? Also why is there a need for new logic to handle substitutions? Can't we just convert the BVE substitutions into already existing format?

cpp/src/mip_heuristics/presolve/probing_cache.cu
```diff
@@ -704,36 +713,52 @@ void apply_substitution_queue_to_problem(
...
   }
+  std::sort(batch_recs.begin(),
+            batch_recs.end(),
+            [](const var_postsolve_t<i_t, f_t>& a, const var_postsolve_t<i_t, f_t>& b) {
+              return a.sub.substituted_var < b.sub.substituted_var;
+            });
+  auto& recs = problem.presolve_data.var_postsolve;
+  recs.insert(recs.end(),
+              std::make_move_iterator(batch_recs.begin()),
+              std::make_move_iterator(batch_recs.end()));
```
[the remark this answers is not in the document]
akifcorduk: @aliceb-nv  sanitize_graph only handles deduplicationn of A->B and B->A. Three-way+ deduplications are not handled. A workaround might be to leave it but run are_exclusive before passing to the problem.subtitute_variables.

cpp/src/mip_heuristics/presolve/probing_cache.cu
```diff
@@ -997,6 +1027,54 @@ bool compute_probing_cache(bound_presolve_t<i_t, f_t>& bound_presolve,
...
+  i_t n_contradicted = 0;
+  for (const auto& forcing : forcings) {
+    cuopt_assert(forcing.var != forcing.forced_var, "self-forcing is not a projection finding");
+    auto entry_it = probing_cache.find(forcing.var);
+    if (entry_it == probing_cache.end()) { continue; }
+    const f_t probed_val = forcing.value ? f_t(1) : f_t(0);
+    const f_t forced_val = forcing.forced_value ? f_t(1) : f_t(0);
+    for (cache_entry_t<i_t, f_t>& entry : entry_it->second) {
+      if (entry.var_to_cached_bound_map.empty()) { continue; }
+      if (entry.val_interval.interval_type != interval_type_t::EQUALS) { continue; }
```
akifcorduk, line 1063: Doesn't a cache entry with `EQUALS` interval and `entry.val_interval.val == probed_val`cause a conflict unless it is the same forced value?

cpp/src/mip_heuristics/problem/presolve_data.cu
```diff
@@ -135,19 +135,44 @@ void presolve_data_t<i_t, f_t>::post_process_assignment(
...
-    CUOPT_LOG_DEBUG("Post-process substitution: x[%d] = %f + %f * x[%d] = %f",
-                    sub.substituted_var,
-                    sub.offset,
-                    sub.coefficient,
-                    sub.substituting_var,
-                    h_assignment[sub.substituted_var]);
+  // Reverse-append undo of the unified GPU-presolve reconstruction log
+  for (auto it = var_postsolve.rbegin(); it != var_postsolve.rend(); ++it) {
+    const auto& rec = *it;
+    switch (rec.kind) {
```
akifcorduk: The logic is a bit complicated here. Do you think if there is a way to convert the bve substitutions into a standardized substitutions?

cpp/src/utilities/integer_scaling.hpp
```diff
@@ -0,0 +1,217 @@
...
+
+#include <utilities/macros.cuh>
+
+namespace cuopt {
+
+namespace detail {
+
+// Best rational approximation p/q to x with q <= max_denom, via continued fractions. Returns the
+// last valid convergent if the denominator limit is reached.
+inline std::pair<int64_t, int64_t> rational_approximation(double x,
```
akifcorduk, line 27: I think there is already something similarr to this in: `cpp/src/cuts/rational.hpp`

akifcorduk (commented):
Thanks Alice! Great results! I couldn't review the BVE reductions in detail but I reviewed the surrounding logic. Have you tested the probing_cache probe counts, runtime of cuopt presolve and the impact of this to primal gap/integral ? Since this is some heavy presolve logic, I would run a benchmark with asserts to see if there is any obvious issue, later it is quite hard to find presolve related issues.

ramakrishnap-nv (approved):

cpp/src/mip_heuristics/diversity/diversity_manager.cu
aliceb-nv: Darn, oversight! No, it's not, but it should be _DEBUG regardless, I will fix

cpp/src/mip_heuristics/presolve/probing_cache.cu
aliceb-nv: It's the postsolve stack for cuOpt reductions (subsitutions only previously, now the blockBVE reconstructions as well). Agreed that it's a terrible name :) Will rename to postsolve_reconstructions or something.

For the question, the BVE reconstructions are a bit more complicated in that they consider the values of multiple binaries (the "boundary"), turn this into a bit pattern, and index it into a precomputed table to reconstruct another set of binaries (the "interior"). This can't be expressed with affine substitutions nor with Papilo's postsolve tools, so I had to add a new type of postsolve operation for this

cpp/src/utilities/integer_scaling.hpp
aliceb-nv: This is existing code, I just moved it into a shared header :) I'll see if it's possible to unify everything cleanly

cpp/src/mip_heuristics/presolve/probing_cache.cu
aliceb-nv: Yes, and if there's a conflict, we conclude this forces a fixing to the other value for the binary variable (or the problem is infeasible)

cpp/src/mip_heuristics/problem/presolve_data.cu
aliceb-nv: It's not really possible I think :/ They're not affine substitutions at all. I will do a bit of cleanup for readability though

cpp/src/utilities/integer_scaling.hpp
aliceb-nv: I looked, I could unify them, but maybe in a later PR because I'm concerned of accidentally changing semantics and breaking cut generation

aliceb-nv:
Thanks for the review work Akif :) I've added benchmark numbers in the PR description. We're slightly better in feasibility, primal gap, and primal integral, mostly thanks to the bnatt400 solve. SGM doesn't really change, but as far as I can see this is all within typical run to run variance. I saw no obvious regression

cpp/src/mip_heuristics/presolve/probing_cache.cu
akifcorduk: Okay, you are doing what i described already:) But I couldn't see how you report infeasibility with `n_contradicted`, do you report infeasibility with it somewhere?

akifcorduk (approved):

cpp/src/mip_heuristics/presolve/probing_cache.cu
aliceb-nv: This is done downstream in "apply_bve_fixings"

PR #1690, opened 2026-08-07
ci: update NVSkills CI request template and add require-nvskills-status workflow

ramakrishnap-nv:
## Summary

Updates \`.github/workflows/request-nvskills-ci.yml\` to the latest NVSkills CI request template as requested by the NVSkills team.

**Changes vs the old template:**
- Adds \`pull_request\` trigger so NVSkills CI validation status is shown directly on PRs
- \`require-nvskills-ci\` job calls the centrally managed \`NVIDIA/skills/.github/workflows/require-nvskills-status.yml@main\` — no local copy needed in this repo
- \`request\` job is now inline (removes dependency on the external reusable workflow call that was causing 403 dispatch failures)
- Only dispatches when \`skills/\`, \`team-skills/\`, \`rules/team-rules/\`, or \`plugins/\` files are changed — including renames out of watched paths (\`previous_filename\` check added to match downstream behavior and prevent merge deadlock)
- Permission check updated from \`OWNER/MEMBER/COLLABORATOR\` to \`admin|maintain\` role
- Bot actor updated from \`nv-nvskill-ci[bot]\` to \`nv-skills-ci[bot]\`

## Testing

Pre-commit passes (yamllint, zizmor, verify-copyright).

## Docs

No doc changes needed.


.github/workflows/request-nvskills-ci.yml
```diff
@@ -10,22 +10,201 @@ name: Request NVSkills CI
...
+          GH_TOKEN: ${{ github.token }}
+          REPO: ${{ github.repository }}
+          ACTOR: ${{ github.actor }}
+        run: |
+          set -euo pipefail
+          role_name="$(curl -fsSL \
+            -H "Authorization: Bearer ${GH_TOKEN}" \
+            -H "Accept: application/vnd.github+json" \
+            "https://api.github.com/repos/${REPO}/collaborators/${ACTOR}/permission" \
+            | jq -r '.role_name // ""')"
```
jameslamb, line 56: You might find it easier to use `gh api` with with the GitHub CLI than this `curl`. That'd keep `GH_TOKEN` out of the code entirely (`gh` would just silently read the variable and use it).

.github/workflows/request-nvskills-ci.yml
```diff
@@ -10,22 +10,201 @@ name: Request NVSkills CI
...
+            exit 1
+          fi
+
+          owner="${REPO%%/*}"
+          repo="${REPO#*/}"
+
+          curl -fsSL -X POST \
+            -H "Authorization: Bearer ${DISPATCH_TOKEN}" \
+            -H "Accept: application/vnd.github+json" \
+            "https://api.github.com/repos/NVIDIA/nvskills-ci/actions/workflows/nvskills-ci.yml/dispatches" \
```
jameslamb, line 188: The GitHub API has a `gh workflow run` which would be simpler than `curl`-ing against `api.github.com`, I think you should consider that.

jameslamb (approved):
Looks fine, do what you need to. I left some small suggestions.

This is also quite a lot of shell code to have in workflow files... IMO it'd be better in scripts in the repo, where it'd go through `shellcheck` and other linting.

PR #1693, opened 2026-08-07
Add cooperative cancel for LP/MIP solves

tmckayus:
Poll a shared cancel flag like time limits so mid-solve cancel can unwind and avoid as much as possible the need for SIGKILL especially in the gRPC server, since SIGKILL can leave the GPU in an uninterruptible sleep in some cases. Remap limit statuses to Cancelled at solution finalization, preempt MIP heuristics on cancel.


tmckayus:
engine team: I did this with AI so please check carefully and treat this as a rough cut -- there very well may be improvements that can be made.  

I reviewed and kept the changes as minimal as possible.  I will call out a few places specifically with comments where I'm uncertain what the best approach is.

The basic idea is to test a cancel flag, settable from a thread in the same process as the solver, every place that a time limit or iteration limit is currently checked.  Much of the code change is the modification of timer objects which now take a cancel flag at initialization.

Also, please feel free to take over this PR and carry it across the goal line :)

cpp/src/branch_and_bound/branch_and_bound.cpp
```diff
@@ -1733,6 +1738,14 @@ void branch_and_bound_t<i_t, f_t>::plunge_with(bfs_worker_t<i_t, f_t>* worker,
       break;
     }
 
+    if (settings_.cancel_requested != nullptr &&
```
tmckayus, line 1741: The code here matches the behavior in the timer limit case afaik.  Maybe they can be combined?

cpp/src/branch_and_bound/branch_and_bound.cpp
```diff
@@ -2800,6 +2813,15 @@ lp_status_t branch_and_bound_t<i_t, f_t>::solve_root_relaxation(
   // Wait for the root relaxation solution to be sent by the diversity manager or dual simplex
   while (!root_crossover_solution_set_.load(std::memory_order_acquire) &&
          *get_root_concurrent_halt() == 0) {
+    // Dual may early-return on time/cancel without crossover; also poll limits
```
tmckayus, line 2816: In testing I hit a point where after a cancel, a MIP would never complete. It seemed to be the case that one of the threads in the OpenMP team was halted while the other wasn't, and so the join never completed.  After drilling in to the failure with the agent, this seemed like the fix and no hangs were ever observed afterward.

cpp/src/dual_simplex/phase2.cpp
```diff
@@ -2573,6 +2573,16 @@ dual_status_t dual_phase2_with_advanced_basis(i_t phase,
 
   dual_status_t status = dual_status_t::UNSET;
 
+  // Concurrent root MIP: dual and barrier/PDLP/crossover race. On any phase-2
+  // exit (optimal, time limit, cancel→TIME_LIMIT early return, etc.), raise
+  // concurrent_halt so waiters in solve_root_relaxation are not stuck forever
+  // when dual returns before the success-path epilogue below.
```
tmckayus, line 2584: related to the same deadlock mentioned above

cpp/src/grpc/server/grpc_job_management.cpp
```diff
@@ -232,7 +232,7 @@ void delete_log_file(const std::string& job_id)
 
 int cancel_job(const std::string& job_id, JobStatus& job_status_out, std::string& message)
 {
-  std::lock_guard<std::mutex> lock(tracker_mutex);
```
tmckayus, line 235: this was changed because we need the ability to unlock, since we return the cancel status immediately but launch a background thread to do a SIGKILL if the worker fails to cancel the job.

cpp/src/grpc/server/grpc_worker.cpp
```diff
@@ -394,6 +397,13 @@ static SolveResult run_mip_solve(DeserializedJob& dj,
     auto gpu_solution = cuopt::mathematical_optimization::solve_mip(*gpu_problem, dj.mip_settings);
     SERVER_LOG_INFO("[Worker] solve_mip done");
 
+    if (gpu_solution.get_termination_status() ==
```
tmckayus, line 400: the point of this is, in the grpc server, always set the status as "cancelled" if we know it was cancelled regardless of the status that comes back. It may be hard to properly return "cancel" vs "time limit" throughout the solver.

cpp/src/mip_heuristics/diversity/diversity_manager.cu
```diff
@@ -402,6 +406,12 @@ void diversity_manager_t<i_t, f_t>::generate_quick_feasible_solution()
 template <typename i_t, typename f_t>
 bool diversity_manager_t<i_t, f_t>::check_b_b_preemption()
 {
+  if (context.settings.cancel_requested != nullptr &&
```
tmckayus, line 409: This is one of those places that needs close engine attention :)

cpp/src/mip_heuristics/feasibility_jump/early_cpufj.cu
```diff
@@ -36,7 +38,12 @@ void early_cpufj_t<i_t, f_t>::start()
   this->preemption_flag_.store(false);
   this->start_time_ = std::chrono::steady_clock::now();
 
-  fj_cpu_ = init_fj_cpu_standalone(*this->problem_ptr_, *this->solution_ptr_, preemption_flag_);
```
tmckayus, line 39: this should be checked closely as well. It substitutes the cancel flag for the local preempt flag if the cancel flag has been wired in.  If the solve was not configured with a cancel flag, it's unchanged.

cpp/src/mip_heuristics/local_search/local_search.cu
```diff
@@ -145,7 +145,11 @@ void local_search_t<i_t, f_t>::stop_cpufj_scratch_threads()
 {
   if (omp_get_num_threads() < CUOPT_MIP_FJ_REQUIRED_THREAD_COUNT) return;
 
+  // Climbers are null until the matching start_* fills them (and start_lptopt
+  // is independent of start_scratch). Skip unset slots so stop is safe if a
+  // start never ran.
   for (size_t i = 0; i < scratch_cpu_fj.size(); ++i) {
```
tmckayus, line 151: This is just an extra safety. In one iteration of the code, it was possible that this  might be null. Left the check because it seemed like a good one to have.

cpp/src/mip_heuristics/presolve/third_party_presolve.cpp
```diff
@@ -745,11 +745,20 @@ void set_presolve_parameters(
     };
     // Papilo has work unit measurements for probing. Because of this when the first batch fails to
     // produce any reductions, the algorithm stops. To avoid stopping the algorithm, we set a
```
tmckayus, line 747: This is another place that needs close attention. Cursor suggested adding a max badge size for presolve so that the cancel check was sure to be polled more often.  However, I don't know what affect this may have on performance. The alternative is to make the grace period for SIGKILL longer in threads that monitor cancel (as in the gRPC server).

cpp/src/mip_heuristics/solver.cu
```diff
@@ -497,6 +502,14 @@ solution_t<i_t, f_t> mip_solver_t<i_t, f_t>::run_solver()
     sol                           = dm.run_solver();
   }  // implicit barrier for all tasks created in B&B and heuristics
 
+  const bool cancel_requested = cuopt::cancel_flag_set(context.settings.cancel_requested) ||
```
tmckayus, line 505: another place to check correctness.

chris-maes (changes requested):
I'm in favor of adding the ability to stop the solver. Thanks for looking into this.

An important use case is the user pressing Ctrl-C to stop a MIP solve and still getting back some solution info in this case.

But this is a giant PR. Somethings seems off it takes ~3K lines of code to stop the solver.

Could we take a step back and work out the basic design of how we want this to work? For example, what is the termination code for canceled solves etc? Do we want to piggy back off CONCURRENT_HALT or do we want to separate this?

Then lets figure out how to implement this. I don't think we want to put this in the timer class. I also don't think we want to have to remap status codes.

PR #1697, opened 2026-08-10
Root Heuristics (RINS, CPU FJ)

nguidotti:
- Refactored the root heuristics (more specifically, the CPU FJ launches at the root node) to be more modular, while allowing to use more workers. 
- Extended recursive RINS to run on the root node
- Fixed incorrect objective passed to the `set_solution_from_submip` (the objective was on the submip space, which is not comparable with the solution space of the B&B).

## Checklist

- Testing
   - [ ] Added tests
   - [ ] NA
- Documentation
   - [ ] Added new documentation
   - [ ] NA

### DIFF
--- a/cpp/src/branch_and_bound/branch_and_bound.cpp
+++ b/cpp/src/branch_and_bound/branch_and_bound.cpp
@@ -41,6 +41,7 @@
 #include <cstdio>
 #include <cstdlib>
 #include <limits>
+#include <list>
 #include <string>
 #include <vector>
 
@@ -654,13 +655,12 @@ void branch_and_bound_t<i_t, f_t>::set_solution_from_cpu_fj(f_t obj,
 // user space.
 template <typename i_t, typename f_t>
 void branch_and_bound_t<i_t, f_t>::set_solution_from_submip(
-  const std::vector<f_t>& solution,
-  const third_party_presolve_t<i_t, f_t>& presolver,
-  f_t fixrate,
-  f_t obj)
+  const std::vector<f_t>& solution, const third_party_presolve_t<i_t, f_t>& presolver, f_t fixrate)
 {
   std::vector<f_t> leaf_sol;
   presolver.uncrush_primal_solution(solution, leaf_sol);
+  f_t obj = compute_objective(original_lp_, leaf_sol);
+
   std::vector<f_t> user_sol;
   mutex_original_lp_.lock();
   uncrush_primal_solution(original_problem_, original_lp_, leaf_sol, user_sol);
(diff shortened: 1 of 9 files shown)


cpp/src/branch_and_bound/branch_and_bound.cpp
```diff
@@ -2766,8 +2800,68 @@ void branch_and_bound_t<i_t, f_t>::rins(diving_worker_t<i_t, f_t>* rins_worker,
...
+{
+  assert(worker_count);
+  if (settings_.deterministic || *worker_count >= settings_.num_threads - 1) return;
+
+  bool is_root_heuristic                 = true;
+  i_t id                                 = heuristics.size();
+  root_heuristics_t<i_t, f_t>& heuristic = heuristics.emplace_back(Arow_, var_types_);
+
+  if (settings_.submip_settings.rins != 0 && incumbent_.has_incumbent &&
+      (*worker_count < settings_.num_threads - 1)) {
```
akifcorduk, line 2827: `*worker_count < settings_.num_threads - 1` is unnecessary in the two if blocks, as we are already returning earlier.

cpp/src/branch_and_bound/branch_and_bound.cpp
```diff
@@ -2766,8 +2800,68 @@ void branch_and_bound_t<i_t, f_t>::rins(diving_worker_t<i_t, f_t>* rins_worker,
...
-  rins_worker_pool_.return_worker_to_pool(rins_worker);
+  if (!is_root_heuristic) {
+    rins_worker_pool_.return_worker_to_pool(worker);
+  } else {
+    worker->set_inactive();
+  }
+}
+
+template <typename i_t, typename f_t>
+void branch_and_bound_t<i_t, f_t>::launch_root_heuristics(
```
akifcorduk, line 2813: Do you think we can also add some diving workers here?

nguidotti: Yup. I will add then later (maybe in an another PR).

nguidotti: I will close this PR and convert into a PR stack

nguidotti:
Close this PR to transform into a PR stack

cpp/src/dual_simplex/solve.cpp
```diff
@@ -104,6 +104,12 @@ f_t compute_user_objective(const lp_problem_t<i_t, f_t>& lp, f_t obj)
   return user_obj;
 }
 
+template <typename i_t, typename f_t>
+f_t compute_solver_objective(const lp_problem_t<i_t, f_t>& lp, f_t user_obj)
```
chris-maes, line 108: Let's not use the term `solver_objective`

cpp/src/branch_and_bound/branch_and_bound.cpp
```diff
@@ -2766,8 +2800,68 @@ void branch_and_bound_t<i_t, f_t>::rins(diving_worker_t<i_t, f_t>* rins_worker,
...
+  const lp_problem_t<i_t, f_t>& lp,
+  const std::vector<f_t>& sol,
+  std::list<root_heuristics_t<i_t, f_t>>& heuristics,
```
chris-maes, line 2816: Do we need a list of heuristics here?

cpp/src/branch_and_bound/branch_and_bound.cpp
```diff
@@ -2730,21 +2751,33 @@ void branch_and_bound_t<i_t, f_t>::rins(diving_worker_t<i_t, f_t>* rins_worker,
...
+                                             worker->leaf_solution.x,
                                              settings_,
                                              std::format("{} [CPU FJ]", log_prefix),
-                                             rins_worker->rng.next_i64());
+                                             worker->rng.next_i64());
           submip_fj_cpu_worker.run_sync(time_limit, work_limit);
         }
 
-        dive_with(rins_worker, 5);
+        // We need the pseudocost to do the DFS, which we do not have during the cut passes.
```
chris-maes, line 2763: Why don't we do a fast psuedocost init based on a single dual simplex pivot.

cpp/src/dual_simplex/solve.cpp
nguidotti: Maybe "internal" objective?

cpp/src/branch_and_bound/branch_and_bound.cpp
nguidotti: It is either a list or a vector of `unique_ptr` since `fj_cpu_worker` is not copyable or movable. List seems the nicer option

cpp/src/branch_and_bound/branch_and_bound.cpp
nguidotti: It is a good suggestion. I will do this as another layer in the stack

PR #1698, opened 2026-08-11
[BUG] Fixed uninitialized FJ constraint weights when the problem grows

nguidotti:
When the feasibility pump adds the objective cutting plane to the problem, `fj.solve()` resizes the constraint weights, but leave the new weights unintialized. This trips the assertion that checks for infinite (and a negative) weights in the GPU FJ. The solution is to set the new weights to `1.0`, which matches `copy_weights` and `cstr_weights` fixup in `do_fj_solve`. Note that `fj.solve()` bypasses `do_fj_solve`.

## Checklist

- Testing
   - [ ] Added tests
   - [ ] NA
- Documentation
   - [ ] Added new documentation
   - [ ] NA


aliceb-nv (approved):
Thanks for the quick fix Nicolas!

akifcorduk (approved):

PR #1700, opened 2026-08-11
Root heuristics

nguidotti:
- Refactored the root heuristics (more specifically, the CPU FJ launches at the root node) to be more modular, while allowing to use more workers. 
- Extended recursive RINS to run on the root node
- Fixed incorrect objective passed to the `set_solution_from_submip` (the objective was on the submip space, which is not comparable with the solution space of the B&B).

MIPLIB 2017:
GH200, 10min

```
 main-2026-08-13 (1) vs root-heuristics (2)

| Imported                                                 240                  240                   +0                 --- |
| Feasible                                                 226                  225                   -1                 --- |
| Optimal                                                   87                   87                   +0                 --- |
| Solutions with <0.1% primal gap                          145                  143                   -2                 --- |
| Nodes explored (mean)                              1.306e+07            1.322e+07            +1.63e+05               +1.25 |
| Nodes explored (shifted geomean)                   1.113e+04                 8201                -2930               -26.3 |
| Relative MIP gap (mean)                               0.2626               0.2695            +0.006939               +2.64 |
| Relative MIP gap (shifted geomean)                   0.08603               0.0867           +0.0006709               +0.78 |
| Solve time (mean)                                      412.7                415.6               +2.922              +0.708 |
| Solve time (shifted geomean)                           182.3                179.6               -2.652               -1.46 |
| Primal gap (mean)                                      9.867                10.28              +0.4097               +4.15 |
| Primal gap (shifted geomean)                          0.3855               0.4062             +0.02071               +5.37 |
| Primal integral (mean)                                    22                22.11              +0.1057              +0.481 |
| Primal integral (shifted geomean)                      2.274                2.074              -0.1999               -8.79 |


----------------------------------------------------------------------
|             Name             |     status 1     |     status 2     |
----------------------------------------------------------------------
| app1-2                                  optimal           feasible |
| cbs-cta                                feasible            optimal |
| cmflsp50-24-8-8                         optimal           feasible |
| csched007                               optimal           feasible |
| map16715-04                            feasible            optimal |
| neos-1171737                           feasible            optimal |
| neos-5104907-jarama                    feasible            timeout |
----------------------------------------------------------------------
```

## Checklist

- Testing
   - [ ] Added tests
   - [ ] NA
- Documentation
   - [ ] Added new documentation
   - [ ] NA

### DIFF
--- a/cpp/src/branch_and_bound/branch_and_bound.cpp
+++ b/cpp/src/branch_and_bound/branch_and_bound.cpp
@@ -2200,14 +2200,12 @@ bool branch_and_bound_t<i_t, f_t>::launch_rins_worker(const std::vector<f_t>& so
   current_incumbent = incumbent_.x;
   mutex_upper_.unlock();
 
-  // Note that this node does not have the vstatus (it was clear at the start of B&B exploration)
+  // Note that this node does not have the vstatus (it was cleared at the start of B&B exploration)
   worker->start_node         = mip_node_t<i_t, f_t>(root_objective_, root_vstatus_);
   worker->leaf_vstatus       = root_vstatus_;
   worker->leaf_problem.lower = original_lp_.lower;
   worker->leaf_problem.upper = original_lp_.upper;
   worker->leaf_solution.x    = sol;
-  worker->recompute_bounds   = false;
-  worker->recompute_basis    = true;
   worker->search_strategy    = search_strategy_t::RINS;
   worker->set_active();
 
@@ -2227,7 +2225,7 @@ bool branch_and_bound_t<i_t, f_t>::launch_rins_worker(const std::vector<f_t>& so
 template <typename i_t, typename f_t>
 void branch_and_bound_t<i_t, f_t>::solve_submip(diving_worker_t<i_t, f_t>* worker,
                                                 const std::vector<f_t>& current_incumbent,
-                                                const std::vector<variable_type_t>& var_type,
+                                                const std::vector<variable_type_t>& var_types,
                                                 i_t num_var_fixed,
                                                 i_t num_integers,
                                                 i_t submip_level,
@@ -2309,7 +2307,7 @@ void branch_and_bound_t<i_t, f_t>::solve_submip(diving_worker_t<i_t, f_t>* worke
   // there is only equality rows (the range row vector is empty) and it contains
   // structural + slacks + cuts constraints/variables.
   user_problem_t<i_t, f_t> submip_problem(original_problem_.handle_ptr);
-  simplex::convert_lp_to_user_problem(worker->leaf_problem, var_type, settings_, submip_problem);
+  simplex::convert_lp_to_user_problem(worker->leaf_problem, var_types, settings_, submip_problem);
 
   third_party_presolve_t<i_t, f_t> presolver;
   f_t presolve_time_limit = std::min(0.1 * submip_settings.time_limit, 60.0);
@@ -2376,7 +2374,7 @@ void branch_and_bound_t<i_t, f_t>::solve_submip(diving_worker_t<i_t, f_t>* worke
   submip_bnb.set_initial_guess(presolved_incumbent);
 
   const f_t user_upper    = compute_user_objective(worker->leaf_problem, upper_bound_.load());
-  const f_t submip_cutoff = compute_internal_objective(submip_bnb.original_lp_, user_upper);
+  const f_t submip_cutoff = compute_presolved_objective(submip_bnb.original_lp_, user_upper);
   submip_bnb.set_initial_upper_bound(submip_cutoff);
 
   if (!is_root_heuristic)
@@ -2595,7 +2593,8 @@ void branch_and_bound_t<i_t, f_t>::rins(diving_worker_t<i_t, f_t>* worker,
   ++rins_stats_.total_calls;
 
   bool has_submip          = false;
-  const f_t abs_fathom_tol = settings_.absolute_mip_gap_tol / 10;
+  worker->recompute_bounds = false;
+  worker->recompute_basis  = true;
 
   branch_and_bound_stats_t<i_t, f_t> rins_stats;
   mip_node_t<i_t, f_t>& node        = worker->start_node;
@@ -2778,7 +2777,6 @@ void branch_and_bound_t<i_t, f_t>::rins(diving_worker_t<i_t, f_t>* worker,
         // We need the pseudocost to do the DFS, which we do not have during the cut passes.
         if (!is_root_heuristic) {
           DEBUG_SUBMIP("{} Running a quick DFS for the submip!", log_prefix);
-
           dive_with(worker, settings_.submip_settings.dfs_max_backtrack);
         }
       }
@@ -2836,7 +2834,8 @@ void branch_and_bound_t<i_t, f_t>::launch_root_heuristics(
   // oldest set of heuristics launched. Leave 2 threads for the cut passes and the clique
   // table generation. Add the number of workers that will be launched (1 submip worker +
   // 1 CPU FJ worker).
-  if (worker_count + 4 > settings_.num_threads && !heuristics.empty()) {
+  i_t clique_table_generation = cut_pass == 0 ? 1 : 0;
+  if (worker_count + 3 + clique_table_generation > settings_.num_threads && !heuristics.empty()) {
     heuristics.erase(heuristics.begin());
   }
 
@@ -3663,7 +3662,7 @@ mip_status_t branch_and_bound_t<i_t, f_t>::solve(mip_solution_t<i_t, f_t>& solut
     settings_.benchmark_info_ptr->cut_generation_time_sec = cut_generation_time;
   }
   if (cut_info.has_cuts()) {
-    settings_.log.printf("Cut generation time: %.2f seconds\n", cut_generation_time);
+    settings_.log.printf("Root cut passes time: %.2f seconds\n", cut_generation_time);
     settings_.log.printf("Cut pool size  : %d\n", cut_pool_size);
     settings_.log.printf("Size with cuts : %d constraints, %d variables, %d nonzeros\n",
                          original_lp_.num_rows,
--- a/cpp/src/branch_and_bound/branch_and_bound.cpp
+++ b/cpp/src/branch_and_bound/branch_and_bound.cpp
@@ -1005,7 +1005,7 @@ branch_variable_t<i_t> branch_and_bound_t<i_t, f_t>::variable_selection(
         branch_var = pc_.variable_selection(fractional, solution);
       }
 
-      round_dir = martin_criteria(solution[branch_var], root_relax_soln_.x[branch_var]);
+      round_dir = martin_criteria(solution[branch_var], worker->root_solution[branch_var]);
 
       return {branch_var, round_dir};
 
@@ -1014,10 +1014,10 @@ branch_variable_t<i_t> branch_and_bound_t<i_t, f_t>::variable_selection(
         original_lp_, fractional, solution, var_up_locks_, var_down_locks_, log);
 
     case search_strategy_t::LINE_SEARCH_DIVING:
-      return line_search_diving(fractional, solution, root_relax_soln_.x, log);
+      return line_search_diving(fractional, solution, worker->root_solution, log);
 
     case search_strategy_t::PSEUDOCOST_DIVING:
-      return pseudocost_diving(pc_, fractional, solution, root_relax_soln_.x, log);
+      return pseudocost_diving(pc_, fractional, solution, worker->root_solution, log);
 
     case search_strategy_t::GUIDED_DIVING:
       assert(incumbent_.has_incumbent);
@@ -1034,7 +1034,7 @@ branch_variable_t<i_t> branch_and_bound_t<i_t, f_t>::variable_selection(
 
     case search_strategy_t::RINS:  // This is used for solving the DFS of the sub-MIP.
       branch_var = pc_.variable_selection(fractional, solution);
-      round_dir  = martin_criteria(solution[branch_var], root_relax_soln_.x[branch_var]);
+      round_dir  = martin_criteria(solution[branch_var], worker->root_solution[branch_var]);
       return {branch_var, round_dir};
   }
 
@@ -1187,7 +1187,7 @@ struct deterministic_bfs_policy_t
                                                 const std::vector<f_t>& x) override
   {
     i_t var  = this->worker.pc_snapshot.variable_selection(fractional, x);
-    auto dir = martin_criteria(x[var], this->bnb.root_relax_soln_.x[var]);
+    auto dir = martin_criteria(x[var], this->worker.root_solution[var]);
     return {var, dir};
   }
 
@@ -1523,7 +1523,8 @@ dual_status_t branch_and_bound_t<i_t, f_t>::solve_node_lp(
   branch_and_bound_worker_t<i_t, f_t>* worker,
   branch_and_bound_stats_t<i_t, f_t>& stats,
   logger_t& log,
-  i_t iter_limit)
+  i_t iter_limit,
+  std::atomic<int>* halt)
 {
   raft::common::nvtx::range scope("BB::solve_node");
 #ifdef DEBUG_BRANCHING
@@ -1562,7 +1563,7 @@ dual_status_t branch_and_bound_t<i_t, f_t>::solve_node_lp(
 #endif
 
   simplex_solver_settings_t lp_settings = settings_;
-  lp_settings.concurrent_halt           = &node_concurrent_halt_;
+  lp_settings.concurrent_halt           = halt ? halt : &node_concurrent_halt_;
   lp_settings.set_log(false);
   f_t cutoff = upper_bound_.load();
   if (original_lp_.objective_step.has_step()) {
@@ -1609,7 +1610,7 @@ dual_status_t branch_and_bound_t<i_t, f_t>::solve_node_lp(
 
   bool feasible           = worker->set_lp_variable_bounds(node_ptr, settings_);
   dual_status_t lp_status = dual_status_t::DUAL_UNBOUNDED;
-  worker->leaf_edge_norms = edge_norms_;
+  worker->leaf_edge_norms = worker->root_edge_norm;
   if (worker->recompute_bounds && worker->orbital_fixing &&
       worker->search_strategy == search_strategy_t::BEST_FIRST) {
     worker->orbital_fixing->reset(symmetry_, node_ptr);
@@ -2352,6 +2353,13 @@ void branch_and_bound_t<i_t, f_t>::solve_submip(diving_worker_t<i_t, f_t>* worke
     return;
   }
 
+  if (toc(exploration_stats_.start_time) > settings_.time_limit) {
+    solver_status_ = mip_status_t::TIME_LIMIT;
+    return;
+  }
+
+  if (halt && halt->load(std::memory_order::acquire)) { return; }
+
   submip_settings.heuristic_preemption_callback   = nullptr;
   submip_settings.dual_simplex_objective_callback = nullptr;
   submip_settings.set_simplex_solution_callback   = nullptr;
@@ -2621,7 +2629,8 @@ void branch_and_bound_t<i_t, f_t>::rins(diving_worker_t<i_t, f_t>* worker,
   i_t min_var_fixed = min_fixrate * num_integers;
   i_t num_var_fixed = 0;
 
-  while (solver_status_ == mip_status_t::UNSET && is_running_ && !this->is_halted()) {
+  while (solver_status_ == mip_status_t::UNSET && is_running_ && !this->is_halted() &&
+         (halt ? !halt->load() : true)) {
     // RINS neighbourhood 1: Fix all the integer variables where the starting solution matches the
     // current incumbent, considering only the fractional values in the current node
     i_t prev_num_fixed = num_var_fixed;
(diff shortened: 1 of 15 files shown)


cpp/src/branch_and_bound/branch_and_bound.cpp
```diff
@@ -2730,21 +2764,34 @@ void branch_and_bound_t<i_t, f_t>::rins(diving_worker_t<i_t, f_t>* rins_worker,
...
+                                             worker->rng.next_i64());
           submip_fj_cpu_worker.run_sync(time_limit, work_limit);
         }
 
-        dive_with(rins_worker, 5);
+        // We need the pseudocost to do the DFS, which we do not have during the cut passes.
+        if (!is_root_heuristic) {
+#ifdef DEBUG_SUBMIP
+          settings_.log.print_format("{} Running a quick DFS for the submip!", log_prefix);
+#endif
```
aliceb-nv: Any reason why we're not using debug_format here? Reducing #ifdef clutter is good for readability IMO

cpp/src/branch_and_bound/branch_and_bound.cpp
```diff
@@ -2220,10 +2249,15 @@ void branch_and_bound_t<i_t, f_t>::solve_submip(diving_worker_t<i_t, f_t>* worke
...
   submip_settings.strong_branching_simplex_iteration_limit = 50;
   submip_settings.submip_settings.level                    = submip_level;
-  submip_settings.log.log                                  = false;
   submip_settings.benchmark_info_ptr                       = nullptr;
 
 #ifdef DEBUG_SUBMIP
+  submip_settings.log.log = true;
+#else
+  submip_settings.log.log = false;
+#endif
```
aliceb-nv: Seems like leftover debug code right? If possible let's ensure such code is out of major function bodies

aliceb-nv (approved):
LGTM, minor nits, thanks Nicolas! As always, let's get Chris' eyes on this before the merge

cpp/src/branch_and_bound/branch_and_bound.cpp
nguidotti: This is mostly to enable/disable the logs from sub-MIP quickly. With log set to `debug`, things can get pretty chaotic hahah
(maybe we could have a way to enable/disable the debug logs from each part of the solver separately)

cpp/src/branch_and_bound/branch_and_bound.cpp
nguidotti: Same as above. Without the `DEBUG_SUBMIP`, the sub-MIP logs should be silent.

cpp/src/branch_and_bound/branch_and_bound.cpp
aliceb-nv: Maybe not for this PR, but perhaps we could later add a scope/type argument to the debug logs to allow for more fine-grained filtering :)

nguidotti: Wrapped the submip logs calls with a macro

cpp/src/branch_and_bound/branch_and_bound.cpp
```diff
@@ -215,8 +224,10 @@ inline char feasible_solution_symbol(search_strategy_t strategy, bool show_divin
     case search_strategy_t::GUIDED_DIVING: return 'G';
     case search_strategy_t::FARKAS_DIVING: return 'F';
     case search_strategy_t::VECTOR_LENGTH_DIVING: return 'V';
-    default: return 'U';
+    case search_strategy_t::RINS: return 'S';
```
akifcorduk, line 227: Why S and not R?

cpp/src/branch_and_bound/branch_and_bound.cpp
```diff
@@ -2164,19 +2178,34 @@ bool branch_and_bound_t<i_t, f_t>::launch_rins_worker(const std::vector<f_t>& so
...
 
   if (settings_.inside_submip) {
     // LLVM libomp's GOMP compatibility path skips GCC's firstprivate copy
     // function for included tasks.
-    rins(worker, sol);
+    rins(worker, current_incumbent, var_types_, is_root_heuristic);
   } else {
-#pragma omp task priority(CUOPT_DEFAULT_TASK_PRIORITY) affinity(worker) firstprivate(worker, sol)
-    rins(worker, sol);
+#pragma omp task priority(CUOPT_DEFAULT_TASK_PRIORITY) affinity(worker) \
```
akifcorduk, line 2216: Shouldn't we check if there are enough threads available at this point?

cpp/src/branch_and_bound/branch_and_bound.cpp
```diff
@@ -2767,7 +2801,89 @@ void branch_and_bound_t<i_t, f_t>::rins(diving_worker_t<i_t, f_t>* rins_worker,
...
+  i_t cut_pass,
+  std::list<root_heuristics_t<i_t, f_t>>& heuristics,
+  omp_atomic_t<i_t>& worker_count)
+{
+  if (settings_.deterministic) return;
+  if (settings_.num_threads < 2) return;
+
+  // If we already exhausted all threads for the root heuristics, stop workers for the
+  // oldest set of heuristics launched. Leave 2 threads for the cut passes and the clique
+  // table generation. Add the number of workers that will be launched (1 submip worker +
```
akifcorduk: Clique table generation only happens in the first cut pass. So later it is freed. It complicates the logic, but in case you want to use that thread too :)

cpp/src/branch_and_bound/branch_and_bound.cpp
```diff
@@ -2767,7 +2801,89 @@ void branch_and_bound_t<i_t, f_t>::rins(diving_worker_t<i_t, f_t>* rins_worker,
...
+    diving_worker_t<i_t, f_t>* worker =
+      heuristic->create_submip_worker(id, lp, settings_, root_objective_, root_vstatus_, sol);
+
+    std::vector<f_t> current_incumbent;
+    mutex_upper_.lock();
+    current_incumbent = incumbent_.x;
+    mutex_upper_.unlock();
+
+    if (settings_.inside_submip) {
+      // LLVM libomp's GOMP compatibility path skips GCC's firstprivate copy
```
akifcorduk, line 2876: Is this comment still necessary?

akifcorduk (approved):
Thanks Nicolas, just a few nitpicks.

cpp/src/branch_and_bound/branch_and_bound.cpp
aliceb-nv: Iirc this is because LLVM has an upstream bug where they don't honor the firstprivate copy constructor in their GCC OpenMP compat layer. It caused issues a few weeks back because some wheel builds had to move to LLVM's openmp lib due to the GCC openmp in their environment being outdated

cpp/src/branch_and_bound/branch_and_bound.cpp
nguidotti: I used `S` for `SUBMIP`, but I guess `R` is also valid

cpp/src/branch_and_bound/branch_and_bound.cpp
nguidotti: This is controlled by the `submip` worker pool. If no worker is available, then we exit early (see the if some lines prior). Although we may need a better way to control the total number of threads, each level of the sub-MIP recursion can use more than one thread

cpp/src/branch_and_bound/branch_and_bound.cpp
nguidotti: Should we switching to LLVM's openmp lib completely? Or even use clang?

aliceb-nv: Probably a P2 task for much later :) Although it may be worth benchmarking GOMP vs llvmomp

cpp/src/branch_and_bound/branch_and_bound.cpp
```diff
@@ -2164,19 +2178,34 @@ bool branch_and_bound_t<i_t, f_t>::launch_rins_worker(const std::vector<f_t>& so
...
+  bool is_root_heuristic            = false;
   diving_worker_t<i_t, f_t>* worker = rins_worker_pool_.pop_idle_worker();
   if (!worker) return false;
 
+  std::vector<f_t> current_incumbent;
+  mutex_upper_.lock();
+  current_incumbent = incumbent_.x;
+  mutex_upper_.unlock();
+
+  // Note that this node does not have the vstatus (it was clear at the start of B&B exploration)
```
chris-maes: Nit: `it was clear at the start` -> `it was cleared at the start`

cpp/src/branch_and_bound/branch_and_bound.cpp
```diff
@@ -2185,10 +2214,12 @@ bool branch_and_bound_t<i_t, f_t>::launch_rins_worker(const std::vector<f_t>& so
 template <typename i_t, typename f_t>
 void branch_and_bound_t<i_t, f_t>::solve_submip(diving_worker_t<i_t, f_t>* worker,
                                                 const std::vector<f_t>& current_incumbent,
+                                                const std::vector<variable_type_t>& var_type,
```
chris-maes: Nit: `var_type` -> `var_types`. As this is the types of the variables.

cpp/src/branch_and_bound/branch_and_bound.cpp
```diff
@@ -2325,24 +2362,24 @@ void branch_and_bound_t<i_t, f_t>::solve_submip(diving_worker_t<i_t, f_t>* worke
   presolver.crush_primal_solution(submip_problem, current_incumbent, presolved_incumbent);
   submip_bnb.set_initial_guess(presolved_incumbent);
 
-  const f_t user_upper = compute_user_objective(original_lp_, upper_bound_.load());
-  const f_t submip_cutoff =
-    user_upper / submip_bnb.original_lp_.obj_scale - submip_bnb.original_lp_.obj_constant;
+  const f_t user_upper    = compute_user_objective(worker->leaf_problem, upper_bound_.load());
+  const f_t submip_cutoff = compute_internal_objective(submip_bnb.original_lp_, user_upper);
```
chris-maes: Ah. Not another different name `internal_objective`. Let's try to standardize on names:
1) user objective for the objective associated with the original problem
2) presolved objective for the objective associated with the presolved problem. 

Let's try not to introduce more words for the same things.

cpp/src/branch_and_bound/branch_and_bound.cpp
```diff
@@ -2730,21 +2753,32 @@ void branch_and_bound_t<i_t, f_t>::rins(diving_worker_t<i_t, f_t>* rins_worker,
...
+                                             worker->leaf_solution.x,
                                              settings_,
                                              std::format("{} [CPU FJ]", log_prefix),
-                                             rins_worker->rng.next_i64());
+                                             worker->rng.next_i64());
           submip_fj_cpu_worker.run_sync(time_limit, work_limit);
         }
 
-        dive_with(rins_worker, 5);
+        // We need the pseudocost to do the DFS, which we do not have during the cut passes.
```
chris-maes, line 2781: What are you doing if you don't have the psuedocosts?

cpp/src/branch_and_bound/branch_and_bound.cpp
```diff
@@ -2767,7 +2801,89 @@ void branch_and_bound_t<i_t, f_t>::rins(diving_worker_t<i_t, f_t>* rins_worker,
...
+
+  bool is_root_heuristic                 = true;
+  i_t id                                 = cut_pass;
+  root_heuristics_t<i_t, f_t>* heuristic = &heuristics.emplace_back(Arow_, var_types_);
+
+  constexpr bool is_cpufj_enabled = true;
+  if (is_cpufj_enabled) {
+    fj_cpu_worker_t<i_t, f_t>* worker = &heuristic->fj_cpu_worker_;
+
+    f_t work_limit = 2.0;
```
chris-maes: What work_limit is this? That is what does 2.0 represent?

cpp/src/branch_and_bound/branch_and_bound.cpp
```diff
@@ -3485,8 +3595,6 @@ mip_status_t branch_and_bound_t<i_t, f_t>::solve(mip_solution_t<i_t, f_t>& solut
     }
 
     cut_pass_result_t cut_pass_result;
-    root_fj_cpu_worker.run_async(settings_.time_limit - toc(exploration_stats_.start_time));
-
     cut_pass_result = do_cut_pass(cut_pass,
```
chris-maes, line 3601: Are new sub-MIP heuristics being launched in the cut passes? It's hard to tell from the PR.

chris-maes: Oh. I see. They are launched below.

cpp/src/branch_and_bound/branch_and_bound.cpp
```diff
@@ -3541,18 +3650,6 @@ mip_status_t branch_and_bound_t<i_t, f_t>::solve(mip_solution_t<i_t, f_t>& solut
...
-    if (incumbent_.has_incumbent && incumbent_.x.size() != original_lp_.num_cols) {
-      std::vector<f_t> uncrushed_incumbent;
-      uncrush_primal_solution(original_problem_, original_lp_, incumbent_.x, uncrushed_incumbent);
-      crush_primal_solution(
-        original_problem_, original_lp_, uncrushed_incumbent, new_slacks_, incumbent_.x);
-    }
-
-    mutex_upper_.unlock();
-
     settings_.log.printf("Cut generation time: %.2f seconds\n", cut_generation_time);
```
chris-maes: This isn't Cut generation time. This is time for all the root cut pass. Should probably read:
"Root cut passes time: 10.7 seconds"

cpp/src/branch_and_bound/branch_and_bound.hpp
```diff
@@ -368,24 +370,32 @@ class branch_and_bound_t {
...
+                                const std::vector<f_t>& solution,
                                 const third_party_presolve_t<i_t, f_t>& presolver,
-                                f_t fixrate,
-                                f_t obj);
+                                f_t fixrate);
 
   // Solve the RINS sub-MIP
   void solve_submip(diving_worker_t<i_t, f_t>* worker,
                     const std::vector<f_t>& current_incumbent,
+                    const std::vector<simplex::variable_type_t>& var_type,
```
chris-maes: Nit: `var_type` -> `var_types`

cpp/src/branch_and_bound/branch_and_bound.hpp
```diff
@@ -368,24 +370,32 @@ class branch_and_bound_t {
...
                     i_t num_var_fixed,
                     i_t num_integers,
                     i_t submip_level,
-                    std::string_view log_prefix);
+                    std::string_view log_prefix,
+                    bool is_root_heuristic);
```
chris-maes: Can we eliminate the need for the `is_root_heuristics`? Maybe doing something at the root, before we call submip?

cpp/src/branch_and_bound/constants.hpp
```diff
@@ -27,15 +27,31 @@ enum class heuristics_origin_t {
...
+{
+  switch (search_strategy) {
+    case search_strategy_t::BEST_FIRST: return "BEST_FIRST";
+    case search_strategy_t::PSEUDOCOST_DIVING: return "PSEUDOCOST_DIVING";
+    case search_strategy_t::LINE_SEARCH_DIVING: return "LINE_SEARCH_DIVING";
+    case search_strategy_t::GUIDED_DIVING: return "GUIDED_DIVING";
+    case search_strategy_t::COEFFICIENT_DIVING: return "COEFFICIENT_DIVING";
+    case search_strategy_t::FARKAS_DIVING: return "FARKAS_DIVING";
+    case search_strategy_t::VECTOR_LENGTH_DIVING: return "VECTOR_LENGTH_DIVING";
+    case search_strategy_t::RINS: return "RINS";
```
chris-maes, line 51: If you are going to call the search strategy "RINS" you should probably print 'R' in the logs.

cpp/src/branch_and_bound/worker.hpp
```diff
@@ -243,6 +243,8 @@ class diving_worker_t : public branch_and_bound_worker_t<i_t, f_t> {
   // The best-first worker that is associated with this diving worker. Used for controlling the
   // number of active diving workers.
   bfs_worker_t<i_t, f_t>* bfs_worker{nullptr};
+
+  std::atomic<int> halt = false;
```
chris-maes, line 257: We can't use the halt in simplex_solver_settings? I'm not sure it's good to have too many different ways to halt solves.

cpp/src/dual_simplex/solve.cpp
```diff
@@ -104,6 +104,12 @@ f_t compute_user_objective(const lp_problem_t<i_t, f_t>& lp, f_t obj)
   return user_obj;
 }
 
+template <typename i_t, typename f_t>
+f_t compute_internal_objective(const lp_problem_t<i_t, f_t>& lp, f_t user_obj)
```
chris-maes: See comment above. Let's not introduce new names.

cpp/src/dual_simplex/solve.hpp
```diff
@@ -63,6 +63,9 @@ f_t compute_user_objective(const lp_problem_t<i_t, f_t>& lp, const std::vector<f
 template <typename i_t, typename f_t>
 f_t compute_user_objective(const lp_problem_t<i_t, f_t>& lp, f_t obj);
 
+template <typename i_t, typename f_t>
+f_t compute_internal_objective(const lp_problem_t<i_t, f_t>& lp, f_t user_obj);
```
chris-maes: See comment above. Let's not introduce new names.

chris-maes (changes requested):
Overall looks good. The only two things I'd like to avoid before merging is:
1) introducing a new "internal_objective". 
2) introducing a new way to halt 

Other than those two issue. The only other thing that stands out is the special case code if we are calling sub-mip from the root. It might be worth quickly getting some psuedocost estimates (say from a single dual simplex pivot) so that you don't need this special case.

cpp/src/branch_and_bound/branch_and_bound.cpp
nguidotti: Renamed in [9e79502](https://github.com/NVIDIA/cuopt/pull/1700/commits/9e79502cc9b4cb4e80976261ae822b4b8467da7e)

cpp/src/branch_and_bound/branch_and_bound.cpp
nguidotti: @aliceb-nv for viz

nguidotti: `2.0` is an arbitrary number (Alice set it to `0.5` in the previous version). With that said, we are now stopping the CPU FJ after the root passes anyway, so maybe this is limit is not needed.

cpp/src/branch_and_bound/constants.hpp
nguidotti: I renamed `SUBMIP` to `RINS`, so it is distinguish from `RENS`. Both shares the same code path with some minor differences (I use this flag to select the path to take)

cpp/src/branch_and_bound/worker.hpp
nguidotti: Unfortunately it is not as simple:
1. Each submip gets it own `settings` that is local to `solve_submip`. So it not accessible from outside.
2. Let say we use the `concurrent_halt_`, then it also stops the simplex for the cuts.
3. We may also want to stop some of the sub-MIP, but not all. For instances, if the number of threads is limited, then we want to stop just the oldest sub-MIP when launching a new one (effectively replacing it with a newer version)

cpp/src/branch_and_bound/branch_and_bound.cpp
nguidotti: We only run the CPU FJ, but not the DFS for the sub-MIP.

cpp/src/branch_and_bound/branch_and_bound.cpp
chris-maes: Prefer 'R' if you are going to call the search strategy RINS

cpp/src/branch_and_bound/branch_and_bound.cpp
```diff
@@ -2338,24 +2376,24 @@ void branch_and_bound_t<i_t, f_t>::solve_submip(diving_worker_t<i_t, f_t>* worke
   presolver.crush_primal_solution(submip_problem, current_incumbent, presolved_incumbent);
   submip_bnb.set_initial_guess(presolved_incumbent);
 
-  const f_t user_upper = compute_user_objective(original_lp_, upper_bound_.load());
-  const f_t submip_cutoff =
-    user_upper / submip_bnb.original_lp_.obj_scale - submip_bnb.original_lp_.obj_constant;
+  const f_t user_upper    = compute_user_objective(worker->leaf_problem, upper_bound_.load());
+  const f_t submip_cutoff = compute_presolved_objective(submip_bnb.original_lp_, user_upper);
```
chris-maes, line 2381: Why do you need `compute_presolve_objective` here? 
Shouldn't this just be
`const f_t submip_cutoff = upper_bound_.load()` ?

chris-maes (approved):
Overall looks good. I'm still not sure I understand the purpose of the new function `compute_presolved_objective`? Is it somehow trying to compensate for different objective constants between the original problem and the sub-MIP problem? 

If so, maybe this should be: `compute_submip_objective`?

cpp/src/branch_and_bound/branch_and_bound.cpp
nguidotti: No. Since we are apply presolve, the solution space in the sub-MIP is different from the solution of the parent B&B (Papilo may change the object offset). Hence, we need to crush the objective into the sub-MIP space so it is a valid cutoff.

nguidotti: We will have a similar issue when implementing restarts

PR #1701, opened 2026-08-11
Describe solver settings and defaults in the gRPC codegen field registry

ramakrishnap-nv:
Adds `description:` and `default:` to `field_registry.yaml` for all 85 settings fields across `pdlp_settings`/`mip_settings` — previously nothing tied a setting's meaning and default to the field itself (`constants.h` is bare `#define`s, `docs/*.rst` keys prose to C API names, not fields). `default:` values come from the C++ member initializers, not the docs, which disagree in several places.

Also adds two checks that were previously missing:
- `./build.sh codegen` now fails if a non-optional settings field's `default:` doesn't textually match the proto3 zero value — the exact class of bug a reviewer would otherwise have to catch by eye.
- `cpp/src/grpc/codegen/lint_registry_defaults.py` — manual, best-effort — diffs simple C++ initializers against the registry to catch drift the check above can't (a changed default with no matching registry update).

And `skills/cuopt-developer/references/grpc_codegen.md`, covering the codegen pipeline so an agent doesn't hand-edit `generated/` or skip regeneration.

## How this reaches the MCP server

This PR only touches `field_registry.yaml`/`generate_conversions.py`; no MCP code lives here. But `description:`/`default:` have a live consumer once #1916 lands on top of this branch:

```
field_registry.yaml (this PR)
        |
        v
generate_conversions.py  (./build.sh codegen)
        |
        +--> cuopt_remote_data.proto, generated_*.inc --> libcuopt / cuopt_grpc_server
        |         (wire types, C++ <-> proto converters)
        |
        +--> cuopt_mcp_schema.json --> cuopt_mcp (#1916) --> MCP client
                  (JSON Schema `description`/`default` an agent reads
                   before deciding whether to set a solver parameter)
```

So a `description:`/`default:` edit here reaches an MCP agent's tool schema through `cuopt_mcp_schema.json`, with no MCP-specific work needed on that side — which is also why the optional-vs-zero-value check in this PR matters even without cuopt_mcp in view: a wrong default here is a wrong default an agent sees.


cpp/src/grpc/codegen/generated/cuopt_remote_data.proto
```diff
@@ -161,94 +161,329 @@ message OptimizationProblem {
...
   optional int64 iteration_limit = 10;
+  // Whether the solver writes log output to the console. Logs may still be
+  // written to a file when this is false. (default: true)
   optional bool log_to_console = 11;
+  // Whether PDLP attempts to detect infeasibility. Detection is not always
+  // accurate — some problems reported infeasible converge under a different
+  // tolerance — and costs roughly 3-7% runtime and 10-20% memory. Dual
+  // simplex always detects infeasibility regardless of this setting.
+  // (default: false)
   bool detect_infeasibility = 12;
```
mlubin, line 175: (pre-existing) Is the bool not optional in the proto just because the default is false? What happens if/when the default changes?

cpp/src/grpc/codegen/field_registry.yaml
```diff
@@ -581,52 +761,103 @@ mip_settings:
...
+        description: >-
+          How close to an integer a variable must be to count as integral.
+        default: "1e-5"
         field_num: 4
         optional: true
     - absolute_tolerance:
+        description: MIP absolute tolerance.
+        default: "1e-6"
         field_num: 5
         optional: true
```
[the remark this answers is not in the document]
ramakrishnap-nv: Fixed in dd074cda — both absolute_tolerance and relative_tolerance now describe the constraint-feasibility check (`get_cstr_tolerance`: abs_tol + rel_tol * bound_magnitude) and call out that this is distinct from absolute_mip_gap/relative_mip_gap, which control solve termination. Regenerating this also surfaced a real bug — `_json_schema_property` still unpacked `parse_enum_values()` as 2-tuples while every other call site expects the 3-tuple `parse_enum_entry` actually returns — fixed in the same commit.

cpp/src/grpc/codegen/generate_conversions.py
```diff
@@ -334,6 +334,46 @@ def _array_wire_type_comment(f):
...
+    for word in body.split():
+        candidate = f"{current} {word}" if current else word
+        if current and len(candidate) > width:
+            lines.append(f"{prefix}{current}")
+            current = word
+        else:
+            current = candidate
+    if current:
+        lines.append(f"{prefix}{current}")
+    return lines
```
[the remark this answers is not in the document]
ramakrishnap-nv: This one's stale: the doc-comment-in-proto feature this refers to (`_field_doc_comment`) was removed by a later commit in this same PR (5846d9ba, "Drop proto comments; the MCP schema is the real consumer") before this review ran against an intermediate state. There's nothing left to add tests for — the MCP schema (the actual consumer now) is covered by test_schema.py.

skills/cuopt-developer/SKILL.md
```diff
@@ -167,6 +167,18 @@ cuopt/
 - Never suggest `--no-verify` or skipping checks
 - All PRs must pass CI
 
+### Never Hand-Edit Generated Files
+Edit the source, run the generator, commit both. A generated file usually says so in its first line or two — check before editing anything unfamiliar.
+
+| Generated | Source | Regenerate with |
+|-----------|--------|-----------------|
+| `cpp/src/grpc/codegen/generated/` | `cpp/src/grpc/codegen/field_registry.yaml` | `./build.sh codegen` |
+| `conda/environments/*.yaml`, `pyproject.toml` | `dependencies.yaml` | `pre-commit run --all-files` |
```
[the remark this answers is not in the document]
ramakrishnap-nv: Fixed in 5e84bf61 — added generate_conversions.py to the source column. Left the dependencies.yaml row out of the suggested diff since that's a pre-existing, unrelated table row.

cpp/src/grpc/codegen/generated/cuopt_remote_data.proto
ramakrishnap-nv: Yes — that's exactly the rule (documented in references/grpc_codegen.md's `optional` bullet): it's only added when the C++ default differs from the proto3 zero value. detect_infeasibility defaults to false, matching proto3's implicit zero, so an omitting client already gets the right default without presence tracking.

If the default ever flips to true, optional would become required — otherwise an omitting client would silently get false (the wire zero value) instead of the new true default. Worth flagging: nothing currently checks that automatically. description/default in the registry are free text the generator doesn't validate against the C++ initializer (noted in the same doc), so a future default change without a matching optional: true addition would be a silent correctness bug, not a caught one. Filing this as a follow-up for the registry docs rather than fixing here, since it's pre-existing and out of scope for this PR.

cpp/src/grpc/codegen/field_registry.yaml
```diff
@@ -38,6 +38,30 @@
...
+#   description        – human-readable meaning of the field. Emitted into
+#                          cuopt_mcp_schema.json as the JSON Schema
+#                          `description`, which is what an MCP client shows a
+#                          model deciding whether to set the field. Consumed
+#                          for settings sections only.
+#   default            – what the field does when omitted, written as a string
+#                          describing the C++ member initializer (e.g. "1e-4",
+#                          "-1 (automatic)"). Appended to the emitted
+#                          description as "Default: ...". Documentation only —
+#                          the generator neither derives nor validates it, and
```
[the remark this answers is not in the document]
ramakrishnap-nv: Fixed in 1c9ab7ca — corrected to describe the partial validation (optional-vs-zero-value check) added in 179f1e95, instead of the now-stale 'neither derives nor validates' claim.

cpp/src/grpc/codegen/generate_conversions.py
```diff
@@ -339,6 +340,131 @@ def _array_wire_type_comment(f):
...
+        # A model must express "no limit" by omitting the field, never by
+        # sending the reserved value.
+        prop["description"] = (
+            prop.get("description", "") + " Omit for the default limit."
+        ).strip()
+    return prop
+
+
+def generate_mcp_schema(registry):
+    """Build the MCP tool-input JSON Schema for the solver settings.
```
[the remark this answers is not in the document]
ramakrishnap-nv: Fixed in 1c9ab7ca — added type hints and Args/Returns/Raises.

cpp/src/grpc/codegen/lint_registry_defaults.py
```diff
@@ -0,0 +1,175 @@
...
+            else f"C++ default is {cpp_value}, default: doesn't match"
+        )
+
+    cpp_num = _NUMBER_RE.search(cpp_value)
+    if cpp_num is None:
+        return None  # unrecognized expression (e.g. a function call) -- skip
+    reg_num = _NUMBER_RE.search(reg_lower)
+    if reg_num is None:
+        return None  # default: is pure prose (e.g. context-dependent) -- skip
+    if abs(float(cpp_num.group()) - float(reg_num.group())) > 1e-12:
```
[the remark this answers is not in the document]
ramakrishnap-nv: Fixed in 1c9ab7ca — switched to exact Decimal equality (still treats '1.0e-4' == '1e-4', but no longer misses e.g. 1e-13 vs 0).

python/cuopt_mcp/cuopt_mcp/__init__.py
```diff
@@ -0,0 +1,13 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""MCP server for NVIDIA cuOpt."""
+
+__all__ = ["main"]
+
+
+def main():
+    """Console-script entry point (``cuopt-mcp``)."""
```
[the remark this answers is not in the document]
ramakrishnap-nv: Fixed in 1c9ab7ca.

python/cuopt_mcp/cuopt_mcp/client.py
```diff
@@ -0,0 +1,86 @@
...
+
+    return TlsConfig(
+        root_certs=os.environ.get("CUOPT_TLS_ROOT_CERT"),
+        client_cert=os.environ.get("CUOPT_TLS_CLIENT_CERT"),
+        client_key=os.environ.get("CUOPT_TLS_CLIENT_KEY"),
+    )
+
+
+def get_client():
+    """Return a process-wide gRPC client, connecting on first use.
```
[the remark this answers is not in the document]
ramakrishnap-nv: Fixed in ff12e300 — type hints (using TYPE_CHECKING for the lazily-imported Client type) and docstrings.

python/cuopt_mcp/cuopt_mcp/client.py
```diff
@@ -0,0 +1,86 @@
...
+def describe_connection_error(exc: Exception) -> CuOptMCPError:
+    host, port = endpoint()
+    text = str(exc)
+    if "UNAVAILABLE" in text or "failed to connect" in text.lower():
+        return CuOptMCPError(
+            f"cuOpt gRPC server unreachable at {host}:{port}. Start it with "
+            f"`cuopt_grpc_server --port {port}`, or set CUOPT_REMOTE_HOST / "
+            "CUOPT_REMOTE_PORT to point at a running server."
+        )
+    return CuOptMCPError(text)
```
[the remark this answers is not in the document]
ramakrishnap-nv: Real and confirmed reachable: grpc_service_impl.cpp:789 raises 'Failed to open log file: ' + a server-derived path, which would flow through describe_connection_error unmodified. Fixed in ff12e300 by redacting filesystem paths from non-connectivity error text before it reaches the caller, rather than suppressing error detail wholesale — the client only ever gets a plain message string (no structured status code survives the C++ -> Cython -> here hop), so a caller-facing solve/parse error and a server-internal one can't be told apart by type, and blanket-genericizing all backend errors would break the tool's core diagnostic value.

python/cuopt_mcp/cuopt_mcp/schema.py
```diff
@@ -0,0 +1,97 @@
...
+    / "cpp"
+    / "src"
+    / "grpc"
+    / "codegen"
+    / "generated"
+    / SCHEMA_FILENAME
+)
+
+
+def schema_path() -> Path:
```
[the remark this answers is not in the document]
ramakrishnap-nv: Fixed in ff12e300 — docstrings on schema_path, load, settings_schema, known_parameters, validate_settings. (tools.py's docstrings landed in 1c9ab7ca.)

python/cuopt_mcp/cuopt_mcp/schema.py
```diff
@@ -0,0 +1,97 @@
...
+        expected = prop.get("type")
+        if expected == "string" and "enum" in prop:
+            if value not in prop["enum"]:
+                raise ValueError(
+                    f"{name} must be one of {prop['enum']}, got {value!r}"
+                )
+        elif expected == "integer" and not isinstance(value, int):
+            raise ValueError(f"{name} must be an integer, got {value!r}")
+        elif expected == "number" and not isinstance(value, (int, float)):
+            raise ValueError(f"{name} must be a number, got {value!r}")
```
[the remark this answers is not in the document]
ramakrishnap-nv: Confirmed real — fixed in ff12e300.

python/cuopt_mcp/cuopt_mcp/server.py
```diff
@@ -0,0 +1,185 @@
...
+    )
+
+
+@server.tool(structured_output=True)
+def cuopt_status(job_id: str) -> dict[str, Any]:
+    """Report whether a cuOpt job is queued, running, or finished.
+
+    Cheap to call repeatedly. Returns terminal=true once the job has
+    reached COMPLETED, FAILED, CANCELLED, or NOT_FOUND.
+    """
```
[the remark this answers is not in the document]
ramakrishnap-nv: Fixed in 1c9ab7ca — documented params/returns for all five, including the {"error": ...} shape _guard returns on failure (added a docstring to _guard itself as the single place that's specified, referenced from each tool rather than repeated in full each time).

python/cuopt_mcp/cuopt_mcp/server.py
```diff
@@ -0,0 +1,185 @@
...
+    kind: "pdlp_settings" for LP, "mip_settings" for MILP.
+    name: a single parameter to describe in full, instead of listing names.
+
+    The catalogue is generated from cuOpt's field registry, so it always
+    matches the solver build being talked to.
+    """
+    return _guard(tools.list_settings, kind=kind, name=name)
+
+
+def main() -> None:
```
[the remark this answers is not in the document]
ramakrishnap-nv: Fixed in 1c9ab7ca.

python/cuopt_mcp/cuopt_mcp/tools.py
```diff
@@ -0,0 +1,262 @@
...
+INLINE_SOLUTION_LIMIT = 200
+
+
+def _solution_dir() -> Path:
+    path = Path(
+        os.environ.get(
+            "CUOPT_MCP_SOLUTION_DIR", Path(tempfile.gettempdir()) / "cuopt-mcp"
+        )
+    )
+    path.mkdir(parents=True, exist_ok=True)
```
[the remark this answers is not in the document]
ramakrishnap-nv: Confirmed real (CWE-377) — fixed in ff12e300: the directory is now created 0700 (chmod'd explicitly since mkdir's mode is subject to umask), and a pre-existing directory is only trusted if it's already private to this user; otherwise it's rejected rather than silently reused.

python/cuopt_mcp/cuopt_mcp/tools.py
```diff
@@ -0,0 +1,262 @@
...
+        summary["num_nonzero"] = len(selected)
+
+    if len(selected) <= limit:
+        summary["variables"] = {k: float(v) for k, v in selected.items()}
+    else:
+        summary["variables_truncated"] = True
+        summary["variables_shown"] = limit
+        summary["variables"] = {
+            k: float(v) for k, v in list(selected.items())[:limit]
+        }
```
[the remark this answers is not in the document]
ramakrishnap-nv: Confirmed real (CWE-400) — fixed in ff12e300: limit now rejects bool/non-int/out-of-range values before any backend call. Didn't add Field(ge=..., le=...) schema metadata — this codebase's MCP tools don't currently use pydantic Field/Annotated anywhere, and the server-side validation added here is the actual enforcement boundary regardless of what the schema advertises, so introducing that pattern felt like scope beyond this finding. Happy to add it if there's a reason to want earlier client-side rejection too.

python/cuopt_mcp/cuopt_mcp/tools.py
```diff
@@ -0,0 +1,262 @@
...
+    except Exception as exc:
+        raise describe_connection_error(exc) from exc
+    lines = (text or "").splitlines()
+    truncated = len(lines) > tail_lines
+    return {
+        "job_id": job_id,
+        "truncated": truncated,
+        "lines": lines[-tail_lines:],
+        "next_byte": from_byte + len(text or ""),
+    }
```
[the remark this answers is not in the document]
ramakrishnap-nv: Same fix and same reasoning as the limit finding above — fixed in ff12e300, tail_lines now bounded 1..MAX_TAIL_LINES (2000) before the backend call; no schema-level Field metadata added for the same reason.

python/cuopt_mcp/tests/test_end_to_end.py
```diff
@@ -0,0 +1,183 @@
...
+    sub = await _call(
+        session,
+        "cuopt_solve_lp",
+        problem_path=mps_file,
+        settings={"time_limit": 30.0},
+    )
+    while not (await _call(session, "cuopt_status", job_id=sub["job_id"]))[
+        "terminal"
+    ]:
+        await __import__("asyncio").sleep(0.3)
```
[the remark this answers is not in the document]
ramakrishnap-nv: Fixed in 1c9ab7ca — bounded the same way the adjacent test already does (120 iterations of 0.3s, fail with the last observed state on timeout).

python/cuopt_mcp/tests/test_schema.py
```diff
@@ -0,0 +1,108 @@
...
+
+@pytest.mark.parametrize("kind", ["pdlp_settings", "mip_settings"])
+def test_every_parameter_states_a_default(kind):
+    props = schema.settings_schema(kind)["properties"]
+    missing = [
+        n
+        for n, p in props.items()
+        if "Default:" not in p.get("description", "")
+    ]
+    assert missing == []
```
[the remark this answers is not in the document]
ramakrishnap-nv: The suggested diff assumes a separate machine-readable default key exists in the schema — it doesn't, by design: _json_schema_property's own docstring explains why (`default` is untyped registry prose, so it's rendered into `description` only, never a typed JSON Schema `default`). Applying that fix as written would break a correct design. Fixed in 1c9ab7ca by tightening what the test actually checks instead — it now requires content after the 'Default:' label, not just the bare literal, so a dropped value would still be caught.

skills/cuopt-developer/references/grpc_codegen.md
```diff
@@ -0,0 +1,120 @@
+# gRPC Wire Fields and the Codegen Registry
```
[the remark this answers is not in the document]
ramakrishnap-nv: Re-triggered — commented /nvskills-ci again for the latest push (1c9ab7ca).

skills/cuopt-developer/references/grpc_codegen.md
```diff
@@ -0,0 +1,120 @@
...
+  the explicitly-sent case, `optional` covers the omitted case. `iteration_limit`
+  and `node_limit` need both.
+- **`description`** / **`default`** — documentation for a settings field,
+  emitted into `cuopt_mcp_schema.json`. `description` becomes the JSON Schema
+  `description`, which is what an MCP client shows a model deciding whether to
+  set the field, so write it for someone choosing a value rather than for
+  someone reading the struct. `default` is a free-text string describing the
+  C++ member initializer (`"1e-4"`, `"-1 (automatic)"`); the generator neither
+  derives nor validates it, so take it from the C++ struct — `docs/` is known
+  to disagree in several places.
```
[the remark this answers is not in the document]
ramakrishnap-nv: Fixed in 1c9ab7ca — same correction as the field_registry.yaml header comment: describes the partial (optional-vs-zero-value) validation instead of claiming none exists.

cpp/src/grpc/codegen/generate_conversions.py
```diff
@@ -339,6 +339,46 @@ def _array_wire_type_comment(f):
...
+    does when omitted.  Returns [] when the field carries neither
+    attribute, so undocumented fields emit exactly as before.
+
+    `default:` is rendered verbatim from the registry — it is a string
+    describing the C++ member initializer, not a value the generator
+    derives or validates.
+    """
+    description = f.get("description")
+    default = f.get("default")
+    if not description and default is None:
```
KyleFromNVIDIA: Why is `description` treated as a bool while `default` is checked by identity? Do we expect `description` to be an empty string?

cpp/src/grpc/codegen/generate_conversions.py
```diff
@@ -339,6 +339,46 @@ def _array_wire_type_comment(f):
...
+        return []
+    body = " ".join(str(description).split()) if description else ""
+    if default is not None:
+        suffix = f"(default: {default})"
+        body = f"{body} {suffix}" if body else suffix
+    prefix = f"{indent}// "
+    width = max(_DOC_COMMENT_WIDTH - len(prefix), 20)
+    lines, current = [], ""
+    for word in body.split():
```
KyleFromNVIDIA: Nit: This splits the description a second time. We could save the results from the first split instead, perhaps in a variable called `words`.

cpp/src/grpc/codegen/lint_registry_defaults.py
```diff
@@ -0,0 +1,176 @@
...
+    (
+        "mip_settings",
+        "cpp/include/cuopt/mathematical_optimization/mip/heuristics_hyper_params.hpp",
+        "mip_heuristics_hyper_params_t",
+        "heuristic_params.",
+    ),
+]
+
+_INIT_RE = re.compile(
+    r"^\s*(?:f_t|i_t|bool)\s+(\w+)\s*(?:=\s*(?P<eq>[^;]+)|\{(?P<br>[^}]+)\})\s*;",
```
KyleFromNVIDIA, line 69: Using regexes is fine for now, but if it ever runs into C++ declarations that it can't handle, we could use libclang instead. [Here](https://github.com/NVIDIA/cuvs/blob/be8ab314d044aee0f80fbe2c2277a893561288de/ci/check_c_abi/check_c_abi/abi.py) is an example of libclang being used in a Python script in cuvs.

KyleFromNVIDIA (approved):
Approved with a few minor comments

cpp/src/grpc/codegen/generate_conversions.py
ramakrishnap-nv: Good catch — not intentional, just leftover asymmetry. Fixed in b2e4cfba: both description and default now build into the same words list, so an empty-string description and a missing one behave identically (verified: description="" still correctly produces no comment lines).

cpp/src/grpc/codegen/generate_conversions.py
ramakrishnap-nv: Fixed in b2e4cfba, same change as above — words is built once and reused for wrapping instead of splitting body a second time.

cpp/src/grpc/codegen/lint_registry_defaults.py
ramakrishnap-nv: Agreed, and thanks for the cuvs pointer. Leaving it as regex for now per its own docstring (best-effort, warn-only, explicitly skips what it can't confidently parse) — happy to revisit with libclang if it starts missing real drift in practice.

KyleFromNVIDIA (approved):

skills/cuopt-developer/references/grpc_codegen.md
```diff
@@ -0,0 +1,111 @@
+# gRPC Wire Fields and the Codegen Registry
+
+Read this before touching anything that crosses the cuOpt gRPC wire: problem
+fields, solution fields, or solver settings.
+
+## The one rule
+
+**Never hand-edit files under `cpp/src/grpc/codegen/generated/`.** They are
```
mlubin: Do we need this to be in the skill? Normally a comment at the top of the file would suffice.

skills/cuopt-developer/references/grpc_codegen.md
```diff
@@ -0,0 +1,111 @@
...
+generated from `cpp/src/grpc/codegen/field_registry.yaml` by
+`cpp/src/grpc/codegen/generate_conversions.py`, and CI will reject a mismatch.
+
+Every field that crosses the wire is declared once in the registry. One entry
+drives ~28 generated artifacts: the `.proto`, the C++ to/from-proto converters,
+the chunked upload/download paths, and the size estimator.
+
+## Workflow
```
mlubin: The workflow should be documented for all, not just for agents.

skills/cuopt-developer/references/grpc_codegen.md
```diff
@@ -0,0 +1,111 @@
...
+| `pdlp_settings` / `mip_settings` | `PDLPSolverSettings` / `MIPSolverSettings` |
+| `lp_solution` / `mip_solution` | solution messages and converters |
+| `chunked_result_header` | `ChunkedResultHeader` for the chunked download path |
+
+## Attributes worth knowing
+
+Full reference: `cpp/src/grpc/codegen/FIELD_REGISTRY_REFERENCE.md`. The ones
+that bite:
+
+- **`optional`** — emits proto3 presence tracking. Required whenever the C++
```
mlubin: Agents don't know how proto3 works?

skills/cuopt-developer/references/grpc_codegen.md
```diff
@@ -0,0 +1,111 @@
...
+  third-party client reading only the `.proto` should learn what a settings
+  field means and what omitting it does. `default` is a free-text string
+  describing the C++ member initializer (`"1e-4"`, `"-1 (automatic)"`); the
+  generator doesn't derive it from the C++ struct or detect initializer drift
+  (only the optional/zero-value check above), so take it from the C++ struct
+  — `docs/` is known to disagree in several places.
+
+## Field numbers are permanent
+
+`field_num` and `array_id` are wire identifiers. Never renumber or reuse a
```
mlubin: Should be documented in-line

skills/cuopt-developer/SKILL.md
```diff
@@ -167,6 +167,18 @@ cuopt/
 - Never suggest `--no-verify` or skipping checks
 - All PRs must pass CI
 
+### Never Hand-Edit Generated Files
```
mlubin: Why do we need these instructions? The headers of the generated files should be sufficient.

mlubin (approved):

skills/cuopt-developer/references/grpc_codegen.md
ramakrishnap-nv: Agreed. Deleted the skill file in 46a4f408 -- turned out most of it duplicated cpp/src/grpc/GRPC_CODE_GENERATION.md and FIELD_REGISTRY_REFERENCE.md, which already existed and are already the developer-facing home for this.

skills/cuopt-developer/references/grpc_codegen.md
ramakrishnap-nv: Same fix (46a4f408) -- the workflow already lived in GRPC_CODE_GENERATION.md's Quick start section, so the skill copy was just a redundant agent-only fork of it. Pointed there instead.

skills/cuopt-developer/references/grpc_codegen.md
ramakrishnap-nv: Fair -- fixed by deleting it (46a4f408). The registry-specific parts of that bullet (the new optional-vs-zero-value validation check, lint_registry_defaults.py) moved into FIELD_REGISTRY_REFERENCE.md's existing §2.4/§4.3, which already covered proto3 optional semantics properly; no need for a second, more basic explanation.

skills/cuopt-developer/references/grpc_codegen.md
ramakrishnap-nv: Done in 46a4f408 -- the description:/default: attributes are now documented in-line in FIELD_REGISTRY_REFERENCE.md §2.1 (as table rows alongside every other attribute), not in the skill.

skills/cuopt-developer/SKILL.md
ramakrishnap-nv: Agreed, removed the whole section in 46a4f408. Confirmed while at it: cuopt_remote_data.proto and every .inc file do carry a DO NOT EDIT / regenerate-with header (I also found and fixed a small drift bug where that header text had been hand-duplicated in two places with different wording -- 8f462b63 -- now both derive from one source).

ramakrishnap-nv:
@mlubin Addressed all 5 — replied to each individually, but the short version: deleted `references/grpc_codegen.md` entirely and moved what wasn't duplicate into `cpp/src/grpc/GRPC_CODE_GENERATION.md`/`FIELD_REGISTRY_REFERENCE.md` (46a4f408), so it's documented once, for everyone, where it already lived rather than a second agent-only copy. Also caught and fixed a real drift bug while consolidating: the generated-file header notice was hand-duplicated between the `.proto` and the `.inc` files with different wording (8f462b63).

mlubin (approved):

cpp/src/grpc/codegen/field_registry.yaml
```diff
@@ -664,88 +911,181 @@ mip_settings:
...
+        Reliability branching mode. -1 automatic (the solver picks whether to
+        use it and with what factor), 0 disabled, k > 0 enables it and treats a
+        variable as reliable once it has been branched on k times.
+      default: "-1 (automatic)"
       field_num: 16
       type: int32
       optional: true
   - mip_batch_pdlp_strong_branching:
+      description: >-
+        Whether strong branching at the root evaluates candidates with a single
```
tmckayus: description has this as boolean, but solver_settings.hpp comments has this as 0, 1 or 2. Which one is correct?

cpp/src/grpc/codegen/field_registry.yaml
```diff
@@ -664,88 +911,181 @@ mip_settings:
...
+        batched PDLP solve instead of solving them in parallel with dual
+        simplex, which can cut strong-branching time when dual simplex
+        struggles. 0 disabled, 1 enabled.
+      default: "0 (disabled)"
       field_num: 17
       type: int32
   - mip_batch_pdlp_reliability_branching:
```
tmckayus, line 966: same comment as above for mip_batch_pdlp_strong_branching

cpp/src/grpc/codegen/field_registry.yaml
```diff
@@ -757,6 +1097,13 @@ mip_settings:
       # `optional`, the proto3 wire zero (`false`) would silently overwrite
       # the C++ default. See also `dual_postsolve` and
       # `barrier_iterative_refinement` in pdlp_settings.
+      description: >-
+        Whether the probing-cache step of MIP presolve runs. Probing evaluates
+        variable fixings to discover implications later used by branch-and-bound
+        and the rounding heuristics. Only meaningful when presolve is otherwise
+        enabled — presolver=0 disables the whole pipeline. Probing is also
```
tmckayus: comment from AI reviewer:

The last sentence of this `description:` looks copied from `docs/cuopt/source/mip-settings.rst`, not from the solver:

> Probing is also skipped in deterministic mode and on LP-only solves.

The probing-cache gate in `cpp/src/mip_heuristics/diversity/diversity_manager.cu` only honors `settings.probing` (and a `constexpr` `fj_only_run` that is `false`). Deterministic mode skips other work (e.g. early CPUFJ); I don’t see it skipping this probing step.

This PR’s rule is that C++ is ground truth and the rst defaults/prose can drift. Please drop the deterministic-mode claim (keep the presolver=0 part), or point to the code that actually skips probing when `determinism_mode` is set. Then `./build.sh codegen` so the proto comment matches.

cpp/src/grpc/codegen/generate_conversions.py
```diff
@@ -339,6 +351,45 @@ def _array_wire_type_comment(f):
...
+    """Render a field's `description:` / `default:` registry attributes as
+    proto leading comment lines.
+
+    The generated proto is part of the public wire contract (see
+    GRPC_INTERFACE.md, "Custom Clients"), so a third-party client reading
+    only the .proto should learn what a settings field means and what it
+    does when omitted.  Returns [] when the field carries neither
+    attribute, so undocumented fields emit exactly as before.
+
+    `default:` is rendered verbatim from the registry — it is a string
```
tmckayus, line 367: comment from AI reviewer -- changes in this PR make it untrue that the generator does not validate defaults (it does at least somewhat):

This docstring is stale relative to the rest of the PR.

`_field_doc_comment` still says `default:` is “not a value the generator derives or validates,” but `_validate_registry_uniqueness` now fails codegen when a non-optional settings field’s `default:` isn’t the proto3 zero value.

Please update this docstring to match that: rendered verbatim, not derived from C++, but checked for non-optional settings against the proto3 zero.

Same wording is still in `cpp/src/grpc/codegen/lint_registry_defaults.py` (module docstring, ~L10–12: “the generator neither derives nor validates”). Align that too so a later reader doesn’t skip the new check.

cpp/src/grpc/codegen/field_registry.yaml
ramakrishnap-nv: Good catch, real bug -- confirmed in solver_settings.hpp:143-144 these are 0/1/2 tri-state (dual simplex only / cooperative / batched-PDLP-only), not boolean. Fixed in 749851a3.

cpp/src/grpc/codegen/field_registry.yaml
ramakrishnap-nv: Same fix, 749851a3.

cpp/src/grpc/codegen/field_registry.yaml
ramakrishnap-nv: Confirmed the AI reviewer's read: traced the only gate on this (diversity_manager.cu:314, settings.probing, plus a constexpr fj_only_run=false that has no dynamic effect) and found nothing checking determinism_mode or an LP-only path there. Dropped the claim in 749851a3, kept the presolver=0 part which does hold.

cpp/src/grpc/codegen/generate_conversions.py
ramakrishnap-nv: Fixed both docstrings in 749851a3 -- _field_doc_comment now describes the actual check (textual match against proto3 zero for non-optional settings fields, not full validation), and lint_registry_defaults.py's module docstring matches. Also fixed a dangling reference there to references/grpc_codegen.md, which no longer exists (deleted earlier today).

tmckayus (approved):
left a few comments on consistency found with an AI review, overall looks good to me

Issue #1703, opened 2026-08-12
[FEA] Expand cuopt_c.h with problem-model accessors for non-Python language bindings


ramakrishnap-nv:
**Is your feature request related to a problem? Please describe.**

The Java bindings (#1524) cannot reach parts of the problem model through the C API, so `java/cuopt/src/main/native/cuopt_jni.cpp` includes the private `pdlp/cuopt_c_internal.hpp` and reaches into `problem_and_stream_view_t` directly. There are 18 such call sites.

That couples `libcuopt_jni.so` to a specific `libcuopt` build rather than to a stable ABI. It is fine while the module is built from the same source tree, but it blocks shipping the jar as a binary artifact: a cuOpt point release can break it in ways no version range expresses. Any other C-ABI binding (Go, Rust, C#) would hit the same wall, and a `jextract`-based binding could not work around it at all, since jextract can only bind the C header.

This is the problem-model counterpart to #1202, which covers solver statistics.

**Describe the solution you'd like**

Roughly 10 functions. Two of them are already acknowledged as `TODO` in `cuopt_c.h`:

```
 * TODO: there is no getter for the quadratic objective matrix (Q)
 * or the quadratic constraint rows.
```

*Name setters — the C API reads names but cannot set them*
- `cuOptSetVariableNames(problem, const char* const* names, cuopt_int_t count)`
- `cuOptSetRowNames(problem, const char* const* names, cuopt_int_t count)`
- `cuOptSetProblemName(problem, const char* name)`
- `cuOptGetProblemName(problem, char* buf, cuopt_int_t size)` — no getter exists either

*Quadratic objective (Q is CSR, so this mirrors `cuOptGetConstraintMatrix`)*
- a non-zero count attribute, plus `cuOptGetQuadraticObjective(problem, offsets, indices, values)`

*Quadratic constraint rows — the awkward one*

`quadratic_constraint_t` is a variable-length array of variable-length records: each row carries a name, sense, RHS, a linear part (values + indices), and a COO `Q` (rows/cols/vals). Copy-out cannot express nested ragged data in one call, so it needs a shape-query-then-fill pattern, roughly:

```c
cuOptGetNumQuadraticConstraints(problem, &n);
cuOptGetQuadraticConstraintSizes(problem, i, &n_linear, &n_quad);
cuOptGetQuadraticConstraint(problem, i, /* arrays */);
cuOptGetQuadraticConstraintName(problem, i, buf, size);
```

[...]

PR #1524, opened 2026-07-07
Java bindings for LP, MIP and QP

nvidiacbrissette:
## Description

Java bindings for LP, MIP and QP, built as hand-written JNI over the public C API, with build and CI integration. Contributes to #1535 and #860.

The API is deliberately small. Following review, it covers building a problem, solving it, and reading the result — nothing beyond that. Anything further can be added later with a case for it.

## API

`Problem` builds and solves; `Variable` and `Constraint` carry the model and the values from the last solve; `SolverSettings` configures the solver; `Solution` reports the outcome.

```java
try (Problem problem = new Problem("burglar")) {
  Variable x = problem.addVariable(0.0, 1.0, 15.0, VariableType.INTEGER, "take_item_0");
  Variable y = problem.addVariable(0.0, 1.0, 100.0, VariableType.INTEGER, "take_item_1");
  problem.addConstraint(LinearExpression.of(x, 2).plus(y, 20).le(102.0), "capacity");
  problem.setObjective(LinearExpression.of(x, 15).plus(y, 100), ObjectiveSense.MAXIMIZE);

  try (SolverSettings settings = new SolverSettings().setSetting(CuOptConstants.CUOPT_TIME_LIMIT, 10.0);
       Solution solution = problem.solve(settings)) {
    System.out.println(solution.getTerminationStatus());
    System.out.println(x.getValue());
  }
}
```

Solution values are read from the model — `Variable.getValue`, `Variable.getReducedCost`, `Constraint.getDualValue`, `Constraint.getSlack` — rather than as bulk arrays.

Solver statistics are read as scalar solution attributes, keyed by a `CuOptConstants.CUOPT_SOLUTION_ATTR_*` value:

```java
double gap   = solution.getFloatAttribute(CuOptConstants.CUOPT_SOLUTION_ATTR_LP_GAP);
int    nodes = solution.getIntAttribute(CuOptConstants.CUOPT_SOLUTION_ATTR_MIP_NUM_NODES);
```

This mirrors the problem attribute accessors. Because the selectors are generated into `CuOptConstants` from `constants.h`, a statistic added later is a new constant rather than new Java. A selector that does not apply to the solution, or that does not have the requested value type, raises `CuOptException`.

## Build integration

`java` is an opt-in target of the top-level `build.sh`:

```bash
./build.sh libcuopt                 # once
./build.sh java                     # build libcuopt_jni.so and package the jar
./build.sh java --run-java-tests    # the same, then run the suite
```

[...]

PR #1715, opened 2026-08-13
Expose solver statistics as scalar solution attributes

### LINKED ISSUE #1202: [FEA] Expand cuopt_c.h with solver stats getters for non-Python language bindings

The C API (`cpp/include/cuopt/linear_programming/cuopt_c.h`) does not expose all of the solver statistics that the C++ structs `lp_solution_t` and `mip_ret_t` carry. Python sidesteps this gap by binding directly to the C++ structs via Cython (see `python/cuopt/cuopt/linear_programming/solver/solver.pxd:181-205`), so Python users see fields like `nb_iterations`, residuals, gap, `solved_by`, MIP nodes, simplex iterations, presolve time, and violation metrics.

Non-Python language bindings that go through the C ABI cannot see these fields. The first concrete consumer is **cuopt-java** (PR #1192), which binds via Java FFM (jextract on `cuopt_c.h`). The Java records `LpStats`, `MilpStats`, and `SolverMethod` were defined with fields matching Python's surface but currently report `NaN` / `-1` / `UNSET` for anything not exposed by the C API. Future bindings (Go, Rust, C#, …) would hit the same wall.

Add thin C getters to `cuopt_c.h` (and corresponding implementations in `cuopt_c.cpp`) that read the existing fields off `lp_solution_t` / `mip_ret_t`. Approximately 11 new functions, each a one-line wrapper around an existing struct field.

LP (read from `lp_solution_t`):
- `cuOptGetPrimalResidual(cuOptSolution, cuopt_float_t*)` → `l2_primal_residual_`
- `cuOptGetDualResidual(cuOptSolution, cuopt_float_t*)` → `l2_dual_residual_`
- `cuOptGetGap(cuOptSolution, cuopt_float_t*)` → `gap_`
- `cuOptGetIterationCount(cuOptSolution, cuopt_int_t*)` → `nb_iterations_`
- `cuOptGetSolvedBy(cuOptSolution, cuopt_int_t*)` → `solved_by_` (method_t)

MIP (read from `mip_ret_t`):
- `cuOptGetPresolveTime(cuOptSolution, cuopt_float_t*)` → `presolve_time_`
- `cuOptGetNumNodes(cuOptSolution, cuopt_int_t*)` → `nodes_`
- `cuOptGetNumSimplexIterations(cuOptSolution, cuopt_int_t*)` → `simplex_iterations_`
- `cuOptGetMaxConstraintViolation(cuOptSolution, cuopt_float_t*)` → `max_constraint_violation_`
- `cuOptGetMaxIntViolation(cuOptSolution, cuopt_float_t*)` → `max_int_violation_`
- `cuOptGetMaxVariableBoundViolation(cuOptSolution, cuopt_float_t*)` → `max_variable_bound_violation_`

Naming/signatures should match the existing `cuOptGet*` convention (return `cuopt_int_t` error code, take an output pointer). The exact function names are open for review.

1. **Bind language bindings directly to C++** — what Python does. Not viable for FFM/CFFI/cgo style bindings that target the C ABI, since C++ name mangling and ABI stability aren't guaranteed.
2. **Skip these fields in Java** — current state of PR #1192. Acceptable v1, but users coming from Gurobi/CPLEX expect these in stats output.
3. **Add a single bulk `cuOptGetStats` returning a struct** — possible but heavier ABI footprint; individual getters are smaller and easier to evolve.

**Additional context**

- C++ struct definitions: `cpp/src/linear_programming/...` (where `lp_solution_t` / `mip_ret_t` are defined)
- Python's view of the same fields: `python/cuopt/cuopt/linear_programming/solver/solver.pxd:181-205`
- Python's Solution class consuming them: `python/cuopt/cuopt/linear_programming/solution/solution.py:206-222`
- Java placeholder code awaiting these getters: `java/cuopt-java/src/main/java22/com/nvidia/cuopt/internal/CuOptProviderImpl.java:425-441`
- PR that uncovered the gap: #1192 (cuopt-java)

Once the C API ships these, the Java side becomes ~30 lines of straightforward FFM wiring.

ramakrishnap-nv:
## Description

Adds two C API accessors so solver statistics are reachable from the C ABI, and closes #1202.

Split out of #1524 at @mlubin's request, so the C API can be reviewed without the Java diff attached. The Java bindings are the first consumer but nothing here refers to them.

### The gap

The C API reports the *outcome* of a solve — `cuOptGetTerminationStatus`, `cuOptGetObjectiveValue`, `cuOptGetMIPGap`, `cuOptGetSolutionBound`, `cuOptGetSolveTime`, and the primal/dual/reduced-cost arrays — but none of the diagnostics the C++ solution interfaces already carry. Of the eleven fields listed in #1202, **zero** are reachable today.

Python sidesteps this by binding to the C++ structs through Cython, so the C ABI is a second-class path and every non-Python binding hits the same wall.

Concretely, what these unlock:

- **Judging a non-optimal answer.** When a solve stops on a time limit, residuals and gap are the only way to distinguish a nearly-converged solution from a useless one.
- **Verifying a returned MIP solution.** The three violation magnitudes are what a caller checks before acting on a solution.
- **Knowing which algorithm solved it.** Under `CUOPT_METHOD_CONCURRENT` the caller otherwise cannot tell PDLP from dual simplex.
- **Performance work.** Iteration counts, nodes, simplex iterations, and presolve time.

### Why attributes rather than one getter per statistic

#1202 proposed eleven individual getters. This uses the attribute model instead, following review feedback from @chris-maes:

- it matches how problem data is already read (`cuOptGetProblemIntAttribute` and friends);
- a future statistic is a new constant rather than a new exported symbol, so existing callers need no relink;
- bindings that generate from `constants.h` — as the Java bindings already do for `CuOptConstants.java` — pick up new statistics with no hand-written code at all.

Structured data stays on dedicated functions, which is the existing convention: `cuOptGetConstraintMatrix` is a function because CSR is three parallel arrays, and the same will apply to the quadratic constraint rows in #1703. The rule is scalars and homogeneous arrays as attributes, ragged or multi-output as functions.

### Safety

Solution selectors are numbered in their own range (300+), so a problem selector passed to a solution accessor, or the reverse, is rejected rather than silently read. LP selectors require an LP solution and MIP selectors require a MIP solution, since the two come from different solvers; `CUOPT_ATTR_IS_MIP` on the originating problem says which set applies.

The values are read straight off `lp_solution_interface_t` / `mip_solution_interface_t`, so this is exposure only — no computation, no solve-time cost.

### Tests

`c_api.lp_solution_attributes` and `c_api.mip_solution_attributes` cover both solvers and the ways a caller can get it wrong: a float selector through the integer accessor and the reverse, the other solver's selectors, unknown selectors, and null arguments.

Float outputs are seeded with `NaN` rather than a numeric sentinel, since the solver cannot legitimately produce `NaN` — so an accessor that never writes its output is caught, where a numeric sentinel would be indistinguishable from a real result.

Verified locally: **69/69** `C_API_TEST` cases pass with the full LP/MIP/QP dataset.

## Checklist

- Testing
   - [x] Added tests
- Documentation
   - [x] Added new documentation

### DIFF
--- a/cpp/include/cuopt/mathematical_optimization/constants.h
+++ b/cpp/include/cuopt/mathematical_optimization/constants.h
@@ -274,8 +274,7 @@
 
 /* @brief Scalar solution attribute selectors
  * (see cuOptGetSolution{Int,Float}Attribute). Passed as cuopt_int_t; numbered in a separate
- * range from the problem selectors so a problem selector passed to a solution accessor, or the
- * reverse, is rejected rather than silently read.
+ * range from the problem selectors.
  *
  * Which of these a solution carries depends on the class of problem that produced it; the
  * accessors return CUOPT_INVALID_ARGUMENT for one that does not apply. Use CUOPT_ATTR_IS_MIP on
--- a/cpp/include/cuopt/mathematical_optimization/constants.h
+++ b/cpp/include/cuopt/mathematical_optimization/constants.h
@@ -272,4 +272,25 @@
 #define CUOPT_STRING_ARRAY_VARIABLE_NAMES 200
 #define CUOPT_STRING_ARRAY_ROW_NAMES      201
 
+/* @brief Scalar solution attribute selectors
+ * (see cuOptGetSolution{Int,Float}Attribute). Passed as cuopt_int_t; numbered in a separate
+ * range from the problem selectors so a problem selector passed to a solution accessor, or the
+ * reverse, is rejected rather than silently read.
+ *
+ * LP attributes require an LP solution and MIP attributes require a MIP solution; the accessors
+ * return CUOPT_INVALID_ARGUMENT otherwise. Use CUOPT_ATTR_IS_MIP on the originating problem to
+ * tell which set applies.
+ */
+#define CUOPT_SOLUTION_ATTR_LP_PRIMAL_RESIDUAL               300
+#define CUOPT_SOLUTION_ATTR_LP_DUAL_RESIDUAL                 301
+#define CUOPT_SOLUTION_ATTR_LP_GAP                           302
+#define CUOPT_SOLUTION_ATTR_LP_NUM_ITERATIONS                303
+#define CUOPT_SOLUTION_ATTR_LP_SOLVED_BY                     304
+#define CUOPT_SOLUTION_ATTR_MIP_PRESOLVE_TIME                305
+#define CUOPT_SOLUTION_ATTR_MIP_NUM_NODES                    306
+#define CUOPT_SOLUTION_ATTR_MIP_NUM_SIMPLEX_ITERATIONS       307
+#define CUOPT_SOLUTION_ATTR_MIP_MAX_CONSTRAINT_VIOLATION     308
+#define CUOPT_SOLUTION_ATTR_MIP_MAX_INT_VIOLATION            309
+#define CUOPT_SOLUTION_ATTR_MIP_MAX_VARIABLE_BOUND_VIOLATION 310
+
 #endif  // CUOPT_CONSTANTS_H
--- a/cpp/include/cuopt/mathematical_optimization/constants.h
+++ b/cpp/include/cuopt/mathematical_optimization/constants.h
@@ -277,9 +277,9 @@
  * range from the problem selectors so a problem selector passed to a solution accessor, or the
  * reverse, is rejected rather than silently read.
  *
- * LP attributes require an LP solution and MIP attributes require a MIP solution; the accessors
- * return CUOPT_INVALID_ARGUMENT otherwise. Use CUOPT_ATTR_IS_MIP on the originating problem to
- * tell which set applies.
+ * Which of these a solution carries depends on the class of problem that produced it; the
+ * accessors return CUOPT_INVALID_ARGUMENT for one that does not apply. Use CUOPT_ATTR_IS_MIP on
+ * the originating problem to determine the class.
  */
 #define CUOPT_SOLUTION_ATTR_LP_PRIMAL_RESIDUAL               300
 #define CUOPT_SOLUTION_ATTR_LP_DUAL_RESIDUAL                 301
--- a/cpp/include/cuopt/mathematical_optimization/cuopt_c.h
+++ b/cpp/include/cuopt/mathematical_optimization/cuopt_c.h
@@ -1111,6 +1111,54 @@ cuopt_int_t cuOptGetDualObjectiveValue(cuOptSolution solution,
 cuopt_int_t cuOptGetReducedCosts(cuOptSolution solution, cuopt_float_t* reduced_cost_ptr);
 
 /* -------------------------------------------------------------------------- */
+/* Solution attributes                                                        */
+/* -------------------------------------------------------------------------- */
+
+/*
+ * Solver statistics are read as scalar attributes, using the CUOPT_SOLUTION_ATTR_* selectors in
+ * constants.h. New statistics can then be added as constants rather than as new exported
+ * functions, which keeps the ABI stable for existing callers.
+ *
+ * LP and MIP statistics come from different solvers, so an LP selector requires an LP solution
+ * and a MIP selector requires a MIP solution. Query CUOPT_ATTR_IS_MIP on the originating problem
+ * to decide which set applies.
+ */
+
+/** @brief Get a scalar integer solution attribute (a CUOPT_SOLUTION_ATTR_* with an integer
+ * value: iteration counts, node counts, or the method that solved the problem).
+ *
+ * @param[in] solution - The solution object.
+ *
+ * @param[in] attribute - The attribute selector.
+ *
+ * @param[out] value_out - A pointer to a cuopt_int_t that on output will contain the value.
+ *
+ * @return A status code indicating success or failure. Returns CUOPT_INVALID_ARGUMENT if the
+ *  selector is unknown, does not have an integer value, or does not apply to this solution's
+ *  solver.
+ */
+cuopt_int_t cuOptGetSolutionIntAttribute(cuOptSolution solution,
+                                         cuopt_int_t attribute,
+                                         cuopt_int_t* value_out);
+
+/** @brief Get a scalar floating-point solution attribute (a CUOPT_SOLUTION_ATTR_* with a
+ * floating-point value: residuals, gap, presolve time, or violation magnitudes).
+ *
+ * @param[in] solution - The solution object.
+ *
+ * @param[in] attribute - The attribute selector.
+ *
+ * @param[out] value_out - A pointer to a cuopt_float_t that on output will contain the value.
+ *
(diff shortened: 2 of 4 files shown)


cpp/include/cuopt/mathematical_optimization/cuopt_c.h
```diff
@@ -1110,6 +1110,54 @@ cuopt_int_t cuOptGetDualObjectiveValue(cuOptSolution solution,
  */
 cuopt_int_t cuOptGetReducedCosts(cuOptSolution solution, cuopt_float_t* reduced_cost_ptr);
 
+/* -------------------------------------------------------------------------- */
+/* Solution attributes                                                        */
+/* -------------------------------------------------------------------------- */
+
+/*
+ * Solver statistics are read as scalar attributes, using the CUOPT_SOLUTION_ATTR_* selectors in
+ * constants.h. New statistics can then be added as constants rather than as new exported
```
mlubin: The comment should describe what the code does, not what alternative designs could have been.

cpp/include/cuopt/mathematical_optimization/cuopt_c.h
```diff
@@ -1110,6 +1110,54 @@ cuopt_int_t cuOptGetDualObjectiveValue(cuOptSolution solution,
...
+ * functions, which keeps the ABI stable for existing callers.
+ *
+ * LP and MIP statistics come from different solvers, so an LP selector requires an LP solution
```
mlubin: Do we ever present "LP" and "MIP" as different solvers in the C API? It sounds like we're leaking an implementation detail. How about simply saying that not all attributes are present for all solutions, depending on the problem class?

cpp/include/cuopt/mathematical_optimization/cuopt_c.h
```diff
@@ -1110,6 +1110,54 @@ cuopt_int_t cuOptGetDualObjectiveValue(cuOptSolution solution,
  */
 cuopt_int_t cuOptGetReducedCosts(cuOptSolution solution, cuopt_float_t* reduced_cost_ptr);
 
+/* -------------------------------------------------------------------------- */
+/* Solution attributes                                                        */
```
mlubin, line 1114: We're introducing a new concept of a solution attribute without a clearly documented distinction on what makes them different from parameters.

ramakrishnap-nv:
Thanks @mlubin — all three taken, in `e6870345`.

**Rationale in the header.** Removed. The point about statistics being addable as constants rather than exported functions is an argument for the design; it belongs in the PR description, which is where it now lives. The block describes what the accessors do.

**"Different solvers."** Agreed that it leaks an implementation detail — and checking, "solver" was not vocabulary this header used anywhere else for this purpose; the phrase appeared exactly once, in my own text. Reworded as you suggested: not every attribute applies to every solution, depending on the class of problem that produced it. Same change applied to the selector block in `constants.h`.

**Attribute vs parameter.** Fair — the header introduced the concept without defining it. Now stated explicitly:

> Attributes are distinct from parameters. A parameter is an input, set on a `cuOptSolverSettings` before solving with `cuOptSetParameter` and read back with `cuOptGetParameter`. An attribute is an output, read from a solved `cuOptSolution`, or from a `cuOptOptimizationProblem` in the case of the problem attributes further below.

The existing problem-attributes block does not define the term either, so if this wording works I am happy to fold a shared definition into one place in a follow-up rather than have two blocks each half-explaining it.

---

@coderabbitai the leak is real and is fixed. The checks use `ASSERT`, which returns early, so `cuOptDestroySolution` at the end of the test was skipped precisely when a test failed. Rather than downgrade to `EXPECT` — the assertions are load-bearing, since a failed accessor makes the following value check meaningless and would report a confusing second failure — the solution is now held by a scoped guard, so it is destroyed on every exit path and the fail-fast behaviour is kept.

69/69 `C_API_TEST` cases pass.

cpp/include/cuopt/mathematical_optimization/constants.h
```diff
@@ -272,4 +272,25 @@
 #define CUOPT_STRING_ARRAY_VARIABLE_NAMES 200
 #define CUOPT_STRING_ARRAY_ROW_NAMES      201
 
+/* @brief Scalar solution attribute selectors
+ * (see cuOptGetSolution{Int,Float}Attribute). Passed as cuopt_int_t; numbered in a separate
+ * range from the problem selectors so a problem selector passed to a solution accessor, or the
+ * reverse, is rejected rather than silently read.
```
mlubin:
```suggestion
 * range from the problem selectors.
```

cpp/src/pdlp/cuopt_c.cpp
```diff
@@ -1362,6 +1365,114 @@ cuopt_int_t cuOptGetReducedCosts(cuOptSolution solution, cuopt_float_t* reduced_
...
 }
 
+namespace {
+
+// Solution attribute plumbing. Each selector names one scalar on the LP or MIP solution
+// interface; adding a statistic later means adding a constant and one case, not a new symbol.
+
+bool is_lp_solution_attribute(cuopt_int_t attribute)
+{
+  return attribute >= CUOPT_SOLUTION_ATTR_LP_PRIMAL_RESIDUAL &&
```
mlubin: This approach doesn't leave any room for us to define new LP solution attributes in the future. Maybe we don't really need `is_lp_solution_attribute` and can instead define macros like:
```
PROCESS_MIP_INT_ATTR(CUOPT_SOLUTION_ATTR_MIP_NUM_NODES, get_num_nodes)
```
which translates to:
```
case CUOPT_SOLUTION_ATTR_MIP_NUM_NODES:
    auto* mip = as_mip_solution(solution);
    if (mip == nullptr) { return CUOPT_INVALID_ARGUMENT; }
    *value_out = static_cast<cuopt_int_t>(mip->get_num_nodes());
    return CUOPT_SUCCESS;
```

ramakrishnap-nv:
Both taken, in `7a43536a`.

**Selector range.** You're right that the range test made the numeric values load-bearing — a new LP attribute would have had to be inserted inside 300–304, renumbering everything after it and breaking anyone already compiled against those constants, or be appended after the MIP block where `is_lp_solution_attribute` would have misclassified it as MIP. Neither is acceptable for a constant that ships.

`is_lp_solution_attribute` is gone, replaced by the per-case macros you described:

```c
CUOPT_READ_MIP_ATTRIBUTE(CUOPT_SOLUTION_ATTR_MIP_NUM_NODES, get_num_nodes, cuopt_int_t)
```

expanding to a case that resolves the solution kind, rejects a mismatch, and reads the field. Each case now states which kind it reads, so selector values carry no meaning beyond identity and new ones can be appended anywhere in the block.

Behaviour is unchanged, which the existing tests confirm: a selector naming the other kind of solution still returns `CUOPT_INVALID_ARGUMENT`, now because the corresponding accessor yields `nullptr` rather than because of a range test, and a selector of the wrong value type still falls through to `default`.

**Comment.** Applied your suggestion verbatim.

69/69 `C_API_TEST` cases pass locally.

mlubin (approved):

Issue #1635, opened 2026-07-28
Split libcuopt wheel into separate routing/LP packages for PyPI publication


ramakrishnap-nv:
## Background

Publishing cuOpt to pypi.org (instead of only pypi.nvidia.com) would simplify installation, remove the need for `--extra-index-url`, and unblock security-conscious users who restrict installs to the main PyPI index.

The blocker is package size. The current `libcuopt-cu12` wheel is ~542 MB compressed. PyPI enforces a 1 GB per-package limit and a soft 100 GiB per-project limit; headroom exists today but is not comfortable enough to commit to indefinitely, especially with features like multi-GPU PDLP added recently.

## What this issue tracks

Split the single `libcuopt` wheel into separate wheels along solver boundaries:

- `libcuopt-routing-cu12` — routing/VRP engine (`libcuopt_routing.so`)
- `libcuopt-mathopt-cu12` — LP/MILP/QP/SOCP solver (`libcuopt_mathopt.so`)
- `libcuopt-grpc-cu12` — gRPC client (`libcuopt_grpc.so`); depends on both engines, see below
- `libcuopt-client-cu12` — CUDA-free host-side client (`libcuopt_client.so`), see #1872
- `libcuopt-cu12` — metapackage depending on all of the above, so existing installs are unaffected

`cuopt_base` no longer exists. It was removed from #1622 after measurement showed a single consumer: `libcuopt_mathopt.so` referenced 9 symbols from it (`seed_generator::seed_`, `print_version_info`, six `work_unit_scheduler_t` methods, `printTimestamp`), while routing, grpc, client, `cuopt_cli` and `cuopt_grpc_server` referenced none. Its original rationale — shared stateful infrastructure — had eroded: the logger became header-only per component in #1778, linear algebra moved into mathopt after review, and #1809 replaced the process-wide seed generator with per-worker RNGs. `UTIL_SRC_FILES` now builds into `cuopt_mathopt`.

That removes a package from this split and the question of whether the routing wheel should depend on base.

The corresponding Python meta-packages (`cuopt-vrp`, `cuopt-lp`, etc.) would declare only the lib package they need, so users who only use routing don't pull in the LP solver and vice versa.

## Component boundaries (measured on #1622, stripped)

Re-measured 2026-09-09 on #1622 merged with main, stripped with `strip --strip-unneeded`:

| Component | Size | External deps beyond rmm / rapids_logger |
|---|---|---|
[...]

PR #1724, opened 2026-08-14
python(grpc): build the LP and routing clients as one extension module

ramakrishnap-nv:
Stacked on #1597 (base is `routing-grpc-vrp-e2e`, since this needs the routing client that PR adds).

`cuopt.grpc.linear_programming.Client` and `cuopt.grpc.routing.RoutingClient` were two extension modules wrapping the same C++ object (`cuopt::cython::grpc_python_client_t`). This builds them as one unit.

Motivation is the client/engine split discussed in #1635: if the gRPC client is detached from the solver engines as a GPU-free package, two units would each need their own copy of the transport layer. One unit keeps it a single package.

Both public classes and both import paths are unchanged — `Client` is already released and documented. Merging the declarations also collapses a drift between the two `.pxd` files, which described the same C++ class differently (routing omitted `is_mip` and most methods, and used `bool` where LP used `bint`).

### Testing
Built and tested in a local `.cuopt_env` on an RTX 8000:
- VRP end-to-end against a live `cuopt_grpc_server` — remote solve matches local; submit/wait/result/delete lifecycle passes. Server log confirms the real path (`VRP problem reconstructed` → `RoutingSolution`).
- LP/MIP gRPC client tests: 11 passed, 6 skipped (TLS certs absent locally).
- Serialization coverage tests pass.
- Verified at runtime that both import paths resolve to the merged `.so` and no stale per-domain extension remains.

Not covered locally: TLS/mTLS and chunked >2 GiB upload (both LP-only, unrelated to the merge).

### Docs
No user-facing API change, so no doc updates. Module docstrings explain why the arms share a unit.


ramakrishnap-nv:
Folding this into #1597 instead.

Keeping it separate meant #1597 would add `grpc/routing/grpc_client.{pyx,pxd}` and this PR would immediately delete them — churn in `main` for no benefit, and review effort spent on files already slated for removal. The unification also has no standalone value, since it depends on the routing client #1597 introduces.

The commit is now on `routing-grpc-vrp-e2e`, and #1597's net diff shows only the end state: one `cuopt/grpc/client/` extension module holding both arms, with the two public import paths preserved as shims.

PR #1726, opened 2026-08-14
Mod-2 cut separator for zero-half cuts

akifcorduk:
- Add general mod-2 zero-half separation. The separator finds combinations of constraint rows whose odd coefficients cancel modulo 2 while their aggregated right-hand side remains odd. Halving and rounding this aggregate yields a valid cut.
- Use sparse GF(2) elimination to find these row dependencies, then transform, lift, score, and add violated candidates to the cut pool.
- Propagate B&B timeouts to active node solves and improve work unit checks for some cuts.
- Update PaPILO pin for the coefficient-capacity guard fix.

On H100, two runs each:
Main vs mod-2
**Feasible** | 224.0 | 226.5 | +2.5
**Average error gap** | 12.285 | 11.655 | -0.630
**Optimal instances** | 76.5 | 77.5 | +1.0
**Root gap closed average** | 29.7518% | 31.0948% | +1.3429%
**Root gap closed shifted geomean** (+1) | 10.4875% | 11.1626% | +0.6750%
MIP gap unchanged, due to some cut passes taking longer.


akifcorduk:
/ok to test [cb33aff](https://github.com/NVIDIA/cuopt/pull/1726/commits/cb33afff1ac41cea388e966159a27bf4ecda8da8)

cpp/src/branch_and_bound/branch_and_bound.cpp
```diff
@@ -1727,7 +1728,8 @@ void branch_and_bound_t<i_t, f_t>::plunge_with(bfs_worker_t<i_t, f_t>* worker,
     }
 
     if (now > settings_.time_limit) {
-      solver_status_ = mip_status_t::TIME_LIMIT;
+      node_concurrent_halt_ = 1;
```
aliceb-nv: I haven't triple checked the surrounding code, but do you think this is something that could be RAII'd? e.g. with a scope_guard
My worry is that we'll forget to set the concurrent halt in later changes otherwise

cpp/src/cuts/zero_half_mod2.cpp
```diff
@@ -0,0 +1,774 @@
...
+  f_t start_time,
+  f_t time_limit,
+  f_t& work_estimate,
+  f_t max_work_estimate,
+  bool& work_limit_reached)
+{
+  constexpr i_t max_integral_scale       = 1000;
+  const i_t max_integer_row_length       = 1000 + lp.num_cols / 10;
+  constexpr f_t row_tight_tol            = (f_t)1e-6;
+  constexpr f_t coefficient_integral_tol = (f_t)1e-6;
```
aliceb-nv: Should these coefficients be tighter? I seem to recall using 1e-10 in other cuts code

cpp/src/cuts/zero_half_mod2.cpp
```diff
@@ -0,0 +1,774 @@
...
+                        const std::vector<variable_type_t>& var_types,
+                        const std::vector<f_t>& transformed_xstar,
+                        i_t max_integral_scale,
+                        f_t row_tight_tol,
+                        f_t coefficient_integral_tol,
+                        f_t start_time,
+                        f_t time_limit,
+                        f_t& work_estimate,
+                        f_t max_work_estimate,
+                        bool& work_limit_reached)
```
aliceb-nv, line 173: I think we already have existing functions to do this, in problem_t at least (to do the objective row scaling), there may be an opportunity for reuse

aliceb-nv (approved):
LGTM, minor nits, thanks Akif!

cpp/src/branch_and_bound/branch_and_bound.cpp
akifcorduk: There are few other break/return reasons, so a single RAII object might be a bit confusing because there are two different logic for checking time limit (lp_status, toc(exploration_stats_.start_time) > settings_.time_limit). We would still need some logic within each break block. What I can do is to leave setting `solver_status_        = mip_status_t::TIME_LIMIT;` and do the concurrent halt at the end of the function, if it is time limit.

aliceb-nv: if we have a single point of exit in this function, it's probably the better way

cpp/src/cuts/zero_half_mod2.cpp
akifcorduk: I tried to reuse what other cuts were doing. It seems 1e-10 is mostly used for dual tolerance and when some ratio was involved.
https://github.com/NVIDIA/cuopt/blob/main/cpp/src/cuts/cuts.cpp#L5276-L5277
https://github.com/NVIDIA/cuopt/blob/main/cpp/src/cuts/cuts.cpp#L1696
https://github.com/NVIDIA/cuopt/blob/main/cpp/src/cuts/cuts.cpp#L5048
https://github.com/NVIDIA/cuopt/blob/main/cpp/src/cuts/cuts.cpp#L5051

cpp/src/cuts/cuts.cpp
```diff
@@ -3095,19 +3102,27 @@ void cut_generation_t<i_t, f_t>::generate_implied_bound_cuts(
 {
   if (probing_implied_bound_.zero_offsets.empty()) { return; }
 
-  const f_t tol      = 1e-4;
-  i_t num_cuts       = 0;
-  const i_t pib_cols = static_cast<i_t>(probing_implied_bound_.zero_offsets.size()) - 1;
-  const i_t n_cols   = std::min(lp.num_cols, pib_cols);
+  const f_t tol                  = 1e-4;
+  i_t num_cuts                   = 0;
+  const i_t pib_cols             = static_cast<i_t>(probing_implied_bound_.zero_offsets.size()) - 1;
```
nguidotti: NIT: remove `static_cast` here

cpp/src/cuts/zero_half_mod2.cpp
```diff
@@ -0,0 +1,774 @@
...
+  candidates.reserve(lp.num_rows);
+  for (i_t row = 0; row < lp.num_rows; ++row) {
+    if (toc(start_time) >= time_limit || work_limit_reached) { break; }
+    const i_t slack = complemented_mir.slack_cols(row);
+    if (slack < 0 || transformed_xstar[slack] > row_tight_tol) { continue; }
+
+    const i_t row_length = Arow.row_start[row + 1] - Arow.row_start[row];
+    if (row_length > max_integer_row_length) { continue; }
+    const f_t row_work =
+      (f_t)(8 * row_length + 5) + (f_t)row_length * std::log2((f_t)row_length + (f_t)1.0);
```
nguidotti: I think you can avoid the explicit conversion here. It should convert everything to double

nguidotti (approved):
Looks good to me! Thanks Akif!
My only comment is that you are doing some explicit cast when it is not needed.

cpp/src/cuts/cuts.cpp
akifcorduk: Thanks, AI tends to do that. Will fix it.

cpp/src/cuts/zero_half_mod2.cpp
akifcorduk: I think they are slightly different:
- There is filtering of continuous variables and integral valued variables from the check.
- Work unit accounting.
- Relative tolerance.

It is a simple function, I think trying to reuse might complicate the logic with more parameters and ifs.

aliceb-nv: okay, makes sense. just a bit concerned because I think this is the third or fourth time we implement this sort of operation in the solver :)

akifcorduk: Yeah, I understand and it makes sense. But I feel specializing logic in simple functions is kind of more confusing.

cpp/src/branch_and_bound/branch_and_bound.cpp
akifcorduk: Handled it for both time_limit and optimal. Now we set it at the end of the function.

akifcorduk:
/ok to test [75117f1](https://github.com/NVIDIA/cuopt/pull/1726/commits/75117f10757f7488fde06e912d5eada2a132327d)

ramakrishnap-nv (approved):

akifcorduk:
/ok to test [d11984c](https://github.com/NVIDIA/cuopt/pull/1726/commits/d11984ca33dde98f4f8bc3ee6669cc1cf91ec6ee)

cpp/src/branch_and_bound/branch_and_bound.cpp
```diff
@@ -1840,6 +1845,9 @@ void branch_and_bound_t<i_t, f_t>::plunge_with(bfs_worker_t<i_t, f_t>* worker,
 template <typename i_t, typename f_t>
 void branch_and_bound_t<i_t, f_t>::launch_bfs_worker(bfs_worker_t<i_t, f_t>* worker)
 {
+  // The status may change after the caller checks its search-loop condition.
+  if (solver_status_ != mip_status_t::UNSET) { return; }
+
```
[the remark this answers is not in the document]
akifcorduk: It requires more sychronization and logic. This is just a loose check.

akifcorduk:
/ok to test [f799e24](https://github.com/NVIDIA/cuopt/pull/1726/commits/f799e244662c3b16b23125f0871082c6b75f2034)

PR #1727, opened 2026-08-14
fix: pin rapids-logger to 0.3.* to match rapids-cmake

ramakrishnap-nv:
`conda-cpp-build` and `wheel-build-libcuopt` are currently failing on `main` for every matrix entry:

```
-- CPM: Adding package rapids_logger@0.3.0 (0.3.0)
CMake Error at build/_deps/rapids_logger-src/CMakeLists.txt:96 (add_library):
  add_library cannot create ALIAS target "rapids_logger::rapids_logger"
  because another target with the same name already exists.
-- Configuring incomplete, errors occurred!
```

`rapids_cpm_rapids_logger()` resolves 0.3.0, but the environment pinned `0.2.*`. The installed 0.2.x CMake config is imported transitively via rmm, CPM then fails its version check and adds a downloaded 0.3.0 as a subdirectory, so the ALIAS target is defined twice. cudf pins `rapids-logger==0.3.*` for the same reason.

Nothing else changed — every modified line in the 7 files is this pin, with `conda/environments/` and the two `pyproject.toml` files regenerated by `rapids-dependency-file-generator`.

Verified locally: the regenerated env solves to `rapids-logger 0.3.0`, and configure then reports `CPM: Using local package rapids_logger@0.3.0` and completes.


ramakrishnap-nv:
Closing — redundant. rapids-logger was already bumped to 0.3 in #1659 (6df36e2f); this branch was cut from a stale local main that predated it, so the change nets to zero against current main.

PR #1731, opened 2026-08-14
Build cuopt as an abi3 wheel using the CPython Limited API

### LINKED ISSUE #1725: [FEA] python: use CPython Limited API, produce ABI3 wheels

This project currently builds 1 `cuopt-cu{12,13}` wheel per each combination of Python `{major}.{minor}` (4) and CPU architecture (2).

Image

ref: https://pypi.nvidia.com/cuopt-cu13/

Those 8 wheels require:

* 8 CI jobs on every commit
* new builds to support new Python versions

That could be reduced to 2 wheels (1 per CPU architecture) by adopting CPython's Limited API (https://docs.python.org/3/c-api/stable.html).

This was done across many of the RAPIDS projects in https://github.com/rapidsai/build-planning/issues/42 and we saw it notably reduce CI end-to-end time, CI resource usage, and the effort required to add support for new Python versions.

N/A

**Additional context**

The PRs linked from https://github.com/rapidsai/build-planning/issues/42 show examples of how to implement this.

ramakrishnap-nv:
Closes #1725.

Builds `cuopt` against the CPython Limited API so we ship one wheel per (CUDA major, arch) instead of one per Python version. The `cuopt` wheel build drops from 8 CI jobs to 2, and adding a new Python version no longer requires a new build.

- `python/cuopt`: `wheel.py-api = "cp311"` (matches `requires-python`), overridden per-job by `RAPIDS_PY_API`
- cython bumped to `>=3.2.2,<3.3.0a0` for its Limited API support
- cuopt wheel artifacts named with `rapids-artifact-name --stable` on both the producing and consuming sides
- `wheel-build-cuopt` matrix filtered to the minimum supported Python

`libcuopt` is already `py3-none` and the server/client wheels are pure Python, so this only touches `python/cuopt`. Wheel tests still run the full Python matrix, now all installing the single abi3 wheel.

Approach mirrors the RAPIDS rollout in https://github.com/rapidsai/build-planning/issues/42 (rmm/cudf `main`). rapids-cmake already supports `SKBUILD_SABI_VERSION`, so no CMake changes were needed.

Verified locally that all 9 Cython modules compile clean under `Py_LIMITED_API=0x030b0000` and build as `*.abi3.so`; no `.pyx` changes were required. Full wheel link/test happens in CI.

The second commit is a small `skills/` doc update, which is why this PR comes from an upstream branch rather than a fork.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -157,6 +157,8 @@ jobs:
       script: ci/build_wheel_cuopt.sh
       package-name: cuopt
       package-type: python
+      # abi3 wheel: build one per CUDA x ARCH, using the minimum supported Python
+      matrix_filter: group_by({CUDA_VER, ARCH}) | map(min_by(.PY_VER | split(".") | map(tonumber)))
   wheel-publish-cuopt:
     needs: wheel-build-cuopt
     permissions:
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -157,8 +157,10 @@ jobs:
       script: ci/build_wheel_cuopt.sh
       package-name: cuopt
       package-type: python
-      # abi3 wheel: build one per CUDA x ARCH, using the minimum supported Python
-      matrix_filter: group_by({CUDA_VER, ARCH}) | map(min_by(.PY_VER | split(".") | map(tonumber)))
+      # abi3 wheel: build one per ARCH + CUDA major, using the minimum supported Python.
+      # Grouping on the CUDA major (not the full version) keeps one job per published
+      # artifact name, which only encodes the major.
+      matrix_filter: group_by([.ARCH, (.CUDA_VER|split(".")|map(tonumber)|.[0])]) | map(min_by((.PY_VER|split(".")|map(tonumber)), (.CUDA_VER|split(".")|map(-tonumber))))
   wheel-publish-cuopt:
     needs: wheel-build-cuopt
     permissions:
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -157,10 +157,8 @@ jobs:
       script: ci/build_wheel_cuopt.sh
       package-name: cuopt
       package-type: python
-      # abi3 wheel: build one per ARCH + CUDA major, using the minimum supported Python.
-      # Grouping on the CUDA major (not the full version) keeps one job per published
-      # artifact name, which only encodes the major.
-      matrix_filter: group_by([.ARCH, (.CUDA_VER|split(".")|map(tonumber)|.[0])]) | map(min_by((.PY_VER|split(".")|map(tonumber)), (.CUDA_VER|split(".")|map(-tonumber))))
+      # Build a wheel for each CUDA x ARCH x minimum supported Python version
+      matrix_filter: group_by({CUDA_VER, ARCH}) | map(min_by(.PY_VER | split(".") | map(tonumber)))
   wheel-publish-cuopt:
     needs: wheel-build-cuopt
     permissions:
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -216,8 +216,8 @@ To build all libraries and tests, simply run
 
 - **Note**: if Cython files (`*.pyx` or `*.pxd`) have changed, the Python build must be rerun.
 - **Note**: the `cuopt` wheel is built against the CPython Limited API (abi3), so Cython code must
-  avoid CPython APIs outside it. Using one surfaces as a compile error under `-DPy_LIMITED_API`,
-  not at runtime.
+  use only APIs the Limited API exposes. An unsupported API fails to compile under
+  `-DPy_LIMITED_API` rather than failing at runtime.
 
 To run the C++ tests, run
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -215,6 +215,9 @@ To build all libraries and tests, simply run
 ```
 
 - **Note**: if Cython files (`*.pyx` or `*.pxd`) have changed, the Python build must be rerun.
+- **Note**: the `cuopt` wheel is built against the CPython Limited API (abi3), so Cython code must
+  avoid CPython APIs outside it. Using one surfaces as a compile error under `-DPy_LIMITED_API`,
+  not at runtime.
 
 To run the C++ tests, run
 
--- a/ci/validate_wheel.sh
+++ b/ci/validate_wheel.sh
@@ -46,3 +46,13 @@ rapids-logger "validate packages with 'twine'"
 twine check \
     --strict \
     "$(echo "${wheel_dir_relative_path}"/*.whl)"
+
+rapids-logger "validate packages with 'abi3audit'"
+
+# 'abi3audit' fails on wheels with DSOs that lack an ABI tag (e.g. 'lib*' wheels).
+# Filtering by '*abi*' avoids those.
+find \
+    "${wheel_dir_relative_path}" \
+    -type f \
+    -name '*abi*' \
+    -exec abi3audit --strict --summary --verbose '{}' \+
--- a/ci/validate_wheel.sh
+++ b/ci/validate_wheel.sh
@@ -49,10 +49,21 @@ twine check \
 
 rapids-logger "validate packages with 'abi3audit'"
 
-# 'abi3audit' fails on wheels with DSOs that lack an ABI tag (e.g. 'lib*' wheels).
-# Filtering by '*abi*' avoids those.
-find \
-    "${wheel_dir_relative_path}" \
-    -type f \
-    -name '*abi*' \
-    -exec abi3audit --strict --summary --verbose '{}' \+
+# 'abi3audit' fails on wheels with DSOs that lack an ABI tag, so only the abi3 wheels
+# are audited. Of the packages sharing this script, only 'cuopt' builds one; the rest
+# are 'py3-none'.
+abi3_wheels=()
+while IFS= read -r -d '' wheel; do
+    abi3_wheels+=("${wheel}")
+done < <(find "${wheel_dir_relative_path}" -type f -name '*-abi3-*.whl' -print0)
+
+# Guard against 'cuopt' silently losing its abi3 tag: without this, dropping
+# 'wheel.py-api' would skip the audit entirely and leave CI green.
+if [[ "${package_dir}" == "python/cuopt" ]] && [[ "${#abi3_wheels[@]}" -eq 0 ]]; then
+    rapids-echo-stderr "expected an abi3 wheel in '${wheel_dir_relative_path}', found none"
+    exit 1
+fi
+
+if [[ "${#abi3_wheels[@]}" -gt 0 ]]; then
+    abi3audit --strict --summary --verbose "${abi3_wheels[@]}"
+fi
--- a/ci/validate_wheel.sh
+++ b/ci/validate_wheel.sh
@@ -49,21 +49,10 @@ twine check \
 
 rapids-logger "validate packages with 'abi3audit'"
 
-# 'abi3audit' fails on wheels with DSOs that lack an ABI tag, so only the abi3 wheels
-# are audited. Of the packages sharing this script, only 'cuopt' builds one; the rest
-# are 'py3-none'.
-abi3_wheels=()
-while IFS= read -r -d '' wheel; do
-    abi3_wheels+=("${wheel}")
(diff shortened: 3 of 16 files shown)


skills/cuopt-developer/references/contributing.md
```diff
@@ -109,6 +109,7 @@ Follow YAGNI strictly here — flags, fallbacks, env-var overrides, and config k
 A few non-YAGNI points worth keeping in mind:
 
 - Prefer extending an existing script over adding a new one.
+- For build/CI conventions shared across RAPIDS (wheel packaging, artifact naming, matrix filters), read the reference implementation in `rapidsai/rmm` or `rapidsai/cudf` on `main` before writing anything. cuOpt's `ci/` scripts and shared-workflow inputs are near-copies of theirs, and gha-tools expects exact conventions — an equivalent-but-different local invention silently breaks the download side, which looks for the name the build side wrote.
```
jameslamb:
```suggestion
- For build/CI conventions shared across RAPIDS (wheel packaging, artifact naming, matrix filters), read the reference implementation in `rapidsai/rmm` or `NVIDIA/cudf` on `main` before writing anything. cuOpt's `ci/` scripts and shared-workflow inputs are near-copies of theirs, and gha-tools expects exact conventions — an equivalent-but-different local invention silently breaks the download side, which looks for the name the build side wrote.
```

This is a good note to add to the skill, but `cudf` has moved to the NVIDIA org: https://github.com/NVIDIA/cudf

jameslamb (approved):
This is looking great!!! I spot-checked the logs, all looking the way I'd expect. It's great that this didn't require any Python code changes.

One other small request... can you please also add `abi3audit` to `ci/validate_wheel.sh`? You could follow the examples from the PRs linked to https://github.com/rapidsai/build-planning/issues/315

skills/cuopt-developer/references/contributing.md
ramakrishnap-nv: Good catch, thanks — fixed in a409e23f.

CONTRIBUTING.md
```diff
@@ -215,6 +215,9 @@ To build all libraries and tests, simply run
 ```
 
 - **Note**: if Cython files (`*.pyx` or `*.pxd`) have changed, the Python build must be rerun.
+- **Note**: the `cuopt` wheel is built against the CPython Limited API (abi3), so Cython code must
+  avoid CPython APIs outside it. Using one surfaces as a compile error under `-DPy_LIMITED_API`,
+  not at runtime.
```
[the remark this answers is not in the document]
ramakrishnap-nv: Reworded in a409e23f. I went with "use only APIs the Limited API exposes. An unsupported API fails to compile under `-DPy_LIMITED_API` rather than failing at runtime." — the suggested version repeated "CPython Limited API" twice in consecutive clauses.

ramakrishnap-nv:
Thanks @jameslamb! Added `abi3audit` to `ci/validate_wheel.sh` in a409e23f, following the cudf/rmm implementation — same `*abi*` name filter so the `py3-none` wheels (`libcuopt`, `cuopt-server`, `cuopt-sh-client`) are skipped rather than failing on DSOs without an ABI tag.

Also addressed the CodeRabbit findings in the same commit, and verified the two it flagged as unverified:

**Matrix grouping (valid, fixed).** The filter grouped on the full `CUDA_VER`, but artifact names encode only the CUDA major. With two minors in one major that would have produced two jobs writing the same artifact name. Now grouped on `[.ARCH, CUDA major]` with a deterministic minor tiebreak, matching the existing `libcuopt` filter. Checked against both matrices — identical output today (4 jobs, all 3.11), 8 → 4 on a matrix with two minors per major.

**`build_wheel.sh` arg interface (no change needed).** All four callers pass exactly two positionals; only `build_wheel_cuopt.sh` adds `--stable`, so nothing hits the `Unknown option` path.

**`wheel.py-api` override reaching scikit-build-core (no change needed).** `--config-settings skbuild.wheel.py-api=` is passed in the same `pip wheel` invocation, which is how cudf does it. Confirmed locally: the build dir becomes `build/cp311-abi3-linux_x86_64`, `Py_LIMITED_API=0x030b0000` appears in the compile lines, and modules build as `*.abi3.so`. In CI the override happens to equal the default since the matrix pins the build to the minimum Python — the uploaded artifacts came out as `cuopt_wheel_python_cuopt_{x86_64,aarch64}_abi3_cu{12,13}`.

ci/validate_wheel.sh
```diff
@@ -46,3 +46,13 @@ rapids-logger "validate packages with 'twine'"
...
+
+rapids-logger "validate packages with 'abi3audit'"
+
+# 'abi3audit' fails on wheels with DSOs that lack an ABI tag (e.g. 'lib*' wheels).
+# Filtering by '*abi*' avoids those.
+find \
+    "${wheel_dir_relative_path}" \
+    -type f \
+    -name '*abi*' \
+    -exec abi3audit --strict --summary --verbose '{}' \+
```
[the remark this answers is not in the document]
ramakrishnap-nv: Good catch on the silent-skip, fixed in 63de0b26 — but I had to scope it, because the suggested patch as written would have broken the build.

`ci/validate_wheel.sh` is shared by four packages, and only `cuopt` produces an abi3 wheel:

| package | wheel tag | `*-abi3-*.whl` matches |
|---|---|---|
| `python/cuopt` | `cp311-abi3-linux_x86_64` | 1 |
| `python/libcuopt` | `py3-none-manylinux_2_28_*` | 0 |
| `python/cuopt_server` | `py3-none-any` | 0 |
| `python/cuopt_self_hosted` | `py3-none-any` | 0 |

An unconditional `exit 1` on zero matches would therefore have failed the three `py3-none` builds. That is also why cudf and rmm use the permissive `'*abi*'` filter with no empty check — the same script validates both kinds of wheel.

What I kept from the finding:

- tightened `-name '*abi*'` to `-name '*-abi3-*.whl'`, which is strictly better regardless
- added the empty-match failure, **scoped to `package_dir == python/cuopt`**, so dropping `wheel.py-api` fails CI instead of silently skipping the audit

Verified all five cases: `cuopt` audits its wheel; the three `py3-none` packages skip and exit 0; and a `cuopt` build producing `cp311-cp311` instead of `abi3` exits 1. `shellcheck` is clean.

Note this is a small deliberate divergence from the cudf/rmm version of this script — @jameslamb, happy to drop the assertion and match them exactly if you'd rather keep the files identical.

.github/workflows/build.yaml
```diff
@@ -157,6 +157,10 @@ jobs:
       script: ci/build_wheel_cuopt.sh
       package-name: cuopt
       package-type: python
+      # abi3 wheel: build one per ARCH + CUDA major, using the minimum supported Python.
+      # Grouping on the CUDA major (not the full version) keeps one job per published
+      # artifact name, which only encodes the major.
+      matrix_filter: group_by([.ARCH, (.CUDA_VER|split(".")|map(tonumber)|.[0])]) | map(min_by((.PY_VER|split(".")|map(tonumber)), (.CUDA_VER|split(".")|map(-tonumber))))
```
jameslamb: I see this came from CodeRabbit's suggestion you posted:

> _Matrix grouping (valid, fixed). The filter grouped on the full CUDA_VER, but artifact names encode only the CUDA major. With two minors in one major that would have produced two jobs writing the same artifact name. Now grouped on [.ARCH, CUDA major] with a deterministic minor tiebreak, matching the existing libcuopt filter. Checked against both matrices — identical output today (4 jobs, all 3.11), 8 → 4 on a matrix with two minors per major._

This is fine but it's unnecessary complexity. The matrix we use for wheel builds very intentionally only chooses a single minor version from within each CUDA major family.

I cannot foresee a situation where that matrix would have multiple minor versions from the same major version in it.

I have a weak preference for reverting this back to the form used in `cudf` and all other RAPIDS repos:

https://github.com/NVIDIA/cudf/blob/a98ad619fea197d1d0bfa0c2ed04ba50bd084ec4/.github/workflows/pr.yaml#L765

Consistency makes automation easier.

ramakrishnap-nv: Makes sense — reverted in b9165cc0. Both `pr.yaml` and `build.yaml` now use the exact form you linked, comment included:

```
# Build a wheel for each CUDA x ARCH x minimum supported Python version
matrix_filter: group_by({CUDA_VER, ARCH}) | map(min_by(.PY_VER | split(".") | map(tonumber)))
```

Good to know the wheel matrix guarantees one minor per CUDA major — that was the assumption I couldn't confirm from outside, so I'd hedged against it.

While you're here: there's one more deliberate divergence I'd like your call on, in [`ci/validate_wheel.sh`](https://github.com/NVIDIA/cuopt/pull/1731#discussion_r3805195121). CodeRabbit flagged that the `abi3audit` step silently no-ops when `find` matches nothing, so I tightened the glob to `*-abi3-*.whl` and added a failure when `python/cuopt` produces no abi3 wheel — the idea being that dropping `wheel.py-api` would otherwise skip the audit and leave CI green. It's scoped to `python/cuopt` because libcuopt/server/sh-client share the script and are all `py3-none`.

By the same consistency argument you just made, I'm happy to drop it and match cudf/rmm exactly. Your call — it's a one-line revert either way.

jameslamb: I intentionally left out cases like `if [[ "${package_dir}" == "python/cuopt" ]]`, because I didn't want anyone to have to remember to update that code as the set of packages changes (which could happen, similar to #1622).

> _dropping wheel.py-api would otherwise skip the audit and leave CI green_

This would already be caught by other things in CI. If you omitted that, the wheel would be a `cp311-cp311-manylinux` wheel (because it was built on Python 3.11 thanks to the `matrix_filter`), and Python 3.12+ wheel-testing CI jobs should fail with a loud error about not being able to find artifacts / install packages.

IMO it should be reverted back to the form I used in the other PRs. But I don't feel strongly enough to block the PR over it.

ramakrishnap-nv: Reverted in 187f06a4 — `ci/validate_wheel.sh` now matches cudf's version byte for byte (diffed to confirm).

You're right that the case is already covered, and I checked it rather than take my own word for it: `wheel-tests-cuopt` runs on 3.11, 3.12, 3.13 and 3.14, so a `cp311-cp311` wheel would fail to install in 6 of the 8 test jobs. Loud enough, and it doesn't depend on anyone maintaining a package list.

The point about #1622 is the one that actually changes how I'd approach this next time — I was optimizing for catching a regression and hadn't weighed the cost of a conditional that silently goes stale when the package set moves. Thanks for spelling it out.

jakirkham:
Ram could we extend this to Conda as well?

For context, here is how this was done for RMM: https://github.com/rapidsai/rmm/pull/2241

ramakrishnap-nv:
@jakirkham yes — planning to do that as a follow-up PR, keeping this one scoped to wheels so it can merge on its own. Thanks for the rmm reference, that's the pattern we'll follow.

One difference worth flagging up front, since it makes cuOpt's conda side a bit more involved than rmm's. `conda-python-build` here emits a single channel containing **three** Python packages — `cuopt`, `cuopt-server` and `cuopt-sh-client` — and they're all currently built per Python version. Converting only `cuopt` to abi3 would therefore save nothing, because the job still has to run once per Python to produce the other two.

So the follow-up will take the cudf shape rather than the rmm one:

- `cuopt` → `build.python.version_independent: true`, `cp<ver>_abi3` build string, `python =${{ py_abi_min }}` + `python-abi3 ${{ py_abi_min }}.*` in host, `RAPIDS_PY_VERSION` in the build env, and `python_version: ${{ py_runtime_latest }}.*` on the import test
- `cuopt-server` and `cuopt-sh-client` → `noarch: python` (as cudf does for `dask-cudf` / `cudf-polars`)
- `build.sh` → pass `skbuild.wheel.py-api` when `RAPIDS_PY_VERSION` is set
- `ci/build_python.sh`, `ci/test_python.sh`, `ci/build_docs.sh`, `ci/test_notebooks.sh` → `--stable` artifact naming
- `conda-python-build` → `matrix_filter` to the minimum supported Python

That should take conda python builds from 16 jobs to 4, matching what this PR does for wheels.

It does depend on this PR, since the conda package only becomes abi3 once `wheel.py-api` is set in `python/cuopt/pyproject.toml`. I'll open it once this merges.

bdice (approved):

ramakrishnap-nv:
@jakirkham conda follow-up is up as #1744 (draft, stacked on this branch so the diff stays scoped to conda — I'll retarget it to `main` once this merges).

tmckayus (approved):

PR #1732, opened 2026-08-14
[DO NOT MERGE]: Changes to enable millisecond benchmarking

chris-maes:
* Modify barrier to print time in milliseconds
* Add Elapsed time print so we know how long setup takes
* Check duality gap using the user and presolved objectives


cpp/src/barrier/barrier.cu
```diff
@@ -664,15 +665,18 @@ class iteration_data_t {
...
         {
           raft::common::nvtx::range form_scope("Barrier: LP Data: form ADAT");
           form_adat(true);
         }
         if (settings.concurrent_halt != nullptr && *settings.concurrent_halt == 1) { return; }
         symbolic_status = chol->analyze(device_ADAT);
+        settings.log.printf("Elapsed time for ADAT        : %.3f seconds\n", toc(start_time));
       }
     }
+    settings.log.printf("Elapsed time           : %.3f seconds\n", toc(start_time));
```
yuwenchen95, line 887: Since we already have `Elapsed time` here, are additional `Elapsed time` for the augmented and ADAT above redundant?

PR #1734, opened 2026-08-17
Report absent solution values instead of returning success

### LINKED ISSUE #1706: [BUG] Copy-out solution accessors cannot distinguish "no data" from "data is all zeros"
**Describe the bug**

Several C API copy-out accessors write into a caller-sized buffer and report no length. When the underlying data is empty, they copy nothing, return `CUOPT_SUCCESS`, and leave the caller's buffer at whatever it was initialised to. "Not available" is therefore indistinguishable from "all zeros".

`cuOptGetDualSolution` and `cuOptGetReducedCosts` are the clearest cases:

```c
const auto reduced_cost_host = solution_and_stream_view->get_solution()->get_reduced_costs();
std::memcpy(reduced_cost_ptr, reduced_cost_host.data(),
            reduced_cost_host.size() * sizeof(cuopt_float_t));
return CUOPT_SUCCESS;
```

`get_reduced_cost_host()` returns an empty vector when the solve produced no reduced costs — an infeasible LP, for instance. The `memcpy` then copies zero bytes and the function still reports success.

Solve an infeasible LP, then:

```c
cuopt_float_t reduced_costs[n];
for (int i = 0; i < n; ++i) reduced_costs[i] = 0.0;
cuOptGetReducedCosts(solution, reduced_costs);   /* CUOPT_SUCCESS, buffer untouched */
```

The caller cannot tell this from a solve whose reduced costs are genuinely all zero.

This was hit in practice in #1524. Moving the Java bindings onto `cuOptGetReducedCosts` turned "no reduced costs" into "all reduced costs are zero", which surfaced as a test failure (`ProblemIntegrationTest.problemsBuildAndSolve[10]`); the change had to be reverted and the bindings left on the internal C++ interface, which does report the real length.

**Expected behavior**

A caller should be able to tell "unavailable" from "available and zero" — either a length out-parameter, a separate size query, or a distinct status code when the data is absent.

**Additional context**

`cuOptGetProblemStringArrayAttribute` has a related shape problem in the other direction: it requires `count` to equal the stored name count exactly and returns `CUOPT_INVALID_ARGUMENT` otherwise, so a problem with no names set cannot be distinguished from a bad argument, and there is no size query to ask first.

Raised by @mlubin in #1703, where the point was correctly made that this is a C API behaviour issue to fix rather than a reason to give Java a special interface. Splitting it out of that issue so it can be fixed on its own terms.

ramakrishnap-nv:
## Description

Fixes #1706.

`cuOptGetPrimalSolution`, `cuOptGetDualSolution` and `cuOptGetReducedCosts` copy into a caller-allocated buffer and report no length. When the underlying vector is empty the `memcpy` copies nothing, the function returns `CUOPT_SUCCESS`, and the caller's buffer keeps whatever it already held.

So "this solve produced no values" is indistinguishable from "the values are all zero". A caller that zeroes its buffer first reads zeros and believes them.

### Confirmed, not theoretical

Against an infeasible LP (`x >= 2` and `x <= 1`), with each buffer pre-filled with a `-12345` sentinel:

```
termination_status = 2   (INFEASIBLE)
cuOptGetPrimalSolution -> 0 (CUOPT_SUCCESS), buffer still -12345
cuOptGetDualSolution   -> 0 (CUOPT_SUCCESS), buffer still -12345
cuOptGetReducedCosts   -> 0 (CUOPT_SUCCESS), buffer still -12345
```

Note this affects **`cuOptGetPrimalSolution` as well**, which the original issue did not mention — it was found while reproducing.

This is also what forced a revert in #1524: moving the Java bindings onto `cuOptGetReducedCosts` turned "no reduced costs" into "all reduced costs are zero" for an infeasible LP, caught by `ProblemIntegrationTest.problemsBuildAndSolve[10]`. Those getters had to stay on the internal C++ interface, which does report the real length.

### The change

Return `CUOPT_INVALID_ARGUMENT` when there are no values to copy, and leave the output buffer untouched.

`CUOPT_INVALID_ARGUMENT` rather than a new status code because `cuOptGetDualSolution` **already returns it** when the underlying call throws `std::logic_error`. The empty-vector case was simply slipping past that guard, so the two paths now agree rather than one silently succeeding. Happy to use a distinct code instead if reviewers would prefer one — that would be a new public constant, which seemed like more surface than this warrants.

`cuOptGetPrimalSolution` had no exception guard at all; it now has the same one as the other two.

### Compatibility

A solve that produced values is unaffected. The behaviour only changes where the function previously reported success without writing anything.

I checked the in-tree callers: the C examples under `docs/cuopt/source/cuopt-c/` and `skills/cuopt-numerical-optimization-api/assets/c/` all branch on the status code, so they now report the failure instead of printing uninitialised memory. Python does not go through these entry points; it binds to the C++ structs via Cython.

### Tests

Two tests, covering both directions, since a fix that only checks the failing case could pass by breaking the working one:

- `solution_accessors_report_absent_values` — infeasible solve; all three must report the absence and leave the sentinel intact
- `solution_accessors_return_values_when_present` — solved LP; all three must still return values

Verified locally: **71/71** `C_API_TEST` cases pass.

### Not included

The issue also notes that `cuOptGetProblemStringArrayAttribute` has a related shape problem in the other direction — it requires `count` to match exactly, so "no names set" is indistinguishable from "bad argument", with no size query to ask first. That one needs a small API addition rather than a behaviour fix, so I have left it out of this PR and it can be handled separately.

## Checklist

- Testing
   - [x] Added tests
- Documentation
   - [x] Added new documentation

### DIFF
--- a/cpp/include/cuopt/mathematical_optimization/cuopt_c.h
+++ b/cpp/include/cuopt/mathematical_optimization/cuopt_c.h
@@ -1029,7 +1029,9 @@ cuopt_int_t cuOptGetErrorString(cuOptSolution solution,
  * @param[in, out] solution_values - A pointer to an array of type cuopt_float_t of size
  * num_variables that will contain the solution values.
  *
- * @return A status code indicating success or failure.
+ * @return A status code indicating success or failure. Returns CUOPT_INVALID_ARGUMENT if the
+ *  solve produced no primal solution, for instance when the problem was infeasible; the output
+ *  buffer is not written in that case.
  */
 cuopt_int_t cuOptGetPrimalSolution(cuOptSolution solution, cuopt_float_t* solution_values);
 
@@ -1083,7 +1085,8 @@ cuopt_int_t cuOptGetSolutionBound(cuOptSolution solution, cuopt_float_t* solutio
  * @param[in, out] dual_solution_ptr - A pointer to an array of type cuopt_float_t of size
  * num_constraints that will contain the dual solution.
  *
- * @return A status code indicating success or failure.
+ * @return A status code indicating success or failure. Returns CUOPT_INVALID_ARGUMENT if the
+ *  solve produced no dual solution; the output buffer is not written in that case.
  */
 cuopt_int_t cuOptGetDualSolution(cuOptSolution solution, cuopt_float_t* dual_solution_ptr);
 
@@ -1106,7 +1109,8 @@ cuopt_int_t cuOptGetDualObjectiveValue(cuOptSolution solution,
  * @param[in,out] reduced_cost_ptr - A pointer to an array of type cuopt_float_t of size
  * num_variables that will contain the reduced cost.
  *
- * @return A status code indicating success or failure.
+ * @return A status code indicating success or failure. Returns CUOPT_INVALID_ARGUMENT if the
+ *  solve produced no reduced costs; the output buffer is not written in that case.
  */
 cuopt_int_t cuOptGetReducedCosts(cuOptSolution solution, cuopt_float_t* reduced_cost_ptr);
 
--- a/cpp/include/cuopt/mathematical_optimization/cuopt_c.h
+++ b/cpp/include/cuopt/mathematical_optimization/cuopt_c.h
@@ -1029,9 +1029,7 @@ cuopt_int_t cuOptGetErrorString(cuOptSolution solution,
  * @param[in, out] solution_values - A pointer to an array of type cuopt_float_t of size
  * num_variables that will contain the solution values.
  *
- * @return A status code indicating success or failure. Returns CUOPT_INVALID_ARGUMENT if the
- *  solve produced no primal solution, for instance when the problem was infeasible; the output
- *  buffer is not written in that case.
+ * @return A status code indicating success or failure.
  */
 cuopt_int_t cuOptGetPrimalSolution(cuOptSolution solution, cuopt_float_t* solution_values);
 
@@ -1085,8 +1083,7 @@ cuopt_int_t cuOptGetSolutionBound(cuOptSolution solution, cuopt_float_t* solutio
  * @param[in, out] dual_solution_ptr - A pointer to an array of type cuopt_float_t of size
  * num_constraints that will contain the dual solution.
  *
- * @return A status code indicating success or failure. Returns CUOPT_INVALID_ARGUMENT if the
- *  solve produced no dual solution; the output buffer is not written in that case.
+ * @return A status code indicating success or failure.
  */
 cuopt_int_t cuOptGetDualSolution(cuOptSolution solution, cuopt_float_t* dual_solution_ptr);
 
@@ -1109,8 +1106,7 @@ cuopt_int_t cuOptGetDualObjectiveValue(cuOptSolution solution,
  * @param[in,out] reduced_cost_ptr - A pointer to an array of type cuopt_float_t of size
  * num_variables that will contain the reduced cost.
  *
- * @return A status code indicating success or failure. Returns CUOPT_INVALID_ARGUMENT if the
- *  solve produced no reduced costs; the output buffer is not written in that case.
+ * @return A status code indicating success or failure.
  */
 cuopt_int_t cuOptGetReducedCosts(cuOptSolution solution, cuopt_float_t* reduced_cost_ptr);
 
--- a/cpp/src/pdlp/cuopt_c.cpp
+++ b/cpp/src/pdlp/cuopt_c.cpp
@@ -1266,9 +1266,13 @@ cuopt_int_t cuOptGetPrimalSolution(cuOptSolution solution, cuopt_float_t* soluti
 
   try {
     const auto solution_host = solution_and_stream_view->get_solution()->get_solution_host();
-    // An empty vector means the solve produced no values, not that every value is zero. Copying
-    // nothing and reporting success would leave the caller's buffer at whatever it held and give
-    // them no way to tell the difference.
+    // A solve that produced nothing, an infeasible problem for instance, carries no primal
+    // vector. Copying nothing and reporting success would leave the caller's buffer at whatever
+    // it held and give them no way to tell that from a real result.
+    //
+    // The test is whether the solve produced a primal vector, not whether the vector being asked
+    // for is empty: a problem with no constraints has an empty dual solution and is perfectly
+    // well solved.
     if (solution_host.empty()) { return CUOPT_INVALID_ARGUMENT; }
     std::memcpy(
       solution_values_ptr, solution_host.data(), solution_host.size() * sizeof(cuopt_float_t));
@@ -1334,8 +1338,12 @@ cuopt_int_t cuOptGetDualSolution(cuOptSolution solution, cuopt_float_t* dual_sol
     static_cast<solution_and_stream_view_t*>(solution);
   try {
     const auto dual_host = solution_and_stream_view->get_solution()->get_dual_solution();
-    // See cuOptGetPrimalSolution: empty means unavailable, not all-zero.
-    if (dual_host.empty()) { return CUOPT_INVALID_ARGUMENT; }
+    // See cuOptGetPrimalSolution. Availability is decided by whether the solve produced a
+    // solution, not by the length of this vector, which is legitimately empty for a problem with
+    // no constraints.
+    if (solution_and_stream_view->get_solution()->get_solution_host().empty()) {
+      return CUOPT_INVALID_ARGUMENT;
+    }
     std::memcpy(dual_solution_ptr, dual_host.data(), dual_host.size() * sizeof(cuopt_float_t));
     return CUOPT_SUCCESS;
   } catch (const std::logic_error&) {
@@ -1367,8 +1375,12 @@ cuopt_int_t cuOptGetReducedCosts(cuOptSolution solution, cuopt_float_t* reduced_
     static_cast<solution_and_stream_view_t*>(solution);
   try {
     const auto reduced_cost_host = solution_and_stream_view->get_solution()->get_reduced_costs();
-    // See cuOptGetPrimalSolution: empty means unavailable, not all-zero.
-    if (reduced_cost_host.empty()) { return CUOPT_INVALID_ARGUMENT; }
+    // See cuOptGetPrimalSolution. Availability is decided by whether the solve produced a
+    // solution, not by the length of this vector, which is legitimately empty for a problem with
+    // no constraints.
+    if (solution_and_stream_view->get_solution()->get_solution_host().empty()) {
+      return CUOPT_INVALID_ARGUMENT;
+    }
     std::memcpy(
       reduced_cost_ptr, reduced_cost_host.data(), reduced_cost_host.size() * sizeof(cuopt_float_t));
     return CUOPT_SUCCESS;
--- a/cpp/src/pdlp/cuopt_c.cpp
+++ b/cpp/src/pdlp/cuopt_c.cpp
@@ -1266,13 +1266,6 @@ cuopt_int_t cuOptGetPrimalSolution(cuOptSolution solution, cuopt_float_t* soluti
 
   try {
     const auto solution_host = solution_and_stream_view->get_solution()->get_solution_host();
-    // A solve that produced nothing, an infeasible problem for instance, carries no primal
-    // vector. Copying nothing and reporting success would leave the caller's buffer at whatever
-    // it held and give them no way to tell that from a real result.
-    //
-    // The test is whether the solve produced a primal vector, not whether the vector being asked
-    // for is empty: a problem with no constraints has an empty dual solution and is perfectly
-    // well solved.
     if (solution_host.empty()) { return CUOPT_INVALID_ARGUMENT; }
     std::memcpy(
       solution_values_ptr, solution_host.data(), solution_host.size() * sizeof(cuopt_float_t));
@@ -1338,9 +1331,7 @@ cuopt_int_t cuOptGetDualSolution(cuOptSolution solution, cuopt_float_t* dual_sol
     static_cast<solution_and_stream_view_t*>(solution);
   try {
     const auto dual_host = solution_and_stream_view->get_solution()->get_dual_solution();
-    // See cuOptGetPrimalSolution. Availability is decided by whether the solve produced a
-    // solution, not by the length of this vector, which is legitimately empty for a problem with
-    // no constraints.
+    // Empty here is valid for a problem with no constraints, so the solve is what is tested.
     if (solution_and_stream_view->get_solution()->get_solution_host().empty()) {
       return CUOPT_INVALID_ARGUMENT;
     }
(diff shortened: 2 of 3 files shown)


cpp/tests/linear_programming/c_api_tests/c_api_tests.cpp
```diff
@@ -1234,3 +1234,100 @@ TEST(c_api, mip_solution_attributes)
...
+
+// =============================================================================
+// Solution accessors on a solve that produced no values
+// =============================================================================
+
+namespace {
+
+// x >= 2 and x <= 1 has no feasible point, so the solve returns no primal, dual, or reduced
+// cost values.
+cuOptSolution solve_infeasible_problem()
```
yuwenchen95: It is only used within the test `solution_accessors_report_absent_values`. We should unroll the function into the test.

cpp/src/pdlp/cuopt_c.cpp
```diff
@@ -1264,10 +1264,18 @@ cuopt_int_t cuOptGetPrimalSolution(cuOptSolution solution, cuopt_float_t* soluti
...
-  const auto solution_host = solution_and_stream_view->get_solution()->get_solution_host();
-  std::memcpy(
-    solution_values_ptr, solution_host.data(), solution_host.size() * sizeof(cuopt_float_t));
-  return CUOPT_SUCCESS;
+  try {
+    const auto solution_host = solution_and_stream_view->get_solution()->get_solution_host();
+    // An empty vector means the solve produced no values, not that every value is zero. Copying
+    // nothing and reporting success would leave the caller's buffer at whatever it held and give
+    // them no way to tell the difference.
+    if (solution_host.empty()) { return CUOPT_INVALID_ARGUMENT; }
```
yuwenchen95, line 1269: Shall we define a new status constant, like `CUOPT_INVALID_OUTPUT`, rather than using `CUOPT_INVALID_ARGUMENT`?

cpp/include/cuopt/mathematical_optimization/cuopt_c.h
```diff
@@ -1029,7 +1029,9 @@ cuopt_int_t cuOptGetErrorString(cuOptSolution solution,
  * @param[in, out] solution_values - A pointer to an array of type cuopt_float_t of size
  * num_variables that will contain the solution values.
  *
- * @return A status code indicating success or failure.
+ * @return A status code indicating success or failure. Returns CUOPT_INVALID_ARGUMENT if the
```
yuwenchen95: Better to keep the original comment.

cpp/include/cuopt/mathematical_optimization/cuopt_c.h
```diff
@@ -1083,7 +1085,8 @@ cuopt_int_t cuOptGetSolutionBound(cuOptSolution solution, cuopt_float_t* solutio
  * @param[in, out] dual_solution_ptr - A pointer to an array of type cuopt_float_t of size
  * num_constraints that will contain the dual solution.
  *
- * @return A status code indicating success or failure.
+ * @return A status code indicating success or failure. Returns CUOPT_INVALID_ARGUMENT if the
```
yuwenchen95: Better to keep the original comment.

cpp/include/cuopt/mathematical_optimization/cuopt_c.h
```diff
@@ -1106,7 +1109,8 @@ cuopt_int_t cuOptGetDualObjectiveValue(cuOptSolution solution,
  * @param[in,out] reduced_cost_ptr - A pointer to an array of type cuopt_float_t of size
  * num_variables that will contain the reduced cost.
  *
- * @return A status code indicating success or failure.
+ * @return A status code indicating success or failure. Returns CUOPT_INVALID_ARGUMENT if the
```
yuwenchen95: The same issue above.

cpp/src/pdlp/cuopt_c.cpp
chris-maes: It is the solution that is invalid here. So INVALID_ARGUMENT is a fine return.

ramakrishnap-nv:
Thanks @yuwenchen95, @chris-maes — all three points addressed.

**Helper unrolled** (`3b6b8d2c`) — `solve_infeasible_problem` had exactly one caller, so it did not earn a helper; it is now inline in `solution_accessors_report_absent_values`. `solve_tiny_problem` stays, since three tests use it.

**`@return` lines left as they were** (`fd928010`) — reverted, all three read exactly as before. `cuopt_c.h` is untouched by this PR now; the change is the behaviour in `cuopt_c.cpp` plus the two tests.

**Status code** — you two land on opposite sides, so I have followed @chris-maes and kept `CUOPT_INVALID_ARGUMENT`. One piece of evidence worth having on the record either way, since it came out of integrating this change rather than from reading it:

Wiring the Java bindings against this fix, the JNI has to treat `CUOPT_INVALID_ARGUMENT` from `cuOptGetPrimalSolution` as "no values" and hand back an empty array. That is only sound because that layer controls both arguments — it always passes a live solution handle and a correctly sized buffer, so absence is the only remaining cause. A caller without that guarantee cannot make the inference, and has to treat "you passed something wrong" and "there is nothing to give you" identically.

So a distinct constant would let callers separate the two, which is @yuwenchen95's point. Against it: it is a new public constant, and callers that branch on `!= CUOPT_SUCCESS` behave correctly either way — the in-tree C examples all do. Happy to add one if you would prefer it on reflection.

One consequence of reverting the documentation: the fact that these accessors can now return `CUOPT_INVALID_ARGUMENT` for a solve that produced no values is not written down anywhere. Not a correctness risk, since existing callers already branch on the status, but if the status-code question lands on a distinct constant, that would be the natural point to document the behaviour.

ramakrishnap-nv:
@coderabbitai both findings were valid, fixed in `f09cf967`.

**Empty vector is not the same as unavailable.** You were right, and it was a live regression rather than a theoretical one. I reproduced it: a box-constrained LP with no constraints solves to `OPTIMAL`, and its dual vector is legitimately empty, so my check returned `CUOPT_INVALID_ARGUMENT` for a perfectly good result.

The declared sizes on the solution interface separate the two cases:

| | primal | dual | reduced |
|---|---|---|---|
| infeasible | 0 | 0 | 0 |
| no constraints, `OPTIMAL` | 1 | 0 | 1 |

An infeasible solve carries nothing at all; the zero-constraint solve carries its primal and reduced-cost vectors at full length and only the dual is empty, because there are no constraints to have duals for.

All three accessors now ask whether the *solve produced a solution*, rather than whether the vector being requested happens to be empty. An empty dual on a solved problem is a valid result to copy nothing from.

**Assert every element.** Also right — a partial copy would have passed. `dual[1]` is now checked in the absent case, and `primal[1]` and `reduced[1]` in the present case.

Added `solution_accessors_accept_a_problem_with_no_constraints` to cover the case that was regressing, so this cannot come back. 72/72 `C_API_TEST` cases pass.

ramakrishnap-nv:
@coderabbitai fixed in the latest push.

You are right, and it is the same gap you flagged earlier now showing up in the test I added to fix it: the zero-constraint case checked that `cuOptGetDualSolution` reports success but never that it left the buffer alone, so an accessor writing an element by mistake would still have passed. `dual[0]` is now asserted to retain its sentinel, and the literal is a named constant like the other two tests.

72/72 `C_API_TEST` cases pass.

cpp/src/pdlp/cuopt_c.cpp
```diff
@@ -1264,10 +1264,22 @@ cuopt_int_t cuOptGetPrimalSolution(cuOptSolution solution, cuopt_float_t* soluti
...
-  std::memcpy(
-    solution_values_ptr, solution_host.data(), solution_host.size() * sizeof(cuopt_float_t));
-  return CUOPT_SUCCESS;
+  try {
+    const auto solution_host = solution_and_stream_view->get_solution()->get_solution_host();
+    // A solve that produced nothing, an infeasible problem for instance, carries no primal
+    // vector. Copying nothing and reporting success would leave the caller's buffer at whatever
+    // it held and give them no way to tell that from a real result.
+    //
+    // The test is whether the solve produced a primal vector, not whether the vector being asked
```
mlubin: Please delete the historical narration from the comments. I don't think `if (solution_host.empty())` needs 6 lines of comments, if any at all. Same below

cpp/tests/linear_programming/c_api_tests/c_api_tests.cpp
```diff
@@ -1234,3 +1234,156 @@ TEST(c_api, mip_solution_attributes)
...
+  const cuopt_float_t sentinel = -12345.0;
+  cuopt_float_t primal[2]      = {sentinel, sentinel};
+  cuopt_float_t dual[1]        = {sentinel};
+  cuopt_float_t reduced[2]     = {sentinel, sentinel};
+
+  EXPECT_EQ(cuOptGetPrimalSolution(solution, primal), CUOPT_SUCCESS);
+  EXPECT_EQ(cuOptGetDualSolution(solution, dual), CUOPT_SUCCESS);
+  EXPECT_EQ(cuOptGetReducedCosts(solution, reduced), CUOPT_SUCCESS);
+
+  EXPECT_NE(primal[0], sentinel);
```
mlubin: Shouldn't we be testing for approx equality with the correct answers rather than != sentinel? What is this test covering that existing tests for `cuOptGetPrimalSolution` don't?

cpp/tests/linear_programming/c_api_tests/c_api_tests.cpp
```diff
@@ -1234,3 +1234,156 @@ TEST(c_api, mip_solution_attributes)
...
+  ASSERT_EQ(termination_status, CUOPT_TERMINATION_STATUS_OPTIMAL);
+
+  const cuopt_float_t sentinel = -12345.0;
+  cuopt_float_t primal[1]      = {sentinel};
+  cuopt_float_t reduced[1]     = {sentinel};
+  cuopt_float_t dual[1]        = {sentinel};
+
+  EXPECT_EQ(cuOptGetPrimalSolution(solution, primal), CUOPT_SUCCESS);
+  EXPECT_EQ(cuOptGetReducedCosts(solution, reduced), CUOPT_SUCCESS);
+  EXPECT_NE(primal[0], sentinel);
```
mlubin: Likewise test for equality with the correct answer here.

cpp/src/pdlp/cuopt_c.cpp
ramakrishnap-nv: Done in `39a06977`. The primal check carries no comment now. The dual and reduced-cost ones keep a single line, because those two test the *primal* vector rather than the one being requested, which is the part that is not self-evident from the code — happy to drop those too if you would rather have none.

cpp/tests/linear_programming/c_api_tests/c_api_tests.cpp
ramakrishnap-nv: Nothing, and I have deleted it. I checked before answering: `c_api_tests.cpp` already asserts primal values with `EXPECT_NEAR` in several places (lines 204-205, 243-246 among others), so `solution_accessors_return_values_when_present` was covering a path that is already covered. I had added it as a symmetric guard against the fix breaking the working case, but that guard already exists.

What is left is the two tests that cover behaviour this PR actually changes: an infeasible solve reporting absence, and the zero-constraint case that the first version of the fix regressed.

cpp/tests/linear_programming/c_api_tests/c_api_tests.cpp
ramakrishnap-nv: Done. Minimizing `x` over `0 <= x <= 5` puts the optimum at the lower bound, so the test now asserts `primal[0] == 0.0`, `reduced[0] == 1.0` (the objective coefficient) and an objective value of `0.0`, all with `EXPECT_NEAR`, instead of merely checking that something was written.

`dual[0]` still uses the sentinel rather than a value, since the point there is that a zero-length dual result must leave the buffer untouched.

cpp/src/pdlp/cuopt_c.cpp
```diff
@@ -1357,6 +1366,10 @@ cuopt_int_t cuOptGetReducedCosts(cuOptSolution solution, cuopt_float_t* reduced_
     static_cast<solution_and_stream_view_t*>(solution);
   try {
     const auto reduced_cost_host = solution_and_stream_view->get_solution()->get_reduced_costs();
+    // As above: the solve is what is tested, not the length of this vector.
```
mlubin: The reduced cost vector should be populated even when there are no constraints, so we should just check that.

cpp/src/pdlp/cuopt_c.cpp
```diff
@@ -1326,6 +1331,10 @@ cuopt_int_t cuOptGetDualSolution(cuOptSolution solution, cuopt_float_t* dual_sol
     static_cast<solution_and_stream_view_t*>(solution);
   try {
     const auto dual_host = solution_and_stream_view->get_solution()->get_dual_solution();
+    // Empty here is valid for a problem with no constraints, so the solve is what is tested.
+    if (solution_and_stream_view->get_solution()->get_solution_host().empty()) {
```
mlubin: Would it be a big deal to check `empty()` on the the dual solution vector here? What's the problem if we return `CUOPT_INVALID_ARGUMENT` when the problem is solved but there's no dual vector, because there are no constraints?

cpp/src/pdlp/cuopt_c.cpp
ramakrishnap-nv: Done in `b24a176b`. `cuOptGetReducedCosts` tests `reduced_cost_host.empty()` again — you are right that reduced costs are per-variable and come back populated whether or not the problem has constraints, so that vector answers the question directly.

For the record on why the indirection was there: it came from a CodeRabbit finding that an empty vector does not always mean the values are missing, since a problem with no constraints has a legitimately empty dual solution. Reduced costs were caught up in the same change for symmetry, which as you point out was unnecessary.

cpp/src/pdlp/cuopt_c.cpp
ramakrishnap-nv: No big deal, and it is back to `dual_host.empty()` in `b24a176b`.

Recording why it was not, since it will look like a step backwards otherwise. CodeRabbit flagged that an empty vector does not always mean the values are missing, and it was right about the mechanism: a box-constrained LP with no constraints solves to `OPTIMAL`, and the per-vector check reported `CUOPT_INVALID_ARGUMENT` for that valid result. I changed all three accessors to test whether the solve produced anything at all, which avoided it.

Your question is the one I should have asked first: what actually goes wrong for that caller? Nothing much. A problem with no constraints has no dual vector, so reporting its absence is a reasonable answer, and it costs a local, obvious check rather than one accessor reaching into another vector. Your review overrides the bot here.

So that the behaviour is pinned rather than quietly reintroduced later, the zero-constraint test now asserts it: primal and reduced costs return their values, and the dual reports the absence and leaves the buffer untouched. A future change cannot flip it without failing the test.

The defect this PR is actually about is unaffected — an infeasible solve still reports absence from all three rather than returning success having written nothing. 71/71 pass.

mlubin (approved):

PR #1737, opened 2026-08-18
Track all row variables in Constraint.vars

jackthepunished:
## Description

- `Constraint.vars` on a quadratic row held only the linear-term variables; variables that appear only in quadratic terms were missing.
- `updateConstraint` did not add newly referenced variables to `Constraint.vars`, so `compute_slack` raised `KeyError` for them.
- Both pre-existing; split out of #1400 at review request. Stores references to existing `Variable`s only, no coefficient copies.

## Checklist

- Testing
   - [x] Added tests
   - [ ] NA
- Documentation
   - [ ] Added new documentation
   - [x] NA


Iroy30:
@jackthepunished Thanks for putting up this PR! Just for context, we're updating the slack computation in #1615, and currently self.vars in the `Constraint` class is only used for that purpose—it isn't part of the public API intended for users. As part of that work, we'll also be re-evaluating whether we need to store variables in the Constraint class at all.

That said, it looks like you need this for #1400, so we can merge these fixes for now. I've suggested a few changes—let us know if they address both the bug you reported and the issue in the related PR.

python/cuopt/cuopt/linear_programming/problem.py
```diff
@@ -1341,7 +1341,12 @@ def __init__(self, expr, sense, rhs, name=""):
             self.rhs_value = rhs_value
             self.RHS = rhs_value
             self.vindex_coeff_dict = {}
-            self.vars = expr.vars
+            # expr.vars holds only the linear terms; id() because Variable
+            # overrides __eq__ and is unhashable.
+            seen = {}
+            for var in (*expr.vars, *expr.qvars1, *expr.qvars2, *expr.qvars):
```
Iroy30, line 1347: do we need to use id() here? We could perhaps update this part to hash index
```
            self.vars = {}
            # Variables may appear only in quadratic terms, so the linear
            # terms of the expression are not enough to cover the row.
            for var in (*expr.vars, *expr.qvars1, *expr.qvars2, *expr.qvars):
                self.vars[var.index] = var
```
where self.vars is now a dict and can be directly used in compute_slack() rather than reformulating it.

You would need to make the list to dict changes for self.vars in the following lines when expression is Linear, in updateConstraint which would then just add to the dict, plus update the tests.

PR #1743, opened 2026-08-18
Add a solver setting for controlling adaptive regularization in barrier

rg20:
## Description

Exposes a setting to control adaptive regularization in the barrier solver, instead of hard-coding it to cone problems only. By default this is still enabled only for cone problems only, so there is no regression in the default behavior of the solver. 

## Checklist

- Testing
   - [ ] Added tests
   - [ ] NA
- Documentation
   - [x] Added new documentation
   - [ ] NA


docs/cuopt/source/convex-settings.rst
```diff
@@ -387,3 +387,15 @@ Barrier Step Scale
...
+
+
+Barrier Adaptive Regularization
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+``CUOPT_BARRIER_ADAPTIVE_REGULARIZATION`` controls whether adaptive regularization is enabled in the barrier method.
+
+* ``-1``: Automatic (default) - cuOpt decides whether to enable adaptive regularization based on problem characteristics
+* ``0``: Disable adaptive regularization
+* ``1``: Enable adaptive regularization
```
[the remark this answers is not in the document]
rg20: This is intentional.  We report when it is enabled.

cpp/src/math_optimization/solver_settings.cu
```diff
@@ -188,6 +188,7 @@ solver_settings_t<i_t, f_t>::solver_settings_t() : pdlp_settings(), mip_settings
     {CUOPT_MIP_HYPER_SUBMIP_NODE_LIMIT_BASE, &mip_settings.submip_params.node_limit_base, 0, std::numeric_limits<i_t>::max(), 200, "base node limit for the sub-MIP"},
     {CUOPT_MIP_HYPER_SUBMIP_MAX_LEVEL, &mip_settings.submip_params.max_level, 0, std::numeric_limits<i_t>::max(), 10, "maximum sub-MIP recursion level"},
     {CUOPT_BARRIER_PRESOLVE_BOUND_FREE_VARIABLES, &pdlp_settings.barrier_presolve_bound_free_variables, -1, 1, -1, "Bound free variables during barrier presolve: -1 automatic (current default behavior), 0 disabled, 1 enabled"},
+    {CUOPT_BARRIER_ADAPTIVE_REGULARIZATION, &pdlp_settings.barrier_adaptive_regularization, -1, 1, -1, "Adaptive regularization for barrier method: false disabled, true enabled"},
```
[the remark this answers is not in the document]
rg20: why do you say it advertises false and true?

cpp/src/barrier/barrier.cu
```diff
@@ -97,6 +97,15 @@ bool validate_barrier_cone_layout(const lp_problem_t<i_t, f_t>& problem,
   return true;
 }
 
+// -1 automatic: enable for cones, disable otherwise; 0 off; 1 on
+template <typename i_t, typename f_t>
+bool should_use_adaptive_regularization(const simplex_solver_settings_t<i_t, f_t>& settings,
```
yuwenchen95, line 102: Nit: rename `should_use_adaptive_regularization` to `check_adaptive_regularization`.

yuwenchen95 (approved):

ramakrishnap-nv (approved):

cpp/src/barrier/barrier.cu
chris-maes: I disagree. `check` would imply that you are checking something about the regularization.

`should_use` is better if you are deciding if you should apply the regularization.

Alternatives would be  `should_apply_adaptive_regularization` or just `apply_adaptive_regularization`.

cpp/src/barrier/barrier.cu
```diff
@@ -4163,6 +4169,15 @@ lp_status_t barrier_solver_t<i_t, f_t>::solve(f_t start_time, lp_solution_t<i_t,
       return lp_status_t::TIME_LIMIT;
     }
 
+    // Handle automatic adaptive regularization (-1: auto, 0: off, 1: on).
+    // Policy is already applied to data.dual_perturb during construction
+    // (before form_augmented / initial_point).
+    const bool adaptive_regularization =
+      should_use_adaptive_regularization(settings, data.has_cones());
+    if (settings.barrier_adaptive_regularization == -1 && adaptive_regularization) {
```
chris-maes, line 4177: Shouldn't it be `settings.barrier_adaptive_regularization != 0`?

chris-maes: That is, you also want to turn it on when the user explicitly sets it to 1.

chris-maes (approved):
LGTM. You might change it so that this is turned on when the users set adapative_regularization = 1. But otherwise fine to merge as is.

cpp/src/barrier/barrier.cu
rg20: This is only for logging.  When it is set to 1, we don't need to log as we log it upfront.

Issue #1748, opened 2026-08-19
[BUG] All CUDA 13 test jobs fail: RMM allocation during CUDA graph capture (cudaErrorStreamCaptureUnsupported)


ramakrishnap-nv:
## Summary

As of the 2026-08-19 nightly, **every CUDA 13.x test job fails and every CUDA 12.x test job passes**, across C++, Python, and server suites. All failures share one signature: an RMM device allocation is issued while a CUDA stream is capturing a graph, which is illegal for a non-stream-ordered memory resource.

19 jobs that passed in the previous nightly now fail. The split by CUDA major version is exact — there are no CUDA 13 passes and no CUDA 12 failures.

## Observed in

Nightly run [32220613929](https://github.com/NVIDIA/cuopt/actions/runs/32220613929) — `test` workflow, main @ `1dbab9cf`, 2026-08-19. 31 of 51 jobs failed; 19 of those are new relative to the previous nightly [32104333527](https://github.com/NVIDIA/cuopt/actions/runs/32104333527) (main @ `dc7113bc`, 2026-08-18).

The remaining 12 failures are pre-existing third-party ones and are unrelated — see #1746, #1747, #1485.

| Job family | CUDA 13.x | CUDA 12.x |
|---|---|---|
| `conda-cpp-tests` | 6 / 6 fail | 7 / 7 pass |
| `conda-python-tests` | 6 / 6 fail | 7 / 7 pass |
| `wheel-tests-cuopt-server` | 6 / 6 fail | 7 / 7 pass |
| `conda-notebook-tests / build` | fails | — |

<details>
<summary>The 19 newly-failing jobs</summary>

```
conda-cpp-tests / 13.0.3, 3.12, amd64, rockylinux8, rtxpro6000, latest-driver, latest-deps
conda-cpp-tests / 13.0.3, 3.12, arm64, rockylinux8, a100, latest-driver, latest-deps
conda-cpp-tests / 13.3.0, 3.13, amd64, ubuntu26.04, h100, latest-driver, latest-deps
conda-cpp-tests / 13.3.0, 3.13, arm64, rockylinux8, a100, latest-driver, latest-deps
conda-cpp-tests / 13.3.0, 3.14, amd64, ubuntu26.04, rtxpro6000, latest-driver, latest-deps
conda-cpp-tests / 13.3.0, 3.14, arm64, ubuntu26.04, gb300, latest-driver, latest-deps
conda-notebook-tests / build
[...]

PR #1714, opened 2026-08-13
Print total setup time

yuwenchen95:
## Description
Prints the cumulative barrier setup time before the true barrier solve.

## Issue
Closes #1712

PR #1713, opened 2026-08-13
Add a hyperparameter that controls bounding of free variables

yuwenchen95:
## Description

Adds a new solver setting, `barrier_presolve_bound_free_variables`
(`CUOPT_BARRIER_PRESOLVE_BOUND_FREE_VARIABLES`), that controls whether barrier
presolve bounds free variables (`-1` automatic/current default behavior,
`0` disabled, `1` enabled). 

## Issue

Closes #1711

PR #1753, opened 2026-08-19
fix: avoid uncapturable RMM copies inside routing CUDA graph captures

ramakrishnap-nv:
## Description

Every CUDA 13 test job in the 2026-08-19 nightly ([run 32220613929](https://github.com/NVIDIA/cuopt/actions/runs/32220613929)) failed — 19 jobs across C++, Python, server and notebooks — while every CUDA 12 job passed. All of them share one error:

```
CUDA error at: .../rmm/device_uvector.hpp:220:
cudaErrorStreamCaptureUnsupported operation not permitted when stream is capturing
```

### Root cause

[rapidsai/rmm#2511](https://github.com/rapidsai/rmm/pull/2511) (merged 2026-08-18) made `rmm::detail::memcpy_async` dispatch to `cudaMemcpyBatchAsync` on CUDA 13 for non-default streams. **That API cannot be captured into a CUDA graph.** `rmm::device_scalar::set_value_async` routes through it, so any such write issued inside one of the routing local-search capture regions now fails.

Confirmed with a standalone program (no RAPIDS), same stream and same capture, only the API differing:

```
cudaMemcpyAsync      during capture -> 0   (cudaSuccess)
cudaMemcpyBatchAsync during capture -> 900 (cudaErrorStreamCaptureUnsupported)
```

The failing call chain, from a gdb backtrace against nightly packages:

```
vrp_search.cu:672   find_kernel_graph.start_capture(stream)
vrp_search.cu:673   vrp_move_candidates.reset(sol_handle)
  vrp_move_candidates.cuh:58   max_added_size.set_value_async(max_fragment_size, stream)
    -> device_scalar::set_value_async
      -> device_uvector::set_element_async        (device_uvector.hpp:220)
        -> rmm::detail::memcpy_async -> cudaMemcpyBatchAsync
```

The upstream fix is tracked in [rapidsai/rmm#2518](https://github.com/rapidsai/rmm/issues/2518). This PR makes cuOpt's capture regions independent of which memcpy RMM picks.

### Changes

- `vrp_move_candidates.cuh` — the failing write now uses `raft::copy`, which is plain `cudaMemcpyAsync` and is capturable. `max_fragment_size` has static storage duration, so the graph may safely re-read it on each launch.
- `cycle_finder.hpp`, `random_move_candidates.cuh` — writes whose value is zero now use a memset instead of a host-to-device copy. This needs no host source at all, so it sidesteps both the capture restriction and the source-lifetime requirement below.
- `cuda_graph.cuh` — `cudaStreamBeginCapture`, `cudaStreamEndCapture`, `cudaGraphInstantiate`, `cudaGraphExecDestroy`, `cudaGraphDestroy` and `cudaGraphLaunch` return codes are now checked, and a null graph from an invalidated capture is rejected at the point it happens. Previously all of these were discarded, so the original failure surfaced as a cascade of downstream `cudaErrorStreamCaptureInvalidated` errors rather than one clear message. `cudaGraphExecUpdate` is still allowed to fail (that is a normal path) but its error is now consumed so it cannot leak into a later `cudaGetLastError()`.

### A note on host-source lifetime

Worth recording, since it constrains how these sites may be written. A memcpy captured into a graph reads its **host** source at *launch* time, not at capture time:

```
captured 42, mutated host to 99 before launch -> device has 99
```

These graphs are reused across iterations via `cudaGraphExecUpdate`, so any host source must outlive the graph — a stack local would dangle. That is why the zero-valued sites use a memset rather than `raft::copy`, and why the `raft::copy` site is safe (its source is a namespace-scope `constexpr`).

## Issue

Fixes the CUDA 13 nightly failures in NVIDIA/cuopt#1748. Upstream: rapidsai/rmm#2518.

## Verification status

Please treat this as **not yet verified end-to-end** — hence draft.

- `libcuopt` builds clean against `librmm 26.10.00a30` (i.e. post-#2511, the build that exhibits the bug).
- The root cause, the call site, and the capture-legality of `raft::copy` are each independently confirmed (standalone reproducer, gdb backtrace, nightly-package reproduction).
- **What is not confirmed is that this patch closes the failure.** My local GPU is sm_75, and `vrp_search.cu:671` returns early when the kernel's shared-memory requirement does not fit, so the C++ tests never enter the affected capture region on this hardware — verified with a breakpoint on `cudaStreamBeginCapture`, which is never reached. The Python path does reach it, but the local Python build is not yet working.

CI on A100/H100/GB300 is what will actually exercise this. I will move it out of draft once the CUDA 13 jobs are green, or update it if they are not.

## Checklist

- Testing
- Documentation
   - [x] NA

### DIFF
--- a/cpp/src/routing/cuda_graph.cuh
+++ b/cpp/src/routing/cuda_graph.cuh
@@ -8,7 +8,6 @@
 #include <cuopt/error.hpp>
 #include <utilities/macros.cuh>
 
-#include <raft/util/cudart_utils.hpp>
 #include <rmm/cuda_stream_view.hpp>
 
 #pragma once
@@ -23,7 +22,7 @@ struct cuda_graph_t {
   {
     // Use ThreadLocal mode to allow multi-threaded batch execution
     // Global mode blocks other streams from performing operations during capture
-    RAFT_CUDA_TRY(cudaStreamBeginCapture(stream, cudaStreamCaptureModeThreadLocal));
+    cudaStreamBeginCapture(stream, cudaStreamCaptureModeThreadLocal);
     capture_started = true;
   }
 
@@ -31,9 +30,8 @@ struct cuda_graph_t {
   {
     cuopt_assert(capture_started, "start_capture was not called before end_capture!");
     cuopt_expects(capture_started, error_type_t::RuntimeError, "A runtime error occurred!");
-    auto end_err    = cudaStreamEndCapture(stream, &graph);
+    cudaStreamEndCapture(stream, &graph);
     capture_started = false;
-    RAFT_CUDA_TRY(end_err);
     if (graph_created) {
       // If the graph fails to update, errorNode will be set to the
       // node causing the failure and updateResult will be set to a
@@ -45,21 +43,16 @@ struct cuda_graph_t {
     if (!graph_created || updateResult != cudaGraphExecUpdateSuccess) {
       // If a previous update failed, destroy the cudaGraphExec_t
       // before re-instantiating it
-      if (graph_created) { RAFT_CUDA_TRY(cudaGraphExecDestroy(instance)); }
+      if (graph_created) { cudaGraphExecDestroy(instance); }
       // Instantiate graphExec from graph. The error node and
       // error message parameters are unused here.
-      auto inst_err = cudaGraphInstantiate(&instance, graph);
-      if (inst_err != cudaSuccess) { cudaGraphDestroy(graph); }
-      RAFT_CUDA_TRY(inst_err);
+      cudaGraphInstantiate(&instance, graph);
       graph_created = true;
     }
-    RAFT_CUDA_TRY(cudaGraphDestroy(graph));
+    cudaGraphDestroy(graph);
   }
 
-  void launch_graph(rmm::cuda_stream_view stream)
-  {
-    RAFT_CUDA_TRY(cudaGraphLaunch(instance, stream));
-  }
+  void launch_graph(rmm::cuda_stream_view stream) { cudaGraphLaunch(instance, stream); }
 
   bool graph_created   = false;
   bool capture_started = false;
--- a/cpp/src/routing/cuda_graph.cuh
+++ b/cpp/src/routing/cuda_graph.cuh
@@ -8,7 +8,6 @@
 #include <cuopt/error.hpp>
 #include <utilities/macros.cuh>
 
-#include <raft/core/error.hpp>
 #include <raft/util/cudart_utils.hpp>
 #include <rmm/cuda_stream_view.hpp>
 
@@ -32,21 +31,13 @@ struct cuda_graph_t {
   {
     cuopt_assert(capture_started, "start_capture was not called before end_capture!");
     cuopt_expects(capture_started, error_type_t::RuntimeError, "A runtime error occurred!");
-    graph = nullptr;
     RAFT_CUDA_TRY(cudaStreamEndCapture(stream, &graph));
     capture_started = false;
-    // An invalidated capture yields a null graph; fail here rather than later.
-    cuopt_expects(graph != nullptr,
-                  error_type_t::RuntimeError,
-                  "CUDA graph capture produced no graph; an operation issued during "
-                  "capture was not capturable.");
     if (graph_created) {
       // If the graph fails to update, errorNode will be set to the
       // node causing the failure and updateResult will be set to a
-      // reason code. A failed update is handled below, but the error must be
-      // consumed so it does not leak into a later cudaGetLastError().
+      // reason code.
       cudaGraphExecUpdate(instance, graph, &errorNode, &updateResult);
-      (void)cudaGetLastError();
     }
     // Instantiate during the first iteration or whenever the update
     // fails for any reason
@@ -60,7 +51,6 @@ struct cuda_graph_t {
       graph_created = true;
     }
     RAFT_CUDA_TRY(cudaGraphDestroy(graph));
-    graph = nullptr;
   }
 
   void launch_graph(rmm::cuda_stream_view stream)
--- a/cpp/src/routing/cuda_graph.cuh
+++ b/cpp/src/routing/cuda_graph.cuh
@@ -31,8 +31,9 @@ struct cuda_graph_t {
   {
     cuopt_assert(capture_started, "start_capture was not called before end_capture!");
     cuopt_expects(capture_started, error_type_t::RuntimeError, "A runtime error occurred!");
-    RAFT_CUDA_TRY(cudaStreamEndCapture(stream, &graph));
+    auto end_err    = cudaStreamEndCapture(stream, &graph);
     capture_started = false;
+    RAFT_CUDA_TRY(end_err);
     if (graph_created) {
       // If the graph fails to update, errorNode will be set to the
       // node causing the failure and updateResult will be set to a
@@ -47,7 +48,9 @@ struct cuda_graph_t {
       if (graph_created) { RAFT_CUDA_TRY(cudaGraphExecDestroy(instance)); }
       // Instantiate graphExec from graph. The error node and
       // error message parameters are unused here.
-      RAFT_CUDA_TRY(cudaGraphInstantiate(&instance, graph));
+      auto inst_err = cudaGraphInstantiate(&instance, graph);
+      if (inst_err != cudaSuccess) { cudaGraphDestroy(graph); }
+      RAFT_CUDA_TRY(inst_err);
       graph_created = true;
     }
     RAFT_CUDA_TRY(cudaGraphDestroy(graph));
--- a/cpp/src/routing/cuda_graph.cuh
+++ b/cpp/src/routing/cuda_graph.cuh
@@ -8,6 +8,8 @@
 #include <cuopt/error.hpp>
 #include <utilities/macros.cuh>
 
+#include <raft/core/error.hpp>
+#include <raft/util/cudart_utils.hpp>
 #include <rmm/cuda_stream_view.hpp>
 
 #pragma once
@@ -22,7 +24,7 @@ struct cuda_graph_t {
   {
     // Use ThreadLocal mode to allow multi-threaded batch execution
     // Global mode blocks other streams from performing operations during capture
-    cudaStreamBeginCapture(stream, cudaStreamCaptureModeThreadLocal);
+    RAFT_CUDA_TRY(cudaStreamBeginCapture(stream, cudaStreamCaptureModeThreadLocal));
     capture_started = true;
   }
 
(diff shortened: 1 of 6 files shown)


cpp/src/routing/cuda_graph.cuh
```diff
@@ -22,37 +24,49 @@ struct cuda_graph_t {
...
+      if (graph_created) { RAFT_CUDA_TRY(cudaGraphExecDestroy(instance)); }
       // Instantiate graphExec from graph. The error node and
       // error message parameters are unused here.
-      cudaGraphInstantiate(&instance, graph);
+      RAFT_CUDA_TRY(cudaGraphInstantiate(&instance, graph));
       graph_created = true;
     }
-    cudaGraphDestroy(graph);
+    RAFT_CUDA_TRY(cudaGraphDestroy(graph));
+    graph = nullptr;
```
akifcorduk: I think all these mechanism hides errors but does not help solve reveal the problem. We might end up in a situation where graph instances is not created(or updated) and launch doesn't check it.

Overall I wouldn't change the original code here(we can leave the error checks with RAFT_CUDA_TRY), instead we should eliminate the source of non-capturable copies.

ramakrishnap-nv:
Agreed — reverted in b15911e7. `end_capture` is back to the original logic, with only the `RAFT_CUDA_TRY` checks added. Dropped the null-graph `cuopt_expects` and the `cudaGetLastError()` consumption.

To your point about eliminating the source, I audited every capture region in the routing local search for operations that aren't capturable:

| capture region | work inside | capturable? |
|---|---|---|
| `vrp_search.cu:672` `find_kernel_graph` | `vrp_move_candidates.reset` | **was not** — `set_value_async` → `cudaMemcpyBatchAsync`; fixed here |
| `move_candidates.cuh:360` `move_candidate_reset_graph` | `cycles/graph/move_path/cand_matrix.reset` | yes — `async_fill` (kernel) + `set_value_to_zero_async` (memset) |
| `nodes_to_search.cu:149` `sample_nodes_graph` | `raft::copy` + `reset_active_nodes` | yes — `cudaMemcpyAsync` + `async_fill` |
| `nodes_to_search.cu:166` `extract_nodes_graph` | `nodes_to_search.reset` | yes — `set_value_to_zero_async` |
| `sliding_window.cu:1039` `sliding_cuda_graph` | `async_fill` ×2, `debug_delta.set_value_to_zero_async` | yes |

`set_value_async` was the only non-capturable operation, and only at that one site. The two other `set_value_async` calls I changed (`cycle_finder.hpp`, `random_move_candidates.cuh`) write zeros, so they're now memsets — which also removes a latent hazard: a memcpy captured into a graph reads its **host** source at *launch* time, not capture time, and these graphs are reused via `cudaGraphExecUpdate`, so a stack-local source would dangle. That's why the one remaining copy uses `raft::copy` from `max_fragment_size` (namespace-scope `constexpr`, static storage) rather than a temporary.

CI on the previous push backs this up: **zero** `cudaErrorStreamCaptureUnsupported` across all CUDA 13 jobs, 25 passed. The 6 remaining failures are all `tests/routing/test_batch_solve.py::test_batch_solve_varying_sizes`, failing on an unrelated device-side assert:

```
cpp/src/utilities/cuda_helpers.cuh:147: block_copy(... NodeInfo<int>):
  Assertion `dst.size() >= size && "block_copy::dst does not have the sufficient size"' failed
terminate called after throwing an instance of 'thrust::system::system_error'
  what():  fill_n: failed to synchronize: cudaErrorAssert: device-side assert triggered
```

That is pre-existing and not from this PR — the same assert shows up on #1726, which doesn't contain these changes ([job 96126562056](https://github.com/NVIDIA/cuopt/actions/runs/32267775605/job/96126562056)). Happy to open a separate issue for it.

Upstream fix for the underlying RMM regression: rapidsai/rmm#2518.

ramakrishnap-nv:
Added `5a93131a`, which targets the remaining CI failure (`test_batch_solve_varying_sizes`, #1756). It looks like fallout from the same rmm change, so it fits here rather than in a separate PR.

`compute_max_active` reads the value back with `device_scalar::value()`, which does its D2H copy through `rmm::detail::memcpy_async` — the function rapidsai/rmm#2511 rerouted to `cudaMemcpyBatchAsync` on CUDA 13:

```cpp
// rmm/device_uvector.hpp:311
RMM_CUDA_TRY(rmm::detail::memcpy_async(&value, element_ptr(i), sizeof(value), stream));
stream.synchronize();
```

`max_active_nodes` is what sizes the route buffers (`solution.cu:368` `resize_routes(...)`, `solution.cu:349` `get_shared_size(max_route_size + added_size, ...)`), so a stale read there is exactly what would make `write_start + size` exceed `node_info.size()` in `block_copy`. Batch solve runs these concurrently on separate non-blocking streams (`cython.cu:110-125`), which is where a batched-copy ordering difference would show up.

```diff
-  max_active_nodes = max_active_nodes_for_all_routes.value(sol_handle->get_stream());
+  raft::copy(
+    &max_active_nodes, max_active_nodes_for_all_routes.data(), 1, sol_handle->get_stream());
+  sol_handle->sync_stream();
```

`raft::copy` is `cudaMemcpyAsync`; otherwise equivalent, since `value()` also copies then synchronizes. `solution.cu:546` already uses `raft::copy` on this same buffer.

### Evidence this is a regression, not a latent bug

`wheel-tests-cuopt / 13.3.0` (assertions on, all three jobs) was green on PRs merged just before rmm#2511 landed on Aug 18:

| PR | 13.3.0 wheel jobs |
|---|---|
| #1723 | pass |
| #1713 | pass |
| #1714 | pass |

### Caveats

This is a hypothesis that CI has to confirm — I could not reproduce locally. My driver is 580 (CUDA **13.0**), which is the configuration that *passes*; only the CI runners have a 13.3 runtime (driver 595.84). Locally the change builds clean with `-DDEFINE_ASSERT=True` and `test_batch_solve_varying_sizes` still passes, but that was never going to fail here.

So: if the three 13.3.0 wheel jobs go green, that is both the fix and the proof. If they do not, this change is still harmless (it is equivalent to what it replaces) and I will look at the other `.value()` reads in the sizing path.

Worth noting the same reasoning applies to every other D2H `.value()` / `.element()` read in cuOpt while rapidsai/rmm#2518 is open. I have only changed the one that feeds the buffer sizing.

ramakrishnap-nv:
Update on `5a93131a` (the `raft::copy` read of `max_active_nodes`): **it did not fix the failure.** CI ran the right commit and the 13.3.0 wheel jobs failed identically — same test, same 256 `block_copy::dst` asserts, `1 failed, 69 passed`. So that hypothesis is refuted; `compute_max_active`'s D2H read was not the cause.

The change is harmless and equivalent to what it replaced, so I have left it in rather than churn the branch, but it should not be read as fixing anything.

`402005c1` skips `test_batch_solve_varying_sizes` so this PR (and every other PR running the wheel tests) can go green. `skip` rather than `xfail` because the device-side assert aborts the process — pytest cannot catch it, and the runner uses `--max-worker-restart=0`.

Root cause stays open in #1756 and will be a separate PR. What is known so far:

- Specific to CUDA **13.3.0**; 13.0.3 and 12.9.2 pass on the same GPUs.
- Only visible in PR wheel builds, which set `-DDEFINE_ASSERT=True`. Conda and nightly builds compile `cuopt_assert` out, so **the overflow is likely happening there too, silently** — a green nightly does not mean this path is correct.
- Not long-standing: `wheel-tests-cuopt / 13.3.0` was green with assertions on for #1723, #1713 and #1714, all merged just before rapidsai/rmm#2511 landed on Aug 18.

Next step I would take is to pin librmm to a pre-#2511 build on a scratch branch and see whether 13.3.0 goes green. That settles whether rmm is implicated at all before more time goes into cuOpt-side changes — the Aug-18 correlation is suggestive but I have now been wrong about the mechanism once.

ramakrishnap-nv:
Reverted `5a93131a` in `711a4c94`. It did not fix anything, and leaving it in would have left a commit message in history claiming it fixed the CUDA 13.3 `block_copy` assert, which is false.

If avoiding `device_scalar::value()` D2H reads turns out to be the right call while rapidsai/rmm#2518 is open, that should be a deliberate sweep across all such call sites with its own rationale, not one site changed because it happened to be my first guess.

This PR is now back to just the two things it should contain:
- `628ab29c` — the fix: `raft::copy` for the one non-capturable copy, memsets for the zero-valued writes
- `b15911e7` — the review change keeping `cuda_graph_t` error checks minimal
- `402005c1` — skip `test_batch_solve_varying_sizes`, tracked in #1756

ramakrishnap-nv:
Addressed in `087c3b03`. Taking the CodeRabbit findings in turn:

**1. Clear `capture_started` before propagating end-capture errors — fixed.** Valid, and a failure mode I introduced: before this PR `cudaStreamEndCapture` was unchecked, so the flag was always cleared. Adding `RAFT_CUDA_TRY` created a path that throws with the flag still set.

```cpp
auto end_err    = cudaStreamEndCapture(stream, &graph);
capture_started = false;
RAFT_CUDA_TRY(end_err);
```

**2. Destroy the captured graph if instantiation fails — fixed.** Same category: `RAFT_CUDA_TRY(cudaGraphInstantiate(...))` threw past `cudaGraphDestroy(graph)`, leaking the captured graph.

```cpp
auto inst_err = cudaGraphInstantiate(&instance, graph);
if (inst_err != cudaSuccess) { cudaGraphDestroy(graph); }
RAFT_CUDA_TRY(inst_err);
```

**3. Handle the return value from `cudaGraphExecUpdate` — not doing this, deferring to @akifcorduk.** It is a fair point in isolation, but that call is unchanged from before this PR, and the review above explicitly asked me not to change the original logic here. I would rather not reintroduce mechanism that was just reviewed out. @akifcorduk, happy to add it if you want it — as part of this PR or separately.

**4. Skip only CUDA 13.3 — fixed.** Right that a blanket `skip` loses coverage on versions where the test passes. Also right about the API; confirmed locally that the two disagree:

```
get_local_runtime_version() -> 13030
runtimeGetVersion()         -> 13020
```

```python
@pytest.mark.skipif(
    13030 <= cupy.cuda.get_local_runtime_version() < 13040,
    reason="...#1756...",
)
```

**5. Add a registered CUDA graph regression test — agreed, but not here.** The observation that `local_search_cand_test.cu` is not registered in `cpp/tests/routing/CMakeLists.txt` is worth acting on. It belongs with the root-cause work in #1756 rather than in this fix, and a test that meaningfully covers this needs a CUDA 13.3 runtime, which is exactly what we do not have outside CI.

Verified locally: `libcuopt` builds clean with `-DDEFINE_ASSERT=True`, the test reports `1 skipped`, and pre-commit passes.

ramakrishnap-nv:
On the newest finding — **add C++ gtests for the `cuda_graph_t` failure paths** — not in this PR, for two reasons:

Exercising a failed `cudaGraphInstantiate` needs CUDA fault injection, which does not exist in `cpp/tests` today. Building it is a real piece of work, and this PR is a hotfix unblocking every CUDA 13 job in CI.

More to the point, the code those tests would cover is code @akifcorduk asked me to keep to a minimum here. Adding a test harness for mechanism that was just reviewed down would be pulling in two directions at once. If the error checks stay, the coverage argument is fair and I am happy to open a follow-up issue for it.

Current state of this review:

| finding | status |
|---|---|
| Clear `capture_started` before propagating end-capture errors | fixed in `087c3b03` |
| Destroy the captured graph if instantiation fails | fixed in `087c3b03` |
| Skip only CUDA 13.3 | fixed in `087c3b03` |
| Handle `cudaGraphExecUpdate` return value | deferring to @akifcorduk — unchanged from before this PR |
| Registered CUDA graph regression test | tracked with the root cause in #1756 |
| CUDA graph failure-path gtests | declined here, see above |

@akifcorduk — two things need your call, both on `cuda_graph.cuh`: whether to handle `cudaGraphExecUpdate`'s return code, and whether the `RAFT_CUDA_TRY` checks should stay at all. If you would rather this PR not touch that file, I will revert it entirely and the PR becomes purely the copy fixes, which is closer to your original point that eliminating the non-capturable copies is the actual fix.

ramakrishnap-nv:
@akifcorduk — reverted `cuda_graph.cuh` entirely in `f77a0faf`. It is now byte-identical to `main`.

You were right, and the reason is clearer than I gave at the time: `raft::copy` already wraps its own call in `RAFT_CUDA_TRY` (`raft/util/cudart_utils.hpp:139`), so **none of those error checks were needed for this fix**. I added them because the swallowed errors had made the original bug hard to diagnose — a real problem, but a separate one, and not something to smuggle into a CI hotfix.

That is now #1758, which records the two failure modes the attempt exposed (`capture_started` left set when the error propagates, and the captured graph leaking if instantiation fails), plus your point about `launch_graph` not checking whether the instance was created.

The PR is now only the thing you asked for — eliminating the copies that cannot be captured:

| file | change |
|---|---|
| `vrp_move_candidates.cuh` | `set_value_async` → `raft::copy` (the actual failing site) |
| `cycle_finder.hpp` | zero-valued write → `cudaMemsetAsync` |
| `random_move_candidates.cuh` | zero-valued writes → `set_value_to_zero_async` |
| `test_batch_solve.py` | skip on CUDA 13.3 runtime only, tracked in #1756 |

21 insertions, 9 deletions across 4 files.

This also resolves three of CodeRabbit's findings by removing the code they referred to: the `cudaGraphExecUpdate` return value, the failure-path gtests, and the exception-safety concerns are all moot now, and are captured in #1758 instead.

Verified locally: builds clean with `-DDEFINE_ASSERT=True`, pre-commit passes.

tmckayus (approved):
as far as the pytest change, lgtm

python/cuopt/cuopt/tests/routing/test_batch_solve.py
```diff
@@ -16,6 +18,12 @@ def create_tsp_cost_matrix(n_locations):
     return cudf.DataFrame(cost_matrix)
 
 
+@pytest.mark.skipif(
```
Iroy30, line 21: Is this a test issue or code path issue?

ramakrishnap-nv: Code path, but to unblock CI, I have disabled this test.

Iroy30: Okay, approved for this test updated , thanks!

Iroy30 (approved):

akifcorduk (approved):

PR #1754, opened 2026-08-20
Fix knapsack rational coefficient scaling

aliceb-nv:
Knapsack cover cuts could be invalid: rational_coefficients scales rows in floating point so coefficients land an ulp low, and the separation DP floored weights, so it returned oversized sets whose complement is not a cover.

## Checklist

- Testing
   - [ ] Added tests
   - [ ] NA
- Documentation
   - [ ] Added new documentation
   - [ ] NA


cpp/src/cuts/cuts.cpp
```diff
@@ -2972,10 +2993,12 @@ f_t knapsack_generation_t<i_t, f_t>::solve_knapsack_problem(const std::vector<f_
 
   solution.assign(n, 0.0);
 
-  // dp(j, v) = minimum weight using first j items to get value v
-  dense_matrix_t<i_t, i_t> dp(n + 1, sum_value + 1, INT_INF);
+  // dp(j, v) = minimum weight using first j items to get value v.
+  // The weights are carried at full precision: rounding one down would let the DP return a set
```
chris-maes, line 2997: I'm not sure I understand why we need to use floats here. Where is the "rounding down" occurring that is causing the issue? And why does floats fix it?

cpp/tests/mip/cuts_test.cu
```diff
@@ -934,6 +947,62 @@ TEST(cuts, test_cuts_2)
   EXPECT_EQ(solution.get_num_nodes(), 0);
 }
 
+io::mps_data_model_t<int, double> create_knapsack_cover_floor_problem()
+{
+  // The odd cycle over z1, z2, z3 makes z = (0.5, 0.5, 0.5), w = 0 the unique LP optimum, which
+  // puts y at 0.412078. Integrality moves the optimum to w = 1 with y = 0 and objective 3.
+  //
+  // The capacity coefficients are load bearing: scaling that row to integers multiplies by 100,
```
chris-maes, line 955: Nit: "load bearing" sounds like an AI generated phrase. Could we reword?

cpp/src/cuts/cuts.cpp
```diff
@@ -4629,6 +4660,17 @@ bool rational_coefficients(const std::vector<variable_type_t>& var_types,
 
   rational_inequality.scale(scalar);
 
+  // The scaled product can land an ulp off the integer it represents. Callers rely on the
```
chris-maes, line 4663: This code is a bit strange. First, the `integral_tol` used here, doesn't match our default `integer_tol`. It would probably be best to pass `settings` in and use `settings.integer_tol`. Unless, this should be a separate tolerance.

Second, I think this function is used by more than knapsack cuts. So it's a bit weird to reference knapsack cuts in this code. I think this code is just trying to generate an inequality with *rational coefficients*, whereas knapsack needs an inequality with *integer coefficients*. Maybe this code should be moved into a post-processing step, say in a function called `integer_coefficients` (that first calls `rationalize_coefficients`) 

Finally, rounding introduces some error. That just multiplying by a scalar does not. So, we probably don't want to do this in all cases. Maybe we also want to track the total amount of error introduced (instead of just tracking the error on individual coefficients).

cpp/src/cuts/cuts.cpp
chris-maes: Ah ok. Reading further, it sounds like we tried to convert an inequality with floating point coefficients into a knapsack constraint with integer coefficients, but when we do this we can make some small error if we round down. Then when we do dynamic programming we can think we are under the weight limit, when we are actually over. 

But I'm not sure the fix should be use floating point in the DP table. That can make the table a lot bigger. Maybe we should reject inequalities with large errors when converting to integer coefficients in our classification of knapsack constraints, so we don't arrive at this case in the first place? Or if this occurs when the error is quite small, maybe we just need to do a check at the end to see if the selected weights violate the capacity (similar to what is done in the assert now)

cpp/src/cuts/cuts.cpp
```diff
@@ -2985,8 +3008,7 @@ f_t knapsack_generation_t<i_t, f_t>::solve_knapsack_problem(const std::vector<f_
 
       // Take item j-1 if possible
       if (v >= scaled_values[j - 1]) {
-        i_t candidate =
-          dp(j - 1, v - scaled_values[j - 1]) + static_cast<i_t>(std::floor(weights[j - 1]));
```
chris-maes, line 2989: Seems like the `floor` is the big issue here. Because it causes us to underestimate the weight. Could the issue be solved by just using `std::ceil(weights[j-1])` here? If `weights[k]` is an integer than `ceil(weights[k]) == floor(weights[k])`. But if it isn't exact, we will round up, overestimate how much weight we can carry, and thus still produce a valid solution.

cpp/src/cuts/cuts.cpp
chris-maes: Or just use ceil(weights[j-1]) instead of floor(weights[j-1])

mlubin: Yeah I think the typical solution here is careful rounding.

chris-maes (changes requested):

PR #1759, opened 2026-08-20
fix: keep host sources alive for async route-view copies (#1756)

### LINKED ISSUE #1756: [BUG] test_batch_solve_varying_sizes: block_copy destination too small on CUDA 13.3.0 (blocks all PR wheel CI)
## Summary

`test_batch_solve_varying_sizes` fails on CUDA **13.3.0** with a device-side assert in `block_copy`: the destination span is smaller than the number of elements being copied.

```
cpp/src/utilities/cuda_helpers.cuh:147: void cuopt::block_copy(
  raft::span<T,...>, raft::span<T,...>, unsigned long) [with T = cuopt::routing::detail::NodeInfo<int>]:
  block: [0,0,0], thread: [192,0,0]
  Assertion `dst.size() >= size && "block_copy::dst does not have the sufficient size"` failed.

terminate called after throwing an instance of 'thrust::system::system_error'
  what():  fill_n: failed to synchronize: cudaErrorAssert: device-side assert triggered
[gw2] node down: Not properly terminated
```

The assert fires across hundreds of threads and aborts the process, taking the pytest-xdist worker down with it.

## Observed in

Reproduces on any PR that runs the wheel tests. Two examples on unrelated branches:

- [#1753 job 96487821122](https://github.com/NVIDIA/cuopt/actions/runs/32384305561/job/96487821122) — CUDA graph capture fix
- [#1726 job 96126562276](https://github.com/NVIDIA/cuopt/actions/runs/32267775605/job/96126562276) — Mod-2 cut separator

Neither touches routing batch solve, so this is not caused by either.

## It is specific to CUDA 13.3.0

From a single PR run, `wheel-tests-cuopt`:

| CUDA | result |
|---|---|
| **13.3.0** | 3 / 3 **fail** |
| 13.0.3 | 2 / 2 pass |
| 12.9.2 | pass |

13.0.3 passes, so this is not "CUDA 13" generally — something about the 13.3 toolkit specifically.

## Why only wheel jobs, and why nightly is green

Assertions are enabled only for pull-request wheel builds:

```bash
# ci/build_wheel_libcuopt.sh
if [ "$RAPIDS_BUILD_TYPE" = "pull-request" ]; then
    export SKBUILD_CMAKE_ARGS="${SKBUILD_CMAKE_ARGS};-DDEFINE_ASSERT=True"
```

`cuopt_assert` compiles to nothing without `ASSERT_MODE` (`cpp/src/utilities/macros.cuh:15-22`). Conda builds never set `DEFINE_ASSERT`, and nightly wheel builds take the release branch.

> [!WARNING]
> This means the undersized copy is almost certainly **also happening in release and conda builds** — silently. There the assert is compiled out and `block_copy` writes past the end of the destination span. The failing PR jobs are the only place we currently see it; a green nightly does not mean the code path is correct.

That should be confirmed or ruled out (e.g. `compute-sanitizer` on this test with a release build) before treating this as test-only.

## Call path

```
route.cuh:541  requests().copy_fixed_route_data(orig_route.requests(), start_idx, end_idx, write_start)
  pdp_route.cuh:93   block_copy(node_info.subspan(write_start),
                                orig_route.node_info.subspan(from_idx),
                                to_idx - from_idx)
    cuda_helpers.cuh:147  cuopt_assert(dst.size() >= size, ...)   <-- fails
```

## Suspected cause

The test batch-solves TSPs of interleaved varying sizes:

```python
tsp_sizes = [5, 8, 10, 6, 7, 9, 12, 15, 11, 4, 13, 14, 8, 6, 10, 9, 7, 11, 5, 12]
solutions = routing.BatchSolve(data_models, settings)
```

A route/solution buffer sized for one problem being reused for a larger problem later in the batch would produce exactly this: `write_start + size` exceeding the destination span. Not yet confirmed — the sizing logic in the batch path needs a proper look.

Unclear why CUDA 13.3 specifically surfaces it; that may just be scheduling/timing exposing a latent bug rather than a 13.3 regression.

## Impact on CI

Every PR that runs `wheel-tests-cuopt` on 13.3.0 is red because of this, regardless of what the PR changes. It is currently the only failure on #1753.

## Suggested next steps

- [ ] Confirm whether the out-of-bounds write also occurs in release builds (compute-sanitizer)
- [ ] Find the buffer whose size is computed from the wrong problem in the batch path
- [ ] Determine why 13.3.0 exposes it and 13.0.3 does not
- [ ] Temporarily skip the test to unblock CI (note: `xfail` will not work — the device assert aborts the process, so pytest cannot catch it)

ramakrishnap-nv:
## Description

Fixes #1756 — `test_batch_solve_varying_sizes` aborted with a device-side `block_copy` assert on CUDA 13.3.

`rmm::device_uvector::set_element_async` does not synchronize; the host source must stay valid until the stream is synced (`device_uvector.hpp:200-207`). Four sites in `solution.cu` copied a route `view_t` from a stack local that died each loop iteration, two of them without ever syncing. Since the whole `view_t` is copied, a dead source corrupts the span size and the pointers together — hence both the assert and the `cudaErrorIllegalAddress` seen in CI.

Each site now stages its host sources in a vector reserved up front, so the staging outlives the copy.

Also removes the CUDA 13.3 skip, since the test passes again.

Note: only pull-request builds set `DEFINE_ASSERT`, so nightly and released builds were doing this out-of-bounds copy silently.

## Checklist

- Testing
- Documentation
   - [x] NA

### DIFF
--- a/cpp/src/routing/solution/solution.cu
+++ b/cpp/src/routing/solution/solution.cu
@@ -122,6 +122,18 @@ void solution_t<i_t, f_t, REQUEST>::add_routes(
   n_routes += added_routes;
 
   check_and_allocate_routes(n_routes);
+  // Host sources of the async copies below must stay alive and unmodified until the
+  // sync_stream() after this loop. Stage them here (reserved up front so no reallocation
+  // can invalidate a pending copy) instead of using per-iteration locals. vehicle_id and
+  // route are references into new_routes, which already outlives this function.
+  std::vector<std::vector<NodeInfo<i_t>>> h_node_infos;
+  std::vector<i_t> h_n_nodes;
+  std::vector<i_t> h_route_ids;
+  std::vector<typename route_t<i_t, f_t, REQUEST>::view_t> h_route_views;
+  h_node_infos.reserve(new_routes.size());
+  h_n_nodes.reserve(new_routes.size());
+  h_route_ids.reserve(new_routes.size());
+  h_route_views.reserve(new_routes.size());
   for (const auto& [vehicle_id, route] : new_routes) {
     // depot for the beginning and the end
     i_t new_route_size = route.size() + 2;
@@ -131,7 +143,8 @@ void solution_t<i_t, f_t, REQUEST>::add_routes(
         std::max(max_nodes_per_route, raft::alignTo(new_route_size, base_route_size));
       resize_routes(new_route_size);
     }
-    std::vector<NodeInfo<i_t>> node_info_h;
+    auto& node_info_h = h_node_infos.emplace_back();
+    node_info_h.reserve(route.size());
     for (size_t x = 0; x < route.size(); ++x) {
       // Dummy, will be overriden in set_nodes_data_of_route
       node_info_h.emplace_back(route[x]);
@@ -141,13 +154,14 @@ void solution_t<i_t, f_t, REQUEST>::add_routes(
                node_info_h.data(),
                node_info_h.size(),
                sol_handle->get_stream());
-    const i_t n_nodes = route.size() + 1;
+    const i_t& n_nodes = h_n_nodes.emplace_back(route.size() + 1);
     d_route.n_nodes.set_value_async(n_nodes, sol_handle->get_stream());
-    d_route.route_id.set_value_async(route_id, sol_handle->get_stream());
+    const i_t& stable_route_id = h_route_ids.emplace_back(route_id);
+    d_route.route_id.set_value_async(stable_route_id, sol_handle->get_stream());
     d_route.vehicle_id.set_value_async(vehicle_id, sol_handle->get_stream());
-    i_t route_slot        = route_id_to_idx[route_id];
-    const auto route_view = routes[route_slot].view();
+    i_t route_slot = route_id_to_idx[route_id];
     cuopt_assert(route_id < (int)routes_view.size(), "route id should be in range");
+    const auto& route_view = h_route_views.emplace_back(routes[route_slot].view());
     routes_view.set_element_async(route_id, route_view, sol_handle->get_stream());
     ++route_id;
   }
@@ -294,10 +308,17 @@ void solution_t<i_t, f_t, REQUEST>::set_route_views()
     routes_view.resize(routes.size(), sol_handle->get_stream());
   }
 
+  // The host source of an async copy must stay alive until the stream is synchronized
+  // (rmm::device_uvector::set_element_async does not synchronize). Stage the views in a
+  // vector that outlives the copy instead of copying from a per-iteration stack local.
+  std::vector<typename route_t<i_t, f_t, REQUEST>::view_t> h_route_views;
+  h_route_views.reserve(routes.size());
   for (size_t i = 0; i < routes.size(); ++i) {
-    const auto route_view = get_route(i).view();
-    routes_view.set_element_async(i, route_view, sol_handle->get_stream());
+    h_route_views.push_back(get_route(i).view());
   }
+  raft::copy(
+    routes_view.data(), h_route_views.data(), h_route_views.size(), sol_handle->get_stream());
+  sol_handle->sync_stream();
 }
 
 // init the rotues with the desired number of routes
@@ -374,14 +395,22 @@ void solution_t<i_t, f_t, REQUEST>::resize_routes(i_t new_size)
 {
   raft::common::nvtx::range fun_scope("resize_routes");
 
+  // See set_route_views: the host source must outlive the async copy, so the resized views
+  // are staged here rather than copied from a stack local that dies each iteration.
+  std::vector<typename route_t<i_t, f_t, REQUEST>::view_t> h_route_views;
+  std::vector<i_t> h_route_indices;
   for (i_t i = 0; i < n_routes; ++i) {
     auto& route_i = get_route(i);
     if (route_i.max_nodes_per_route() < new_size) {
       route_i.resize(new_size);
-      const auto route_view = get_route(i).view();
-      routes_view.set_element_async(i, route_view, sol_handle->get_stream());
+      h_route_views.push_back(get_route(i).view());
+      h_route_indices.push_back(i);
     }
   }
+  for (size_t k = 0; k < h_route_indices.size(); ++k) {
+    routes_view.set_element_async(h_route_indices[k], h_route_views[k], sol_handle->get_stream());
+  }
+  if (!h_route_indices.empty()) { sol_handle->sync_stream(); }
   max_nodes_per_route = std::max(max_nodes_per_route, new_size);
 }
 
@@ -512,10 +541,17 @@ void solution_t<i_t, f_t, REQUEST>::copy_device_solution(solution_t<i_t, f_t, RE
   if (!src_sol.n_routes) { return; }
   check_and_allocate_routes(src_sol.n_routes);
 
+  // See set_route_views: the host source must outlive the async copy. copy_routes below
+  // reads these entries, so a dangling source here corrupts the route span sizes it uses.
+  std::vector<typename route_t<i_t, f_t, REQUEST>::view_t> h_route_views;
+  h_route_views.reserve(std::max(0, src_sol.n_routes - n_routes));
   for (i_t i = n_routes; i < src_sol.n_routes; ++i) {
-    const auto route_view = get_route(i).view();
-    routes_view.set_element_async(i, route_view, sol_handle->get_stream());
+    h_route_views.push_back(get_route(i).view());
   }
+  for (size_t k = 0; k < h_route_views.size(); ++k) {
+    routes_view.set_element_async(n_routes + (i_t)k, h_route_views[k], sol_handle->get_stream());
+  }
+  if (!h_route_views.empty()) { sol_handle->sync_stream(); }
 
   n_routes = src_sol.n_routes;
 
--- a/cpp/src/routing/solution/solution.cu
+++ b/cpp/src/routing/solution/solution.cu
@@ -96,7 +96,7 @@ void solution_t<i_t, f_t, REQUEST>::add_route(route_t<i_t, f_t, REQUEST>&& route
   i_t route_slot     = route_id_to_idx[route_id];
   routes[route_slot] = std::move(route);
   cuopt_assert(route_id < (int)routes_view.size(), "route id should be in range");
-  set_route_views();
+  set_route_views(route_id, route_id + 1);
   if (max_nodes_per_route < get_route(route_id).max_nodes_per_route()) {
     resize_routes(raft::alignTo(get_route(route_id).max_nodes_per_route(), base_route_size));
   }
@@ -125,9 +125,6 @@ void solution_t<i_t, f_t, REQUEST>::add_routes(
   // sync_stream() after this loop, so per-iteration locals cannot be used. vehicle_id and
   // route are references into new_routes and already outlive this function; the scalars
   // are staged here, reserved up front so no reallocation can invalidate a pending copy.
-  static_assert(std::is_same_v<NodeInfo<i_t>, NodeInfo<>>,
-                "add_routes copies node info straight out of new_routes, which requires the "
-                "caller's NodeInfo<> to be the same type as NodeInfo<i_t>");
   std::vector<i_t> h_n_nodes;
   std::vector<i_t> h_route_ids;
   h_n_nodes.reserve(new_routes.size());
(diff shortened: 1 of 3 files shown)


ramakrishnap-nv:
Pushed diagnostics (`8ad0cd01`). All three are under `ASSERT_MODE`, so they are active in PR wheel *and* conda builds and compile away in nightly/release. **All temporary — revert before merge.**

1. `cuda_helpers.cuh` — at the failing `block_copy`: prints `dst`/`src`/`size` plus `gridDim`/`blockDim`/`blockIdx`. `gridDim` discriminates the remaining 256-thread candidates (`execute_sliding_move` launches with `gridDim = n_routes` = 1 here; `find_insertions_kernel` with many more).
2. `route.cuh` `copy_from` — prints destination capacity, both `n_nodes` values and route/vehicle ids, separating "buffer sized for a smaller problem" from "source `n_nodes` is garbage".
3. `solution.cu` `add_routes` — verifies the `n_nodes`/`route_id` written via `set_value_async` actually landed. Placed **after** `sync_stream()` on purpose: checking immediately after the write would force the copy to complete while the host source was still alive and mask the problem.

### Working hypothesis

`add_routes` has three host sources of async copies that outlive their safety window:

- `n_nodes` (`solution.cu:147`) — stack local, destroyed at the end of each loop iteration
- `route_id` (`solution.cu:149`) — mutated by `++route_id` before the sync
- `node_info_h` (`solution.cu:137`) — `std::vector` destroyed each iteration, source of the `raft::copy` at :143

All are safe under `cudaMemcpyAsync`, which stages small pageable copies before returning. None are safe if the runtime defers the host read — and #1753 established that rmm changed exactly this dispatch on CUDA 13 (rapidsai/rmm#2511 → `cudaMemcpyBatchAsync`). RMM itself deletes the rvalue overload of `set_value_async` for precisely this hazard (`device_scalar.hpp:199-201`).

A garbage `n_nodes` gives `size = *n_nodes + 1` far exceeding the 64-element (`base_route_size`) route buffer, which is the assert we see.

**Unexplained gap:** #1753 described the rmm dispatch as CUDA-13-wide, yet 13.0.3 passes and 13.3.0 fails. If the prints confirm corruption, that still needs accounting for.

### Prior results

- `-n 2` (run 32413945833): all three 13.3.0 jobs still failed → worker contention is **not** the trigger. Reverted.
- Local sm_75 + CUDA 13.3.73 + asserts: not reproducible. Full suite 129 passed, 0 undersized; batch solve 5/5 clean. A local trace shows the suspect kernel runs 6846×/solve at `gridDim.x=1 blockDim.x=256` (matching the CI signature) without ever overflowing — so the path is reached on Turing, it just does not misbehave there.

### Correction to the issue

#1756 states conda builds never set `DEFINE_ASSERT`. `conda/recipes/libcuopt/recipe.yaml:40` sets `-a --host-lineinfo` for `build_type == pull-request`, and `conda-python-tests / 13.3.0 … rtxpro6000` fails with the same assert. The real split is **pull-request vs nightly**, not wheel vs conda. The silent-corruption concern still stands for nightly and released artifacts.

cpp/src/routing/solution/solution.cu
```diff
@@ -122,6 +121,17 @@ void solution_t<i_t, f_t, REQUEST>::add_routes(
   n_routes += added_routes;
 
   check_and_allocate_routes(n_routes);
+  // Host sources of the async copies below must stay valid and unmodified until the
+  // sync_stream() after this loop, so per-iteration locals cannot be used. vehicle_id and
+  // route are references into new_routes and already outlive this function; the scalars
+  // are staged here, reserved up front so no reallocation can invalidate a pending copy.
+  static_assert(std::is_same_v<NodeInfo<i_t>, NodeInfo<>>,
```
akifcorduk: I don't think we need that. NodeInfo is the same for the solve.

cpp/src/routing/solution/solution.cu
```diff
@@ -93,11 +93,10 @@ void solution_t<i_t, f_t, REQUEST>::add_route(route_t<i_t, f_t, REQUEST>&& route
...
   route.route_id.set_value_async(route_id, sol_handle->get_stream());
   sol_handle->sync_stream();
-  i_t route_slot        = route_id_to_idx[route_id];
-  routes[route_slot]    = std::move(route);
-  const auto route_view = routes[route_slot].view();
+  i_t route_slot     = route_id_to_idx[route_id];
+  routes[route_slot] = std::move(route);
   cuopt_assert(route_id < (int)routes_view.size(), "route id should be in range");
-  routes_view.set_element_async(route_id, route_view, sol_handle->get_stream());
+  set_route_views();
```
akifcorduk: I am not sure why this is necessary, it should be equivalent as before. PR description writes that set_element_async does not sync but we are synchronizing few lines later.

cpp/src/routing/solution/solution.cu
```diff
@@ -374,14 +389,17 @@ void solution_t<i_t, f_t, REQUEST>::resize_routes(i_t new_size)
...
     if (route_i.max_nodes_per_route() < new_size) {
       route_i.resize(new_size);
-      const auto route_view = get_route(i).view();
-      routes_view.set_element_async(i, route_view, sol_handle->get_stream());
+      any_resized = true;
     }
   }
+  // A resize reallocates the route buffers, so the device-side views are stale. Republish
+  // through set_route_views rather than copying each changed view from a local here.
+  if (any_resized) { set_route_views(); }
```
akifcorduk: I think this is also unnecessary. We are already setting the device views up.

cpp/src/routing/solution/solution.cu
```diff
@@ -512,10 +530,8 @@ void solution_t<i_t, f_t, REQUEST>::copy_device_solution(solution_t<i_t, f_t, RE
   if (!src_sol.n_routes) { return; }
   check_and_allocate_routes(src_sol.n_routes);
 
-  for (i_t i = n_routes; i < src_sol.n_routes; ++i) {
-    const auto route_view = get_route(i).view();
-    routes_view.set_element_async(i, route_view, sol_handle->get_stream());
-  }
+  // copy_routes below reads these entries, so they must be published before it launches.
+  if (src_sol.n_routes > n_routes) { set_route_views(); }
```
akifcorduk: I think this is unnecessary.

cpp/src/routing/solution/solution.cu
```diff
@@ -294,10 +300,19 @@ void solution_t<i_t, f_t, REQUEST>::set_route_views()
...
+  h_routes_view.clear();
+  h_routes_view.reserve(routes.size());
   for (size_t i = 0; i < routes.size(); ++i) {
-    const auto route_view = get_route(i).view();
-    routes_view.set_element_async(i, route_view, sol_handle->get_stream());
+    h_routes_view.push_back(get_route(i).view());
   }
+  raft::copy(
+    routes_view.data(), h_routes_view.data(), h_routes_view.size(), sol_handle->get_stream());
+  sol_handle->sync_stream();
```
akifcorduk, line 317: I am worried about this synchronization. I think we can remove that, I expect the call sites handle the synchronization pretty well. It was not synhronized before, it shouldn't be synchonized now.

cpp/src/routing/solution/solution.cu
akifcorduk: Instead of doing full `set_route_views`, I would pass an argument of range, which is otherwise defaulted to `0` and `n_routes`

cpp/src/routing/solution/solution.cu
ramakrishnap-nv: Agreed, removed.

cpp/src/routing/solution/solution.cu
ramakrishnap-nv: You're right, and I checked before changing it back: in the original, `route_view` was a function-scope local and the `sync_stream()` at the end of `add_route` ran before it went out of scope, so that site was already safe. I changed it while centralizing publication without needing to. Now uses `set_route_views(route_id, route_id + 1)` per your range suggestion below, so it is O(1) as before and still goes through the single path.

cpp/src/routing/solution/solution.cu
ramakrishnap-nv: Adopted. `set_route_views(i_t start = 0, i_t end = -1)`, with `end < 0` meaning `routes.size()`. Call sites now publish only what changed: `add_route` -> `[route_id, route_id+1)`, `add_routes` -> `[prev_route_size, n_routes)`, and the `copy_routes` feeder -> `[n_routes, src_sol.n_routes)`. This also addresses the CodeRabbit note about O(n_routes) republish.

cpp/src/routing/solution/solution.cu
ramakrishnap-nv: I'd like to push back on this one, or at least lay out why I think it is load-bearing.

The unsynchronized version is what caused #1756. Without a sync inside `set_route_views`, the host source has to outlive the copy, which is why the staging moved to a member buffer. That leaves exactly one hazard: a second publish overwriting the buffer while the first copy is still pending. It is reachable in `add_route`:

```cpp
set_route_views(route_id, route_id + 1);            // copy A enqueued, reads h_routes_view
if (max_nodes_per_route < ...) resize_routes(...);  // republishes -> overwrites h_routes_view
```

With the sync removed, `resize_routes` can overwrite the staging before copy A executes. Copy A would then publish the post-resize view, i.e. a pointer to a buffer whose stream-ordered allocation has not run yet. Any kernel between the two dereferences it. Same failure class as the original bug, just narrower.

On "the call sites handle the synchronization pretty well" -- that was the assumption that did not hold. rmm sets `srcAccessOrder = cudaMemcpySrcAccessOrderStream` on CUDA 13 (rapidsai/rmm#2511), so the host bytes are read when the copy runs, not when it is enqueued. Before that change `cudaMemcpyAsync` staged small pageable copies during the API call, which is why the old code worked for years despite being UB by contract.

That said, your perf concern is fair and I do not want a blocking sync in a hot path either. The principled way to get both is to record a `cudaEvent` after the copy and wait on it only before *reusing* the staging buffer. In steady state the event is long complete so it costs nothing, and correctness stops depending on call-site discipline. Happy to implement that instead if you prefer.

Two mitigations already in place: `resize_routes` only republishes when something actually resized, and with the range argument each publish is now bounded, so this is not per-iteration of the solver. I have not profiled it though, and CodeRabbit asked for exactly that -- if you would rather I measure before we decide, I can.

ramakrishnap-nv: @akifcorduk What is your suggestion on this https://github.com/NVIDIA/cuopt/pull/1759#discussion_r3831045167

ramakrishnap-nv:
Applied the `h_routes_view` private nitpick in `b0b31ce`.

On separating allocation from publication in `check_and_allocate_routes()`: the observation is fair, but that `set_route_views()` call predates this PR -- it is already on `main`. What this PR changed is that it now synchronizes.

It is worth doing. Three of the four callers (`add_route`, `add_routes`, and the `copy_routes` feeder) now publish their own bounded range after allocating, so the full publish inside `check_and_allocate_routes` is redundant work plus a sync, and removing it would cut the sync count -- which is the concern raised separately in this review.

I am not doing it in this PR though. It changes allocation/publication semantics that predate the fix, and its safety depends on every caller's bounded publish covering exactly the newly allocated range. `add_route` publishes `[route_id, route_id+1)` while `check_and_allocate_routes` can allocate more than one route, so that invariant is not obviously true on all paths. Getting it wrong reintroduces "kernel reads an unpublished view", the same failure class as #1756, on a change that is currently verified green on the hardware that reproduces it.

Happy to do it as a follow-up where it can be reasoned about and profiled properly. It may also become moot depending on how the synchronization thread resolves -- the event-guarded reuse I proposed there removes the blocking sync at the source.

cpp/src/routing/solution/solution.cu
akifcorduk: okay makes sense. we can keep it.

cpp/src/routing/solution/solution.cu
```diff
@@ -93,11 +93,10 @@ void solution_t<i_t, f_t, REQUEST>::add_route(route_t<i_t, f_t, REQUEST>&& route
...
   route.route_id.set_value_async(route_id, sol_handle->get_stream());
   sol_handle->sync_stream();
-  i_t route_slot        = route_id_to_idx[route_id];
-  routes[route_slot]    = std::move(route);
-  const auto route_view = routes[route_slot].view();
+  i_t route_slot     = route_id_to_idx[route_id];
+  routes[route_slot] = std::move(route);
   cuopt_assert(route_id < (int)routes_view.size(), "route id should be in range");
-  routes_view.set_element_async(route_id, route_view, sol_handle->get_stream());
+  set_route_views(route_id, route_id + 1);
```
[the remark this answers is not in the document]
akifcorduk: I would leave it as it is, the performance impact should be negligible as the total amount of memcpy requests are the same. Size of the copy should not matter a lot.

akifcorduk (approved):

tmckayus (approved):

PR #1760, opened 2026-08-21
B&B Halt

nguidotti:
With this PR, the B&B code can now be stopped via the `concurrent_halt` flag in the `simplex_solver_settings`. This also fixes the improper handle of the `concurrent_halt` flag during the cut passes and replaces the sub-MIP halt mechanism with the newer version.

## Checklist

- Testing
   - [ ] Added tests
   - [ ] NA
- Documentation
   - [ ] Added new documentation
   - [ ] NA

### DIFF
--- a/cpp/src/branch_and_bound/branch_and_bound.cpp
+++ b/cpp/src/branch_and_bound/branch_and_bound.cpp
@@ -2236,11 +2236,11 @@ bool branch_and_bound_t<i_t, f_t>::launch_submip_worker(const std::vector<f_t>& 
   if (settings_.inside_submip) {
     // LLVM libomp's GOMP compatibility path skips GCC's firstprivate copy
     // function for included tasks.
-    recursive_submip(worker, current_incumbent, var_types_);
+    recursive_submip(worker, current_incumbent, var_types_, settings_);
   } else {
 #pragma omp task priority(CUOPT_DEFAULT_TASK_PRIORITY) affinity(worker) \
   firstprivate(worker, current_incumbent)
-    recursive_submip(worker, current_incumbent, var_types_);
+    recursive_submip(worker, current_incumbent, var_types_, settings_);
   }
 
   return true;
@@ -2253,21 +2253,17 @@ void branch_and_bound_t<i_t, f_t>::solve_submip(diving_worker_t<i_t, f_t>* worke
                                                 submip_stats_t& submip_stats,
                                                 f_t fixrate,
                                                 i_t simplex_iter_used,
-                                                bool is_root_heuristic,
-                                                std::atomic<int>* halt)
+                                                simplex_solver_settings_t<i_t, f_t> submip_settings)
 {
   double start_time = tic();
 
-  i_t submip_level = settings_.submip_settings.level + 1;
-  std::string log_prefix =
-    std::format("[{} {}] ", search_strategy_to_string(worker->search_strategy), submip_level);
+  bool is_root_heuristic = submip_settings.inside_root_node;
+  i_t submip_level       = settings_.submip_settings.level + 1;
+  f_t user_lower         = compute_user_objective(worker->leaf_problem, get_lower_bound());
+  f_t user_obj           = compute_user_objective(worker->leaf_problem, upper_bound_.load());
+  f_t rel_gap            = user_relative_gap(user_obj, user_lower);
+  i_t explored           = exploration_stats_.nodes_explored;
 
-  f_t user_lower = compute_user_objective(worker->leaf_problem, get_lower_bound());
-  f_t user_obj   = compute_user_objective(worker->leaf_problem, upper_bound_.load());
-  f_t rel_gap    = user_relative_gap(user_obj, user_lower);
-  i_t explored   = exploration_stats_.nodes_explored;
-
-  simplex_solver_settings_t<i_t, f_t> submip_settings      = settings_;
   submip_settings.print_presolve_stats                     = false;
   submip_settings.num_threads                              = 1;
   submip_settings.reliability_branching                    = 0;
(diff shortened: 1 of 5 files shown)


cpp/src/branch_and_bound/branch_and_bound.cpp
```diff
@@ -2343,6 +2358,8 @@ void branch_and_bound_t<i_t, f_t>::solve_submip(diving_worker_t<i_t, f_t>* worke
     return;
   }
 
+  if (halt != nullptr && halt->load(std::memory_order_acquire)) { return; }
```
chris-maes: Use `submip_settings.concurrent_halt` here for consistency

cpp/src/branch_and_bound/branch_and_bound.cpp
```diff
@@ -2239,7 +2252,8 @@ void branch_and_bound_t<i_t, f_t>::solve_submip(diving_worker_t<i_t, f_t>* worke
                                                 const std::vector<variable_type_t>& var_types,
                                                 submip_stats_t& submip_stats,
                                                 f_t fixrate,
-                                                i_t simplex_iter_used)
+                                                i_t simplex_iter_used,
+                                                std::atomic<int>* halt)
```
chris-maes: It's a bit strange to be passing the halt pointers around. They are meant to be contained within the settings. It might be better to create the settings with the halt and pass the settings in here.

cpp/src/branch_and_bound/branch_and_bound.cpp
```diff
@@ -2644,7 +2661,8 @@ f_t calculate_fixrate(const std::vector<i_t>& integer_list,
 template <typename i_t, typename f_t>
 void branch_and_bound_t<i_t, f_t>::recursive_submip(diving_worker_t<i_t, f_t>* worker,
                                                     const std::vector<f_t>& current_incumbent,
-                                                    const std::vector<variable_type_t>& var_types)
+                                                    const std::vector<variable_type_t>& var_types,
+                                                    std::atomic<int>* halt)
```
chris-maes: Same here. Could you just pass in the settings with the halt set?

cpp/src/branch_and_bound/branch_and_bound.cpp
```diff
@@ -2689,7 +2707,8 @@ void branch_and_bound_t<i_t, f_t>::recursive_submip(diving_worker_t<i_t, f_t>* w
 
   i_t round = 0;
 
-  while (solver_status_ == mip_status_t::UNSET && is_running_ && !worker->halt) {
+  while (solver_status_ == mip_status_t::UNSET && is_running_ &&
+         !(halt && halt->load(std::memory_order::acquire))) {
```
chris-maes: Could we use submip_settings.concurrent_halt here instead of a bare halt.

cpp/src/branch_and_bound/branch_and_bound.cpp
```diff
@@ -3157,7 +3190,7 @@ auto branch_and_bound_t<i_t, f_t>::do_cut_pass(
   f_t& last_objective,
   f_t root_relax_objective,
   i_t& cut_pool_size,
-  [[maybe_unused]] const std::vector<f_t>& saved_solution) -> cut_pass_result_t
+  [[maybe_unused]] const std::vector<f_t>& saved_solution) -> cut_pass_action_t
```
chris-maes, line 3194: Good to see the return type simiplified

cpp/src/branch_and_bound/branch_and_bound.cpp
```diff
@@ -3549,11 +3615,20 @@ mip_status_t branch_and_bound_t<i_t, f_t>::solve(mip_solution_t<i_t, f_t>& solut
   f_t root_relax_elapsed_time            = toc(root_relax_start_time);
   exploration_stats_.total_lp_solve_time = root_relax_elapsed_time;
 
+  scope_guard cliques_scope([&]() {
```
chris-maes, line 3622: I'm not a fan of scope guards. These make it difficult to understand what is happening in the code. Since code is executing that may be defined far away from the actual return statement.

chris-maes: I understand that it is helpful to avoid having to have do this clean up at every return though.  Is it possible to put this in the cliques destructor instead maybe?

chris-maes: If not, could you add a comment here explaining the need for the scope guard; so a reader is aware.

cpp/src/branch_and_bound/branch_and_bound.cpp
chris-maes: Would you mind replacing the `auto` with `cut_pass_action_t` and removing the `-> cut_pass_action_t` here?

cpp/src/mip_heuristics/root_heuristics.hpp
```diff
@@ -19,6 +19,7 @@ struct root_heuristics_t {
   csr_matrix_t<i_t, f_t> Arow_;
   std::vector<f_t> root_solution_;
   std::vector<f_t> root_edge_norm_;
+  std::atomic<int> halt_;
```
chris-maes, line 22: Why do we need a bare halt here? Can we store the halt in settings?

nguidotti: We need to store the actual object somewhere, right? The `concurrent_halt` points to this object.

nguidotti: We need to declare here, since we want to send a halt signal to each individual heuristics during the cut passes.

cpp/src/branch_and_bound/branch_and_bound.cpp
nguidotti: Since `cut_pass_action_t` is declared inside `branch_and_bound_t`, we need to set the return type as `branch_and_bound_t<i_t, f_t>::cut_pass_action_t`. IMO, the intent is clear here with auto and the `-> cut_pass_action_t` at the end.

cpp/src/branch_and_bound/branch_and_bound.cpp
nguidotti: Added a comment

chris-maes: Why couldn't we use a unique_ptr for the clique_table? That doesn't need to be for this PR.

cpp/src/branch_and_bound/branch_and_bound.cpp
```diff
@@ -3637,13 +3697,11 @@ mip_status_t branch_and_bound_t<i_t, f_t>::solve(mip_solution_t<i_t, f_t>& solut
 
   if (num_fractional == 0) {
     if (settings_.benchmark_info_ptr != nullptr) {
-      const double v = static_cast<double>(compute_user_objective(original_lp_, root_objective_));
+      f_t v = compute_user_objective(original_lp_, root_objective_);
```
chris-maes, line 3700: Is the removal of the const intentional?

chris-maes (approved):
LGTM. Thanks for making sure we halt in the cut passes @nguidotti

PR #1761, opened 2026-08-21
Fix QCQP dual solution overrun and constraint count reporting

### LINKED ISSUE #1751: [BUG] cuOptGetDualSolution writes beyond the documented num_constraints buffer
## Summary

For a problem containing a quadratic constraint, `cuOptGetDualSolution`
writes more elements than reported by `cuOptGetNumConstraints`.

This overruns a correctly sized caller-owned buffer.

## Environment

- cuOpt 26.08.0
- Git hash: `400863c1`
- CUDA 13.3
- GPU: NVIDIA RTX PRO 6000 Blackwell Workstation Edition

## Standalone C reproducer

```c
#include <cuopt/mathematical_optimization/cuopt_c.h>

#include <math.h>
#include <stdio.h>

int main(void)
{
    cuOptOptimizationProblem problem = NULL;
    cuOptSolverSettings settings = NULL;
    cuOptSolution solution = NULL;

    /*
     * minimize t
     *
     * subject to
     *     t >= 0
     *     x1^2 + x2^2 - t^2 <= 0
     *     t >= -1
     *     x1 >= 3
     *     x2 >= 4
     */
    const cuopt_int_t num_constraints = 1;
    const cuopt_int_t num_variables = 3;

    const cuopt_float_t objective[] = {1.0, 0.0, 0.0};

    const cuopt_int_t row_offsets[] = {0, 1};
    const cuopt_int_t column_indices[] = {0};
    const cuopt_float_t matrix_values[] = {1.0};
    const char constraint_sense[] = {CUOPT_GREATER_THAN};
    const cuopt_float_t rhs[] = {0.0};

    const cuopt_float_t lower_bounds[] = {-1.0, 3.0, 4.0};
    const cuopt_float_t upper_bounds[] = {
        CUOPT_INFINITY,
        CUOPT_INFINITY,
        CUOPT_INFINITY
    };
    const char variable_types[] = {
        CUOPT_CONTINUOUS,
        CUOPT_CONTINUOUS,
        CUOPT_CONTINUOUS
    };

    cuopt_int_t status = cuOptCreateProblem(
        num_constraints,
        num_variables,
        CUOPT_MINIMIZE,
        0.0,
        objective,
        row_offsets,
        column_indices,
        matrix_values,
        constraint_sense,
        rhs,
        lower_bounds,
        upper_bounds,
        variable_types,
        &problem
    );
    if (status != CUOPT_SUCCESS) {
        printf("cuOptCreateProblem failed: %d\n", status);
        return 1;
    }

    const cuopt_int_t qc_row[] = {0, 1, 2};
    const cuopt_int_t qc_col[] = {0, 1, 2};
    const cuopt_float_t qc_coeff[] = {-1.0, 1.0, 1.0};

    status = cuOptAddQuadraticConstraint(
        problem,
        3,
        qc_row,
        qc_col,
        qc_coeff,
        0,
        NULL,
        NULL,
        CUOPT_LESS_THAN,
        0.0
    );
    if (status != CUOPT_SUCCESS) {
        printf("cuOptAddQuadraticConstraint failed: %d\n", status);
        return 1;
    }

    cuopt_int_t queried_num_constraints = -1;
    status = cuOptGetNumConstraints(problem, &queried_num_constraints);
    if (status != CUOPT_SUCCESS) {
        printf("cuOptGetNumConstraints failed: %d\n", status);
        return 1;
    }

    printf("cuOptGetNumConstraints = %d\n", queried_num_constraints);

    status = cuOptCreateSolverSettings(&settings);
    if (status != CUOPT_SUCCESS) {
        printf("cuOptCreateSolverSettings failed: %d\n", status);
        return 1;
    }

    status = cuOptSetIntegerParameter(
        settings,
        CUOPT_METHOD,
        CUOPT_METHOD_BARRIER
    );
    if (status != CUOPT_SUCCESS) {
        printf("cuOptSetIntegerParameter failed: %d\n", status);
        return 1;
    }

    status = cuOptSolve(problem, settings, &solution);
    if (status != CUOPT_SUCCESS) {
        printf("cuOptSolve failed: %d\n", status);
        return 1;
    }

    /*
     * Use an oversized sentinel buffer so the out-of-bounds writes can be
     * observed without corrupting unrelated memory.
     */
    const cuopt_float_t sentinel = 1.234567890123456e200;
    cuopt_float_t guarded[64];

    for (int i = 0; i < 64; ++i) {
        guarded[i] = sentinel;
    }

    status = cuOptGetDualSolution(solution, guarded);
    printf("cuOptGetDualSolution status = %d\n", status);

    for (int i = 0; i < 64; ++i) {
        if (guarded[i] != sentinel) {
            printf("changed[%d] = %g\n", i, guarded[i]);
        }
    }

    cuOptDestroySolverSettings(&settings);
    cuOptDestroyProblem(&problem);

    return 0;
}
```

## Output
```
cuOptGetNumConstraints = 1
cuOptGetDualSolution status = 0
changed[0] = nan
changed[1] = nan
changed[2] = nan
changed[3] = nan
```

## Recommended fix options:
 - Return invalid argument when querying `cuOptGetDualSolution` in this scenario
 - Copy only numConstraints elements into the solution vector

yuwenchen95:
## Description

`cuOptGetDualSolution`/`cuOptGetReducedCosts` wrote more elements than
`cuOptGetNumConstraints` documented whenever the problem had quadratic
constraints, overrunning a caller-sized buffer. Root cause: quadratic
constraints are internally reformulated into second-order cones for the
barrier solver, which grows the row count beyond the original
constraint count, but the returned dual vector was never resized back
down before being copied into the caller's buffer — and dual recovery
for QCQP isn't supported yet regardless.

Fix: dual solution / reduced cost access for QCQP now returns an error
instead of writing invalid or wrong-sized data, consistently across all
three surfaces:
- **C API**: `cuOptGetDualSolution`/`cuOptGetReducedCosts` return
  `CUOPT_INVALID_ARGUMENT` (matching the existing MIP-solution pattern).
- **Python**: `Solution.get_dual_solution()`/`get_reduced_cost()` raise
  `AttributeError` only when explicitly called — the solve itself still
  succeeds normally.
- **gRPC**: `dual_solution`/`reduced_cost` are simply omitted from the
  response payload (same mechanism already used for other optional
  fields), so a QCQP solve still succeeds end-to-end.

In all cases, a normal solve (primal solution, objective, termination
status, etc.) is unaffected for QCQP problems — only explicit
dual/reduced-cost access is gated.

Also fixes the underlying ambiguity flagged in the issue discussion:
`cuOptGetNumConstraints` / `CUOPT_ATTR_NUM_CONSTRAINTS` previously
reported linear constraints only; they now report linear + quadratic
constraints combined. Added a new `CUOPT_ATTR_NUM_QUADRATIC_CONSTRAINTS`
attribute so callers can still recover the quadratic-only count.

## Issue

closes #1751

## Checklist

- Testing
   - [x] Added tests
   - [ ] NA
- Documentation
   - [x] Added new documentation
   - [ ] NA


mlubin:
This diff seems more invasive than necessary to handle the temporary situation where QCQP doesn't return duals. We already have a documented convention of returning NaN in this situation, we just need the vectors to be of the correct length.  What code do we need to track the appropriate lengths of the vectors to return in the interfaces? We'll need this logic when QCQP duals are implemented anyway.

cpp/include/cuopt/mathematical_optimization/cuopt_c.h
```diff
@@ -1174,11 +1174,13 @@ cuopt_int_t cuOptGetSolutionFloatAttribute(cuOptSolution solution,
  *
  * These accessors use copy-out semantics: the caller allocates the output buffer and cuOpt copies
  * values into it. Array attributes are sized by the problem dimensions: variable-indexed arrays
- * have num_variables entries and constraint-indexed arrays have num_constraints entries (see
- * cuOptGetProblemIntAttribute). The sole exception to copy-out is the
- * string-array getter, which fills a caller-provided array of pointers with borrowed pointers into
- * cuOpt-owned string storage; those pointers are valid until the problem is modified or destroyed
- * and must not be freed.
+ * have num_variables entries and constraint-indexed arrays (CUOPT_ARRAY_ATTR_CONSTRAINT_*) have
+ * one entry per LINEAR constraint only — not the CUOPT_ATTR_NUM_CONSTRAINTS value, which for
```
mlubin: Are you sure CUOPT_ATTR_NUM_CONSTRAINTS includes quadratic constraints?

yuwenchen95: I have updated its computation and now it's the sum of `linear` and `quadratic` constraints. This is to address Chris's comment, which is achievable at the point.

yuwenchen95:
> e already have a documented convention of returning NaN in this situation, we just need the vectors to be of the correct length.



> This diff seems more invasive than necessary to handle the temporary situation where QCQP doesn't return duals. We already have a documented convention of returning NaN in this situation, we just need the vectors to be of the correct length. What code do we need to track the appropriate lengths of the vectors to return in the interfaces? We'll need this logic when QCQP duals are implemented anyway.

Tried to remove unnecessary bool symbol and only resize the dual solution before filling it to NaN.

cpp/include/cuopt/mathematical_optimization/cuopt_c.h
mlubin: I see. I'd propose to add a NUM_LINEAR_CONSTRAINTS attribute in that case.

yuwenchen95: Added. Now we have three attributes now: `CUOPT_ATTR_NUM_LINEAR_CONSTRAINTS `, `CUOPT_ATTR_NUM_QUADRATIC_CONSTRAINTS ` and `CUOPT_ATTR_NUM_QUADRATIC_CONSTRAINTS `.

python/cuopt/cuopt/linear_programming/solver/solver.pxd
```diff
@@ -1,4 +1,4 @@
-# SPDX-FileCopyrightText: Copyright (c) 2023-2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved. # noqa
```
chris-maes, line 1: Revert change to this file

python/cuopt/cuopt/linear_programming/solver/solver_wrapper.pyx
```diff
@@ -1,4 +1,4 @@
-# SPDX-FileCopyrightText: Copyright (c) 2023-2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved. # noqa
```
chris-maes, line 1: Revert change to this file

cpp/tests/linear_programming/c_api_tests/c_api_test.c
```diff
@@ -2504,6 +2504,197 @@ cuopt_int_t test_mip_solution_lp_methods()
...
+  cuopt_int_t status;
+  cuopt_int_t num_constraints;
+  cuopt_int_t num_linear;
+  cuopt_int_t num_quadratic;
+  int i;
+
+  /*
+   * minimize t
+   * subject to
+   *     t >= 0
```
chris-maes: This is the linear constraint? You might want to call that out in the comment.

cpp/tests/linear_programming/c_api_tests/c_api_test.c
```diff
@@ -2504,6 +2504,197 @@ cuopt_int_t test_mip_solution_lp_methods()
...
+  if (status != CUOPT_SUCCESS) {
+    printf("Error solving QCQP: %d\n", status);
+    goto DONE;
+  }
+
+  /* cuOptGetDualSolution on a QCQP solution should return CUOPT_SUCCESS and fill exactly
+   * the num_constraints-sized buffer with NaN (dual recovery not yet supported). A
+   * regression back to issue #1751 (writing more than num_constraints entries) would
+   * overrun this exactly-sized heap allocation. */
+  status = cuOptGetDualSolution(solution, dual_solution);
```
chris-maes, line 2652: So the idea is that you want to return NaN here, until we support getting the dual solution from QCQP problems? Rather than returning a bad status?

chris-maes (approved):
LGTM.

cpp/src/pdlp/cuopt_c.cpp
mlubin: We can avoid duplication and call cuOptGetProblemIntAttribute(problem, CUOPT_ATTR_NUM_CONSTRAINTS, &num_constraints_ptr) here instead.

mlubin (commented):
I was expecting that we'd want `cuOptGetDualSolution` to return a vector of length num_linear_constraints. Combined with the change in definition of getNumConstraints in this PR, we're changing the contract of cuOptGetDualSolution. Do we want to do this? Is this change in line with what the API will look like once we can return duals on quadratic constraints?

cpp/tests/linear_programming/c_api_tests/c_api_test.c
yuwenchen95: Yes, this is what we did now. Miles' issue is the dimensional mismatch caused from QCQP translation to SOCP. After the internal SOCP is solved, we need to resize the dimension of dual consistent with QCQP rather than internal SOCP.

cpp/tests/linear_programming/c_api_tests/c_api_test.c
yuwenchen95: Yes, `t>=0` is addressed as a linear constraint while `t >= -1, x1 >= 3, x2 >= 4` are regarded as variable vounds. New comment has been added to clarify it.

yuwenchen95:
> I was expecting that we'd want `cuOptGetDualSolution` to return a vector of length num_linear_constraints. Combined with the change in definition of getNumConstraints in this PR, we're changing the contract of cuOptGetDualSolution. Do we want to do this? Is this change in line with what the API will look like once we can return duals on quadratic constraints?

Ideally when we support dual recovery for QCQP from SOCP, we will have a scalar dual for quadratic constraint, so it's more reasonable to include count of quadratic constraints in `cuOptGetDualSolution`.

cpp/src/pdlp/cuopt_c.cpp
```diff
@@ -1501,6 +1501,12 @@ cuopt_int_t cuOptGetProblemIntAttribute(cuOptOptimizationProblem problem,
     case CUOPT_ATTR_HAS_QUADRATIC_CONSTRAINTS:
       *value_out = iface->has_quadratic_constraints() ? 1 : 0;
       return CUOPT_SUCCESS;
+    case CUOPT_ATTR_NUM_LINEAR_CONSTRAINTS:
+      *value_out = iface->get_n_constraints();
+      return CUOPT_SUCCESS;
+    case CUOPT_ATTR_NUM_QUADRATIC_CONSTRAINTS:
+      *value_out = static_cast<cuopt_int_t>(iface->get_quadratic_constraints().size());
```
mlubin, line 1508: Nit: do we need the static cast here?

cpp/src/pdlp/cuopt_c.cpp
```diff
@@ -1480,7 +1477,10 @@ cuopt_int_t cuOptGetProblemIntAttribute(cuOptOptimizationProblem problem,
   auto* iface = get_iface(problem);
   switch (attribute) {
     case CUOPT_ATTR_NUM_VARIABLES: *value_out = iface->get_n_variables(); return CUOPT_SUCCESS;
-    case CUOPT_ATTR_NUM_CONSTRAINTS: *value_out = iface->get_n_constraints(); return CUOPT_SUCCESS;
+    case CUOPT_ATTR_NUM_CONSTRAINTS:
+      *value_out = iface->get_n_constraints() +
+                   static_cast<cuopt_int_t>(iface->get_quadratic_constraints().size());
```
mlubin, line 1482: Nit: same here, do we need the static cast?

mlubin (approved):

cpp/src/pdlp/cuopt_c.cpp
yuwenchen95: I kept it since `size()` returns an `unsigned long` while `value_out` is a `cuopt_int_t` pointer.

cpp/src/pdlp/cuopt_c.cpp
yuwenchen95: The same concern as above.

PR #1762, opened 2026-08-21
Revert changes to PCG

nguidotti:
Revert changes to the PCG random number generator introduced in PR #1608. Note that these generators are carefully constructed and small changes can affect the quality of the random number generated. Added a skill to avoid agents to modify this file.

## Checklist

- Testing
   - [ ] Added tests
   - [ ] NA
- Documentation
   - [ ] Added new documentation
   - [ ] NA

### DIFF
--- a/cpp/src/utilities/pcgenerator.hpp
+++ b/cpp/src/utilities/pcgenerator.hpp
@@ -85,7 +85,7 @@ class pcgenerator_t {
     state               = oldstate * 6364136223846793005ULL + stream;
     uint32_t xorshifted = ((oldstate >> 18u) ^ oldstate) >> 27u;
     uint32_t rot        = oldstate >> 59u;
-    ret                 = (xorshifted >> rot) | (xorshifted << ((32u - rot) & 31u));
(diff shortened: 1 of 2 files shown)


akifcorduk (approved):
Approving on engine side. There seems to be lots of SKILL layout, indentation changes. You could maybe revert that.

aliceb-nv:
Have we measured this change? I see identical codegen when compiling for modern C++:
https://godbolt.org/z/8x7oM6Mvz

nguidotti:
> Have we measured this change? I see identical codegen when compiling for modern C++: https://godbolt.org/z/8x7oM6Mvz

Unless there is a strong reason, I would prefer that an agent do not change the implementation of a random number generator. Generally, these codes are carefully constructed and small changes can alter how it generates the sequence of random numbers.

aliceb-nv:
I do agree that the non-style SKILLS changes are useful.

nguidotti:
I am reverting the style changes

ramakrishnap-nv (approved):

chris-maes (approved):
LGTM

PR #1763, opened 2026-08-21
Add Kitmaker Portal wheel publishing via Charon Ferry

ramakrishnap-nv:
Publishes cuOpt wheels to Kitmaker via Charon Ferry: attach wheels to a GitHub Release on version tags, then request a Kitmaker Portal release pointing at them.

- `kitmaker_portal.yaml`: reusable workflow. `smoke-test` mode polls an existing Kitmaker status ID. `release` mode collects matching wheel assets from a GitHub Release, requests a Kitmaker Portal release, and polls it to completion. Directly dispatchable (`workflow_dispatch`) for manual testing/ops.
- `build.yaml`: on a real release tag, `release-github` attaches built wheels to a GitHub Release, then 7 `kitmaker-portal-release-*` jobs (one per package x CUDA-major-version project: `libcuopt`, `cuopt`, `cuopt-server` each x cu12/cu13, plus `cuopt-sh-client`) call Kitmaker.

`upload: false` everywhere for now — a deliberate staged-rollout default. Flip to `true` once the team signs off on going live; requires the 7 `KITMAKER_PROJECT_ID_*` repo variables and `KITMAKER_PORTAL_AUTHORIZATION`/`KITMAKER_PORTAL_ACC_OWNER_EMAIL` secrets to be set (already done on this repo for testing).

## Validation

Dispatched a full real build off this branch (via a disposable, non-numeric `v*` tag so it satisfied Charon's tag-scoped tenant policy without triggering an actual release), through the entire chain: `cpp-build`/`wheel-build-*` → `wheel-publish-*` → `release-github` (real GitHub Release with real wheels attached) → all 7 `kitmaker-portal-release-*` jobs, each confirmed to reach Kitmaker's `completed` status via the real Charon Ferry → Kitmaker Portal round-trip (dry-run, `upload: false`) against the real Kitmaker projects.

Run: https://github.com/NVIDIA/cuopt/actions/runs/32764788538

Along the way, found and fixed: missing `Bearer` auth prefix, swallowed error response bodies, wrong runner (missing `tbot`, fixed by using `teleport-actions/setup`+`application-tunnel`), Charon's ref-scoping requiring a matching `refs/tags/v*` ref, an undocumented `processing` status value, a `pic`-masking gap in `smoke-test`, the wrong `download-artifact` pattern (`cuopt_wheel_*` not `wheel_*`), and Ferry OIDC token expiry during long polls (now refreshed every iteration). Also surfaced and reported an unrelated compile bug on `main` (#1776) while chasing down what turned out to be a stale/flaky build failure, not caused by this PR.

Disposable test releases, tags, and Kitmaker Portal test-project entries used during validation have been cleaned up.

image


gforsyth (approved):
I can't speak to the Charon part of this handoff, but the wheel publishing steps and the payload creation for Kitmaker look good to me.

I was under the impression that Kitmaker _can't_ reach outside of the corpnet to pull artifacts, but I will very happily be proven wrong on that front.

Construction-wise, this looks correct -- we'll just have to test that all the pieces actually communicate with one-another. If you want to create a "test release" to trial this out (and then delete that release), we should make sure to first disable the `upload-conda` workflows since those will automatically upload conda artifacts if it detects that it's running on a release tag/ref

ramakrishnap-nv:
> I can't speak to the Charon part of this handoff, but the wheel publishing steps and the payload creation for Kitmaker look good to me.
> 
> I was under the impression that Kitmaker _can't_ reach outside of the corpnet to pull artifacts, but I will very happily be proven wrong on that front.
> 
> Construction-wise, this looks correct -- we'll just have to test that all the pieces actually communicate with one-another. If you want to create a "test release" to trial this out (and then delete that release), we should make sure to first disable the `upload-conda` workflows since those will automatically upload conda artifacts if it detects that it's running on a release tag/ref

I have tested end to end run with release being created as part of the build.yaml and then teriggering kitmaker via a job which completed successfully https://github.com/NVIDIA/cuopt/actions/runs/32764788538. I had disabled upload so nothing was uploaded. May be once I merge this, look for any bugs and then enable it another PR.

PR #1764, opened 2026-08-21
Improve duplicate cut detection

hlinsen:
Use hash method for check duplicate method:

Instance / call | Legacy main | Current hash median | CSR median | CSR vs current hash
-- | -- | -- | -- | --
k1mushroom / 3 | 80.094 s | 0.138209 s | 0.059410 s | 2.324×
splice1k1 / 7 | 61.152 s | 0.202498 s | 0.104062 s | 1.887×
hypothyroid-k1 / 4 | 11.455 s | 0.055137 s | 0.023810 s | 2.316×

The new method now reproduces main’s retention exactly:

Metric | Result
-- | --
MIPLIB instances | 240/240
Successful runs | 240/240
Instances invoking duplicate checking | 228
Duplicate-check calls | 1,669
Calls with different masks | 0
Row-level differences | 0
Crashes/errors | 0
Legacy/indexed retained rows across calls | 11,109,430 / 11,109,430
Legacy/indexed removed rows across calls | 3,320,516 / 3,320,516

[per_instance_audit.csv](https://github.com/user-attachments/files/31656401/per_instance_audit.csv)


cpp/src/cuts/cuts.cpp
```diff
@@ -1227,113 +1246,139 @@ f_t cut_pool_t<i_t, f_t>::cut_orthogonality(i_t i, i_t j)
...
+    i_t matching_group = no_group;
+    for (i_t group = bucket->second; group != no_group; group = groups[group].next) {
+      if (rows_are_duplicates(groups[group].representative, r)) {
+        matching_group = group;
+        break;
       }
     }
+    if (matching_group == no_group) {
+      groups.push_back({r, r, bucket->second});
+      bucket->second = static_cast<i_t>(groups.size() - 1);
```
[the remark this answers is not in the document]
hlinsen: It impacts performance as it can force exact coefficient comparisons.

cpp/src/cuts/cuts.cpp
```diff
@@ -1227,113 +1229,132 @@ f_t cut_pool_t<i_t, f_t>::cut_orthogonality(i_t i, i_t j)
...
 
-  csc_matrix_t<i_t, f_t> cut_storage_csc(0, 0, 1);
-  cut_storage_.to_compressed_col(cut_storage_csc);
-  i_t n = cut_storage_csc.n;
-  i_t m = cut_storage_csc.m;
+  constexpr f_t duplicate_tolerance = 1e-10;
+  const i_t no_group                = -1;
 
-  const i_t sentinel = std::numeric_limits<i_t>::max();
+  struct duplicate_group_t {
```
akifcorduk: Please put the structs outside of the function.

cpp/src/cuts/cuts.cpp
```diff
@@ -1227,113 +1229,132 @@ f_t cut_pool_t<i_t, f_t>::cut_orthogonality(i_t i, i_t j)
...
+  groups.reserve(m);
+  std::unordered_map<uint64_t, i_t> buckets;
+  buckets.reserve(m);
+
+  auto coefficients_match = [&](f_t a, f_t divisor_a, f_t b, f_t divisor_b) {
+    const f_t ratio = (a / divisor_a) * (divisor_b / b);
+    return ratio >= 1.0 - duplicate_tolerance && ratio <= 1.0 + duplicate_tolerance;
+  };
+
+  auto rows_are_duplicates = [&](i_t first, i_t second) {
```
akifcorduk: For large lambdas, we prefered named functions on couple of previous PRs. Let's keep this convention and put it in a function.

cpp/src/cuts/cuts.cpp
```diff
@@ -1227,113 +1229,132 @@ f_t cut_pool_t<i_t, f_t>::cut_orthogonality(i_t i, i_t j)
...
+      const f_t value_abs = std::abs(cut_storage_.x[p]);
+      if (value_abs > pivot_abs ||
+          (value_abs == pivot_abs && cut_storage_.j[p] < cut_storage_.j[pivot])) {
+        pivot = p;
+      }
+    }
+    const f_t divisor = cut_storage_.x[pivot];
+    divisors[r]       = divisor;
+
+    uint64_t support_hash = splitmix64_mix(static_cast<uint64_t>(row_end - row_start));
```
akifcorduk: I would put this hash and bucket logic in a separate function as it can also be used for other equality checks. Could you also add few comments on the logic?

akifcorduk (commented):
Thanks Hugo! I would get @chris-maes review as well as this replaces his code. Did you check on a broader benchmark if the retention is the same and there are no crashes?

cpp/src/cuts/cuts.cpp
```diff
@@ -1292,50 +1312,42 @@ void cut_pool_t<i_t, f_t>::check_for_duplicate_cuts()
...
-    if (set_r > 0 && set_r < sentinel && cuts_to_remove[r] == 0) {
-      // This cut has a duplicate
-      for (i_t i = r + 1; i < m; i++) {
-        if (sets[i] == set_r) {
-          const f_t f_r     = divisors[r];
-          const f_t f_i     = divisors[i];
-          const f_t theta_r = rhs_storage_[r] / f_r;
-          const f_t theta_i = rhs_storage_[i] / f_i;
-          if (f_r > 0 && f_i > 0) {
-            // We have sum_j d_rj / f_r x_j >= rhs_r / f_r = theta_r
```
chris-maes, line 1310: Here and below. These comments are crucial for a human to understand what the algorithm is doing. AI loves to delete them. Please ask the agent to add the comments back to the code.

cpp/src/cuts/cuts.cpp
```diff
@@ -1227,55 +1229,73 @@ f_t cut_pool_t<i_t, f_t>::cut_orthogonality(i_t i, i_t j)
 template <typename i_t, typename f_t>
 void cut_pool_t<i_t, f_t>::check_for_duplicate_cuts()
 {
-  // Algorithm from Finding Duplicate Rows in a Linear Programming Model
```
chris-maes, line 1230: This was a tricky algorithm to implement. I'm worried about making changes to it. 

Can you give me a high-level summary of what the changes are? And why they are correct?

hlinsen: I've run a checker methods against the 2 methods to validate we remove all duplicates and if not the same that the cuts with hashing are stronger.  The original method was keeping more cuts than necessary.
I'll add some tests to the PR.

chris-maes: Thanks. Can you also explain the changes?

hlinsen: We use the same duplicate row algorithm and preserve the same first match and strongest cut decisions. The main change is the data structure used to find comparison candidates.
Previously, for each cut appearing in a column, we scanned every later cut in that column and checked whether it belonged to the same partition set. The final removal phase similarly scanned every later cut row.
The new implementation builds hash indexes from a partition set ID to the ordered CSC positions or final cut rows belonging to that set. We can therefore visit only the relevant same set candidates while preserving their original order.
This replaces quadratic full tail scans with output sensitive indexed scans. When duplicate groups are small, as they are in our benchmarks, the behavior is effectively near linear. The theoretical worst case remains quadratic if almost every cut belongs to one large group.

Suppose one column contains eight cut rows whose current partition IDs are:
```
Position:  0  1  2  3  4  5  6  7
Set ID:    A  B  C  D  E  F  G  A
```

When processing the first A, the old implementation scans all seven later entries and checks their set IDs:
`B, C, D, E, F, G, A  -> 7 visits, 1 relevant candidate`

The new implementation first builds:
```
A -> [0, 7]
B -> [1]
C -> [2]
```

It then looks up A and visits only position 7:
`A -> 1 visit, 1 relevant candidate`

cpp/src/cuts/cuts.cpp
```diff
@@ -1292,50 +1305,51 @@ void cut_pool_t<i_t, f_t>::check_for_duplicate_cuts()
...
+          cuts_to_remove[r] = 1;  // Remove row r.
+        } else {
+          // theta_r < theta_i, so cut r is strictly stronger than cut i.
+          cuts_to_remove[i] = 1;  // Remove row i.
         }
       }
     }
   }
 
+  const i_t num_cuts_to_remove =
```
chris-maes: This costs O(m) just to find the number of cuts to remove. The previous way (where we tracked the number of cuts to remove inline by incrementing `num_cuts_to_remove` is better.

cpp/src/cuts/cuts.cpp
```diff
@@ -1227,55 +1229,66 @@ f_t cut_pool_t<i_t, f_t>::cut_orthogonality(i_t i, i_t j)
...
   for (i_t j = 0; j < n; j++) {
     i_t r0        = -1;
     i_t new_rows  = 0;
     i_t new_set_0 = new_set;
     new_set++;
     const i_t col_start = cut_storage_csc.col_start[j];
     const i_t col_end   = cut_storage_csc.col_start[j + 1];
+
+    set_groups.build(
+      col_start, col_end, new_set_0, [&](i_t p) { return sets[cut_storage_csc.i[p]]; });
```
chris-maes: Is it possible to do this without the lambda function? These can be slow.

Both calls are just indexing into sets. The only difference is this call does an indirect index through `cut_storage_csc.i[p]` whereas the other call does a direct index.

cpp/src/cuts/cuts.cpp
```diff
@@ -1227,55 +1229,66 @@ f_t cut_pool_t<i_t, f_t>::cut_orthogonality(i_t i, i_t j)
...
+
     for (i_t p = col_start; p < col_end; p++) {
       const i_t r    = cut_storage_csc.i[p];
       const f_t a_rj = cut_storage_csc.x[p];
       const f_t f_r  = divisors[r];
       if (sets[r] == 0) {
-        r0          = r;  // To enable use to find this new set later
```
chris-maes, line 1257: Don't loose this comment.

cpp/src/cuts/cuts.cpp
```diff
@@ -1227,55 +1229,66 @@ f_t cut_pool_t<i_t, f_t>::cut_orthogonality(i_t i, i_t j)
...
+        r0          = r;
         sets[r]     = new_set_0;
         divisors[r] = a_rj;
         new_rows++;
       } else if (sets[r] < new_set_0) {
-        // Look over indices a_ij with i > r
```
chris-maes, line 1262: Add a corresponding comment

chris-maes: Maybe; 
```
// Loop over all indices a_ij with i > r where i is in the same set as r. 
```

cpp/src/cuts/cuts.cpp
```diff
@@ -1227,55 +1229,66 @@ f_t cut_pool_t<i_t, f_t>::cut_orthogonality(i_t i, i_t j)
...
-        for (i_t q = p + 1; q < col_end; q++) {
+        const i_t old_set      = sets[r];
+        bool matched           = false;
+        const auto set_entries = set_groups.entries_in_set(old_set);
+        auto q_position        = std::upper_bound(set_entries.begin(), set_entries.end(), p);
```
chris-maes: I'm not sure I understand what is going on here. p is our pointer into the CSC row indices/values array. 

For each column j, we build a data structure that groups a_ij that belong to the same set. So if a_1j, a_3j, a_7j are in the same set, we would store [p1, p2, p3] such that values[p1] = a_1j, values[p2] = a_3j, and values[p3] = a_7j.  Let's say p > p1 but p < p2, then we look at [p1, p2]. 

Each time it costs us O( log (|set_entries| ) to narrow down what we want to look at. 

I wonder if we can avoid this somehow.

hlinsen: I think we can with one extra buffer that records the positions.

`std::vector<i_t> position_by_entry_;`
While filling the compressed groups:
```
const i_t position = next_entry_in_set_[bucket]++;
entries_[position] = entry;
position_by_entry_[entry - first] = position;
```
Then for the current CSC position p:
```
const i_t position = position_by_entry_[p - first];
const auto later_entries =
  std::span(entries_.begin() + position + 1, entries_.begin() + set_end);
```

chris-maes (approved):
LGTM. Thanks for making all the changes and speeding up this algorithm.

PR #1775, opened 2026-08-24
Fix concurrent clique table publication

### LINKED ISSUE #1774: [BUG] Concurrent initial-clique publication can race during MIP solve
### Describe the bug

A long-lived Python process using cuOpt 26.8.0 aborted on its 24th sequential MIP solve with `free(): invalid size`. glibc raised SIGABRT while an OpenMP initial-clique task destroyed its task-local `simplex::user_problem_t<int, double>` copy. The failure occurred just after the concurrent LP root relaxation started.

The same workload is stable when each solve runs in a fresh process. Disabling both `mip_clique_cuts` and `mip_zero_half_cuts` is the targeted A/B control because that prevents the asynchronous initial-clique task from launching. There was no OOM kill (about 4.9 GB peak on a 61 GB host) and no NVIDIA Xid or GPU error.

### Source-level race on current main

`branch_and_bound_t::solve` launches an OpenMP task that eventually assigns a newly allocated table to the member `clique_table_`. Meanwhile the root-solve thread can copy that same `std::shared_ptr` member into `cut_generation_t`. Concurrent assignment and copy of one `shared_ptr` object is an unsynchronized read/write and therefore undefined behavior. It can also leave cut generation holding `nullptr` even after the task publishes the table.

The table should instead be allocated and assigned to the owning member before task creation. The task can then mutate only the pointee, with the existing signal/task dependency joining mutation before cut generation reads it.

### Native stack

```text
free(): invalid size
malloc_printerr
free
cuopt::mathematical_optimization::simplex::user_problem_t<int,double>::~user_problem_t()
libcuopt.so
OpenMP task worker
```

The destructor is where heap corruption was detected, not necessarily where it originated. The model and evolving separator-cut pool can be provided privately if useful; together they are roughly 36 MB.

### Environment

- cuopt-cu12 26.8.0, git hash 400863c1
- Python 3.14.2
- Arch Linux x86-64, glibc 2.44
- NVIDIA driver 610.57.04, CUDA 12.9 wheels
- GeForce RTX 4070 Ti SUPER
- AMD Ryzen 7 7800X3D

Possibly related to #1768, but this occurrence uses sequential solves in one long-lived process and has a specific initial-clique-task stack and source-level publication race.

derluke:
Closes #1774.

- Allocate and publish the initial clique table before launching its OpenMP producer task, avoiding an unsynchronized read/write of the owning `shared_ptr`.
- Let the task mutate only the pre-published table; the existing signal/dependency still joins mutation before cut generation reads it.
- Exercise 25 sequential solves with concurrent initial-clique generation enabled.

Pre-commit passes. The C++ regression requires the cuOpt build environment and MIP datasets, so it is left for CI.

### DIFF
--- a/cpp/tests/mip/miplib_test.cu
+++ b/cpp/tests/mip/miplib_test.cu
@@ -90,6 +90,28 @@ TEST(mip_solve, low_thread_count_test)
   test_variable_bounds(problem, solution.get_solution(), settings);
 }
 
+TEST(mip_solve, repeated_concurrent_initial_clique_generation)
+{
+  mip_solver_settings_t<int, double> settings;
+  settings.num_cpu_threads = 8;
+  settings.time_limit      = 30;
+  settings.work_limit      = 1;
+  settings.clique_cuts     = 1;
+  settings.zero_half_cuts  = 1;
+
+  const raft::handle_t handle_{};
+
+  auto path    = make_path_absolute("mip/gen-ip054.mps");
+  auto problem = cuopt::mathematical_optimization::io::read_mps<int, double>(path, false);
+  handle_.sync_stream();
+
+  constexpr int num_solves = 25;
+  for (int solve = 0; solve < num_solves; ++solve) {
+    auto solution = solve_mip(&handle_, problem, settings);
+    EXPECT_NE(solution.get_termination_status(), mip_termination_status_t::NoTermination);
+  }
+}
+
 // Verify --node-limit is respected: swath1 normally requires several thousand B&B
 // nodes to prove optimality, so capping at 1000 forces the solver to stop early with
 // a feasible (but not necessarily optimal) solution.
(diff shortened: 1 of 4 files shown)


cpp/tests/mip/miplib_test.cu
```diff
@@ -90,6 +90,28 @@ TEST(mip_solve, low_thread_count_test)
...
+  settings.zero_half_cuts  = 1;
+
+  const raft::handle_t handle_{};
+
+  auto path    = make_path_absolute("mip/gen-ip054.mps");
+  auto problem = cuopt::mathematical_optimization::io::read_mps<int, double>(path, false);
+  handle_.sync_stream();
+
+  constexpr int num_solves = 25;
+  for (int solve = 0; solve < num_solves; ++solve) {
```
akifcorduk, line 109: These tests are prohibitive for the CI. Could you remove this test completely?

akifcorduk (changes requested):
I was not able to reproduce it locally, but this is indeed a potential race. Could you remove the tests?

PR #1778, opened 2026-08-24
refactor(logger): give each component library its own logger

ramakrishnap-nv:
`cuopt::default_logger()` is one process-wide instance, defined in `logger.cpp` and shared by every solver. Splitting libcuopt into component libraries (#1622) means routing and mathopt should log independently, and nothing should have to exist purely to host that state.

## Change

The logger is header-only and **hidden**. Hidden visibility is what does the separating, and it is not optional:

```
6: 000000000000401c  4 OBJECT  UNIQUE DEFAULT  23 _ZZN4demo7counterEvE1c
```

The static local of an inline function is emitted as `STB_GNU_UNIQUE`, which glibc merges across the whole process **regardless of `RTLD_LOCAL`** -- so a header-only logger left at default visibility is still one shared instance, both when linked and when `dlopen`ed the way `load.py` does it. Marking the namespace `CUOPT_EXPORT` would silently undo this PR.

## Configuring a logger you cannot reach

Callers outside the libraries have their own logger and cannot touch a library's. Each component therefore exports a configure entry point, the only logging symbols that cross a boundary:

```
$ nm -DC libcuopt.so | grep logging
T cuopt::mathematical_optimization::configure_logging(...)
T cuopt::mathematical_optimization::reset_logging()
T cuopt::routing::configure_logging(...)
T cuopt::routing::reset_logging()
```

That gives two types with distinct jobs:

- `init_logger_t(file, console)` -- configures the logger of whichever image constructs it. Library code already used it this way, so `pdlp/solve.cu`, `mip_heuristics/solve.cu` and `grpc/client/solve_remote.cpp` each configure their own library's logger with no change.
- `init_component_logger_t(file, console, target = mathopt)` -- reaches a chosen library from outside. It defaults to mathopt because every external caller today is LP or MILP, so all eight existing sites (CLI x2, `dual_simplex` tests x6) keep their meaning and routing is opted into explicitly. The routing branch sits behind `CUOPT_HAS_ROUTING`, since `SKIP_ROUTING_BUILD` means the symbol may not exist.

## Two fixes needed to keep one log file working

Both showed up running `cuopt_cli`, not reading the code.

- The exported entry point takes the same ref-count guard `init_logger_t` takes. Without it the MIP solve path built its own `init_logger_t` mid-run and, with truncate set, cleared the file the CLI had already written to.
- File sinks always open in append mode, with a single explicit truncate up front. A non-appending sink writes from offset 0 and silently overwrites what another logger has appended.

## Routing's errors were being dropped

`routing::solve` logs through `CUOPT_LOG_ERROR` in its catch blocks, but routing never constructed an `init_logger_t`. The default sink is a buffer that is only drained when one is constructed, so those errors went nowhere. Routing now initialises its own logger from `get_error_logging_mode()`. Pre-existing bug, fixed here because per-library logging forces routing to own its configuration.

## Testing

`cuopt_cli` writes both its own and the solver's messages to one file -- 67 lines, against 62 when the CLI's were being silently overwritten -- and two consecutive runs both give 67, so truncation still works and nothing leaks across runs.

For `ctest` I built a baseline by stashing onto clean main and rebuilding: identical results, same 10 failing suites and same 908 gtest failures, 92% both. Those failures are environmental in my setup (CUDA stream-capture errors, a null-offsets validation), not from this change.

## Follow-ups

Routing has no `log_file` / `log_to_console` in `solver_settings_t`, only `set_error_logging_mode`, so a library caller cannot yet send routing's log to a file the way the LP settings allow. Worth adding in the same shape as the seed in #1717.


ramakrishnap-nv:
@coderabbitai review

ramakrishnap-nv:
ok to test

ramakrishnap-nv:
@coderabbitai review

ramakrishnap-nv:
@coderabbitai review

ramakrishnap-nv:
@coderabbitai review

ramakrishnap-nv:
@coderabbitai full review

cpp/src/utilities/logger.hpp
```diff
@@ -15,39 +15,406 @@
...
+  default_logger().sinks().clear();
+  default_logger().sinks().push_back(default_sink());
+#if CUOPT_LOG_ACTIVE_LEVEL >= RAPIDS_LOGGER_LOG_LEVEL_INFO
+  default_logger().set_pattern("%v");
+#else
+  default_logger().set_pattern(default_pattern());
+#endif
+  default_logger().set_level(default_level());
+  default_logger().flush_on(rapids_logger::level_enum::debug);
+}
```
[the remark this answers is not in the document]
ramakrishnap-nv: Real race, but pre-existing and out of scope here. `logger.cpp` on main does the same unsynchronized mutation — `reset_default_logger` at lines 129-130 and the `init_logger_t` body at 160-168 both `clear()` and `push_back()` on the sink vector with no lock, and the emission path has never taken one. This PR moved that code into the header without changing it.

I would rather not take the suggested fix in this PR. Holding `g_guard_mutex` across `CUOPT_LOG_*` would serialize every log call in the codebase on a single mutex, on the hot path of solvers that log per iteration, and the emission path is generated by rapids_logger rather than written here. That is a performance and layering decision that deserves its own change and its own benchmark, not a rider on a visibility refactor.

Worth noting the exposure is narrow in practice: sinks are mutated at solve boundaries, where an `init_logger_t` or a component `configure_logging` is constructed, not concurrently with the solve that is emitting. Happy to open an issue for it if you think it is worth tracking.

cpp/src/utilities/logger.hpp
```diff
@@ -15,39 +15,406 @@
...
+  std::lock_guard<std::mutex> lock(g_guard_mutex);
+
+  auto existing_guard = g_active_guard.lock();
+  if (existing_guard) {
+    // Reuse existing configuration, just hold a reference to keep it alive
+    guard_ = existing_guard;
+    return;
+  }
+
+  apply_logger_config(log_file, log_to_console, truncate);
```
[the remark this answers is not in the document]
ramakrishnap-nv: Fixed in `3f042f8d`, thanks — this was a real gap. `init_logger_t` called `apply_logger_config` unguarded, and since that clears the sinks before installing the new ones, a throw left the logger with none at all and silently dropped everything afterwards. `configure_logging_impl` already handled it; its sibling did not.

Added `logger.failed_init_logger_restores_a_sink` alongside the existing `failed_configure_does_not_wedge_later_ones`, so both entry points now have a regression for the failure.

ramakrishnap-nv:
@coderabbitai full review

cpp/src/utilities/logger.hpp
```diff
@@ -15,39 +15,345 @@
...
+};
+
+// Guard object whose destructor resets the logger
+struct logger_config_guard {
+  ~logger_config_guard() { cuopt::reset_default_logger(); }
+};
+
+// Weak reference to detect if any init_logger_t instance is still alive
+inline std::weak_ptr<logger_config_guard> g_active_guard;
+inline std::mutex g_guard_mutex;
```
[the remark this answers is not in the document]
ramakrishnap-nv: Fixed in `e031c6b6` — real race, thanks.

A guard's refcount reaching zero expires `g_active_guard` *before* its destructor runs, so another thread can see no live configuration, apply its own and install a new guard inside that window. The old destructor then reset unconditionally and wiped the configuration that thread had just installed. It also mutated the sink vector without the mutex, racing `apply_logger_config`.

The guard now records the generation it configured and resets only if that generation is still current, taking `g_guard_mutex` while it checks. Sink mutation on the configure and reset paths is consistently under that mutex now, which also narrows the unsynchronized-sink concern from the previous round to emission-vs-mutation only.

On the requested test, I want to be straight about what it does and does not cover. I wrote one and it **passes with the generation check disabled**, so it does not demonstrate this race. Entering the window needs the refcount to hit zero exactly while another thread is inside `make_logger_config`, and under contention `g_active_guard.lock()` nearly always succeeds instead, so the window is rarely entered at all. Rather than ship a test that looks like a regression test but cannot fail for the reason claimed, I kept it and labelled it a concurrency smoke test — it does exercise concurrent configure/release/emit on a shared path, and is where a sanitizer run would surface the sink mutation. Reproducing the interleaving deterministically would need a test-only hook inside the destructor, which I did not think was worth putting in production code.

ramakrishnap-nv:
@coderabbitai review

cpp/src/utilities/logger.hpp
```diff
@@ -15,39 +15,311 @@
...
+  ~logger_config_guard()
+  {
+    std::lock_guard<std::mutex> lock(g_guard_mutex);
+    if (active_config_generation() != generation_) { return; }
+    cuopt::reset_default_logger();
+  }
+
+ private:
+  uint64_t generation_;
+};
```
aliceb-nv: What's the rationale behind this change? In what scenarios is refcounting necessary for the logger object? It would seem to me explicit lifetime management is both preferrable and simpler here

cpp/src/utilities/logger.hpp
```diff
@@ -15,39 +15,311 @@
...
+    switch (target) {
+      case log_target_t::routing:
+#ifdef CUOPT_HAS_ROUTING
+        handle_ = cuopt::routing::configure_logging(log_file, log_to_console, truncate);
+#else
+        // Silently doing nothing here would look like a working logger that drops every
+        // message, and the cause -- a SKIP_ROUTING_BUILD mismatch -- would be invisible.
+        throw std::runtime_error(
+          "cuOpt was built with SKIP_ROUTING_BUILD, so routing's logger does not exist and "
+          "log_target_t::routing cannot be configured.");
```
aliceb-nv: I believe cuopt_expects() is the preferred pattern in the repo for throwable errors in production. But shouldn't this be an assert()?

cpp/tests/utilities/test_logger.cpp
```diff
@@ -0,0 +1,290 @@
...
+    CUOPT_LOG_ERROR("after_reconfigure");
+  }
+
+  EXPECT_NE(read_file(first).find("before_reconfigure"), std::string::npos);
+  EXPECT_NE(read_file(second).find("after_reconfigure"), std::string::npos)
+    << "the second configuration left the logger reset to the buffer sink";
+
+  std::remove(first.c_str());
+  std::remove(second.c_str());
+}
```
aliceb-nv, line 89: I see we're building to support nested logging configs. Is this something that can actually happen in the solvers? My understanding is that there is always only a single logger object active during a solve lifetime.

cpp/src/utilities/logger.hpp
```diff
@@ -15,39 +15,311 @@
...
+                                                     bool truncate);
+}  // namespace cuopt::mathematical_optimization
+
+#ifdef CUOPT_HAS_ROUTING
+namespace cuopt::routing {
+CUOPT_EXPORT std::shared_ptr<void> configure_logging(const std::string& log_file,
+                                                     bool log_to_console,
+                                                     bool truncate);
+}  // namespace cuopt::routing
+#endif
```
aliceb-nv: I'm curious why we have two explicitely different functions for routing and for mathopt here. Can't we just build one logger object in the routing library, and another in the mathopt library?

aliceb-nv (commented):
Thanks Ram! I'm a little unclear on some of these edits, I feel like we could achieve the same with much fewer changes
The goal is just to ensure that both libraries get their own logger object, right?

ramakrishnap-nv:
> Thanks Ram! I'm a little unclear on some of these edits, I feel like we could achieve the same with much fewer changes The goal is just to ensure that both libraries get their own logger object, right?

Yes, so when we split libs, we don't need to maintain a common lib for routing and lp.

cpp/src/utilities/logger.hpp
```diff
@@ -15,39 +15,235 @@
...
+ * logger instance. This is not optional: an inline function's static local is emitted as an
+ * STB_GNU_UNIQUE symbol, which glibc merges process-wide regardless of RTLD_LOCAL, so default
+ * visibility would collapse every library back into one shared logger. Do not mark this
+ * namespace CUOPT_EXPORT.
  *
- * @return logger& The default logger
+ * Each solver already builds an init_logger_t from its own settings on entry, so passing a
+ * log file through settings configures that library's logger. An executable linking cuopt
+ * has a separate logger for its own messages, and when both write the same file the solver
+ * would truncate it mid-solve and discard what the executable had already written. The one
```
ramakrishnap-nv, line 37: You're right, and it didn't need them.

Every solver already builds an `init_logger_t` from its own settings on entry — pdlp ×3, mip, grpc ×2, cython — so passing a log file through settings already configures that library's logger. The enum, the dispatcher class, the per-component entry points and `CUOPT_HAS_ROUTING` were solving a problem that only one caller actually has. I had that fact in front of me when I designed it and asked "how do external callers reach the library's logger?" instead of "do they need to?"

The one caller is `cuopt_cli`, which writes the same log file as the solver. Their loggers are separate objects once the state is hidden, so the solver's own initializer truncated the file mid-solve and discarded what the CLI had already written. On `main` that never happened because both shared one logger and the second initializer reused the first's configuration. I cut the cross-library API to nothing first and measured that regression before settling on this — CLI messages vanished from `--log-file` entirely.

So it needs one signal, not an API surface:

```cpp
cuopt::mathematical_optimization::configure_logging(log_file, log_to_console, truncate)
```

The CLI establishes the configuration, the solver's own initializer reuses it. Routing needs no entry point at all — the CLI is LP/MIP only, and routing configures its own logger from its settings like everything else.

`f1009328` removes 235 lines: `log_target_t`, `init_component_logger_t`, routing's `logger_entry.cpp`, `CUOPT_HAS_ROUTING` on three targets, and the `dual_simplex` test edits, which now match `main` again. `libcuopt` exports one logging symbol instead of four.

What's left is the actual goal plus its consequences: hidden visibility so each library owns its logger, append-mode sinks with a single truncate, routing initialising its own logger in `solve` (a pre-existing bug — its `CUOPT_LOG_ERROR` calls went into a buffer nothing drained), and a `ci/check_symbols.sh` assertion that the state stays hidden, since otherwise re-exporting that namespace silently collapses every library back to one shared logger with no build or test failure.

The stale-guard race a previous revision fixed is pre-existing on `main`, so it moved out to #1831 rather than riding along here.

ramakrishnap-nv:
@aliceb-nv — you were right, and `f1009328` cuts 235 lines.

Every solver already builds an `init_logger_t` from its own settings on entry, so passing a log file through settings already configures that library's logger. The enum, dispatcher class, per-component entry points and `CUOPT_HAS_ROUTING` were solving a problem only one caller has: `cuopt_cli`, which writes the same log file as the solver, so the solver's own initializer truncated it mid-solve and discarded what the CLI had written. I tried cutting the cross-library API to nothing first and measured that regression — CLI messages disappeared from `--log-file` entirely — so one signal is needed, but one exported function rather than an API surface. Routing needs none, since the CLI is LP/MIP only.

`libcuopt` now exports one logging symbol instead of four, and the `dual_simplex` test edits are gone (back to matching main). Reasoning in more detail [here](https://github.com/NVIDIA/cuopt/pull/1778#discussion_r3896405448); the stale-guard race a previous revision fixed is pre-existing on main and moved to #1831.

aliceb-nv (approved):
Thanks! Let's just make sure this also works with run_mip.cpp, since it is another executable acting as a frontend. I don't remember if the static library is linked into it, or if it calls into the .so (in which case it would need a configure_logging call as well, I imagine?)

ramakrishnap-nv:
Good catch — it was broken, and I've fixed it in `a0a9628e`.

`solve_MIP` links the shared `cuopt`, not the static one, so it has its own logger exactly like `cuopt_cli`. It emits 12 `CUOPT_LOG_*` of its own and sets `settings.log_file`, but never constructed an `init_logger_t` — it didn't need to when one shared logger existed, because the library's own initializer configured that logger and drained the buffer, so those messages surfaced. Once each library owns its logger that stops being true.

I built it with `-DBUILD_MIP_BENCHMARKS=ON` and confirmed the regression before fixing: `running file` and `run_solver` appeared **neither** in the log file nor on stdout. They were going into this image's buffer sink with nothing to drain it. So yes — it needed a `configure_logging` call, as you guessed.

After the fix, the log has both this image's messages and the solver's (67 lines), and truncates once per run.

I checked the other executables while I was there:

| Executable | links | own `CUOPT_LOG_*` | affected |
|---|---|---|---|
| `cuopt_cli` | shared | 16 | already handled |
| `solve_MIP` | shared | 12 | **was broken, now fixed** |
| `solve_LP` (`run_pdlp.cu`) | shared | 0 | no |
| `cuopt_grpc_server` | shared | 0 (own logger) | no |

Worth flagging that nothing in CI would have caught this: `solve_MIP` is behind `BUILD_MIP_BENCHMARKS`, which is `OFF`, so no CI job builds it. Same for `solve_LP`. If those frontends are expected to keep working, a build-only job for the benchmark targets would be worth having — happy to open an issue for it.

tmckayus (approved):
approve for infra

KyleFromNVIDIA (approved):

Issue #1736, opened 2026-08-18
[BUG] Variable.VariableType is not normalized to VType (str/bytes/enum coexist)


jackthepunished:
**Describe the bug**
`Variable.VariableType` is not normalized, so it can hold three different representations of the same type:

- a `VType` member when built through the Python API with `VType.INTEGER` / `INTEGER`;
- a plain `str` (`"I"`) when the problem comes from `Problem.readMPS` / `Problem.read` — the parser builds `var_types` with `dtype='str'` and `_from_data_model` passes each element straight to `addVariable(vtype=...)`;
- `bytes` (`b"I"`) when a caller passes them — `addVariable` and `setVariableType` accept any value unchecked.

Downstream code copes ad hoc: `Problem.IsMIP` tests `in ("I", "S", b"I", b"S")` and `solver.is_mip` branches on `isinstance(var_types[0], bytes)`. `VType("I")` works because `VType` is a `str` enum, but `VType(b"I")` raises `ValueError`.

**Steps/Code to reproduce bug**
```python
from cuopt.linear_programming.problem import Problem, VType, INTEGER

p = Problem()
a = p.addVariable(vtype=INTEGER)
b = p.addVariable(vtype="I")
c = p.addVariable(vtype=b"I")
print([type(v.VariableType) for v in (a, b, c)])
# [<enum 'VType'>, <class 'str'>, <class 'bytes'>]

q = Problem.readMPS("<any MIP>.mps")
print(type(q.getVariables()[0].VariableType))   # <class 'str'>
```

**Expected behavior**
`VariableType` is always a `VType` member, normalized in `Variable.__init__` / `setVariableType` (decoding bytes, then `VType(value)`), so `IsMIP`, `solver.is_mip`, and display code do not each need to handle the variants.

**Environment details**
 - From source on `main` (`Problem.addVariable`, `Problem._from_data_model`, `io/parser_wrapper.pyx`); not environment-specific.

**Additional context**
Surfaced while adding `__repr__` in #1400, which decodes bytes before `VType(...)` until this is normalized.

mlubin:
Solved by #1783

PR #1783, opened 2026-08-24
update variable type

Iroy30:
## Description
Addresses #1736 

## Checklist

- Testing
   - [x] Added tests
   - [ ] NA
- Documentation
   - [ ] Added new documentation
   - [x] NA


python/cuopt/cuopt/linear_programming/problem.py
```diff
@@ -94,8 +115,10 @@ class Variable:
     ----------
     VariableName : str
         Name of the Variable.
-    VariableType : CONTINUOUS, INTEGER, or SEMI_CONTINUOUS
-        Variable type.
+    VariableType : VType
+        Variable type, always normalized to a :py:class:`VType` member
```
mlubin: "always normalized" is redundant with the ": VType" comment above.

python/cuopt/cuopt/linear_programming/problem.py
```diff
@@ -181,13 +212,15 @@ def getUpperBound(self):
     def setVariableType(self, val):
```
mlubin: Nit: it's a bit confusing to have two styles of setters and getters for the same field (setVariableType/getVariableType and the the new property setter/getter above.

python/cuopt/cuopt/linear_programming/problem.py

ramakrishnap-nv (approved):

tmckayus (approved):

python/cuopt/cuopt/linear_programming/problem.py
Iroy30: get/setVariableType is already existing public API. The property itself is what normalizes, because `VariableType` is also a documented public attribute that gets assigned directly and a plain attribute can't coerce those writes.

PR #1792, opened 2026-08-25
ci: pass the build datetime through to image builds

ramakrishnap-nv:
## Description

`build-images` fails with `Expected environment variable 'RAPIDS_DATETIME_STRING' to be set and non-empty` ([run 32811823369](https://github.com/NVIDIA/cuopt/actions/runs/32811823369/job/97697878626)).

`rapids-datetime-string` does not generate that value in CI, it only validates that something upstream exported it.

The value is already available: `build.yaml` passes `build-datetime` from `compute-build-details` to every other build job, and `build-images` was the only one not taking it. This wires it through and uses it directly, so the pin matches the wheels this run published.

Release builds are unaffected — `RAPIDS_VERSION_SUFFIX` is a no-op there.

## Checklist

- Testing
   - [x] NA
- Documentation
   - [x] NA


jameslamb (changes requested):
I don't recommend adding even more complexity to this workflow.

Is `rapids-datetime-string` needed at all here?

https://github.com/NVIDIA/cuopt/blob/b7891b59d7ebad9a76ad7c2679a6bd3b36a11b86/.github/workflows/build_test_publish_images.yaml#L84-L98

In other words... do you actually need to preserve a history of all container images for every branch/nightly build?

If not, then just rely on `rapids-generate-version` + overwriting tags.

e.g. if yesterday's was tagged `26.10.0a123` and a new image is built today and tagged `26.10.0a123`, it could just overwrite those old ones.

jameslamb:
Or if the only place this is invoked is from `build.yaml`, here:

https://github.com/NVIDIA/cuopt/blob/b7891b59d7ebad9a76ad7c2679a6bd3b36a11b86/.github/workflows/build.yaml#L354-L371

then just pass through the already-computed nightly version just like all the other workflows do:

https://github.com/NVIDIA/cuopt/blob/b7891b59d7ebad9a76ad7c2679a6bd3b36a11b86/.github/workflows/build.yaml#L62

ramakrishnap-nv:
@jameslamb you were right that the extra machinery was not needed, and it turned out to be simpler than either of us suggested.

I had claimed the datetime was not reachable from this job. That was wrong. `build.yaml` already runs `compute-build-details.yaml` and passes `build-datetime` to every other build job — `build-images` was the only one not taking it. I checked for a workflow-level `outputs:` block, did not find one, and stopped there, missing that it is a job output already wired into the same file six times over.

So the fix is now: add `build-details` to `build-images`'s `needs`, pass `build-datetime` through, and use it directly. No index lookup, no `curl`, no error handling. Net diff is +13/-5 across two files.

On your question about image history: nightly tags already overwrite. `IMAGE_TAG_PREFIX` is `26.10.0a` for every non-release build and `create_multiarch_manifest.sh` uses only that, so no per-build history exists today. The datetime is not used in any tag — `CUOPT_VER` feeds the `pip install` pin in the Dockerfile, and bare versions are never published, which is the failure #1780 was fixing.

Worth flagging separately: because merge and nightly builds share the `26.10.0a` tag, an image tag does not identify a commit. If that matters for debugging, putting the alpha in the tag would be a separate change.

jameslamb (approved):

PR #1800, opened 2026-08-25
Add conversion routines for LP/MIP cuOpt JSON dictionaries to/from DataModel and Solution

tmckayus:
These changes are part of a migration path from the http server to the gRPC server for users/applications that generate LP/MIP problems in cuOpt JSON dictionary format for submission to the http server and process the response dictionaries the http server returns.

With these converters, dictionary format problems can be converted to DataModel and Settings and passed to the gRPC client. Likewise, Solution objects returned from the client can be converted into dictionary response objects.


tmckayus:
closing this, we will add client-side converters in the future if necessary but for now we plan only to provide a proxy server to support legacy JSON formats on a remote service.

PR #1802, opened 2026-08-25
refactor: make to_optimization_problem a free function

ramakrishnap-nv:
`to_optimization_problem()` was a pure virtual on `optimization_problem_interface_t`, so it occupied a slot in **every** implementer's vtable — including `cpu_optimization_problem_t`, whose vtable then held an entry only `libcuopt` can define.

Vtable relocations resolve **eagerly at load time**, so this cannot be deferred or hidden behind lazy binding: any library carrying that vtable is unloadable without `libcuopt.so`. That blocks the CUDA-free client library (#1804).

It is now a free function declared in `optimization_problem.hpp`, defined in `cpu_optimization_problem_to_gpu.cpp`, dispatching on the concrete type:

```diff
- auto gpu = problem->to_optimization_problem(&handle);
+ auto gpu = to_optimization_problem(*problem, &handle);
```

Semantics are unchanged — the GPU override was a one-line `return nullptr`, so a GPU-backed problem still yields `nullptr`. Unrecognised implementations now throw instead of returning `nullptr`, since the documented fallback `static_cast`s the reference and would otherwise be UB. 5 call sites updated.

**Breaking:** removes a pure virtual from an installed public header. Out-of-tree implementers should delete their override; callers switch to the free function as above.

2 of 4 toward a CUDA-free client library (#1801 merged, #1803, #1804 follow).


ramakrishnap-nv:
@coderabbitai review

ramakrishnap-nv:
@coderabbitai Fixed. The point stands even though only two implementations exist today (`cpu_optimization_problem_t` and `optimization_problem_t`): as a pure virtual, the compiler forced every implementer to provide an override, and turning it into a free function silently dropped that guarantee. Combined with the documented fallback `static_cast<optimization_problem_t&>(problem)`, an unrecognised type would have been undefined behaviour rather than an error.

`to_optimization_problem()` now detects `optimization_problem_t` explicitly and throws for anything else, instead of returning `nullptr` for "not a CPU problem".

ramakrishnap-nv:
@coderabbitai full review

ramakrishnap-nv:
@coderabbitai You're right on both counts — fixed.

**Breaking-change label.** I had this as `non-breaking`, which was wrong. `to_optimization_problem` was a *pure virtual* in the installed public header `optimization_problem_interface.hpp`, so removing it breaks external code two ways: any out-of-tree implementer of the interface loses the override (their `override` keyword becomes a compile error), and any caller of `problem->to_optimization_problem(&handle)` no longer compiles. Relabelled `breaking`.

**Migration:**

```cpp
// before
auto gpu = problem->to_optimization_problem(&handle);

// after -- free function in <cuopt/mathematical_optimization/optimization_problem.hpp>
auto gpu = to_optimization_problem(*problem, &handle);
```

Semantics are unchanged: a `cpu_optimization_problem_t` is converted, a GPU-backed `optimization_problem_t` still yields `nullptr`. Out-of-tree subclasses should simply delete their override — the free function dispatches on the concrete type, and now raises a validation error for an unrecognised implementation rather than returning `nullptr` (which the documented `static_cast` fallback would have turned into undefined behaviour).

**Error convention.** Switched my new validation from `throw std::runtime_error` to `cuopt_expects(..., error_type_t::ValidationError, ...)` to match the codebase. Note the adjacent null-handle `throw std::runtime_error` is pre-existing and moved verbatim from `cpu_optimization_problem.cpp` — I left it alone deliberately so this PR stays a faithful move rather than quietly changing an existing error type.

ramakrishnap-nv:
@coderabbitai full review

cpp/include/cuopt/mathematical_optimization/cpu_optimization_problem.hpp
```diff
@@ -173,9 +173,11 @@ class cpu_optimization_problem_t : public optimization_problem_interface_t<i_t,
    * @param handle_ptr RAFT handle with CUDA resources for GPU memory allocation.
    * @return unique_ptr to new optimization_problem_t with all data copied to GPU
    * @throws std::runtime_error if handle_ptr is null
+   *
+   * Provided as the free function to_optimization_problem() in optimization_problem.hpp,
+   * not as a member: keeping it out of this class's vtable is what lets cuopt_client
+   * load without libcuopt.so.
```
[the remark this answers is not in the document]
ramakrishnap-nv: Already fixed in 63172d01 — this review ran against 97452948.

Agreed on the diagnosis, and the consequence was slightly worse than orphaned text: the block sat immediately before `write_to_mps()`'s own `/** */`, so doxygen was discarding the first of two consecutive blocks. The `@param`/`@return`/`@throws` documentation was already missing from generated docs rather than merely misplaced.

Removed entirely; the free function declaration in `optimization_problem.hpp` carries those docs.

ramakrishnap-nv:
@coderabbitai review

jameslamb (approved):
approving just the CMake changes (the look non-controversial) for `ci-codeowners`. I did not look at anything else in the PR.

hlinsen (approved):

cpp/src/pdlp/cpu_optimization_problem_to_gpu.cpp
```diff
@@ -0,0 +1,149 @@
...
+                  "implementation. Only optimization_problem_t and cpu_optimization_problem_t "
+                  "are supported.");
+    // Already a GPU-backed problem; nothing to convert.
+    return nullptr;
+  }
+  auto& self = *cpu_problem;
+
+  if (handle_ptr == nullptr) {
+    throw std::runtime_error(
+      "cpu_optimization_problem_t::to_optimization_problem(): "
```
tmckayus: nit, the exception message names the old function

tmckayus (approved):
one small nit in the exception message naming the old function

cpp/src/pdlp/cpu_optimization_problem_to_gpu.cpp
ramakrishnap-nv: Good catch — fixed in 144b3082. The message was moved verbatim with the function body, so the `cpu_optimization_problem_t::` prefix came along even though it is no longer a member. Now just `to_optimization_problem():`.

Also merged current `main` in (was 6 behind); clean, build + LP_UNIT_TEST pass.

Iroy30 (approved):

PR #1803, opened 2026-08-25
refactor: keep the GPU warm-start path out of host-only translation units

ramakrishnap-nv:
## Description

`populate_from_data_model_view()` handled the GPU and CPU warm-start directions in one inlined `if/else`. Only the GPU direction needs a device, but the compiler instantiated both into every translation unit that includes the header, pulling `convert_to_gpu_warmstart` and friends into code that never touches a GPU.

Split into three helpers, selected by a `kHostOnly` template parameter dispatched with `if constexpr` so a host-only caller never instantiates the GPU branch:

- `apply_warmstart_gpu_target()` — real handle; defined in libcuopt
- `apply_warmstart_cpu_target_with_device()` — null handle, caller has a device; defined in libcuopt
- `apply_warmstart_cpu_target()` — host-only caller; inline in the header

Two CPU-target variants because a `kHostOnly` caller cannot hold device-resident warm start, while a normal caller passing `handle == nullptr` can (`cython_solve.cu:181`) and needs the D2H copy.

Also moves the trivial warm-start accessors into `solver_settings_accessors.cpp` so host-only consumers resolve them without the CUDA translation unit.

Third of four steps toward a CUDA-free client library, after #1801 and #1802.

## Issue

Follow-up for test coverage: #1867

## Checklist

- Testing
   - [ ] Added tests
   - [x] Created an issue to follow-up
   - [ ] NA
- Documentation
   - [ ] Added new documentation
   - [x] NA


ramakrishnap-nv:
@coderabbitai review

ramakrishnap-nv:
@coderabbitai Good catch — this was a real regression and I've fixed it.

You're right that the device-to-CPU branch was reachable. `cython_solve.cu:181` calls `populate_from_data_model_view(&cpu_problem, data_model, solver_settings, nullptr)` — null handle with the default `kHostOnly=false` — so a caller holding device-resident warm start would have silently lost it on a remote solve. My original comment claiming the case was "rejected" was also wrong: it just fell through and returned.

Fixed as you suggested, with two helpers:

- `apply_warmstart_cpu_target()` stays host-only (already-host data, or a host-span view) and is now only reachable from the `kHostOnly` path.
- `apply_warmstart_cpu_target_with_device()` is declared in the header and defined out-of-line in `optimization_problem.cu`, restoring the `convert_to_cpu_warmstart` D2H copy. The non-host-only null-handle path routes here.

Semantics now match `main` exactly: host data → no-op, view → direct copy, device data → D2H.

ramakrishnap-nv:
## Reworked — the `shared_ptr` mechanism is gone

A code review pass raised several problems with the `pdlp_warm_start_data_t` pimpl, and on checking they were all real. Rather than patch them I've removed that half of the PR entirely; the title and scope are updated to match.

**Why it was wrong.** `shared_ptr` was doing two jobs and I only wanted one. It gave type-erased storage (good), but it also silently turned settings copies into shared aliases. `pdlp_warm_start_data_t` has an explicit deep-copy ctor that `raft::copy`s each buffer, and settings are copied at `solve.cu:1158`, `:1358` and `:1532` — that last one is `run_concurrent`, which runs PDLP/Barrier/DualSimplex **in parallel**. So `set_pdlp_warm_start_data()` (which `resize()`s and `thrust::fill`s) would have mutated every sibling copy. On top of that the lazy init was unsynchronised, allocated device memory as a side effect of read-only `is_populated()` checks, and sat behind a `noexcept` getter that could throw `rmm::bad_alloc` — turning an allocation failure into `std::terminate`.

**Why not just fix it.** Every mechanism that preserves deep-copy semantics needs either a hand-enumerated scalar copy constructor (which silently drops any field added later — a worse trap than the original bug) or the interface restructuring that the header-separation work is going to do anyway.

**What is left**, and it stands on its own: the `kHostOnly` / `if constexpr` split so host-only callers never instantiate the GPU warm-start branch, now with two CPU-target variants — a host-only one, and `apply_warmstart_cpu_target_with_device()` for the null-handle case where the caller does have a device (`cython_solve.cu:181`) and its warm start may need a D2H copy. Plus the trivial accessors moving to `solver_settings_accessors.cpp`.

**Cost:** `libcuopt_client.so` keeps one undefined symbol, `pdlp_warm_start_data_t()`. That is fine for now — the library already has 14 undefined `cuopt::routing::` symbols, so it is not standalone-loadable either way, and both belong to the same follow-up.

ramakrishnap-nv:
@coderabbitai full review

ramakrishnap-nv:
@coderabbitai full review

ramakrishnap-nv:
@coderabbitai three findings, statuses below.

**1. Stream sync at `optimization_problem.cu:1572`** — skipping: not this PR's code. It is pre-existing on `main` (identical there), last touched by #1858, and this PR's diff in that file starts at line 1637. CodeRabbit flagged it as "outside diff range" itself. It may well be a real latent issue — `gpu_cast` runs on `stream` while `other.set_*` copies on `other.stream_view_`, and the `stream_view_.sync()` afterwards syncs the wrong stream if they differ — but it belongs in its own PR against the code's owner rather than riding along here.

**2. D2H conversion for the null-handle path** — already fixed, in an earlier revision than the one reviewed. `apply_warmstart_cpu_target_with_device()` exists in both the header and `optimization_problem.cu`, and the `convert_to_cpu_warmstart` call is restored at `optimization_problem.cu:1698`. This was a real regression when first raised, and the fix follows exactly the shape suggested: keep `apply_warmstart_cpu_target()` host-only for `kHostOnly`, and route the non-host-only null-handle case to a CUDA-capable variant.

**3. Regression tests for warm-start conversion** — fair, and not done. The three paths (host-view → CPU, host-data → GPU, device-data → CPU) need a real GPU fixture and a populated warm start to be meaningful, which is more than a unit test around the predicate. Flagging for reviewers rather than silently skipping: worth doing, better as follow-up than expanding this PR.

cpp/src/pdlp/optimization_problem.cu
```diff
@@ -1637,4 +1637,43 @@ template CUOPT_EXPORT optimization_problem_t<int32_t, float>
...
+    // Host warmstart -> GPU backend: convert H2D.
+    pdlp_warm_start_data_t<i_t, f_t> warm_start =
+      convert_to_gpu_warmstart(pdlp.get_cpu_pdlp_warm_start_data(), handle->get_stream());
+    pdlp.set_pdlp_warm_start_data(warm_start);
+  }
+}
+
+#if MIP_INSTANTIATE_FLOAT
+template void apply_warmstart_gpu_target(solver_settings_t<int, float>*, const raft::handle_t*);
+#endif
```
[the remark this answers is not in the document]
ramakrishnap-nv: Fixed in b8097840. The float guard read `MIP_INSTANTIATE_FLOAT` alone while the surrounding instantiations in this file use `MIP_ || PDLP_`; a PDLP-only build with float enabled would have compiled and then failed at link. Now matches.

cpp/src/pdlp/solver_settings.cu
```diff
@@ -382,37 +385,28 @@ std::optional<i_t> pdlp_solver_settings_t<i_t, f_t>::get_initial_pdlp_iteration(
...
+// These two live here rather than in solver_settings_accessors.cpp: they may have to
+// allocate the device-backed warm-start object, so they need CUDA.
 template <typename i_t, typename f_t>
-cpu_pdlp_warm_start_data_t<i_t, f_t>&
-pdlp_solver_settings_t<i_t, f_t>::get_cpu_pdlp_warm_start_data() noexcept
+const pdlp_warm_start_data_t<i_t, f_t>& pdlp_solver_settings_t<i_t, f_t>::get_pdlp_warm_start_data()
+  const noexcept
 {
-  return cpu_pdlp_warm_start_data_;
+  return ensure_pdlp_warm_start_data();
```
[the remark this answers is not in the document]
ramakrishnap-nv: No longer applies. The `noexcept` concern was against the revision where `get_pdlp_warm_start_data()` routed through a lazily-allocating `ensure_...()`. That pimpl was reverted — it broke the deep-copy semantics settings rely on — so the getter is back to returning the member directly and allocates nothing.

cpp/include/cuopt/mathematical_optimization/optimization_problem_utils.hpp
```diff
@@ -137,6 +137,39 @@ void populate_from_mps_data_model(optimization_problem_interface_t<i_t, f_t>* pr
...
+  auto& pdlp = solver_settings->get_pdlp_settings();
+
+  if (pdlp.get_cpu_pdlp_warm_start_data().is_populated()) { return; }
+
+  // Warmstart view (host spans from Cython) -> CPU backend: copy directly, no CUDA needed.
+  if (solver_settings->get_pdlp_warm_start_data_view()
+        .last_restart_duality_gap_dual_solution_.size() > 0) {
+    pdlp.get_cpu_pdlp_warm_start_data() =
+      cpu_pdlp_warm_start_data_t<i_t, f_t>(solver_settings->get_pdlp_warm_start_data_view());
+  }
```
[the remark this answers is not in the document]
ramakrishnap-nv: Already addressed, in an earlier revision than the one reviewed. `apply_warmstart_cpu_target_with_device()` is defined in `optimization_problem.cu` with the `convert_to_cpu_warmstart` call restored at line 1698, and `apply_warmstart_cpu_target()` stays host-only for the `kHostOnly` path — exactly the split suggested. It was a real regression when raised: `cython_solve.cu:181` passes `handle == nullptr` from a caller that does have a device, so its device-resident warm start would have been silently dropped.

cpp/src/pdlp/optimization_problem.cu
```diff
@@ -1637,4 +1637,77 @@ template CUOPT_EXPORT optimization_problem_t<int32_t, float>
...
+    return;
+  }
+
+  // Device-resident warm start -> CPU backend: convert D2H.
+  auto& gpu_ws = pdlp.get_pdlp_warm_start_data();
+  if (gpu_ws.is_populated()) {
+    pdlp.get_cpu_pdlp_warm_start_data() =
+      convert_to_cpu_warmstart(gpu_ws, gpu_ws.current_primal_solution_.stream());
+  }
+}
```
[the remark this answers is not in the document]
ramakrishnap-nv: Not done, and worth a reviewer's call. The three paths need a GPU fixture with a populated warm start rather than a predicate-level unit test, so I have listed it as a known gap in the PR description instead of expanding scope here.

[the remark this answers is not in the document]
ramakrishnap-nv: Tracked as #1867 rather than left implicit — it needs a GPU fixture with a populated warm start, so it is follow-up work rather than scope here.

cpp/src/pdlp/optimization_problem.cu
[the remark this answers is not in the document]
ramakrishnap-nv: Reverting this. The wider guard was wrong and it broke `wheel-build-libcuopt` on all four configs.

`PDLP_INSTANTIATE_FLOAT` is hardcoded to `1`, so `MIP_INSTANTIATE_FLOAT || PDLP_INSTANTIATE_FLOAT` instantiated both helpers for float unconditionally. Both call `solver_settings_t<i_t, f_t>::get_pdlp_settings()` and `::get_pdlp_warm_start_data_view()`, which `math_optimization/solver_settings.cpp` instantiates only under `MIP_INSTANTIATE_FLOAT`. Since `CUOPT_INSTANTIATE_FLOAT` is `0`, those accessors do not exist for float, so `libcuopt.so` ended up with undefined references and `cuopt_cli` / `cuopt_grpc_server` failed to link:

```
ld: libcuopt.so: undefined reference to `solver_settings_t<int, float>::get_pdlp_warm_start_data_view() const'
ld: libcuopt.so: undefined reference to `solver_settings_t<int, float>::get_pdlp_settings()'
```

Narrowed back to `MIP_INSTANTIATE_FLOAT` in 4bda7e84 so the float instantiation set stays internally consistent with the accessors it depends on. Nothing needs the float helpers: all three call sites of `populate_from_data_model_view` (`cython_solve.cu` x2, `cython_grpc_client.cpp`) are `<int, double>`.

A genuine PDLP-float build would need the float accessors in `solver_settings.cpp` widened too, which is a separate change and not one this PR needs.

cpp/include/cuopt/mathematical_optimization/optimization_problem_utils.hpp
```diff
@@ -137,6 +137,54 @@ void populate_from_mps_data_model(optimization_problem_interface_t<i_t, f_t>* pr
...
 
+/**
+ * @brief Move warm-start data into the form a GPU solve needs (H2D / view->device_uvector).
+ *
+ * Declared here, defined in libcuopt (optimization_problem.cu): it touches device memory,
+ * so keeping it out-of-line is what lets CUDA-free consumers of this header link without
+ * a CUDA runtime. Only call it with a real handle.
+ */
+template <typename i_t, typename f_t>
+void apply_warmstart_gpu_target(solver_settings_t<i_t, f_t>* solver_settings,
```
rg20: Naming is not intuitive. 

You can use either `sync_warmstart_data_to_gpu` or `copy_warmstart_data_to_device`

cpp/include/cuopt/mathematical_optimization/optimization_problem_utils.hpp
```diff
@@ -137,6 +137,54 @@ void populate_from_mps_data_model(optimization_problem_interface_t<i_t, f_t>* pr
...
+/**
+ * @brief Move warm-start data into the form a CPU / remote solve needs, including a
+ * device-to-host copy when the warm start is device-resident.
+ *
+ * Declared here, defined in libcuopt (optimization_problem.cu). This is the null-handle
+ * path for a normal (non host-only) caller: it has a device, so its settings may hold
+ * device_uvector-backed warm start that must be brought to host before a remote solve.
+ */
+template <typename i_t, typename f_t>
+void apply_warmstart_cpu_target_with_device(solver_settings_t<i_t, f_t>* solver_settings);
```
rg20: same here

cpp/include/cuopt/mathematical_optimization/optimization_problem_utils.hpp
```diff
@@ -137,6 +137,54 @@ void populate_from_mps_data_model(optimization_problem_interface_t<i_t, f_t>* pr
...
+ * Handles the two cases reachable without a device: warm start already in host form
+ * (nothing to do), and a warm-start view over host spans (copy it).
+ *
+ * Deliberately does NOT handle device-resident warm start -- that needs a D2H copy and
+ * therefore CUDA. Callers that might be holding device data must use
+ * apply_warmstart_cpu_target_with_device() instead; only a kHostOnly caller, which by
+ * construction has no device to have populated it, may use this one.
+ */
+template <typename i_t, typename f_t>
+void apply_warmstart_cpu_target(solver_settings_t<i_t, f_t>* solver_settings)
```
rg20: same here

cpp/include/cuopt/mathematical_optimization/optimization_problem_utils.hpp
rg20:
```suggestion
void apply_warmstart_gpu_target(solver_settings_t<i_t, f_t>& solver_settings,
```

cpp/include/cuopt/mathematical_optimization/optimization_problem_utils.hpp
rg20:
```suggestion
void apply_warmstart_cpu_target_with_device(solver_settings_t<i_t, f_t>& solver_settings);
```

cpp/include/cuopt/mathematical_optimization/optimization_problem_utils.hpp
ramakrishnap-nv: Renamed in 33e33389, using your direction-based form across all three:

- `apply_warmstart_gpu_target` → `copy_warmstart_data_to_device`
- `apply_warmstart_cpu_target_with_device` → `copy_warmstart_data_to_host`
- `apply_warmstart_cpu_target` → `copy_warmstart_view_to_host`

I went with `copy_` over `sync_` since each is a one-way transfer, not a reconcile.

cpp/include/cuopt/mathematical_optimization/optimization_problem_utils.hpp
ramakrishnap-nv: Done — `copy_warmstart_data_to_host` in 33e33389.

This is the device-capable variant: null handle, but the caller still has a device, so it also covers the D2H case that `copy_warmstart_view_to_host` deliberately cannot.

cpp/include/cuopt/mathematical_optimization/optimization_problem_utils.hpp
ramakrishnap-nv: Done — this one is `copy_warmstart_view_to_host` in 33e33389.

Named for its source rather than something like `_no_device`: a `kHostOnly` caller has no device, so a host-span view is the only warm start it can ever be holding. That is also exactly why it stays inline in the header and needs no CUDA.

cpp/include/cuopt/mathematical_optimization/optimization_problem_utils.hpp
ramakrishnap-nv: Applied in 33e33389, along with the rename. Left `handle` as a pointer since the caller passes it through as a nullable value.

cpp/include/cuopt/mathematical_optimization/optimization_problem_utils.hpp
ramakrishnap-nv: Applied in 33e33389. The sole caller null-checks `solver_settings` before dispatching, so the reference is safe here.

cpp/include/cuopt/mathematical_optimization/optimization_problem_utils.hpp
ramakrishnap-nv: Following up on the one part of your suggestion I did not take: `handle` is still a pointer.

```cpp
void copy_warmstart_data_to_device(solver_settings_t<i_t, f_t>& solver_settings,
                                   const raft::handle_t* handle);
```

The only call site already guards it, so it is non-null by construction here:

```cpp
if (handle != nullptr) {
  copy_warmstart_data_to_device(*solver_settings, handle);
} else {
  copy_warmstart_data_to_host(*solver_settings);
}
```

So `const raft::handle_t&` would encode that invariant the same way the settings reference now does. I left it as a pointer only to keep the diff to what you flagged — happy to convert it too if you would rather have both non-null params expressed as references. Your call.

rg20 (approved):

tmckayus (approved):

PR #1804, opened 2026-08-25
build: add the CUDA-free cuopt_client library

ramakrishnap-nv:
## Description

Adds `cuopt_client`, a CPU-only library holding the host-side problem representation (parsers, `data_model_view`, `mps_data_model`, writers), the gRPC wire protocol, and the LP/MIP gRPC client. `libcuopt` and `cuopt_grpc_server` both link it, so there is one implementation rather than a client-side fork.

This is the library-level half of letting a remote client talk to `cuopt_grpc_server` without `cudf`, `cupy`, `rmm` or `pylibraft`. The packaging half is not here: `libcuopt_client.so` still ships inside the `libcuopt` package, which depends on CUDA, so a GPU-free install is not yet possible. Tracked in #1872.

Two notes for reviewers:

- The routing gRPC arm stays in `libcuopt`. Its mappers call routing accessors that live in CUDA translation units, so moving it down would create a `libcuopt -> cuopt_client -> libcuopt` cycle.
- Some public API changes which library exports it — `solver_settings_t::get_mip_callbacks()` and siblings now come from `libcuopt_client.so`. `cuopt` links `cuopt::cuopt_client` as `PUBLIC` and both are in `cuopt-exports`, so CMake consumers resolve them transitively; a bare `-lcuopt` link would also need `-lcuopt_client`.

Verified: `libcuopt_client.so` has no CUDA, rmm or raft in `NEEDED`, and no undefined `cuopt::` symbols.

Last of four steps toward a CUDA-free client library, after #1801, #1802 and #1803.

## Checklist

- Testing
   - [ ] Added tests
   - [ ] NA
- Documentation
   - [ ] Added new documentation
   - [x] NA


ramakrishnap-nv:
@coderabbitai review

ramakrishnap-nv:
## Update: two runtime bugs found by running the suite, both fixed

Building alone was not enough — running the tests surfaced two problems that `ldd -r` reports as clean, because it resolves against the full load set while these only bite at call time.

**1. Routing gRPC arm created a library cycle.** I had put it in `cuopt_client` on the grounds that its mappers contain no raft/rmm/thrust. True, but CUDA-free is not the same as dependency-free: they call `routing::solver_settings_t` and `routing::assignment_t` accessors that live in CUDA TUs, so `libcuopt -> cuopt_client -> libcuopt`. It showed up as `HEURISTICS_HYPER_PARAMS_TEST` dying with `undefined symbol: routing::solver_settings_t::get_time_limit`. The routing arm now stays in libcuopt until its host-only accessors are split out, exactly as was done for LP/MIP.

**2. `template class` pulled the constructor into the CUDA-free TU.** `solver_settings.cpp` used `template class solver_settings_t<...>`, which instantiates *every* member including the ctor — and the ctor builds a `pdlp_solver_settings_t` holding a `pdlp_warm_start_data_t` by value, whose ctor is CUDA-side. Members are now instantiated individually, and the ctor itself moved to `solver_settings_gpu.cu`.

This also corrects something I stated earlier in this PR: I described the leftover undefined symbols as a deferrable gap. They were not — **any** undefined symbol in `libcuopt_client.so` breaks at runtime for anything that loads it, which is every test binary.

```
undefined cuopt symbols : 0     (was 15)
NEEDED with cuda/rmm/raft : 0
```

**Test results:** 119/125 pass. The 6 failures (`WAYPOINT_MATRIXTEST`, `INCUMBENT_CALLBACK_TEST`, `DOC_EXAMPLE_TEST`, `C_API_TEST`, `NUMOPT_INTERNAL_TEST`, `CLI_TEST`) are **not regressions** — I built unmodified `main` (`f63e4df9`) in a separate worktree and it fails the identical six. They are missing downloaded datasets (`square41.mps` and friends, fetched by `ci/test_cpp.sh` via the `download_*_test_dataset.sh` scripts, which I did not run locally).

ramakrishnap-nv:
@coderabbitai full review

cpp/src/math_optimization/solver_settings.cpp
```diff
@@ -594,8 +441,43 @@ bool solver_settings_t<i_t, f_t>::dump_parameters_to_file(const std::string& pat
...
+template CUOPT_EXPORT const std::vector<std::string>
+solver_settings_t<int, float>::get_parameter_names() const;
+template CUOPT_EXPORT const std::vector<parameter_info_t<std::string>>&
+solver_settings_t<int, float>::get_string_parameters() const;
+template CUOPT_EXPORT const pdlp_warm_start_data_view_t<int, float>&
+solver_settings_t<int, float>::get_pdlp_warm_start_data_view() const noexcept;
+template CUOPT_EXPORT void solver_settings_t<int, float>::load_parameters_from_file(
+  const std::string&);
+template CUOPT_EXPORT bool solver_settings_t<int, float>::dump_parameters_to_file(
+  const std::string&, bool) const;
```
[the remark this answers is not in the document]
ramakrishnap-nv: Already addressed — this was reviewed against an earlier revision. Both instantiations are present in `solver_settings.cpp` on the current head (58b7093e):

```
462:  solver_settings_t<int, float>::get_mip_callbacks() const;    # inside #if MIP_INSTANTIATE_FLOAT
507:  solver_settings_t<int, double>::get_mip_callbacks() const;   # inside #if MIP_INSTANTIATE_DOUBLE
```

The underlying concern was real, though — an earlier revision of this branch dropped both and `libcuopt.so` silently stopped exporting a method that `main` exports. Nothing in-tree calls it, so no build caught it.

Verified on the current build:

```
$ nm -D --defined-only libcuopt_client.so | c++filt | grep -c get_mip_callbacks
2
```

Worth noting where it now resolves from: the symbol is exported by `libcuopt_client.so` rather than `libcuopt.so`, since `solver_settings.cpp` is part of the client library. That is intended — `cuopt` links `cuopt::cuopt_client` as `PUBLIC` and both are installed into `cuopt-exports`, so `target_link_libraries(app cuopt::cuopt)` still resolves it transitively.

conda/recipes/libcuopt/recipe.yaml
```diff
@@ -114,6 +114,7 @@ outputs:
         ignore:
           # See https://github.com/rapidsai/build-planning/issues/160
           - lib/libcuopt.so
+          - lib/libcuopt_client.so
```
ramakrishnap-nv, line 117: Applied in ae24d754. Verified rather than taking it as given — `libcuopt_client.so` does embed the build prefix, same as `libcuopt.so`:

```
strings -a cpp/build/libcuopt.so        | grep -c "$PREFIX"   # 34
strings -a cpp/build/libcuopt_client.so | grep -c "$PREFIX"   # 15
```

Same `BUILD_RPATH`/`INSTALL_RPATH "$ORIGIN"` and same install rule, so it needed the same ignore entry.

tmckayus (approved):

cpp/src/math_optimization/solver_settings_gpu.cu
```diff
@@ -105,76 +107,178 @@ void solver_settings_t<i_t, f_t>::add_initial_mip_solution(const f_t* solution,
   mip_settings.add_initial_solution(solution, size, stream);
 }
 
+// The constructor lives here, not in solver_settings.cpp, even though its body is pure
```
rg20: Why is this `solver_settings_gpu.cu` ?  none of this code is related to gpu

ramakrishnap-nv: Fair question — the body really is pure host code. The constraint is not the body, it is what the constructor *calls*.

`solver_settings_t::solver_settings_t()` default-constructs `pdlp_settings`, and `pdlp_solver_settings_t` holds the warm start by value:

```cpp
// pdlp/solver_settings.hpp:389
pdlp_warm_start_data_t<i_t, f_t> pdlp_warm_start_data_;
```

That type's default constructor is declared in the header but defined in `cpp/src/pdlp/pdlp_warm_start_data.cu`, and it constructs nine device buffers:

```cpp
// pdlp_warm_start_data.cu:68
pdlp_warm_start_data_t<i_t, f_t>::pdlp_warm_start_data_t()
  : current_primal_solution_{rmm::device_uvector<f_t>(
      0, cuda::stream_ref{cudaStream_t{cudaStreamDefault}})},
    ... 8 more
```

So the chain is `solver_settings_t()` → `pdlp_settings` → `pdlp_warm_start_data_t()` → `rmm::device_uvector` + `cudaStreamDefault`. Compiling the constructor into `cuopt_client` would leave `libcuopt_client.so` with an undefined reference to that ctor, which only surfaces at call time. Keeping it here is why the client library currently resolves standalone:

```
$ nm -D --undefined-only libcuopt_client.so | c++filt | grep -c pdlp_warm_start_data_t
0
```

Making it genuinely host-side would mean giving `pdlp_warm_start_data_t` a default constructor that does not touch device memory. That is worth doing, but it is a change to the warm-start type rather than to this file, and it would pull `libcudart` into the client library if done naively.

On the naming: you are right that `_gpu` oversells it. The file is really "the members that require CUDA to compile", which is mostly device code plus this one transitive case. Happy to rename it — `solver_settings_cuda.cu` would be more honest — or to move just the constructor into its own TU named for the reason. Let me know which you prefer and I will fold it in.

ramakrishnap-nv: Renamed to `solver_settings.cu` in 464ba8b5 — agreed, `.cu` already carries the meaning and `_gpu` oversold the contents.

Two things it also gets us: it restores the file name from before the host/device split, so `git blame` stays continuous, and it matches the pairing `mip_heuristics` already uses:

```
cpp/src/math_optimization/solver_settings.cpp   cpp/src/mip_heuristics/solver_settings.cpp
cpp/src/math_optimization/solver_settings.cu    cpp/src/mip_heuristics/solver_settings.cu
```

No object-name collision — the two go to different targets (`cuopt_client_objs` and `cuopt_objs`).

I also rewrote the comment above the constructor. It sat at exactly the line you commented on, so it clearly was not doing its job. It now names the member that forces the placement rather than gesturing at "CUDA-side", and records the fix that would let the constructor move to the host TU:

```cpp
// The constructor is here, not in solver_settings.cpp, and its body is a red herring: it
// only builds parameter tables. What forces the placement is the member it default-
// constructs. pdlp_solver_settings_t holds a pdlp_warm_start_data_t by value, and that
// type's default ctor -- defined in pdlp/pdlp_warm_start_data.cu -- constructs nine
// rmm::device_uvectors on cudaStreamDefault. ...
//
// Giving pdlp_warm_start_data_t a default ctor that does not touch device memory would let
// this move to the host translation unit.
```

Rebuilt clean, and `libcuopt_client.so` still has no CUDA in `NEEDED` and no undefined `cuopt::` symbols.

rg20 (approved):

PR #1807, opened 2026-08-26
Add parameters for tuning barrier primal and dual regularization

yuwenchen95:
Adds two new barrier-method solver parameters, `barrier_primal_regularization` and `barrier_dual_regularization`, that let the caller set the initial regularization values used for the barrier method's augmented KKT system on the first factorization:

- `-1` (default): automatic, preserving today's existing heuristics (`has_soc ? 1e-8 : 1e-6` for primal, `adaptive_reg ? 1e-8 : 0` for dual).
- Any non-negative value: used verbatim as the starting regularization instead of the heuristic.

If `--barrier-adaptive-regularization` is enabled, these values are just the *starting point* — the adaptive policy still scales them up/down on later iterations based on iterative-refinement quality.


cpp/include/cuopt/mathematical_optimization/constants.h
```diff
@@ -51,6 +51,8 @@
 #define CUOPT_BARRIER_PRESOLVE_BOUND_FREE_VARIABLES "barrier_presolve_bound_free_variables"
 #define CUOPT_BARRIER_ITERATIVE_REFINEMENT          "barrier_iterative_refinement"
 #define CUOPT_BARRIER_ADAPTIVE_REGULARIZATION       "barrier_adaptive_regularization"
+#define CUOPT_BARRIER_PRIMAL_PERTURB                "barrier_primal_perturb"
```
chris-maes: I think `CUOPT_BARRIER_PRIMAL_REGULARIZATION` would be better a better name.

cpp/include/cuopt/mathematical_optimization/constants.h
```diff
@@ -51,6 +51,8 @@
...
+#define CUOPT_BARRIER_DUAL_PERTURB                  "barrier_dual_perturb"
```
chris-maes: I think `CUOPT_BARRIER_DUAL_REGULARIZATION` would be a better name. 

The alternative is `CUOPT_BARRIER_DUAL_PERTURBATION`. But I think regularization is better than perturbation here.

chris-maes (approved):
LGTM. Thanks

PR #1808, opened 2026-08-26
Reduce GPU/host synchronization overhead in barrier termination check

yuwenchen95:
## Description

Reduces GPU/host synchronization overhead in the barrier LP/QP/SOCP solver's per-iteration termination-check computation.

Previously, computing residual norms, the barrier parameter `mu`, and the primal/dual objectives at the start of `barrier_solver_t::solve` and at the end of every barrier iteration was split across three separate functions (`compute_residual_norms`, `compute_mu`, `compute_primal_dual_objective`), each of which read its GPU reduction/dot-product results back to the host individually via a blocking `rmm::device_scalar::value(stream)` — i.e. a `cudaMemcpyAsync` + full stream synchronize per value, several times per iteration.

This PR fuses all three into a single `compute_residual_norms_mu_and_objective`: every reduction/dot-product kernel now writes into its own slot of a shared device buffer (`d_reduction_results_`), and the host readback is a single batched copy + one stream sync at the end (`h_reduction_results_`, with `d_reduce_temp_storage_` as shared cub scratch space). New reusable primitives (`enqueue_norm_inf_into`, `enqueue_sum_into`, `enqueue_max_into`) in `vector_math.cuh` support writing reductions into a caller-supplied slot/scratch buffer instead of allocating and blocking per call. The old `gpu_compute_residual_norms`, `compute_residual_norms`, `compute_mu`, and `compute_primal_dual_objective` (including their unused `CHECK_OBJECTIVE_GAP` debug path) are removed as dead code now that both call sites use the fused function.

No behavior change to the solver's numerics — same reductions, same values, just fewer synchronization points. Benchmarked with no regression across LP, QP, QCQP, and SOCP problem classes.


cpp/src/linear_algebra/vector_math.cuh
```diff
@@ -68,6 +69,63 @@ f_t device_custom_vector_norm_inf(InputIteratorT in, i_t size, rmm::cuda_stream_
   return d_out.value(stream_view);
 }
 
+// Same reduction as device_custom_vector_norm_inf, but writes into a caller-supplied device
+// pointer (and reuses a caller-supplied temp-storage buffer) instead of allocating a private
+// rmm::device_scalar and blocking on .value(). Lets callers batch several reductions and defer
+// the host readback to a single copy + sync.
+template <typename i_t, typename f_t, typename InputIteratorT>
+void enqueue_norm_inf_into(
```
rg20:
```suggestion
void enqueue_norm_inf_into_async(
```

cpp/src/linear_algebra/vector_math.cuh
```diff
@@ -68,6 +69,63 @@ f_t device_custom_vector_norm_inf(InputIteratorT in, i_t size, rmm::cuda_stream_
...
+  tmp.resize(temp_storage_bytes, stream_view);
+
+  cub::DeviceReduce::Sum(tmp.data(), temp_storage_bytes, in, out, size, stream_view);
+}
+
+// Max reduction (with a floor of 0, matching this codebase's existing
+// thrust::reduce(..., f_t(0), thrust::maximum<f_t>()) usage) into a caller-supplied device
+// pointer/temp-storage buffer, deferring the host readback (see enqueue_norm_inf_into).
+template <typename i_t, typename f_t, typename InputIteratorT>
+void enqueue_max_into(
```
rg20: Instead of implementing for each operation, why not just templatize the op or sending op as argument? 

Also enqueue does not make sense here, you are already specifying the pointer to which the result should be written right?  

May be lets just call it something like: 
`void reduce_to_async(`

cpp/src/barrier/barrier.cu
```diff
@@ -2079,6 +2076,14 @@ class iteration_data_t {
   transform_reduce_pair_helper_t<f_t> transform_reduce_pair_helper_;
   sum_reduce_helper_t<f_t> sum_reduce_helper_;
 
+  // Staging area for compute_residual_norms_mu_and_objective: several independent GPU
+  // reductions/dot-products write into slots of d_reduction_results_, then a single copy into
+  // h_reduction_results_ + one stream sync reads them all back at once instead of one sync each.
+  static constexpr i_t kNumScalarBatchSlots = 12;
```
rg20: Can you move all this logic/data to a struct called `barrier_reduce_helper_t`  (or something similar)

cpp/src/barrier/barrier.cu
```diff
@@ -3828,182 +3793,149 @@ void barrier_solver_t<i_t, f_t>::compute_next_iterate(iteration_data_t<i_t, f_t>
...
-  const f_t mu_denom = data.complementarity_degree(data.x.size(), data.n_upper_bounds);
-  mu                 = (data.sum_reduce_helper_.sum(data.d_complementarity_xz_residual_.begin(),
-                                    data.d_complementarity_xz_residual_.size(),
-                                    stream_view_) +
-        data.sum_reduce_helper_.sum(data.d_complementarity_wv_residual_.begin(),
-                                    data.d_complementarity_wv_residual_.size(),
-                                    stream_view_)) /
-       mu_denom;
-}
+  constexpr i_t kSlotPrimalResidual = 0;
```
rg20: Let the struct handle this logic

cpp/src/barrier/barrier.cu
```diff
@@ -3828,182 +3793,149 @@ void barrier_solver_t<i_t, f_t>::compute_next_iterate(iteration_data_t<i_t, f_t>
...
+  // single-stream FIFO ordering, so that the single sync at the bottom is enough for every
+  // result to be ready on the host.
+  enqueue_norm_inf_into<i_t, f_t>(data.d_primal_residual_.data(),
+                                  data.d_primal_residual_.size(),
+                                  d_batch + kSlotPrimalResidual,
+                                  data.d_reduce_temp_storage_,
+                                  stream_view_);
+  enqueue_norm_inf_into<i_t, f_t>(data.d_bound_residual_.data(),
+                                  data.d_bound_residual_.size(),
+                                  d_batch + kSlotBoundResidual,
```
rg20:
```suggestion
                                  d_reduction_helper.primal_residual(),
```

cpp/src/barrier/barrier.cu
```diff
@@ -3828,182 +3793,149 @@ void barrier_solver_t<i_t, f_t>::compute_next_iterate(iteration_data_t<i_t, f_t>
...
     RAFT_CUBLAS_TRY(raft::linalg::detail::cublasdot(lp.handle_ptr->get_cublas_handle(),
                                                     data.d_Qx_.size(),
                                                     data.d_Qx_.data(),
                                                     1,
                                                     data.d_x_.data(),
                                                     1,
-                                                    d_xQx.data(),
+                                                    d_batch + kSlotXQx,
                                                     stream_view_));
-    quad_objective = 0.5 * d_xQx.value(stream_view_);
```
rg20, line 3915: Is this intentional?

cpp/src/barrier/barrier.cu
chris-maes: Agree. Putting this in a struct would be cleaner.

cpp/src/barrier/barrier.cu
chris-maes: +1

cpp/src/barrier/barrier.cu
chris-maes: I agree, you don't want to have to know about the right slot. Using a name function here would be cleaner.

cpp/src/barrier/barrier.cu
```diff
@@ -3828,182 +3793,149 @@ void barrier_solver_t<i_t, f_t>::compute_next_iterate(iteration_data_t<i_t, f_t>
...
+
+  const bool has_soc       = data.has_cones();
+  const i_t linear_xz_size = data.linear_xz_size(data.d_complementarity_xz_residual_.size());
+  auto linear_xz_span =
+    raft::device_span<const f_t>(data.d_complementarity_xz_residual_.data(), linear_xz_size);
+
+  // All enqueue calls below must stay on stream_view_: correctness relies on strict
+  // single-stream FIFO ordering, so that the single sync at the bottom is enough for every
+  // result to be ready on the host.
+  enqueue_norm_inf_into<i_t, f_t>(data.d_primal_residual_.data(),
```
chris-maes: As discussed below it would be better if this was named something like `reduce`

cpp/src/barrier/barrier.cu
```diff
@@ -3828,182 +3793,149 @@ void barrier_solver_t<i_t, f_t>::compute_next_iterate(iteration_data_t<i_t, f_t>
...
+                             stream_view_);
 
   RAFT_CUBLAS_TRY(raft::linalg::detail::cublasdot(lp.handle_ptr->get_cublas_handle(),
                                                   data.d_c_.size(),
                                                   data.d_c_.data(),
                                                   1,
                                                   data.d_x_.data(),
                                                   1,
-                                                  d_cx.data(),
+                                                  d_batch + kSlotCx,
```
chris-maes: As discussed above, it would be better to have a way to avoid the `kSlot`

cpp/src/barrier/barrier.cu
chris-maes: I think this is just computed later.

chris-maes, line 3915: Please make sure this is still done later.

cpp/src/barrier/barrier.cu
```diff
@@ -3828,182 +3793,149 @@ void barrier_solver_t<i_t, f_t>::compute_next_iterate(iteration_data_t<i_t, f_t>
...
-    std::abs(objective_gap_1 - (xz + wv)),
-    rdx,
-    rpy,
-    rwv,
-    p,
-    y);
-#endif
+  const f_t* h = data.h_reduction_results_.data();
+
+  primal_residual_norm          = std::max(h[kSlotPrimalResidual], h[kSlotBoundResidual]);
```
chris-maes: Here and below, it's a bit ugly to use this `h[Slot...]` to access these values. If you placed this in a struct say `barrier_reduce_helper_t` you could do something like

```
barrier_reduce_helper_t rh;
// Perform reduction 
primal_residual_norm = std::max(rh.primal_residual_norm, rh.bound_residual_norm);
dual_residual_norm = rh.dual_residual_norm;
```

cpp/src/linear_algebra/vector_math.cuh
chris-maes: Yeah `reduce_async` or `async_reduce` is a good name for this.

chris-maes (changes requested):
Very cool. Some minor requests around making the code cleaner. Thanks for the nice work

cpp/src/linear_algebra/vector_math.cuh
yuwenchen95: Thanks for pointing it out @rg20! I added `_async` to those operations that are synchronized.

cpp/src/barrier/barrier.cu
yuwenchen95: I have moved those changes into a new class `barrier_reduce_helper_t` as suggested.

cpp/src/barrier/barrier.cu
yuwenchen95: I created an Enum within the new class to handle it.

cpp/src/barrier/barrier.cu
yuwenchen95: Now, it is hidden in the member function of the new class.

cpp/src/barrier/barrier.cu
yuwenchen95: Yeah, the computation is reordered and computed after the data move from GPU to CPU.

cpp/src/barrier/barrier.cu
```diff
@@ -230,6 +230,144 @@ static void fill_linear_cc_rhs(raft::device_span<f_t> out,
...
+
+  // Raw device slots for the caller's own cublasdot() calls.
+  f_t* cx_slot() { return d_results_.data() + kCx; }
+  f_t* by_slot() { return d_results_.data() + kBy; }
+  f_t* uv_slot() { return d_results_.data() + kUv; }
+  f_t* xqx_slot() { return d_results_.data() + kXQx; }
+
+  // Single batched device-to-host copy + the one stream synchronize needed before any accessor
+  // below can be read.
+  void sync(rmm::cuda_stream_view stream_view)
```
rg20:
```suggestion
  void sync_to_host(rmm::cuda_stream_view stream_view)
```

or `copy_to_host` 

`sync` itself is misleading

cpp/src/barrier/barrier.cu
```diff
@@ -230,6 +230,144 @@ static void fill_linear_cc_rhs(raft::device_span<f_t> out,
...
+  {
+    raft::copy(h_results_.data(), d_results_.data(), static_cast<i_t>(kCount), stream_view);
+    stream_view.synchronize();
+  }
+
+  f_t primal_residual_norm() const
```
rg20: this assumes that sync is called before, maybe its okay, 

but it is safer to check if the data is in sync.  You can have a boolean `host_is_dirty`  and set it `false` once the sync is called. Whenever a reduce call is made set this to `true`.

cpp/src/barrier/barrier.cu
```diff
@@ -230,6 +230,144 @@ static void fill_linear_cc_rhs(raft::device_span<f_t> out,
   RAFT_CHECK_CUDA(stream);
 }
 
+// Batches the independent GPU reductions/dot-products needed by
+// compute_residual_norms_mu_and_objective (primal/dual/complementarity residual norms, mu, and
+// primal/dual objectives) into one on-device results buffer and one host readback + stream sync.
+template <typename i_t, typename f_t>
+class barrier_reduce_helper_t {
```
chris-maes, line 237: This is cleaner. Thank you. 

One additional suggestion: I think it might be better if this class just handled the async reductions. Rather than doing both the async reductions and the mathematics of how to combine elements. 

I would leave the math in the code itself. So that the reader can see it. And just use this abstraction for pulling down everything at once.

cpp/src/barrier/barrier.cu
```diff
@@ -230,6 +230,144 @@ static void fill_linear_cc_rhs(raft::device_span<f_t> out,
...
+  // below can be read.
+  void sync(rmm::cuda_stream_view stream_view)
+  {
+    raft::copy(h_results_.data(), d_results_.data(), static_cast<i_t>(kCount), stream_view);
+    stream_view.synchronize();
+  }
+
+  f_t primal_residual_norm() const
+  {
+    return std::max(h_results_[kPrimalResidual], h_results_[kBoundResidual]);
```
chris-maes: I would leave this math in the main code. Rather than putting it in this class.

cpp/src/barrier/barrier.cu
```diff
@@ -230,6 +230,144 @@ static void fill_linear_cc_rhs(raft::device_span<f_t> out,
...
+  }
+  f_t dual_residual_norm() const { return h_results_[kDualResidual]; }
+  f_t complementarity_residual_norm() const
+  {
+    f_t result = std::max(h_results_[kComplXzLinear], h_results_[kComplWv]);
```
chris-maes: I would leave this math in the main code. Rather than putting it in this class.

cpp/src/barrier/barrier.cu
```diff
@@ -230,6 +230,144 @@ static void fill_linear_cc_rhs(raft::device_span<f_t> out,
...
+    if (has_soc_) { result = std::max(result, h_results_[kComplCone]); }
+    return result;
+  }
+  f_t mu(f_t mu_denom) const { return (h_results_[kMuXzSum] + h_results_[kMuWvSum]) / mu_denom; }
```
chris-maes: I would leave this math in the main code rather than putting it in this class.

cpp/src/barrier/barrier.cu
```diff
@@ -230,6 +230,144 @@ static void fill_linear_cc_rhs(raft::device_span<f_t> out,
...
+    kMuXzSum,
+    kMuWvSum,
+    kCx,
+    kBy,
+    kUv,
+    kXQx,
+    kCount
+  };
+
+  template <typename ReduceOpT>
```
chris-maes, line 352: Since these are used by barrier_reduce_helper_t I would move them before you introduce barrier_helper_t.

chris-maes: Oh whoops these are private functions. My mistake. This is good as is.

cpp/src/barrier/barrier.cu
```diff
@@ -3828,182 +3929,92 @@ void barrier_solver_t<i_t, f_t>::compute_next_iterate(iteration_data_t<i_t, f_t>
...
-                                                    d_xQx.data(),
+                                                    rh.xqx_slot(),
                                                     stream_view_));
-    quad_objective = 0.5 * d_xQx.value(stream_view_);
   }
 
-  primal_objective = d_cx.value(stream_view_) + quad_objective;
-  dual_objective   = d_by.value(stream_view_) - d_uv.value(stream_view_) - quad_objective;
-
-#ifdef CHECK_OBJECTIVE_GAP
```
chris-maes, line 3921: Please make sure this still works. It looks like you deleted the code for CHECK_OBJECTIVE_GAP. That code is not enabled by default. But it is still useful.

cpp/src/barrier/barrier.cu
```diff
@@ -3828,182 +3929,92 @@ void barrier_solver_t<i_t, f_t>::compute_next_iterate(iteration_data_t<i_t, f_t>
...
+  rh.sync(stream_view_);
+
+  primal_residual_norm          = rh.primal_residual_norm();
+  dual_residual_norm            = rh.dual_residual_norm();
+  complementarity_residual_norm = rh.complementarity_residual_norm();
+
+  const f_t mu_denom = data.complementarity_degree(data.x.size(), data.n_upper_bounds);
+  mu                 = rh.mu(mu_denom);
+
+  const f_t quad_objective = (data.Q.n > 0) ? 0.5 * rh.xqx() : f_t(0);
```
chris-maes: `xqx` is not a great name. Can you change this to `xTQx`?

cpp/src/barrier/barrier.cu
```diff
@@ -230,6 +230,144 @@ static void fill_linear_cc_rhs(raft::device_span<f_t> out,
...
+  }
+  f_t dual_residual_norm() const { return h_results_[kDualResidual]; }
+  f_t complementarity_residual_norm() const
+  {
+    f_t result = std::max(h_results_[kComplXzLinear], h_results_[kComplWv]);
+    if (has_soc_) { result = std::max(result, h_results_[kComplCone]); }
+    return result;
+  }
+  f_t mu(f_t mu_denom) const { return (h_results_[kMuXzSum] + h_results_[kMuWvSum]) / mu_denom; }
+  f_t cx() const { return h_results_[kCx]; }
```
chris-maes: Maybe `cTx`, `bTy`, and `xTQx`?

cpp/src/barrier/barrier.cu
```diff
@@ -230,6 +230,144 @@ static void fill_linear_cc_rhs(raft::device_span<f_t> out,
...
+  void mu_terms_async(const rmm::device_uvector<f_t>& d_xz,
+                      const rmm::device_uvector<f_t>& d_wv,
+                      rmm::cuda_stream_view stream_view)
+  {
+    sum_async(kMuXzSum, d_xz.data(), d_xz.size(), stream_view);
+    sum_async(kMuWvSum, d_wv.data(), d_wv.size(), stream_view);
+  }
+
+  // Raw device slots for the caller's own cublasdot() calls.
+  f_t* cx_slot() { return d_results_.data() + kCx; }
```
chris-maes: `cTx_slot`, `bTy_slot`, `uTv_slot`, and `xTQx_slot` would be better names.

cpp/src/barrier/barrier.cu
yuwenchen95: I only pulled scalar operations out of the new class. Is it expected?

cpp/src/barrier/barrier.cu
yuwenchen95: I can confirm all scalar operations are preserved.

cpp/src/barrier/barrier.cu
```diff
@@ -3828,91 +3960,70 @@ void barrier_solver_t<i_t, f_t>::compute_next_iterate(iteration_data_t<i_t, f_t>
...
+    raft::device_span<f_t> cone_dot = data.cones().scratch.template get_slot<0>();
+    data.cones().segmented_sum(
+      data.d_complementarity_xz_residual_.data() + data.cone_start(), cone_dot, stream_view_);
+    rh.cone_complementarity_residual_async(cone_dot, stream_view_);
+  }
+  rh.mu_terms_async(
+    data.d_complementarity_xz_residual_, data.d_complementarity_wv_residual_, stream_view_);
+
+  cublasHandle_t cublas_handle = lp.handle_ptr->get_cublas_handle();
+  rh.cTx_async(data.d_c_, data.d_x_, cublas_handle, stream_view_);
```
yuwenchen95, line 4004: Also abstract the async computation of `cTx`, `bTy` and `uTv` into the helper class.

cpp/src/barrier/barrier.cu
```diff
@@ -3828,91 +3960,70 @@ void barrier_solver_t<i_t, f_t>::compute_next_iterate(iteration_data_t<i_t, f_t>
...
-    RAFT_CUBLAS_TRY(raft::linalg::detail::cublasdot(lp.handle_ptr->get_cublas_handle(),
-                                                    data.d_Qx_.size(),
-                                                    data.d_Qx_.data(),
-                                                    1,
-                                                    data.d_x_.data(),
-                                                    1,
-                                                    d_xQx.data(),
-                                                    stream_view_));
-    quad_objective = 0.5 * d_xQx.value(stream_view_);
+    rh.xTQx_async(data.d_Qx_, data.d_x_, cublas_handle, stream_view_);
```
yuwenchen95, line 4011: The same abstraction for `xTQx`.

yuwenchen95 (commented):
I added more async functions in the helper class, trying to make math in `compute_residual_norms_mu_and_objective` more clear.

chris-maes (approved):
LGTM. Thanks for the change and the nice performance improvement on small models.

PR #1809, opened 2026-08-26
Replace process-wide seed_generator with per-worker RNGs in MIP heuristics

### LINKED ISSUE #1749: MIP heuristics: replace the process-wide seed_generator with per-worker RNGs
#1717 (merged) gave routing a per-problem seed generator, so routing and the MIP heuristics no longer share one process-wide counter. The heuristics still draw from `cuopt::seed_generator`, whose state is a single `static int64_t` handed out by a plain `seed_++`.

That is a data race under the solver's OpenMP thread pool, and the order in which concurrent callers receive values is not fixed, so seed assignment is not reproducible across runs.

## Proposal

Use the pattern branch and bound already uses, at `cpp/src/branch_and_bound/worker.hpp:110`:

```cpp
rng(settings.random_seed + pcgenerator_t::default_seed + rng_offset + worker_id,
    pcgenerator_t::default_stream ^ (worker_id + rng_offset))
```

Each worker owns its RNG, seeded from the user's `random_seed` and its own id, so no state is shared and the sequence a worker sees does not depend on how it interleaves with the others.

## Scope

32 call sites across 16 files under `cpp/src/mip_heuristics`. The work is not a mechanical substitution: none of those files has a worker id in scope today.

- `mip_solver_context_t` is constructed once, at `cpp/src/mip_heuristics/solver.cuh:28`
- the single OpenMP region, `cpp/src/mip_heuristics/solve.cu:850`, exists to create a thread pool shared across the solver, with the solve itself running under `#pragma omp masked`

So the heuristics have no notion of worker identity. Establishing it is the substance of this issue; updating the call sites afterwards is the easy half.

Once this lands, `cpp/src/utilities/seed_generator.{cuh,cu}` has no remaining users and can be deleted.

ramakrishnap-nv:
## Summary

`cuopt::seed_generator` (`cpp/src/utilities/seed_generator.cuh`) is a single process-wide `static int64_t seed_` counter handed out via `seed_++` in `get_seed()`. The MIP heuristics run under an OpenMP thread pool (B&B worker tasks, CPU/GPU FJ tasks, and local-search CPU-FJ climbers all run concurrently), so concurrent `get_seed()` calls race, and the order in which callers receive values is undefined. That makes deterministic-mode reproducibility best-effort rather than guaranteed.

This PR gives every `mip_heuristics` owner (diversity manager, population, local search, feasibility pump, constraint prop, all 5 recombiners, GPU/CPU FJ, bounds repair) its own persistent RNG, seeded once from `mip_solver_context_t::base_seed` (resolved from `settings.seed`, or a fresh random seed if unset) plus a fixed logical component id — never a runtime thread id, since OMP tasks can migrate between OS threads. This mirrors the pattern `branch_and_bound_worker_t` already uses.

Context-less helpers (`bounds_repair_t`, `lb_bounds_repair_t`, `solution_t::round_nearest`/`round_random_nearest`/`assign_random_within_bounds`, `simple_rounding.cu`'s free functions) now take an explicit seed from their caller instead of reaching for global state.

## Scope / non-goals

- `cpp/src/utilities/seed_generator.cuh` is left in place — routing still depends on it (13 files on `main`), tracked separately by #1717, which is narrowing to a routing-only `seed_generator` under `cpp/src/routing/utilities/`. Once that lands, the shared header can be deleted.
- One intentional exception remains: `fj_cpu.cu`'s `seed >= 0 ? seed : cuopt::seed_generator::get_seed()` fallback, reachable only from `branch_and_bound.cpp`'s non-deterministic-mode root-cut path — that's outside `mip_heuristics` and already a deliberate, documented tradeoff there.

Fixes #1749

## AI-Use Disclosure

- [x] AI tools contributed to the development of this PR
    - [x] AI tools generated code
    - Review: reviewed/verified — built and ran the relevant test suite locally, checked correctness/logic of the RNG-seeding design against the existing `branch_and_bound_worker_t` pattern.


ramakrishnap-nv:
@coderabbitai review

ramakrishnap-nv:
@coderabbitai review

ramakrishnap-nv:
@coderabbitai review

cpp/src/mip_heuristics/local_search/rounding/bounds_repair.cuh
```diff
@@ -111,7 +111,9 @@ struct candidates_t {
 template <typename i_t, typename f_t>
 class bounds_repair_t {
  public:
-  bounds_repair_t(const problem_t<i_t, f_t>& p, bound_presolve_t<i_t, f_t>& bound_presolve);
+  bounds_repair_t(const problem_t<i_t, f_t>& p,
+                  bound_presolve_t<i_t, f_t>& bound_presolve,
+                  int64_t seed);
```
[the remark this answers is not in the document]
ramakrishnap-nv: Already fixed in b6b62b91 — both headers have `#include <cstdint>` (bounds_repair.cuh:19, lb_bounds_repair.cuh:18). This looks like a false positive from your check script's regex, which requires literal backticks around `#include` (`'^`#include` <cstdint>$'`) and will never match real source. The `fj_t::rng` visibility fix from the same commit isn't being re-flagged, so that one's confirmed resolved.

cpp/src/mip_heuristics/diversity/diversity_manager.cu
```diff
@@ -439,7 +445,7 @@ template <typename i_t, typename f_t>
 void diversity_manager_t<i_t, f_t>::run_fj_alone(solution_t<i_t, f_t>& solution)
 {
   CUOPT_LOG_INFO("Running FJ alone!");
-  solution.round_nearest();
+  solution.round_nearest(static_cast<int64_t>(rng()));
```
nguidotti: Just to understand: why we need to do `static_cast` to int64 here?

cpp/src/mip_heuristics/diversity/diversity_manager.cu
```diff
@@ -693,7 +699,7 @@ solution_t<i_t, f_t> diversity_manager_t<i_t, f_t>::run_solver()
   if (ls.lp_optimal_exists) {
     solution_t<i_t, f_t> lp_rounded_sol(*problem_ptr);
     lp_rounded_sol.copy_new_assignment(lp_optimal_solution);
-    lp_rounded_sol.round_nearest();
+    lp_rounded_sol.round_nearest(static_cast<int64_t>(rng()));
```
nguidotti: Same as above

cpp/src/mip_heuristics/diversity/recombiners/line_segment_recombiner.cuh
```diff
@@ -40,7 +42,7 @@ class line_segment_recombiner_t : public recombiner_t<i_t, f_t> {
     i_t n_vars_from_other = remaining_variables;
     if (n_vars_from_other > (i_t)ls_recombiner_config_t::max_n_of_vars_from_other) {
       n_vars_from_other = ls_recombiner_config_t::max_n_of_vars_from_other;
-      thrust::default_random_engine g{(unsigned int)cuopt::seed_generator::get_seed()};
+      thrust::default_random_engine g{(unsigned int)this->rng.next_i64()};
```
nguidotti: If it is being convert to `unsigned int`, then you can probably use `next_u32` directly.

cpp/src/mip_heuristics/diversity/recombiners/sub_mip.cuh
```diff
@@ -57,7 +58,7 @@ class sub_mip_recombiner_t : public recombiner_t<i_t, f_t> {
     i_t n_vars_from_other = n_different_vars;
     if (n_vars_from_other > (i_t)sub_mip_recombiner_config_t::max_n_of_vars_from_other) {
       n_vars_from_other = sub_mip_recombiner_config_t::max_n_of_vars_from_other;
-      thrust::default_random_engine g{(unsigned int)cuopt::seed_generator::get_seed()};
+      thrust::default_random_engine g{(unsigned int)this->rng.next_i64()};
```
nguidotti: Same as before: use `next_u32` instead

cpp/src/mip_heuristics/diversity/recombiners/recombiner.cuh
```diff
@@ -65,8 +66,12 @@ class recombiner_t {
  public:
   recombiner_t(mip_solver_context_t<i_t, f_t>& context_,
                i_t n_integer_vars,
-               const raft::handle_t* handle_ptr)
+               const raft::handle_t* handle_ptr,
+               mip_rng_component_id_t component_id)
     : context(context_),
+      rng(static_cast<uint64_t>(context.base_seed) + cuopt::pcgenerator_t::default_seed +
```
nguidotti: Let us transform this into a routine. Or alternatively, we can use a hash function to generate the seed from the id (which is probably better)

cpp/src/mip_heuristics/feasibility_jump/feasibility_jump.cu
```diff
@@ -43,8 +42,13 @@ static constexpr int iterations_per_graph = 50;
 #endif
 
 template <typename i_t, typename f_t>
-fj_t<i_t, f_t>::fj_t(mip_solver_context_t<i_t, f_t>& context_, fj_settings_t in_settings)
-  : context(context_),
+fj_t<i_t, f_t>::fj_t(mip_solver_context_t<i_t, f_t>& context_,
+                     fj_settings_t in_settings,
+                     mip_rng_component_id_t seed_component_id)
+  : rng(static_cast<uint64_t>(context_.base_seed) + cuopt::pcgenerator_t::default_seed +
```
nguidotti: Similar comment. Let us a hash function to generate the seed and stream

cpp/src/mip_heuristics/feasibility_jump/fj_cpu.cu
```diff
@@ -1880,13 +1880,13 @@ std::unique_ptr<fj_cpu_climber_t<i_t, f_t>> fj_t<i_t, f_t>::create_cpu_climber(
   init_fj_cpu(*fj_cpu, solution, left_weights, right_weights, objective_weight, probing_cache);
   fj_cpu->settings = settings;
   if (randomize_params) {
-    auto rng                 = std::mt19937(cuopt::seed_generator::get_seed());
-    fj_cpu->mtm_viol_samples = std::uniform_int_distribution<i_t>(15, 50)(rng);
-    fj_cpu->mtm_sat_samples  = std::uniform_int_distribution<i_t>(10, 30)(rng);
-    fj_cpu->nnz_samples      = std::uniform_int_distribution<i_t>(2000, 15000)(rng);
-    fj_cpu->perturb_interval = std::uniform_int_distribution<i_t>(50, 500)(rng);
+    auto host_rng            = std::mt19937(rng.next_i64());
```
nguidotti, line 1883: @aliceb-nv @akifcorduk Do you think we can replace the `std::mt19937` generator here with PCG?

cpp/src/mip_heuristics/local_search/local_search.cu
```diff
@@ -65,7 +64,7 @@ void local_search_t<i_t, f_t>::start_cpufj_scratch_threads(population_t<i_t, f_t
   solution.clamp_within_bounds();
   i_t counter = 0;
   for (auto& cpu_fj : scratch_cpu_fj) {
-    if (counter > 0) solution.assign_random_within_bounds(0.4);
+    if (counter > 0) solution.assign_random_within_bounds(static_cast<int64_t>(rng()), 0.4);
```
nguidotti: Similar comment as above (why static_cast?)

cpp/src/mip_heuristics/local_search/local_search.cu
```diff
@@ -117,7 +116,7 @@ void local_search_t<i_t, f_t>::start_cpufj_lptopt_scratch_threads(
   solution_t<i_t, f_t> solution_lp(*context.problem_ptr);
   solution_lp.copy_new_assignment(
     host_copy(lp_optimal_solution, context.problem_ptr->handle_ptr->get_stream()));
-  solution_lp.round_random_nearest(500);
+  solution_lp.round_random_nearest(500, static_cast<int64_t>(rng()));
```
nguidotti: Same

cpp/src/mip_heuristics/mip_constants.hpp
```diff
@@ -39,3 +44,57 @@
...
+namespace cuopt::mathematical_optimization::mip {
+
+// Fixed logical identity used to seed each singleton heuristics component's own RNG stream from
+// mip_solver_context_t::base_seed (see cpp/src/mip_heuristics/solver_context.cuh). These are NOT
+// runtime thread/task ids -- OMP tasks can migrate between OS threads, so identity must be a
+// compile-time-fixed constant to keep seeding reproducible across runs. Components that spawn a
+// variable number of parallel workers (e.g. local-search CPU-FJ climbers) additionally offset by
+// their own fixed slot index on top of the relevant id below.
+enum class mip_rng_component_id_t : uint32_t {
+  diversity_manager = 0,
```
nguidotti: NITPICK: I would start with a larger number than `0` like `100 000`

cpp/src/mip_heuristics/mip_constants.hpp
```diff
@@ -39,3 +44,57 @@
...
+
+// Derives a well-mixed, reproducible 64-bit seed from the solve's base seed plus a fixed logical
+// identity, for owners that need a raw seed value (e.g. to hand to std::mt19937 or a multi-armed
+// bandit) rather than owning a cuopt::pcgenerator_t directly. `index` further distinguishes
+// multiple independent draws made by the same component (e.g. one per parallel worker slot).
+inline int64_t mip_derive_seed(int64_t base_seed,
+                               mip_rng_component_id_t component_id,
+                               uint32_t index = 0)
+{
+  cuopt::pcgenerator_t gen(
```
nguidotti: Maybe we could use a hash function (same comment as above)

nguidotti (commented):
Thanks for this PR, Ram! Let us replace the hand-created function for seed/stream with a good hash function so an id will generate drastically different seeds. I will replace the B&B worker rng with it as well.

nguidotti:
We can also use a random number generator to produce the seeds (a good is `SplitMix64`)

cpp/src/mip_heuristics/feasibility_jump/feasibility_jump.cuh
```diff
@@ -248,6 +252,18 @@ class fj_t {
   // executed after a roudning FJ run if any fractionals remain to eliminate them
   void round_remaining_fractionals(solution_t<i_t, f_t>& solution, i_t climber_idx = 0);
 
+  // Draws the next seed from this instance's persistent RNG (see `rng` below). Used by callers
```
akifcorduk: Let's remove the excessive comments.

cpp/src/mip_heuristics/feasibility_jump/feasibility_jump.cuh
```diff
@@ -248,6 +252,18 @@ class fj_t {
...
+  // outside fj_t (e.g. line_segment_search_t) that need a seed derived from this fj_t's stream
+  // without being able to reseed or otherwise mutate it directly.
+  uint64_t next_seed() { return rng.next_u64(); }
+
+ private:
+  // Persistent RNG seeded once from context.base_seed and this instance's fixed component id,
```
akifcorduk: Here too, remove the comments.

cpp/src/mip_heuristics/feasibility_jump/fj_cpu.cu
akifcorduk: I would leave it to @aliceb-nv but i don't think the RNG perf is critical anywhere in the code. std::mt19937 is standard so it is probably easier to use and maintain.

cpp/src/mip_heuristics/mip_constants.hpp
```diff
@@ -39,3 +45,70 @@
...
+
+namespace cuopt::mathematical_optimization::mip {
+
+// Fixed logical identity used to seed each singleton heuristics component's own RNG stream from
+// mip_solver_context_t::base_seed (see cpp/src/mip_heuristics/solver_context.cuh). These are NOT
+// runtime thread/task ids -- OMP tasks can migrate between OS threads, so identity must be a
+// compile-time-fixed constant to keep seeding reproducible across runs. Components that spawn a
+// variable number of parallel workers (e.g. local-search CPU-FJ climbers) additionally offset by
+// their own fixed slot index on top of the relevant id below.
+enum class mip_rng_component_id_t : uint64_t {
```
akifcorduk: Can you add a skill or review rule for agents and especially coderabbit such that the new added classes also use this rng order registry?

akifcorduk: I think we can reduce the name of the class"
`mip_rng_component_id_t` -> `rng_id_t`

cpp/src/mip_heuristics/solver_context.cuh
```diff
@@ -57,6 +61,11 @@ struct mip_solver_context_t {
   diversity_manager_t<i_t, f_t>* diversity_manager_ptr{nullptr};
   std::atomic<bool> preempt_heuristic_solver_ = false;
   const mip_solver_settings_t<i_t, f_t> settings;
+  // Single source of truth for all MIP heuristics RNGs: settings.seed if the user requested a
+  // specific one (>= 0), otherwise a seed drawn once at solve start. Every worker/component
+  // derives its own independent RNG stream from this plus a fixed logical identity (never a
+  // runtime thread id, since OMP tasks can migrate between OS threads).
```
akifcorduk: Reduce to a single line or remove comment please.

cpp/src/mip_heuristics/diversity/diversity_manager.cu
```diff
@@ -79,10 +79,16 @@ diversity_manager_t<i_t, f_t>::diversity_manager_t(mip_solver_context_t<i_t, f_t
                             context.problem_ptr->handle_ptr),
     sub_mip_recombiner(
       context, population, context.problem_ptr->n_variables, context.problem_ptr->handle_ptr),
-    rng(cuopt::seed_generator::get_seed()),
+    rng(mip_derive_seed(context.base_seed, mip_rng_component_id_t::diversity_manager, 0)),
     stats(context.stats),
-    mab_recombiner(0, cuopt::seed_generator::get_seed(), recombiner_alpha, "recombiner"),
-    mab_ls(mab_ls_config_t<i_t, f_t>::n_of_arms, cuopt::seed_generator::get_seed(), ls_alpha, "ls"),
+    mab_recombiner(0,
+                   mip_derive_seed(context.base_seed, mip_rng_component_id_t::diversity_manager, 1),
```
akifcorduk: We can change the name from
`mip_derive_seed` to `derive_seed`

akifcorduk (commented):
Thanks Ram! Few comments.

```cpp/src/utilities/seed_generator.cuh is left in place — routing still depends on it (13 files on main), tracked separately by https://github.com/NVIDIA/cuopt/pull/1717, which is narrowing to a routing-only seed_generator under cpp/src/routing/utilities/. Once that lands, the shared header can be deleted.```
The PR has landed and seed_generator has been moved to routing. So we can remove the seed_generator in the general root utilizies folder.

File: cpp/src/mip_heuristics/mip_constants.hpp [hunk 18]
```diff
@@ -39,3 +45,43 @@
...
+  recombiner_bound_prop,
+  recombiner_fp,
+  recombiner_line_segment,
+  recombiner_default,
+  recombiner_sub_mip,
+  local_search_cpu_fj,
+  early_cpufj,
+  early_gpufj,
+  line_segment_search,
+};
```
aliceb-nv, line 69: Why are we relying on an enum for this? Genuine question. It seems to me it'd be easier to just instanciate a rng_source_t object or similar per use site
With this design, everytime a new rng user appears, it has to be added to this monolith enum which sits in a very common header

aliceb-nv (commented):
Looks fine, just a question before I approve

cpp/src/mip_heuristics/mip_constants.hpp
nguidotti: From my understanding, we use this `enum` to derive the seed for each component based on a base `seed`,i.e., we pass the `base_seed` and the `component_id` to `SplitMix64` and it generates a seed for that component

nguidotti: What is your idea of `rng_source_t`? It is not something similar?

aliceb-nv: An object whose lifetime matches the task or whomever is using it. Modulo, say, the file:line location of that object in the code to provide a unique salt to the seed
But it may be a bit more complicated than it is worth. Let me approve the design as it stands

aliceb-nv (approved):

Issue #1817, opened 2026-08-27
[FEA] Publish self-contained Java classifier JARs with a statically linked libcuopt, as cuDF does


ramakrishnap-nv:
## Is your feature request related to a problem? Please describe.

The Java bindings (#1524) have no installation path. Nothing is published — no jar, no `libcuopt_jni.so` — so a user must build the JNI shim from source. That needs `libcuopt`, its headers, a C++20 compiler, CMake, Ninja, Maven and a JDK.

Conda is effectively mandatory, and not by choice: the JNI build needs the *exact* rmm and raft headers `libcuopt` was compiled against, because rmm encodes its version in an inline namespace (`rmm::_RMM_26_10`). A mismatched copy links cleanly and then fails at `dlopen` with an undefined symbol. Those headers only ship as conda packages, so a pip wheel is not sufficient.

Verified end to end in a clean `nvidia/cuda:13.0.3-base-ubuntu24.04` container: miniforge -> conda env -> `build_native.sh` -> `mvn package` -> compile and run a user program. It works, but that is the floor for "hello world".

## Describe the solution you'd like

Adopt the cuDF model: publish **self-contained per-classifier JARs** that bundle a statically linked native library, so a user needs only a Maven dependency and a driver.

cuDF did this in [rapidsai/cudf#23261](https://github.com/rapidsai/cudf/pull/23261) (merged 2026-07-21):

> Link against a static libcudf built from source per CUDA version rather than a conda shared libcudf.

It builds with `-DBUILD_SHARED_LIBS=OFF` and splits the work into three composable stages — static native build, per-classifier JAR packaging, Maven-repo gather.

The result is genuinely self-contained. `libcudf.so` inside the published jar has these `DT_NEEDED` entries and nothing else:

```
librt, libz, libdl, libpthread, libstdc++, libm, libgcc_s, libc, ld-linux
```

No `libcudart`, no `librmm`, no conda. Published classifiers on Maven Central for `ai.rapids:cudf:26.08.1`:

```
cudf-26.08.1-cuda12.jar          707.9 MB
cudf-26.08.1-cuda12-arm64.jar    702.6 MB
cudf-26.08.1-cuda13.jar          321.5 MB
cudf-26.08.1-cuda13-arm64.jar    317.5 MB
```

[...]

PR #1818, opened 2026-08-27
Explore self-contained Java classifier JARs with a statically linked libcuopt

ramakrishnap-nv:
## Description

Draft, opening the work for #1817. It now produces a **working self-contained classifier JAR**: a solve runs from the JAR alone, with no `libcuopt`, no conda environment and no `-Dcuopt.native.dir`. CI now builds and tests all four classifiers (CUDA 12/13 × amd64/arm64) end to end.

```
=== jar + CUDA math libs only ===
status    = OPTIMAL
objective = 5.0
take_0    = 1.0
```

## Result

Real multi-arch CI build, all classifiers, with routing and gRPC excluded:

| classifier | size |
|---|---|
| `cuopt-26.10.0-cuda12.jar` (amd64) | 944 MB |
| `cuopt-26.10.0-cuda12-arm64.jar` | 942 MB |
| `cuopt-26.10.0-cuda13.jar` (amd64) | 609 MB |
| `cuopt-26.10.0-cuda13-arm64.jar` | 658 MB |

CUDA 12 classifiers carry a larger fatbin and a larger `libnccl.so.2`/`libcudss.so.0`, so they land close to Sonatype's [1 GB per-bundle limit](https://central.sonatype.org/publish/publish-portal-upload/) on their own. Bundling more than one classifier together is not viable at this size; see open question 2.

Packaged libraries beside `libcuopt_jni.so` (amd64/cuda12, representative): `libnccl.so.2` (672 MB), `libcudss.so.0` (103 MB), `libstdc++.so.6` (22 MB), `libgomp.so.1`/`librmm.so` (~1 MB each), `libcudss_mtlayer_gomp.so.0`/`libgcc_s.so.1`/`libtbb.so.12`/`librapids_logger.so` (<1 MB each).

## What static linking actually removed, and what it did not

`-DBUILD_SHARED_LIBS=OFF` does not work here: `cuopt` is declared `add_library(cuopt SHARED ...)`, so the flag is ignored. `cuopt_static` already existed but only inside `if (BUILD_TESTS)`; it is now gated on `BUILD_TESTS OR CUOPT_BUILD_STATIC_LIB`.

Linking that archive removed the dependency on `libcuopt.so`, but not on libraries cuOpt itself needs that conda ships only as shared objects, or that are `dlopen()`'d/ABI-sensitive at runtime rather than link-time. Each appeared as a separate failure, one per rebuild or CI environment:

1. `_ZTIN3rmm10_RMM_26_109bad_allocE` — rmm's exception typeinfo
2. `tbb::detail::r1::throw_exception` — TBB, via KaMinPar
3. `ncclGroupStart` — NCCL
4. `cudssDestroy` — cuDSS
5. cuDSS's OpenMP threading backend (`libcudss_mtlayer_gomp.so.0`), `dlopen()`'d by `cudssSetThreadingLayer` rather than linked
6. `libgomp.so.1`, `libstdc++.so.6`, `libgcc_s.so.1` — the build host's GCC runtime, whose symbol versions can be newer than a consumer's own (observed on Rocky Linux 8)

All are linked/packaged beside the JNI library, resolved through its `$ORIGIN` RPATH.

## NCCL

NCCL cannot be reached from Java: `pdhg.hpp` holds `multi_gpu_engine_t*` as a forward-declared pointer defaulting to `nullptr`, and the whole surface is 14 symbols across 5 files under `cpp/src/pdlp/distributed_pdlp/`.

Findings from investigating whether it could be linked statically instead:

- **conda's `nccl` ships no static archive** — only `libnccl.so{,.2,.2.30.7}`, `libnccl_device.bc` and a pkgconfig file. `libnccl_static.a` exists only in apt's `libnccl-dev`, which would mean sourcing one dependency from apt while the rest comes from conda.
- **static is not smaller anyway**: in `nvidia/cuda:13.0.3-devel`, `libnccl_static.a` is 190 MB against 181 MB for the shared library. NCCL's bulk is per-architecture device code, which comes along either way.
- **NCCL is not part of the CUDA toolkit.** `nvidia/cuda:13.0.3-base` has none; the devel image gets it from the `libnccl2` apt package at `/lib/x86_64-linux-gnu`, not `/usr/local/cuda/lib64`. So it cannot be treated as consumer-supplied the way cuBLAS can.
- It is a hard dependency in `dependencies.yaml` for both paths — `nccl >=2.19` under `build_cpp`, `nvidia-nccl-cu1{2,3}>=2.19` under `cuda_wheels`. There is no build in which cuOpt does without it.

So it is bundled as a shared library for now. The size win is not in *how* it is linked but in **not needing it**: a switch to compile out distributed PDLP, or loading NCCL lazily through `dlopen`, would cut roughly a third off every classifier's size. `cpp/CMakeLists.txt` has no such switch today, unlike `SKIP_ROUTING_BUILD` and `SKIP_GRPC_BUILD`.

## CI

`java-static-build` runs `ci/build_java_static.sh` (static libcuopt, static link, package, verify) across a CUDA 12/13 × amd64/arm64 matrix; `java-static-test` then runs the full suite against each packaged classifier JAR on a GPU, with nothing but a JDK and Maven installed and no `libcuopt` present. Both are wired into `build.yaml`/`pr.yaml`, gated on the same file groups as `java-build`, and part of the `pr-builder` aggregator.

`verify_jar_dependencies.sh` is the check worth having: it reads `DT_NEEDED` for every packaged library and allows only what is inside the JAR, provided by the CUDA toolkit, or part of the base system — resolving against a library directory instead would make any JAR look self-contained just because the build environment happens to have everything installed.

```
ERROR: the JAR is not self-contained. Unsatisfied dependencies:
  libcuopt_jni.so needs libnccl.so.2
```

It also fails if `libcuopt.so` reappears in `DT_NEEDED` (the static link silently falling back to shared), or if `readelf` can't read anything at all.

Maven Central rate-limits NVIDIA's shared CI IP range org-wide ([build-infra#370](https://github.com/rapidsai/build-infra/issues/370)); this now prefers the GCS mirror of Central, the same fix [kvikio#992](https://github.com/rapidsai/kvikio/pull/992) applied.

## Not regressed

The shared `libcuopt` path is untouched. `CUOPT_STATIC_BUILD_DIR` and `CUOPT_BUILD_STATIC_LIB` are both empty/off by default, `-Dcuopt.native.dir` is still the loader's first strategy, and the existing (unclassified-JAR) `java-build` suite passes unchanged in CI.

## Open questions

1. Should distributed PDLP be compilable out, or NCCL loaded lazily? Biggest size win, and it fixes `libcuopt` failing to load on a plain CUDA runtime image.
2. Do classifier JARs upload to Maven Central separately, or as one bundle? At ~940 MB, even two CUDA-12 classifiers together exceed the 1 GB limit — tracked on [build-infra#379](https://github.com/rapidsai/build-infra/issues/379).

## Checklist

- Testing
   - [x] Self-contained JAR built, verified, and run -- all 4 classifiers, in CI
   - [x] `verify_jar_dependencies.sh` tested against a deliberately broken JAR
   - [x] Multi-arch (amd64/arm64) measured and passing in CI
- Documentation
   - [ ] Deferred until the approach is confirmed


ramakrishnap-nv:
@coderabbitai review

ramakrishnap-nv:
Addressed in 651e791f:
- Maven tarball download now verified against Apache's published SHA-512 before extraction.
- `console_log_callback.cpp`: mutex moved to function-local static storage.
- `verify_jar_dependencies.sh`: now fails if `readelf` is missing or yields zero DT_NEEDED entries, instead of silently passing.
- `assemble_maven_repo.sh`: version-probe JAR lookup now excludes sources/javadoc JARs, matching the classifier lookup below it.
- `NativeLibraryLoader.java`: extraction directory is now checked for symlinks/ownership and created with owner-only permissions; cached native libraries are now verified by SHA-256 digest instead of size alone.

Not addressed: pinning the three `rapidsai/shared-workflows@main` references in build.yaml to commit SHAs. All 17 other references to those same reusable workflows in this file already use `@main`, which matches the convention every other RAPIDS repo (cuDF, cuVS, kvikio) uses for these same workflows. Pinning only the 3 new lines would be inconsistent without meaningfully improving security, since the other 17 remain floating.

ci/test_java_static.sh
```diff
@@ -0,0 +1,107 @@
...
+# installing libcuopt itself would defeat the test, since the JAR is supposed to carry its own
+# copy of everything else it needs. See #1817 and the java-static-classifiers PR discussion for
+# why this moved off a fresh `conda create`: that env-solve was slow and consistently synced up
+# concurrent matrix jobs' cold Maven Central resolution, which is what triggered repeated 429s.
+rapids-logger "Installing a JDK (dnf's own maven package is too old; see MAVEN_VERSION below)"
+MAVEN_VERSION="3.9.9"
+dnf install -y java-11-openjdk-devel
+export JAVA_HOME=/usr/lib/jvm/java-11-openjdk
+MAVEN_HOME="$(mktemp -d)"
+curl -fsSL "https://archive.apache.org/dist/maven/maven-3/${MAVEN_VERSION}/binaries/apache-maven-${MAVEN_VERSION}-bin.tar.gz" \
```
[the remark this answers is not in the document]
ramakrishnap-nv: Fixed in 651e791f: the tarball is now downloaded to a file and verified against Apache's published SHA-512 before extraction, failing closed if it doesn't match.

cpp/src/math_optimization/console_log_callback.cpp
```diff
@@ -0,0 +1,31 @@
...
+/* clang-format on */
+
+#include <utilities/logger.hpp>
+
+#include <mutex>
+
+namespace cuopt {
+
+namespace {
+std::mutex g_console_callback_mutex;
```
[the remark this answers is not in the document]
ramakrishnap-nv: Fixed in 651e791f: moved the mutex into a function-local static accessor (console_callback_mutex()) instead of a namespace-scope global, matching the exact suggested diff.

java/cuopt/ci/verify_jar_dependencies.sh
```diff
@@ -0,0 +1,126 @@
...
+    for allowed in "${allowed_external[@]}"; do
+      if [[ "${stem}" == "${allowed}" ]]; then
+        permitted=true
+        break
+      fi
+    done
+    if [[ "${permitted}" == false ]]; then
+      unsatisfied+=("$(basename "${lib}") needs ${needed}")
+    fi
+  done < <(readelf -d "${lib}" 2>/dev/null | sed -n 's/.*NEEDED.*\[\(.*\)\]/\1/p')
```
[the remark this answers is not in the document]
ramakrishnap-nv: Fixed in 651e791f: exits with an error if readelf is missing, and separately requires at least one DT_NEEDED entry to have been read across all packaged libraries before treating the run as a pass.

java/cuopt/src/main/java/com/nvidia/cuopt/mathematicaloptimization/NativeLibraryLoader.java
```diff
@@ -0,0 +1,148 @@
...
+      Path directory =
+          Path.of(
+              System.getProperty("java.io.tmpdir"),
+              "cuopt-native-" + System.getProperty("user.name", "shared"));
+      Files.createDirectories(directory);
+
+      for (String companion : COMPANION_LIBRARIES) {
+        extractResource(resourcePath(osArch, companion), directory, companion);
+      }
+      return extractResource(resource, directory, fileName);
```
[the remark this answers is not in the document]
ramakrishnap-nv: Fixed in 651e791f: extraction now goes through privateExtractionDirectory(), which refuses to reuse the directory if it's a symlink or owned by a different user, and creates it with owner-only (rwx------) permissions. extractResource() now compares SHA-256 digests instead of file size before deciding to reuse a cached copy.

.github/workflows/build.yaml
```diff
@@ -64,6 +64,104 @@ jobs:
       date: ${{ inputs.date }}
       sha: ${{ inputs.sha }}
       script: ci/build_cpp.sh
+  java-static-build-matrix:
+    permissions:
+      contents: read
+    uses: rapidsai/shared-workflows/.github/workflows/compute-matrix.yaml@main
```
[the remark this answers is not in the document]
ramakrishnap-nv: Not addressed: all 17 other references to rapidsai/shared-workflows reusable workflows in this file already use @main, matching the convention every other RAPIDS repo (cuDF, cuVS, kvikio) uses for these same workflows. Pinning only these 3 new lines to a SHA would be inconsistent with the rest of the file without meaningfully improving security, since the other 17 would remain floating.

java/cuopt/src/main/java/com/nvidia/cuopt/mathematicaloptimization/NativeLibraryLoader.java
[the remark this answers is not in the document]
ramakrishnap-nv: Follow-up in b7497f1a: simplified extractResource back to a size comparison. The digest check meant hashing the full (hundreds-of-MB) resource and any cached copy on every JVM startup, which defeated much of the point of caching. privateExtractionDirectory() already closes the actual attack this guarded against -- a same-size file planted by another local user -- by refusing to reuse the directory unless it's private to the current OS user, so a size check is sufficient on top of that.

tmckayus (approved):
conceptually it makes sense to me, but I don't pretend to be a Java build expert

ramakrishnap-nv:
@coderabbitai review

.github/workflows/build.yaml
```diff
@@ -64,6 +64,104 @@ jobs:
...
+      sha: ${{ inputs.sha }}
+      node_type: "cpu16"
+      arch: ${{ matrix.ARCH }}
+      # Must be CUDA-version-pinned, not the bare "26.10-latest" tag: custom-job.yaml has no
+      # per-matrix env passthrough, so an untagged image would leave RAPIDS_CUDA_VERSION at
+      # that image's single baked-in default for every matrix entry.
+      # ci/build_java_static.sh reads it to pick the conda CUDA toolkit, so every "cuN" classifier
+      # JAR would actually be built against the same CUDA version and just mislabeled. Matches
+      # the pattern conda-cpp-build.yaml itself uses for this image family.
+      container_image: "rapidsai/ci-conda:26.10-cuda${{ matrix.CUDA_VER }}-${{ matrix.LINUX_VER }}-py${{ matrix.PY_VER }}"
```
paul-aiyedun: I think we should be using the wheel/Rocky 8 image here in order to avoid issues like this: https://github.com/NVIDIA/cudf/issues/23753

.github/workflows/pr.yaml
```diff
@@ -477,6 +486,112 @@ jobs:
...
+      build_type: pull-request
+      node_type: "cpu16"
+      arch: ${{ matrix.ARCH }}
+      # Must be CUDA-version-pinned, not the bare "26.10-latest" tag: custom-job.yaml has no
+      # per-matrix env passthrough, so an untagged image would leave RAPIDS_CUDA_VERSION at
+      # that image's single baked-in default for every matrix entry.
+      # ci/build_java_static.sh reads it to pick the conda CUDA toolkit, so every "cuN" classifier
+      # JAR would actually be built against the same CUDA version and just mislabeled. Matches
+      # the pattern conda-cpp-build.yaml itself uses for this image family.
+      container_image: "rapidsai/ci-conda:26.10-cuda${{ matrix.CUDA_VER }}-${{ matrix.LINUX_VER }}-py${{ matrix.PY_VER }}"
```
paul-aiyedun: Same comment about using the wheel/Rocky 8 image here.

.github/workflows/build.yaml
```diff
@@ -64,6 +64,104 @@ jobs:
...
+      script: "ci/build_java_static.sh"
+      artifact-name: "cuopt_java_${{ matrix.ARCH }}_cu${{ matrix.CUDA_MAJOR }}"
+      file_to_upload: "java/cuopt/classifier-jars/"
+
+  # Runs the full Java suite against each classifier JAR on a GPU, with no libcuopt installed,
+  # so a JAR that loads but computes wrong answers fails here rather than at a user.
+  java-static-test:
```
paul-aiyedun: I believe the general pattern is to have the (nightly) test job be defined in the test.yaml file.

java/cuopt/ci/build_cuopt_java_jar.sh
```diff
@@ -0,0 +1,139 @@
...
+
+STAGING="$(mktemp -d)"
+trap 'rm -rf "${STAGING}"' EXIT
+mkdir -p "${STAGING}/${RESOURCE_DIR}"
+cp "${NATIVE_LIB}" "${STAGING}/${RESOURCE_DIR}/libcuopt_jni.so"
+
+echo "Packaging classifier ${CLASSIFIER}"
+echo "  native library -> ${RESOURCE_DIR}/libcuopt_jni.so"
+
+# rmm and rapids_logger define the exception types cuOpt throws and have no static build, so
```
paul-aiyedun: Did you consider building these dependencies as static libraries, similar to what cudf does (for example, https://github.com/NVIDIA/cudf/blob/450b24a03e43de384b5417d09837f3c8bf6bb2b8/cpp/CMakeLists.txt#L355-L362)? What is the size impact of including these libraries?

java/cuopt/src/test/java/com/nvidia/cuopt/mathematicaloptimization/PackagedJarOriginCheck.java
```diff
@@ -0,0 +1,51 @@
...
+
+/**
+ * Confirms the suite is exercising a packaged classifier JAR rather than classes built from
+ * source.
+ *
+ * <p>Excluded by default and run only under {@code -Ppackaged-jar-tests}. Without it a stray
+ * {@code target/classes} on the classpath would shadow the JAR and the run would pass while
+ * testing the wrong thing entirely — which is the failure this whole job exists to rule out.
+ */
+final class PackagedJarOriginCheck {
```
paul-aiyedun, line 24: This does not appear to be doing an origin check in the same way cudf-java does i.e. it is not asserting that this class is coming from the specified JAR (see https://github.com/NVIDIA/cudf/blob/450b24a03e43de384b5417d09837f3c8bf6bb2b8/java/src/test/java/ai/rapids/cudf/PackagedJarOriginCheck.java#L41).

ci/test_java_static.sh
```diff
@@ -0,0 +1,115 @@
...
+MAVEN_TARBALL_URL="https://archive.apache.org/dist/maven/maven-3/${MAVEN_VERSION}/binaries/apache-maven-${MAVEN_VERSION}-bin.tar.gz"
+curl -fsSL "${MAVEN_TARBALL_URL}" -o "${MAVEN_TARBALL}"
+# archive.apache.org is plain HTTPS-authenticated hosting, not a signed package index, so verify
+# the download against Apache's published SHA-512 rather than trusting transport security alone.
+echo "$(curl -fsSL "${MAVEN_TARBALL_URL}.sha512")  ${MAVEN_TARBALL}" | sha512sum --check --status
+tar xz -C "${MAVEN_HOME}" --strip-components=1 -f "${MAVEN_TARBALL}"
+rm -f "${MAVEN_TARBALL}"
+export PATH="${MAVEN_HOME}/bin:${JAVA_HOME}/bin:${PATH}"
+
+if command -v ldconfig >/dev/null 2>&1 && ldconfig -p | grep -q libcuopt.so; then
```
paul-aiyedun, line 70: Do users have to do the additional `LD_LIBRARY_PATH` setup themselves or should things just work when using one of the CUDA runtime images?

java/cuopt/src/main/java/com/nvidia/cuopt/mathematicaloptimization/NativeLibraryLoader.java
```diff
@@ -0,0 +1,196 @@
...
+      return;
+    }
+
+    Path embedded = extractEmbeddedLibraries();
+    if (embedded != null) {
+      System.load(embedded.toString());
+      return;
+    }
+
+    System.loadLibrary(LIBRARY_NAME);
```
paul-aiyedun, line 92: What is the difference between this path and the `extractEmbeddedLibraries()` path i.e. what use case would result in execution of this line?

paul-aiyedun (changes requested):
The overall approach makes sense to me. However, I had questions about some of the implementation details. Also, I don't think we are using the correct image for the build.

java/cuopt/src/main/java/com/nvidia/cuopt/mathematicaloptimization/NativeLibraryLoader.java
ramakrishnap-nv: cuopt.native.dir (this line) is for building/testing against a source build: `./build.sh java` or the standalone `scripts/build_native.sh`/`scripts/test.sh` path produces libcuopt_jni.so dynamically linked against a separately-installed libcuopt.so, and this property points the loader straight at that directory -- no extraction needed, since the .so is already sitting on disk where the build put it.

extractEmbeddedLibraries() is the packaged-classifier-JAR path: no external native dir exists to point at, since the whole point is that the JAR carries its own statically-linked copy. It copies the JAR's embedded resources out to a private temp directory and loads from there instead.

They're mutually exclusive in practice: a real consumer of a classifier JAR would never set cuopt.native.dir, and a source build never has anything embedded in its plain (unclassified) JAR for extractEmbeddedLibraries() to find (see resourcePath's caller returning null in that case). PackagedJarOriginCheck.noNativeDirectoryOverrideIsInEffect() asserts the packaged-JAR test suite specifically isn't accidentally exercising the cuopt.native.dir path.

paul-aiyedun: The two paths you mentioned make sense to me. My question was about what looked like a third path i.e. the `System.loadLibrary(LIBRARY_NAME)` line that is executed when both paths are skipped.

ramakrishnap-nv: Right, that third path is real, not dead code — it fires whenever this JAR is *not* a self-contained classifier build. Two cases:

1. A non-classifier artifact: something resolved the plain `cuopt-<version>.jar` (no native resources baked in) rather than one of the `cuda12`/`cuda13`/`cuda12-arm64`/`cuda13-arm64` classifiers — e.g. a dependency declared without a classifier, or a locally-built jar via `mvn package` without the packaging profile that embeds the native libraries.
2. An unsupported `os.arch`: `resourcePath()` throws for anything other than `amd64`/`x86_64`/`aarch64`/`arm64`, so `extractEmbeddedLibraries()` returns `null` and `load()` falls through, even on a classifier jar for the *other* architecture.

In both cases the consumer has to have put `libcuopt_jni.so` (and everything it needs) on `java.library.path`/`LD_LIBRARY_PATH` themselves — it's an escape hatch for that unusual case, not part of the normal classifier-jar flow. The class-level javadoc already documents this as the three-step order (`-Dcuopt.native.dir` → embedded copy → `System.loadLibrary`); happy to make that comment more explicit about when each one actually fires if that would help.

.github/workflows/build.yaml
```diff
@@ -64,6 +64,86 @@ jobs:
...
+      # gcc_linux-64 conda package -- the same fix cuDF applied in rapidsai/cudf#23766 after
+      # hitting this in production (rapidsai/cudf#23753). See setup_java_static_env.sh.
+      #
+      # Must be CUDA-version-pinned, not the bare "26.10-latest" tag: custom-job.yaml has no
+      # per-matrix env passthrough, so an untagged image would leave RAPIDS_CUDA_VERSION at
+      # that image's single baked-in default for every matrix entry.
+      # ci/build_java_static.sh reads it to pick the CUDA toolkit, so every "cuN" classifier
+      # JAR would actually be built against the same CUDA version and just mislabeled. Matches
+      # the pattern java-static-test (test.yaml) already uses for this image family.
+      container_image: "rapidsai/ci-wheel:26.10-cuda${{ matrix.CUDA_VER }}-${{ matrix.LINUX_VER }}-py${{ matrix.PY_VER }}"
```
ramakrishnap-nv: Done — switched to `rapidsai/ci-wheel` (Rocky 8) here, matching cuDF's own fix for this exact issue (rapidsai/cudf#23766). See the comment just above this line for the full reasoning, and setup_java_static_env.sh for how every C++ build dependency (Boost/TBB/cuDSS via dnf, rmm/rapids_logger via prebuilt wheels, Maven) is sourced without conda in this environment.

.github/workflows/pr.yaml
```diff
@@ -477,19 +486,129 @@ jobs:
...
+      # gcc_linux-64 conda package -- the same fix cuDF applied in rapidsai/cudf#23766 after
+      # hitting this in production (rapidsai/cudf#23753). See setup_java_static_env.sh.
+      #
+      # Must be CUDA-version-pinned, not the bare "26.10-latest" tag: custom-job.yaml has no
+      # per-matrix env passthrough, so an untagged image would leave RAPIDS_CUDA_VERSION at
+      # that image's single baked-in default for every matrix entry.
+      # ci/build_java_static.sh reads it to pick the CUDA toolkit, so every "cuN" classifier
+      # JAR would actually be built against the same CUDA version and just mislabeled. Matches
+      # the pattern java-static-test (test.yaml) already uses for this image family.
+      container_image: "rapidsai/ci-wheel:26.10-cuda${{ matrix.CUDA_VER }}-${{ matrix.LINUX_VER }}-py${{ matrix.PY_VER }}"
```
ramakrishnap-nv: Done — same switch to `rapidsai/ci-wheel` here as build.yaml/test.yaml.

ramakrishnap-nv (commented):
Batching replies to the last few open comments in one review to avoid a separate notification for each.

.github/workflows/test.yaml
```diff
@@ -92,6 +92,48 @@ jobs:
...
+      matrix_name: conda-cpp-build
+      matrix_filter: 'map(. + {CUDA_MAJOR: (.CUDA_VER | split(".") | .[0])}) | unique_by([.ARCH, .CUDA_MAJOR])'
+
+  # Runs the full Java suite against each classifier JAR build.yaml's java-static-build produced,
+  # on a GPU, with no libcuopt installed, so a JAR that loads but computes wrong answers fails
+  # here rather than at a user. Downloads those classifier JARs by artifact name (rapids-download-
+  # from-github, inside ci/test_java_static.sh) rather than through 'needs', since build.yaml
+  # dispatches this as a separate workflow run only after java-static-build has already uploaded
+  # them.
+  java-static-test:
```
ramakrishnap-nv: Done — java-static-test now lives in test.yaml, dispatched from build.yaml's `tests` job the same way the other nightly test jobs are, with its own matrix recomputed here since a separate workflow run can't read build.yaml's job outputs directly.

java/cuopt/ci/build_cuopt_java_jar.sh
```diff
@@ -0,0 +1,189 @@
...
+
+STAGING="$(mktemp -d)"
+trap 'rm -rf "${STAGING}"' EXIT
+mkdir -p "${STAGING}/${RESOURCE_DIR}"
+cp "${NATIVE_LIB}" "${STAGING}/${RESOURCE_DIR}/libcuopt_jni.so"
+
+echo "Packaging classifier ${CLASSIFIER}"
+echo "  native library -> ${RESOURCE_DIR}/libcuopt_jni.so"
+
+# The companion set is read off the linked library's own DT_NEEDED entries rather than
```
ramakrishnap-nv: We did consider it, but TBB, NCCL and cuDSS aren't CPM-fetched-from-source the way cuDF's rapids_logger is in that link — they're prebuilt shared libraries from conda/dnf/NVIDIA's own distribution, and none of the three ships (or is safely linkable as) a static archive we could embed instead: NCCL has no static build at all in this environment, and while cuDSS does ship a static variant, CMakeLists.txt keeps it dynamic deliberately because linking both the static and shared copies in the same process risks two divergent copies of its symbol surface (see the comment on CUOPT_CUDSS_LIBRARY). So this mirrors cuDF's own approach for libraries in the same situation, not a static-deps toggle applied uniformly.

On size: from a recent build, the packaged libraries are libcuopt_jni.so 382 MB, libnccl.so.2 285 MB, libcudss.so.0 66 MB, librmm.so/libstdc++.so.6 ~1 MB each, everything else <1 MB — total classifier JAR 578 MB, under Maven Central's 1 GB bundle cap. NCCL dominates; it's only linked because PDLP's distributed solve path references it unconditionally (cpp/CMakeLists.txt has no switch to compile that out for the Java-scoped build).

java/cuopt/src/test/java/com/nvidia/cuopt/mathematicaloptimization/PackagedJarOriginCheck.java
```diff
@@ -0,0 +1,62 @@
...
+  @Test
+  void classesComeFromTheSpecificPackagedJar() throws Exception {
+    Path expected = packagedJar();
+    URL location = Problem.class.getProtectionDomain().getCodeSource().getLocation();
+    assertNotNull(location, "no code source for Problem");
+    Path actual = Paths.get(location.toURI()).toRealPath();
+    assertEquals(
```
ramakrishnap-nv: Done — rewrote this into `classesComeFromTheSpecificPackagedJar`, which now does the same origin-identity assertion as cudf-java's version: resolves the actual class's code-source location and asserts it equals the specific packaged JAR path (forwarded via `-Dcuopt.jar.path`, wired through pom.xml's surefire config), rather than just checking "is a jar".

.github/workflows/build.yaml
ramakrishnap-nv: Done — switched to `rapidsai/ci-wheel` (Rocky 8) here, matching cuDF's own fix for this exact issue (rapidsai/cudf#23766). See the comment just above this line for the full reasoning, and setup_java_static_env.sh for how every C++ build dependency (Boost/TBB/cuDSS via dnf, rmm/rapids_logger via prebuilt wheels, Maven) is sourced without conda in this environment.

.github/workflows/pr.yaml
ramakrishnap-nv: Done — same switch to `rapidsai/ci-wheel` here as build.yaml/test.yaml.

.github/workflows/build.yaml
ramakrishnap-nv: Done — java-static-test now lives in test.yaml, dispatched from build.yaml's `tests` job the same way the other nightly test jobs are, with its own matrix recomputed there since a separate workflow run can't read build.yaml's job outputs directly.

java/cuopt/ci/build_cuopt_java_jar.sh
ramakrishnap-nv: We did consider it, but TBB, NCCL and cuDSS aren't CPM-fetched-from-source the way cuDF's rapids_logger is in that link — they're prebuilt shared libraries from conda/dnf/NVIDIA's own distribution, and none of the three ships (or is safely linkable as) a static archive we could embed instead: NCCL has no static build at all in this environment, and while cuDSS does ship a static variant, CMakeLists.txt keeps it dynamic deliberately because linking both the static and shared copies in the same process risks two divergent copies of its symbol surface (see the comment on CUOPT_CUDSS_LIBRARY). So this mirrors cuDF's own approach for libraries in the same situation, not a static-deps toggle applied uniformly.

On size: from a recent build, the packaged libraries are libcuopt_jni.so 382 MB, libnccl.so.2 285 MB, libcudss.so.0 66 MB, librmm.so/libstdc++.so.6 ~1 MB each, everything else <1 MB — total classifier JAR 578 MB, under Maven Central's 1 GB bundle cap. NCCL dominates; it's only linked because PDLP's distributed solve path references it unconditionally (cpp/CMakeLists.txt has no switch to compile that out for the Java-scoped build).

java/cuopt/src/test/java/com/nvidia/cuopt/mathematicaloptimization/PackagedJarOriginCheck.java
ramakrishnap-nv: Done — rewrote this into `classesComeFromTheSpecificPackagedJar`, which now does the same origin-identity assertion as cudf-java's version: resolves the actual class's code-source location and asserts it equals the specific packaged JAR path (forwarded via `-Dcuopt.jar.path`, wired through pom.xml's surefire config), rather than just checking "is a jar".

ramakrishnap-nv (commented):
Batching replies to a few open threads in one review to avoid a separate notification for each.

.github/workflows/pr.yaml
```diff
@@ -477,6 +486,117 @@ jobs:
...
+      script: "ci/test_java_static.sh"
+      # Surefire fork-crash diagnostics (dumpstream/hs_err files) are also printed inline by
+      # the script on failure, but keep the raw reports downloadable too -- the JVM sometimes
+      # crashes without a clean dumpstream, and per-matrix-entry artifacts survive independently
+      # of how much of the console log GitHub keeps.
+      artifact-name: "cuopt_java_static_test_${{ matrix.ARCH }}_cu${{ matrix.CUDA_MAJOR }}"
+      file_to_upload: "java/cuopt/target/surefire-reports/"
+
+  java-static-gather:
```
paul-aiyedun: This job can be temporarily added for testing but should not be part of `pr.yaml`/run for every PR.

.github/workflows/build.yaml
```diff
@@ -64,6 +64,86 @@ jobs:
...
+      artifact-name: "cuopt_java_${{ matrix.ARCH }}_cu${{ matrix.CUDA_MAJOR }}"
+      file_to_upload: "java/cuopt/classifier-jars/"
+
+  # java-static-test runs against these classifiers from test.yaml instead of here, matching the
+  # general pattern of defining the (nightly) test job in test.yaml -- see the 'tests' job below,
+  # which needs java-static-build before dispatching test.yaml.
+
+  # Combines every classifier into one Maven-repository-layout artifact, which is the form a
+  # publishing workflow consumes. See rapidsai/build-infra#379.
+  java-static-gather:
```
paul-aiyedun, line 118: The artifact from the last PR job run: https://github.com/NVIDIA/cuopt/actions/runs/35367985039/job/105683600034 has the following directory layout: `com/nvidia/cuopt/cuopt/26.10.0/`. `cuopt` is repeated twice. Also, I noticed that there is no `-SNAPSHOT` suffix for these builds.

.github/workflows/pr.yaml
ramakrishnap-nv: Done — removed `java-static-gather` from pr.yaml (job and its pr-builder `needs` entry). It already exists in build.yaml for nightly/branch-push, and doesn't validate anything java-static-build/java-static-test don't already cover, so it had no reason to run on every PR.

.github/workflows/build.yaml
ramakrishnap-nv: Two separate things here:

1. `cuopt` repeated twice in the path — that's expected Maven convention, not a bug: groupId is `com.nvidia.cuopt`, artifactId is `cuopt`, and a standard repo layout is `<groupId-as-path>/<artifactId>/<version>/`, so the last groupId segment and the artifactId legitimately collide here (same as e.g. Guava's `com/google/guava/guava/<version>/`).
2. Missing `-SNAPSHOT` — this one's real. Fixed: pom.xml now carries `26.10.0-SNAPSHOT`, and update-version.sh appends `-SNAPSHOT` on every bump, matching cuDF's java/pom.xml convention (always a snapshot in the source tree; only an actual Maven Central release would strip it as a separate step). This was pre-existing on main, not introduced by this PR — just newly visible since this is the first thing to actually assemble a Maven repo layout. Also had to fix assemble_maven_repo.sh's version-extraction regex, which split on the first hyphen in the JAR filename and would've truncated at `-SNAPSHOT`; it now anchors on the classifier's own fixed shape (`cuda<major>[-arm64]`) instead.

ramakrishnap-nv (commented):
Addressed both — see individual replies.

.github/workflows/build.yaml
paul-aiyedun: Regarding 1 above, the path is valid for `com.nvidia.cuopt:cuopt`, but `com.nvidia:cuopt` is more consistent with long-standing RAPIDS Java projects such as [com.nvidia:spark-rapids-jni](https://central.sonatype.com/artifact/com.nvidia/spark-rapids-jni), [com.nvidia:rapids-4-spark_2.13](https://central.sonatype.com/artifact/com.nvidia/rapids-4-spark_2.13), [ai.rapids:cudf](https://central.sonatype.com/artifact/ai.rapids/cudf), etc. Guava is structured differently. `com.google.guava` contains related artifacts such as [guava-testlib](https://central.sonatype.com/artifact/com.google.guava/guava-testlib), [guava-gwt](https://central.sonatype.com/artifact/com.google.guava/guava-gwt), etc. Is `com.nvidia.cuopt` intended as a namespace for multiple artifacts (or separately delegated ownership)?

ramakrishnap-nv: Think we might have multiple packages under cuopt namespace, so I think other option is better. Is there a particular reason for RAPIDS to adhere to that standard ?

paul-aiyedun (approved):

PR #1821, opened 2026-08-28
Add solver caching to support re-solves for barrier QP

Iroy30:
## Description
This PR adds QP barrier cache reuse for a sequence of solves that only change/update the linear objective. User can opt-in with settings.sequence_solve = True. After an Optimal first solve, the GPU barrier workspace and barrier maps are cached on the DataModel and the subsequent update_linear_objective + Solve skips convert / presolve / scaling. It is strictly QP-only, no QC / SOC.

**Flow**
settings.sequence_solve = True
First `Solve(dm, settings)`              

- cold solve: convert → presolve → scale → barrier_solve
- On Optimal → store/cache the  transform + iteration_data on DataModel

`dm.update_linear_objective(c_new) `     

- crush user c through saved maps → cache c / d_c_
- set c_dirty

Later `Solve(dm, settings) `                   

- warm solve: skip convert/presolve/scaling
- reset_iterate_state (refresh D / KKT values, keep symbolic)
- Barrier_advanced_solve from a new start then unscale / uncrush

## Checklist

- Testing
   - [ ] Added tests
   - [ ] NA
- Documentation
   - [ ] Added new documentation
   - [ ] NA


cpp/include/cuopt/mathematical_optimization/pdlp/solver_settings.hpp
```diff
@@ -22,6 +22,11 @@
 
 #include <cuda/std/span>
 
+
+namespace cuopt::cython {
+class barrier_cache_t;
```
chris-maes: barrier_cache_t should probably live the `cuopt::mathematical_optimization` namespace

cpp/include/cuopt/mathematical_optimization/utilities/barrier_cache.hpp
```diff
@@ -0,0 +1,87 @@
...
+
+void destroy_iteration_data(iteration_data_t<int, double>* data);
+
+void apply_barrier_linear_objective(iteration_data_t<int, double>& data,
+                                    double const* barrier_c,
+                                    int n);
+}  // namespace cuopt::mathematical_optimization::barrier
+
+namespace cuopt {
+namespace CUOPT_EXPORT cython {
```
chris-maes: Why `cython` namespace here?

cpp/src/barrier/barrier.cu
```diff
@@ -356,6 +354,9 @@ class iteration_data_t {
       transform_reduce_helper_(lp.handle_ptr->get_stream()),
       transform_reduce_pair_helper_(lp.handle_ptr->get_stream()),
       sum_reduce_helper_(lp.handle_ptr->get_stream()),
+      d_scalar_batch_(kNumScalarBatchSlots, lp.handle_ptr->get_stream()),
```
chris-maes: Make sure to update to Yuwen's latest PR.

cpp/include/cuopt/mathematical_optimization/pdlp/solver_settings.hpp
```diff
@@ -357,6 +361,10 @@ class pdlp_solver_settings_t {
   // Used to force batch PDLP to solve a subbatch of the problems at a time
   // The 0 default value will make the solver use its heuristic to determine the subbatch size
   i_t fixed_batch_size{0};
+  /** When true, first GPU barrier/QCQP solve returns a ``barrier_cache_t`` capsule. */
+  bool sequence_solve{false};
+  /** Non-owning cache pointer set by ``call_solve`` for barrier symbolic reuse. */
+  cuopt::cython::barrier_cache_t* barrier_cache{nullptr};
```
chris-maes: Do we want to store `barrier_cache` in solver_settings? Maybe this should live in a solution object or in the model?

cpp/include/cuopt/mathematical_optimization/utilities/barrier_cache.hpp
```diff
@@ -0,0 +1,87 @@
...
+
+struct barrier_transform_t;
+
+/**
+ * @brief GPU solve cache owned by DataModel when sequence_solve is on.
+ *
+ * After an Optimal full solve, holds iteration_data_t and the user↔barrier transform.
```
chris-maes: Nit: Remove <-> non-ascii character

cpp/include/cuopt/mathematical_optimization/utilities/barrier_cache.hpp
```diff
@@ -0,0 +1,87 @@
...
+ * update_q crushes the new linear objective and sets c_dirty so the next Solve
```
chris-maes: `update_q` -> `update_linear_objective`

cpp/src/barrier/barrier.cu
```diff
@@ -653,32 +657,102 @@ class iteration_data_t {
...
+    if (chol != nullptr) {
+      static_cast<sparse_cholesky_base_t<i_t, f_t>*>(chol.get())->rebind_settings(settings_);
+    }
+
+    {
+      raft::common::nvtx::range fun_scope("Barrier: reset diagonal scaling");
+      const bool has_Q = Q.n > 0;
+
+      if (has_cones()) {
+        primal_perturb = 1e-8;
```
chris-maes: These need to come from settings not hard coded

cpp/src/barrier/barrier.cu
```diff
@@ -653,32 +657,102 @@ class iteration_data_t {
...
+  }
+
+  device_csr_matrix_t<i_t, f_t>& augmented_system() { return aug_mat(); }
+  const device_csr_matrix_t<i_t, f_t>& augmented_system() const { return aug_mat(); }
+
+  // Attach this solve's settings and rewind iterate-dependent state so IPM can
+  // Mehrotra-start with the new c. A and Q are unchanged; the previous solve
+  // left D and the KKT values at its last iterate. Reuse is QP-only (no cones),
+  // so form_*(false) updates values in the existing CSR; no symbolic rebuild.
+  bool prepare_for_reuse(const simplex_solver_settings_t<i_t, f_t>& settings)
```
chris-maes: Is this necessary? Could we run the initialization again at the start of the solve?

cpp/src/barrier/barrier.cu
```diff
@@ -653,32 +657,102 @@ class iteration_data_t {
...
+      handle_ptr->sync_stream();
+      if (chol != nullptr) { chol->rebind_csr_matrix(adat_mat()); }
     }
+    if (settings_.concurrent_halt != nullptr && *settings_.concurrent_halt == 1) { return false; }
+
+    reset_for_new_solve();
+    return true;
+  }
+
+  void reset_for_new_solve()
```
chris-maes: Seems like this is something we should do at the beginning of a new solve; rather than at the end of the last solve.

cpp/src/barrier/barrier.cu
```diff
@@ -1378,7 +1455,7 @@ class iteration_data_t {
     solution.z = z_tilde;
 
     dense_vector_t<i_t, f_t> dual_res = z_tilde;
-    dual_res.axpy(-1.0, lp.objective, 1.0);
+    dual_res.axpy(-1.0, c, 1.0);
```
chris-maes: Where is `c` coming from? Don't we have the new model as well?

cpp/src/barrier/barrier.cu
```diff
@@ -1986,13 +2078,13 @@ class iteration_data_t {
   f_t dual_perturb{1e-8};
   f_t primal_perturb{1e-8};
 
-  std::unique_ptr<sparse_cholesky_base_t<i_t, f_t>> chol;
+  std::shared_ptr<sparse_cholesky_base_t<i_t, f_t>> chol;
```
chris-maes: I'm not sure we should use a `shared_ptr` here. Can we transfer ownership of the `unique_ptr` instead?

cpp/src/barrier/barrier.cu
```diff
@@ -2189,8 +2289,8 @@ int barrier_solver_t<i_t, f_t>::initial_point(iteration_data_t<i_t, f_t>& data)
   //   LP block: e = 1,  SOC block: e = (sqrt(2), 0, ..., 0)
   if (data.has_cones()) {
     const i_t cs     = data.cone_start();
-    const f_t norm_b = vector_norm_inf<i_t, f_t>(lp.rhs);
-    const f_t norm_c = vector_norm_inf<i_t, f_t>(lp.objective);
+    const f_t norm_b = vector_norm_inf<i_t, f_t>(data.b);
```
chris-maes: I'm not sure I understand why we can't refer to `lp`. On subsequent solves, we should still have a valid `linear_program_t` model.

cpp/src/barrier/barrier.cu
```diff
@@ -4103,175 +4352,390 @@ lp_status_t barrier_solver_t<i_t, f_t>::check_for_suboptimal_solution(
...
-    for (i_t j = 0; j < n; j++) {
-      if (lp.upper[j] < inf) { num_upper_bounds++; }
+    if (!owned_data) {
+      if (cache != nullptr) { cache->clear(); }
+      settings.log.printf(
+        "Barrier: cache reuse failed; cached iteration_data is missing or invalid\n");
+      return lp_status_t::NUMERICAL_ISSUES;
+    }
+    try {
+      if (!owned_data->prepare_for_reuse(settings)) {
```
chris-maes: Oh I see `prepare_for_reuse` is called at the start. Maybe it's just a naming issue.

cpp/src/barrier/barrier.cu
```diff
@@ -4103,175 +4352,390 @@ lp_status_t barrier_solver_t<i_t, f_t>::check_for_suboptimal_solution(
...
+      return lp_status_t::NUMERICAL_ISSUES;
     }
 
-    csc_matrix_t<i_t, f_t> Q(lp.num_cols, 0, 0);
-    if (lp.Q.n > 0) { create_Q(lp, Q); }
-
-    iteration_data_t<i_t, f_t> data(
-      lp, num_upper_bounds, presolve_info.direct_free_variables, Q, settings);
+    iteration_data_t<i_t, f_t>& data = *owned_data;
+    auto fail_reuse = [&](lp_status_t status) {
```
chris-maes: Prefer to not use lambda functions. Either make this a named function or inline directly in code.

cpp/src/barrier/barrier.cu
```diff
@@ -4103,175 +4352,390 @@ lp_status_t barrier_solver_t<i_t, f_t>::check_for_suboptimal_solution(
...
+    settings.log.printf("Error in barrier_solver_t: %s\n", e.what());
+    return lp_status_t::NUMERICAL_ISSUES;
+  } catch (const std::bad_alloc& e) {
+    settings.log.printf("Out of memory in barrier_solver_t: %s\n", e.what());
+    return lp_status_t::NUMERICAL_ISSUES;
+  }
+}
 
+template <typename i_t, typename f_t>
+lp_status_t barrier_solver_t<i_t, f_t>::run_ipm(
```
chris-maes: Please call the main function containing all the implementation of barrier `solve`.

cpp/src/barrier/barrier.cu
```diff
@@ -4103,175 +4352,390 @@ lp_status_t barrier_solver_t<i_t, f_t>::check_for_suboptimal_solution(
...
+  f_t start_time,
+  lp_solution_t<i_t, f_t>& solution,
+  cuopt::cython::barrier_cache_t* cache,
+  std::unique_ptr<iteration_data_t<i_t, f_t>>& owned_data)
+{
+  auto finish_cache = [&](lp_status_t status) -> lp_status_t {
```
chris-maes: Please do not use lambda functions like this. Either inline the function directly (if small) or create a named member function.

cpp/src/barrier/barrier.cu
```diff
@@ -4103,175 +4352,390 @@ lp_status_t barrier_solver_t<i_t, f_t>::check_for_suboptimal_solution(
...
+    if (cache != nullptr && owned_data) {
+      if (status == lp_status_t::OPTIMAL) {
```
chris-maes: Why does this need to be called at the end of the solve. Shouldn't everything automatically be up to date and ready when we exit?

cpp/src/barrier/barrier.cu
```diff
@@ -4103,175 +4352,390 @@ lp_status_t barrier_solver_t<i_t, f_t>::check_for_suboptimal_solution(
...
+  }
+
+  settings.log.printf("Elapsed time                : %.3fs\n", toc(start_time));
+
+  if (toc(start_time) > settings.time_limit) {
+    settings.log.printf("Barrier time limit exceeded\n");
+    return finish_cache(lp_status_t::TIME_LIMIT);
+  }
+
+  i_t initial_status = initial_point(data);
```
chris-maes: Here and below? Why is so much of the barrier method rewritten? We should try to get a set of minimally invasive code changes.

cpp/src/barrier/barrier.hpp
```diff
@@ -34,9 +41,21 @@ class barrier_solver_t {
   barrier_solver_t(const simplex::lp_problem_t<i_t, f_t>& lp,
                    const simplex::presolve_info_t<i_t, f_t>& presolve,
                    const simplex::simplex_solver_settings_t<i_t, f_t>& settings);
-  simplex::lp_status_t solve(f_t start_time, simplex::lp_solution_t<i_t, f_t>& solution);
+  simplex::lp_status_t solve(f_t start_time,
```
chris-maes, line 44: Whats the difference between: `solve`, `barrier_advanced_solve` and `run_ipm`. We should only need two of these: `solve` and `advanced_solve`.

cpp/src/barrier/sparse_cholesky.cuh
```diff
@@ -143,7 +146,9 @@ class sparse_cholesky_cudss_t : public sparse_cholesky_base_t<i_t, f_t> {
       first_factor(true),
       positive_definite(true),
       A_created(false),
-      settings_(settings),
+      settings_(&settings),
```
chris-maes: Why does settings need to be a pointer. This creates a lot of unnecessary code changes? Can we revert this back, so we avoid these changes?

cpp/src/barrier/sparse_cholesky.cuh
```diff
@@ -866,6 +883,63 @@ class sparse_cholesky_cudss_t : public sparse_cholesky_base_t<i_t, f_t> {
     this->positive_definite = positive_definite;
   }
 
+  void rebind_settings(const simplex::simplex_solver_settings_t<i_t, f_t>& settings) override
+  {
+    settings_ = &settings;
```
chris-maes: `simplex_solver_settings` should just contain a set of `int`, `double`, and `bool` parameters. It is cheap to copy. Prefer a copy over unnecessary code changes.

python/cuopt/cuopt/linear_programming/solver_settings/solver_settings.pyx
```diff
@@ -117,6 +117,7 @@ cdef class SolverSettings:
         self.settings_dict = {}
         self.pdlp_warm_start_data = None
         self.mip_callbacks = []
+        self.sequence_solve = False
```
chris-maes, line 120: Why can't this be just a normal parameter?

python/cuopt/cuopt/linear_programming/solver_settings/solver_settings.pyx
```diff
@@ -459,6 +460,16 @@ cdef class SolverSettings:
                 warm_start_data.iterations_since_last_restart # noqa
             )
 
+        c_solver_settings.get_pdlp_settings().sequence_solve = self.sequence_solve
+
+    def set_sequence_solve(self, enabled):
```
chris-maes: Why are special getter/setters needed?

python/cuopt/cuopt/linear_programming/data_model/data_model_wrapper.pyx
```diff
@@ -158,6 +175,34 @@ cdef class DataModel:
     def set_objective_coefficients(self, c):
         self.c = type_cast(c, np.float64, "c")
 
+    def update_q(self, c):
```
chris-maes: `update_q` -> `update_linear_objective`

python/cuopt/cuopt/linear_programming/data_model/data_model_wrapper.pyx
```diff
@@ -68,6 +77,14 @@ cdef class DataModel:
         self.row_names = np.array([])
         self.quadratic_constraints = []
 
+    def has_barrier_cache(self):
```
chris-maes: Why does the user need to know we have a cache?

python/cuopt/cuopt/linear_programming/data_model/data_model_wrapper.pyx
```diff
@@ -68,6 +77,14 @@ cdef class DataModel:
...
+        """Return whether this data model owns a reusable solver cache."""
+        return self.barrier_cache_capsule is not None
+
+    def clear_barrier_cache(self):
```
chris-maes: Why does a user need to be able to clear the cache?

python/cuopt/cuopt/linear_programming/data_model/data_model.py
```diff
@@ -228,6 +228,24 @@ def set_objective_coefficients(self, c):
         """
         super().set_objective_coefficients(c)
 
+    @catch_cuopt_exception
+    def update_q(self, c):
```
chris-maes: `update_q` -> `update_linear_objective`

python/cuopt/cuopt/linear_programming/data_model/data_model.py
```diff
@@ -228,6 +228,24 @@ def set_objective_coefficients(self, c):
...
+        """
+        Update the linear objective coefficients (c) for a sequence re-solve.
+
+        Writes user-space ``c`` onto this DataModel. If a barrier cache is
```
chris-maes: Let's not user the term `user-space`.

cpp/src/pdlp/CMakeLists.txt
```diff
@@ -39,6 +39,7 @@ set(LP_CORE_FILES
 # C and Python adapter files
 set(LP_ADAPTER_FILES
   ${CMAKE_CURRENT_SOURCE_DIR}/utilities/cython_solve.cu
+  ${CMAKE_CURRENT_SOURCE_DIR}/utilities/barrier_cache.cu
```
chris-maes: `barrier_cache.cu` should live in `barrier` subdirectory.

chris-maes (changes requested):

cpp/include/cuopt/mathematical_optimization/utilities/barrier_cache.hpp
```diff
@@ -0,0 +1,87 @@
...
+#include <cuopt/export.hpp>
+
+#include <memory>
+
+#include <raft/core/handle.hpp>
+#include <rmm/cuda_stream.hpp>
+
+namespace cuopt::mathematical_optimization::barrier {
+template <typename i_t, typename f_t>
+class iteration_data_t;
```
rg20, line 19: `iteration_data_t` is too specific to our implementation of barrier.  

It can be just `solver_cache_t`, which makes it easy to add more things.

cpp/include/cuopt/mathematical_optimization/utilities/barrier_cache.hpp
```diff
@@ -0,0 +1,87 @@
...
+#pragma once
+
+#include <cuopt/export.hpp>
+
+#include <memory>
+
+#include <raft/core/handle.hpp>
+#include <rmm/cuda_stream.hpp>
+
+namespace cuopt::mathematical_optimization::barrier {
```
rg20, line 17: I don't think we should expose this namespace at this level

cpp/include/cuopt/mathematical_optimization/pdlp/solver_settings.hpp
yuwenchen95: I agree since users should ideally utilize the resolve under any API.

cpp/src/barrier/barrier.cu
```diff
@@ -2165,12 +2245,27 @@ class iteration_data_t {
...
   rmm::device_uvector<i_t> d_augmented_diagonal_indices_;
+
+  device_csr_matrix_t<i_t, f_t>& aug_mat() { return device_augmented; }
+  const device_csr_matrix_t<i_t, f_t>& aug_mat() const { return device_augmented; }
+  device_csr_matrix_t<i_t, f_t>& adat_mat() { return device_ADAT; }
+  const device_csr_matrix_t<i_t, f_t>& adat_mat() const { return device_ADAT; }
+  device_csr_matrix_t<i_t, f_t>& a_mat() { return device_A; }
+  device_csc_matrix_t<i_t, f_t>& ad_mat() { return device_AD; }
+  rmm::device_uvector<f_t>& original_a_values() { return d_original_A_values; }
+  rmm::device_uvector<f_t>& a_x_values() { return device_A_x_values; }
```
yuwenchen95: `a_x_values` is not used and could be deleted.

cpp/src/barrier/barrier.cu
yuwenchen95: I agree. Transferring ownership is better.

cpp/src/barrier/barrier.cu
yuwenchen95: We have introduced new parameters for setting primal and dual regularization. Better to update the PR with current main to address it.

cpp/src/barrier/barrier.cu
```diff
@@ -2288,7 +2383,7 @@ class iteration_data_t {
 
   rmm::cuda_stream_view stream_view_;
 
-  const simplex_solver_settings_t<i_t, f_t>& settings_;
```
yuwenchen95, line 2291: Keep `simplex_solver_settings_t` as a `const` since the setting itself should be immutable.

cpp/src/pdlp/solve.cu
```diff
@@ -72,6 +75,33 @@
 
 namespace cuopt::mathematical_optimization {
 
+namespace {
+
+template <typename i_t, typename f_t>
+simplex::user_problem_t<i_t, f_t> user_problem_from_transform(
```
yuwenchen95, line 80: This part looks fragile and incomplete. Can we skip it for the reuse path?

cpp/src/barrier/barrier.hpp
```diff
@@ -17,9 +17,16 @@
 #include <linear_algebra/sparse_matrix.hpp>
 #include <math_optimization/tic_toc.hpp>
 
+#include <memory>
+
 #include <rmm/device_uvector.hpp>
 
 #include <utility>
+
+namespace cuopt::cython {
```
yuwenchen95: Do we need `cuopt::cython`? I think `barrier_cache_t` should be part of C++ source code, which allows other APIs to use the resolve.

cpp/src/dual_simplex/solve.cpp
```diff
@@ -358,16 +416,57 @@ lp_status_t solve_linear_program_with_barrier(const user_problem_t<i_t, f_t>& us
...
+  }
+
   lp_problem_t<i_t, f_t> original_lp(handle_ptr, 1, 1, 1);
 
   // Convert the user problem to a linear program with only equality constraints
   std::vector<i_t> new_slacks;
-  simplex_solver_settings_t<i_t, f_t> barrier_settings = settings;
   dualize_info_t<i_t, f_t> dualize_info;
-  convert_user_problem(user_problem, barrier_settings, original_lp, new_slacks, dualize_info);
+  {
```
yuwenchen95: Nit: bracket looks redundant here.

cpp/src/dual_simplex/solve.cpp
```diff
@@ -377,7 +476,10 @@ lp_status_t solve_linear_program_with_barrier(const user_problem_t<i_t, f_t>& us
   // Presolve the linear program
   presolve_info_t<i_t, f_t> presolve_info;
   lp_problem_t<i_t, f_t> presolved_lp(handle_ptr, 1, 1, 1);
-  const i_t ok = presolve(original_lp, barrier_settings, presolved_lp, presolve_info);
+  i_t ok;
+  {
```
yuwenchen95: Nit: bracket looks redundant here.

cpp/src/dual_simplex/solve.cpp
```diff
@@ -389,13 +491,56 @@ lp_status_t solve_linear_program_with_barrier(const user_problem_t<i_t, f_t>& us
                                     presolved_lp.A.col_start[presolved_lp.num_cols]);
   std::vector<f_t> column_scales;
   std::vector<f_t> row_scales;
-  scaling(presolved_lp, barrier_settings, barrier_lp, column_scales, row_scales);
+  {
```
yuwenchen95: Nit: bracket looks redundant here.

cpp/src/barrier/barrier_cache.cu
```diff
@@ -0,0 +1,129 @@
...
+
+bool barrier_cache_t::c_dirty() const
+{
+  return impl_->c_dirty && impl_->transform != nullptr && impl_->iteration_data.get() != nullptr;
+}
+
+void barrier_cache_t::update_linear_objective(double const* c, int n)
+{
+  cuopt_expects(impl_->transform != nullptr,
+                error_type_t::ValidationError,
```
yuwenchen95, line 99: Nit: `update_q` should be updated to `update_linear_objective`.

cpp/src/barrier/barrier.cu
Iroy30: updated

python/cuopt/cuopt/linear_programming/data_model/data_model_wrapper.pyx
Iroy30: removed in cleanup

python/cuopt/cuopt/linear_programming/data_model/data_model_wrapper.pyx
Iroy30: removed in cleanup

cpp/src/dual_simplex/solve.cpp
Iroy30: removed

cpp/src/dual_simplex/solve.cpp
Iroy30: removed

cpp/src/dual_simplex/solve.cpp
Iroy30: removed

cpp/src/barrier/barrier.cu
Iroy30: removed

cpp/include/cuopt/mathematical_optimization/pdlp/solver_settings.hpp
Iroy30: moved

cpp/include/cuopt/mathematical_optimization/utilities/barrier_cache.hpp
Iroy30: updated

cpp/src/barrier/barrier.cu
Iroy30: updated to new parameters. Thanks for catching!

cpp/src/barrier/barrier.hpp
Iroy30: - **solve**: called on first cold solve. Allocates a new iteration_data_t (Q, ADAT vs augmented, symbolic analysis, first form_*) and then calls run_ipm.
- **barrier_advanced_solve**: subsequent solve with cache reuse. Takes the cached iteration_data_t, runs prepare_for_reuse (to be renamed), then calls the same run_ipm.
- **run_ipm**: Mehrotra IPM (initial_point + iterations). On Optimal it stores the workspace back on the cache; otherwise it clears it.

python/cuopt/cuopt/linear_programming/data_model/data_model_wrapper.pyx
Iroy30: updated

python/cuopt/cuopt/linear_programming/data_model/data_model.py
Iroy30: updated

cpp/src/barrier/barrier_cache.cu
Iroy30: updated

cpp/src/pdlp/CMakeLists.txt
Iroy30: moved

cpp/src/barrier/barrier.cu
Iroy30: renamed to reset_iterate_state

cpp/src/barrier/barrier.cu
Iroy30: moved inside reset_iterate_state which is called at the start of a subsequent solve

cpp/src/barrier/barrier.cu
Iroy30: renamed to reset_iterate_state

cpp/src/barrier/barrier.cu
Iroy30: inlined

cpp/src/barrier/barrier.cu
Iroy30: changed into function - "store_or_clear_cache"

cpp/src/barrier/sparse_cholesky.cuh
Iroy30: chol needs a C++ settings object (logs, halt, etc.). It used to refer to the settings for the solve previously. We now keep chol for the next solve, but that C++ settings object is gone when the first call returns. Since a C++ reference couldn't be pointed at a new object, a pointer and rebind_settings was used .. hence all the `settings->` changes.

However, the saved cache already keeps its own copy of settings, I've updated chol to refer to that copy instead so that on the next Solve we overwrite the copy with the new settings and chol still looks at the same object, now with the new values.

cpp/src/barrier/barrier.hpp
Iroy30: updated

cpp/src/pdlp/CMakeLists.txt
Iroy30: moved

cpp/include/cuopt/mathematical_optimization/utilities/barrier_cache.hpp
Iroy30: moved into mathrmatical_optimization namespace

cpp/src/barrier/sparse_cholesky.cuh
Iroy30: removed the need for rebind_settings. See https://github.com/NVIDIA/cuopt/pull/1821#discussion_r3910205054

python/cuopt/cuopt/linear_programming/solver_settings/solver_settings.pyx
Iroy30: removed in cleanup

python/cuopt/cuopt/linear_programming/solver_settings/solver_settings.pyx
Iroy30: If we have `CUOPT_SEQUENCE_SOLVE`, cuOptSetParameter, the CLI, dump/load, and the server config can all set it. However, those paths right now have no place to hold barrier cache itself, meaning, users could turn the flag on and get nothing useful.

Iroy30: That being said, we could have it in our public API but mention we can set and meaningfully reuse it only via python Datamodel flow

cpp/include/cuopt/mathematical_optimization/utilities/barrier_cache.hpp
```diff
@@ -0,0 +1,85 @@
...
+struct barrier_transform_t;
+
+/**
+ * @brief GPU solve cache owned by DataModel when sequence_solve is on.
+ *
+ * After an Optimal full solve, holds iteration_data_t and the user-barrier transform.
+ * update_linear_objective crushes the new linear objective and sets c_dirty so the next Solve
+ * reuses that workspace (skip convert/presolve/scaling).
+ */
+class barrier_cache_t {
```
chris-maes, line 40: I think this should probably be 
```
class barrier_cache_t`
```
And then within the class you should use i_t for int and f_t for double.

cpp/src/barrier/barrier.cu
```diff
@@ -29,6 +29,12 @@
 
 #include <linear_algebra/vector_math.cuh>
 
+#include <cuopt/mathematical_optimization/utilities/barrier_cache.hpp>
+
+#include <algorithm>
```
chris-maes: Nit: these should go with the rest of C++ headers at line 46

cpp/src/barrier/barrier.cu
```diff
@@ -855,32 +858,95 @@ class iteration_data_t {
...
-          raft::common::nvtx::range form_scope("Barrier: LP Data: form ADAT");
-          form_adat(true);
+      }
+    }
+  }
+
+  device_csr_matrix_t<i_t, f_t>& augmented_system() { return aug_mat(); }
+  const device_csr_matrix_t<i_t, f_t>& augmented_system() const { return aug_mat(); }
+
+  // Attach this solve's settings and rewind iterate-dependent state so IPM can
```
chris-maes: We don't use "IPM" in the code. We refer to it as barrier. Please remove all instances of "IPM" from the PR .

cpp/src/barrier/barrier.cu
```diff
@@ -855,32 +858,95 @@ class iteration_data_t {
...
+  // Mehrotra-start with the new c. A and Q are unchanged; the previous solve
```
chris-maes, line 889: Mehrotra-start -> start

cpp/src/barrier/barrier.cu
```diff
@@ -2165,12 +2233,26 @@ class iteration_data_t {
...
-  const csc_matrix_t<i_t, f_t>& A;
+  csc_matrix_t<i_t, f_t> A;
 
-  const csc_matrix_t<i_t, f_t>& Q;
+  csc_matrix_t<i_t, f_t> Q;
   std::vector<f_t> Qdiag;
   bool Q_diagonal;
   rmm::device_uvector<i_t> d_augmented_diagonal_indices_;
+
+  device_csr_matrix_t<i_t, f_t>& aug_mat() { return device_augmented; }
```
chris-maes: why are we creating functions that just return these matrices? Why not just keep using the original names in the code?

cpp/src/barrier/barrier.cu
```diff
@@ -4265,129 +4347,325 @@ lp_status_t barrier_solver_t<i_t, f_t>::check_for_suboptimal_solution(
...
+    raft::common::nvtx::range fun_scope("Barrier: barrier_advanced_solve");
 
     i_t n = lp.num_cols;
     i_t m = lp.num_rows;
-
     solution.resize(m, n);
     settings.log.printf(
       "Barrier solver: %d constraints, %d variables, %ld nonzeros\n", m, n, lp.A.col_start[n]);
-
-    settings.log.printf("\n");
```
chris-maes, line 4281: Are these added somewhere else?

chris-maes: If not please put this back.

python/cuopt/cuopt/linear_programming/solver_settings/solver_settings.pyx
chris-maes: Let's keep it a regular parameter. We will document that it only works within the Python interface.

cpp/include/cuopt/mathematical_optimization/pdlp/solver_settings.hpp
Iroy30: On first solve return, cache comes back on the solution (lp_ret.barrier_cache) & python owns/stores it on DataModel.
Next solve: cache from python DataModel -> call_solve -> call_solve_lp -> solve_lp -> solve_qcqp -> barrier_advanced_solve. This way all APIs stay unchanged.

Alternatives:
1) Put cache on optimization_problem_t and user_problem_t and trickle it down.
2) Or, add a additional cache argument to call_solve_lp, solve_lp, solve_qcqp.

Settings is the smallest and least invasive API change. we can do 1 or 2 as well

cpp/src/barrier/barrier.hpp
chris-maes: Can you eliminate `run_ipm`? I don't think it is necessary. `solve` can create everything that is need for `barrier_advanced_solve`. And `barrier_advanced_solve` can contain the implementation. 

We do *not* use the term IPM in the code base. So if you absolutely must create three functions please use the term barrier.

mlubin:
How specialized is the code to the python interface / how easy would it be to expose the same functionality to C? Not suggesting this should be done now, just asking as a sanity check.

cpp/src/barrier/barrier.cu
Iroy30: renamed to store_or_clear_cache. If Optimal , we store, if not , we clear

cpp/src/barrier/barrier.cu
Iroy30: renamed

cpp/src/barrier/barrier.hpp
Iroy30: renamed . We now have 3 functions: `solve()` & `solve_with_cache()` which both call `barrier_advanced_solve `(which is the core barrier implementation)

cpp/src/barrier/barrier.cu
Iroy30: It was a git diff issue falsely marking code relocation as changes. I've restructured the function calls, it should be easier to review now

cpp/src/barrier/barrier.cu
Iroy30: Resolved. The reason for the change initially was that on a `update_linear_objective` call we were updating new linear obj within iteration_data_t in cache. The stored barrier_lp still had old/stale objective coefficients in `lp.objective`. 
Updated `update_linear_objective` to now edit both copies

cpp/src/barrier/barrier.cu
Iroy30: resolved

cpp/src/barrier/barrier.cu
Iroy30: resolved. For cause and fix see https://github.com/NVIDIA/cuopt/pull/1821#discussion_r3921357011

cpp/src/pdlp/solve.cu
Iroy30: There is a ~2 ms addition in the extra host copy of A/Q/bounds in cuopt_optimization_problem_to_user_problem. The stub skips that. Agreed it's not ideal. Maybe we can find a better way to skip that

Iroy30:
> How specialized is the code to the python interface / how easy would it be to expose the same functionality to C? Not suggesting this should be done now, just asking as a sanity check.

@mlubin It is for now specialized to python datamodel interface. The ease of exposing it to C is adding a way to hold the cache in optimization_problem_t. It should be an easy structure/plumbing change. For exposing it to python API, it is a simpler approach - we would need to wrap it up in the problem model.

cpp/src/barrier/barrier.cu
```diff
@@ -843,9 +848,6 @@ class iteration_data_t {
                  handle_ptr->get_stream());
       // For efficient scaling of AD col we form the col index array
       device_AD.form_col_index(handle_ptr->get_stream());
-      device_A_x_values.resize(device_AD.x.size(), handle_ptr->get_stream());
```
Iroy30, line 846: Deleted since never used - As per @yuwenchen95 review

cpp/include/cuopt/mathematical_optimization/utilities/barrier_cache.hpp
chris-maes: It might need to be exposed due to Cython? @Iroy30 can you comment?

cpp/include/cuopt/mathematical_optimization/utilities/barrier_cache.hpp
chris-maes: I think this is just forward declaring `iteration_data_t` so that it can be referenced in the `barrier_cache_t` definition.

cpp/src/barrier/barrier.cu
```diff
@@ -4701,6 +4727,153 @@ lp_status_t barrier_solver_t<i_t, f_t>::solve(f_t start_time, lp_solution_t<i_t,
...
+    solution.resize(m, n);
+    settings.log.printf(
+      "Barrier solver: %d constraints, %d variables, %ld nonzeros\n", m, n, lp.A.col_start[n]);
+    if (lp.Q.n > 0) {
+      settings.log.printf("Quadratic objective matrix  : %d nonzeros\n", lp.Q.row_start[lp.Q.n]);
+    }
+
+    std::unique_ptr<iteration_data_t<i_t, f_t>> owned_data;
+    if (cache != nullptr) {
+      if (auto* cached = cache->release_iteration_data()) { owned_data.reset(cached); }
```
chris-maes, line 4754: This would be clearer if you wrote it as:
```
auto* cached = cache->release_iteration_data();
if (cached) {
 owned_data.reset(cached);
}
```
It's not great to stick assignment in the predicate of if statements.

cpp/src/barrier/barrier.cu
```diff
@@ -4701,6 +4727,153 @@ lp_status_t barrier_solver_t<i_t, f_t>::solve(f_t start_time, lp_solution_t<i_t,
...
+    }
+    if (!owned_data) {
+      if (cache != nullptr) { cache->clear(); }
+      settings.log.printf(
+        "Barrier: cache reuse failed; cached iteration_data is missing or invalid\n");
```
chris-maes, line 4759: When would this happen?

cpp/src/barrier/barrier.cu
```diff
@@ -4713,6 +4886,38 @@ lp_status_t barrier_solver_t<i_t, f_t>::solve(f_t start_time, lp_solution_t<i_t,
   }
 }
 
+// Optimal: persist iteration_data_t on the cache. Otherwise drop it.
+template <typename i_t, typename f_t>
+lp_status_t store_or_clear_cache(cuopt::mathematical_optimization::barrier_cache_t* cache,
```
chris-maes, line 4891: If this is only used once,  it might be worth inlining. That is, I'm not sure you want to create a separate function for this.

chris-maes: Oh I see, it is called twice. Fine to keep as is.

cpp/src/barrier/barrier.hpp
```diff
@@ -36,9 +41,20 @@ class barrier_solver_t {
   barrier_solver_t(const simplex::lp_problem_t<i_t, f_t>& lp,
                    const simplex::presolve_info_t<i_t, f_t>& presolve,
                    const simplex::simplex_solver_settings_t<i_t, f_t>& settings);
-  simplex::lp_status_t solve(f_t start_time, simplex::lp_solution_t<i_t, f_t>& solution);
+  simplex::lp_status_t solve(f_t start_time,
+                             simplex::lp_solution_t<i_t, f_t>& solution,
+                             cuopt::mathematical_optimization::barrier_cache_t* cache = nullptr);
```
chris-maes, line 46: Why would `solve` be called with `cache != nullptr`?

chris-maes: Would `cache != nullptr` on the first solve in the sequence?

chris-maes (approved):
LGTM. Thanks for all the hard work @Iroy30 .

cpp/src/barrier/barrier.hpp
```diff
@@ -36,9 +41,20 @@ class barrier_solver_t {
...
+  // Cache reuse: cached iteration_data_t already has the updated linear objective.
+  // Reset iterate state, compute a new initial point, run barrier. Same status/solution contract as
+  // solve().
+  simplex::lp_status_t solve_with_cache(f_t start_time,
```
rg20: how is this different from the solve method with non nullptr cache?

cpp/include/cuopt/mathematical_optimization/pdlp/solver_settings.hpp
```diff
@@ -365,6 +367,10 @@ class pdlp_solver_settings_t {
   // Used to force batch PDLP to solve a subbatch of the problems at a time
   // The 0 default value will make the solver use its heuristic to determine the subbatch size
   i_t fixed_batch_size{0};
+  /** When true, first GPU barrier/QCQP solve returns a ``barrier_cache_t`` capsule. */
+  bool sequence_solve{false};
+  /** Non-owning cache pointer set by ``call_solve`` for barrier cache reuse. */
+  barrier_cache_t* barrier_cache{nullptr};
```
yuwenchen95, line 373: It is best to move `barrier_cache` out of settings to `optimization_problem_t`.

cpp/src/barrier/barrier.cu
```diff
@@ -4701,6 +4727,153 @@ lp_status_t barrier_solver_t<i_t, f_t>::solve(f_t start_time, lp_solution_t<i_t,
                      data.cusparse_view_,
                      solution);
     return lp_status_t::ITERATION_LIMIT;
+  }
+}
+
+template <typename i_t, typename f_t>
+lp_status_t barrier_solver_t<i_t, f_t>::solve_with_cache(
```
yuwenchen95, line 4734: Any chance to uniform remaining overlapping parts in `solve_with_cache` and `barrier_solver_t<i_t, f_t>::solve`? It will prevent diverging pre-solve checks in future development and shorten code lines a bit.

cpp/src/barrier/sparse_cholesky.cuh
```diff
@@ -866,6 +878,55 @@ class sparse_cholesky_cudss_t : public sparse_cholesky_base_t<i_t, f_t> {
     this->positive_definite = positive_definite;
   }
 
+  /// Re-point cuDSS CSR wrapper at current device buffers after in-place value refresh.
+  void rebind_csr_matrix(device_csr_matrix_t<i_t, f_t>& Arow) override
```
yuwenchen95, line 882: Do we still need rebinding for cudss matrix for every new problem? If the we assume sparsity is fixed without presolve, the sparsity pattern of cudss matrix should also be unchanged and we only need to refresh numerical values instead of creating a new matrix from the scratch.

yuwenchen95 (changes requested):
Overall it is good to me. Two remaining concerns are the location of `barrier_cache` and necessity of `rebind_csr_matrix`.

yuwenchen95 (approved):

cpp/src/barrier/barrier.hpp
Iroy30: Yes whenever we want to persist the cache and enable sequence_solve, the first solve populates it

cpp/include/cuopt/mathematical_optimization/pdlp/solver_settings.hpp
Iroy30: Noted as a follow-up to address similar concerns 
https://github.com/NVIDIA/cuopt/pull/1821#discussion_r3918378570

cpp/include/cuopt/mathematical_optimization/pdlp/solver_settings.hpp
Iroy30: Noted as a follow-up

cpp/src/barrier/sparse_cholesky.cuh
Iroy30: That is right it is not required to Destroy and recreate cudss matrix once we assert the nnz/sparsity isn't chnaged. Can be skipped. Updated to only refreshing values.

cpp/src/barrier/barrier.cu
Iroy30: Moved overlapping checks into the common `barrier_advanced_solve` routine

cpp/src/barrier/barrier.cu
Iroy30: It's a precautionary guard. It should not happen

cpp/src/barrier/barrier.cu
Iroy30: split

cpp/include/cuopt/mathematical_optimization/utilities/barrier_cache.hpp
Iroy30: Yes when we pass the cache the python/cython needs to know the type

python/cuopt/cuopt/linear_programming/data_model/data_model.py
```diff
@@ -228,6 +228,24 @@ def set_objective_coefficients(self, c):
...
 
+    @catch_cuopt_exception
+    def update_linear_objective(self, c):
+        """
+        Update the linear objective coefficients (c) for a sequence re-solve.
+
+        Writes ``c`` onto this DataModel. If a barrier cache is
+        present, also maps ``c`` into the cached barrier workspace and marks
+        it dirty (quadratic ``Q``, ``A``, and bounds must stay unchanged).
+        Cache reuse is QP-only: quadratic constraints take a full solve.
```
ramakrishnap-nv: May be we can move the note of QP-only to the top.

ramakrishnap-nv: instead of C, should we use coeff? or something bit more elaborate ?

python/cuopt/cuopt/linear_programming/solver/solver_wrapper.pyx
```diff
@@ -79,6 +87,25 @@ cdef extern from "cuopt/mathematical_optimization/utilities/internals.hpp" names
     cdef cppclass base_solution_callback_t
 
 
+cdef extern from *:
```
ramakrishnap-nv: Can't we mention header file directly instead of * ?

ramakrishnap-nv (changes requested):
Have few questions and suggestions, rest looks good. Also lets add tests to the python to cover this API.

python/cuopt/cuopt/linear_programming/data_model/data_model.py
Iroy30: edited

python/cuopt/cuopt/linear_programming/solver/solver_wrapper.pyx
Iroy30: updated

Iroy30:
@ramakrishnap-nv Added a small test. Thanks!

ramakrishnap-nv (approved):

PR #1826, opened 2026-08-28
Enable sccache preprocessor-cache mode for wheel builds

ramakrishnap-nv:
## Summary
Enable scache for wheel build.

### DIFF
--- a/ci/build_wheel.sh
+++ b/ci/build_wheel.sh
@@ -28,6 +28,13 @@ source rapids-configure-sccache
 source rapids-datetime-string
 source rapids-init-pip
 
+# Wheel builds happen in a fresh checkout path on every run, which breaks sccache's default
+# direct-mode cache key (it includes the literal compiler invocation, e.g. include paths).
+# Preprocessor-cache mode keys on the preprocessed source instead, so it survives that. Same
+# pattern as cudf's and rmm's wheel builds.
+export SCCACHE_S3_PREPROCESSOR_CACHE_KEY_PREFIX="${package_name}-${RAPIDS_CONDA_ARCH}-cuda${RAPIDS_CUDA_VERSION%%.*}-wheel-preprocessor-cache"
+export SCCACHE_S3_USE_PREPROCESSOR_CACHE_MODE=true
+
 # Update the version to accomodate nightly and release changes for the wheel name
 RAPIDS_VERSION_SUFFIX=".post${RAPIDS_DATETIME_STRING}" \
   rapids-generate-version > ./VERSION


ci/build_wheel.sh
```diff
@@ -28,6 +28,13 @@ source rapids-configure-sccache
 source rapids-datetime-string
 source rapids-init-pip
 
+# Wheel builds happen in a fresh checkout path on every run, which breaks sccache's default
+# direct-mode cache key (it includes the literal compiler invocation, e.g. include paths).
+# Preprocessor-cache mode keys on the preprocessed source instead, so it survives that. Same
+# pattern as cudf's and rmm's wheel builds.
+export SCCACHE_S3_PREPROCESSOR_CACHE_KEY_PREFIX="${package_name}-${RAPIDS_CONDA_ARCH}-cuda${RAPIDS_CUDA_VERSION%%.*}-wheel-preprocessor-cache"
+export SCCACHE_S3_USE_PREPROCESSOR_CACHE_MODE=true
```
jameslamb, line 36: I don't understand this one. Can you show some evidence of `sccache` not being enabled already in a place you expect it?

For example, on the most recent `libcuopt` build on `main`, I see pretty good cache hit rates building `libcuopt`.

```text
Cache hits rate                            94.44 %
Cache hits rate (c [gcc])                 100.00 %
Cache hits rate (c++ [gcc])                99.42 %
Cache hits rate (cubin [ptxas])           100.00 %
Cache hits rate (cuda [cudafe++])           0.00 %
Cache hits rate (cuda [nvcc])              98.28 %
Cache hits rate (ptx [cicc])                0.00 %
```

([build link](https://github.com/NVIDIA/cuopt/actions/runs/33359460395/job/99387925801#step:13:11463))

jameslamb: If your goal is to get `sccache` benefits for other wheels, turn off build isolation for them so they'll build from a stable path.

https://github.com/NVIDIA/cuopt/blob/d3af87e36869ce0fdfb451cbf1e0c27317a9243c/ci/build_wheel_libcuopt.sh#L91-L93

jameslamb (changes requested):
I don't think this is the right fix and I'm unsure which wheel builds you're trying to enable caching for. Left some suggestions.

ramakrishnap-nv:
Thanks for the review, @jameslamb — you're right, this isn't needed.

We didn't see sccache explicitly enabled for these wheel builds (unlike the conda build, where `rapids-configure-sccache` usage is more visible/direct), so we assumed it wasn't enabled here and added preprocessor-cache mode to compensate for the fresh-path issue we expected. Your data shows `libcuopt` already gets good direct-mode hit rates via the existing `--no-build-isolation` setup, so this change doesn't add value there and we don't have evidence it's needed for the other wheels either.

Closing this PR.

PR #1827, opened 2026-08-28
Show mamba commands directly instead of a conda-then-swap note

ramakrishnap-nv:
## Summary
- `CONTRIBUTING.md` and `skills/cuopt-developer/SKILL.md` both recommended mamba as the preferred package manager, but printed `conda` commands with a "just replace `conda` with `mamba`" aside — so following the recommended path required manually substituting every command.
- Lead the code blocks with `mamba env create ...` directly; kept `conda activate` since mamba doesn't implement `activate`. The mamba-install link and the "swap in conda if mamba isn't installed" fallback note are both preserved.

## Test plan
- [x] `pre-commit run --files CONTRIBUTING.md skills/cuopt-developer/SKILL.md` passes
- [x] Grepped repo for any other "recommend mamba but show conda" occurrences — none found

### DIFF
--- a/skills/cuopt-developer/SKILL.md
+++ b/skills/cuopt-developer/SKILL.md
@@ -183,7 +183,7 @@ Skipping any of these surfaces as confusing runtime errors later. Run them in or
 2. **Create and activate the conda env** before *any* build, test, or `pre-commit` command — this is allowed and expected (see [Refusal Rules](#refusal-rules--read-first)). Use a **local prefix env** (`./.cuopt_env`) per [CONTRIBUTING.md](../../CONTRIBUTING.md), with the env file you picked in step 1 (`mamba` is recommended and faster; swap in `conda` if `mamba` isn't available):
    ```bash
    mamba env create -p ./.cuopt_env --file conda/environments/all_cuda-<ver>_arch-$(uname -m).yaml
-   mamba activate ./.cuopt_env   # or: conda activate ./.cuopt_env (works even without `mamba init`)
+   mamba activate ./.cuopt_env   # or: conda activate ./.cuopt_env
    ```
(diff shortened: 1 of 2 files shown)


skills/cuopt-developer/SKILL.md
```diff
@@ -180,9 +180,9 @@ cuopt/
...
-2. **Create and activate the conda env** before *any* build, test, or `pre-commit` command — this is allowed and expected (see [Refusal Rules](#refusal-rules--read-first)). Use a **local prefix env** (`./.cuopt_env`) per [CONTRIBUTING.md](../../CONTRIBUTING.md), with the env file you picked in step 1 (swap `conda`→`mamba` if available):
+2. **Create and activate the conda env** before *any* build, test, or `pre-commit` command — this is allowed and expected (see [Refusal Rules](#refusal-rules--read-first)). Use a **local prefix env** (`./.cuopt_env`) per [CONTRIBUTING.md](../../CONTRIBUTING.md), with the env file you picked in step 1 (`mamba` is recommended and faster; swap in `conda` if `mamba` isn't available — `activate` always uses `conda`, since `mamba` doesn't implement it):
```
mlubin: mamba (v 2.5.0) supports activate.

ramakrishnap-nv: Good catch, thanks — you're right, that was based on a local mamba 1.4.2 test without `mamba init` run first, not an actual limitation. Fixed in 73c840a9: dropped the incorrect claim, kept `conda activate` in the example since it works regardless of mamba version/init state, and noted `mamba activate` is fine too where supported.

mlubin (approved):

Iroy30 (approved):

PR #1829, opened 2026-08-28
Make MIP root relaxation honor LP method

hlinsen:
Make the MIP root relaxation honor CUOPT_METHOD instead of always using Concurrent. The default remains unchanged, and the configured method is propagated to both LP and MIP settings.
Improves ease of benchmarking the root solve.


chris-maes (approved):
LGTM. This is often requested by users. As long as we always run dual simplex, the user should be able to control the root solve with method.

PR #1833, opened 2026-08-31
Heuristic for arc-flow formulations

aliceb-nv:
This PR adds infrastructure for structural heuristics: heuristics that are specialized to certain problem formulation types in order to generate primal incumbents quickly by taking advantage of known literature on the formulation.

In particular, this PR includes a heuristic for emitting primal solutions for problems following an arc-flow formulation as outlined in the following paper: https://arxiv.org/pdf/1808.10661

This changes nets us +1 feasible on the MIPLIB2017 benchmark set, at a ~0% gap.

Broader benchmark results pending.

## Checklist

- Testing
   - [ ] Added tests
   - [ ] NA
- Documentation
   - [ ] Added new documentation
   - [ ] NA


cpp/src/mip_heuristics/structural/arc_flow.cu
```diff
@@ -0,0 +1,959 @@
...
+    if (is_known(model.slope[l])) {
+      ordered.push_back(l);
+    } else {
+      unidentified.push_back(l);
+    }
+  }
+
+  // Approximate equality is not transitive. Tolerance forms ratio classes before the final sort.
+  std::sort(ordered.begin(), ordered.end(), [&](int a, int b) {
+    const _Float128 lhs = (_Float128)model.slope[a] * (_Float128)model.displacement[b];
```
nguidotti, line 612: Is there a reason why you are using quad-precision here?

cpp/src/mip_heuristics/structural/arc_flow.cu
```diff
@@ -0,0 +1,959 @@
...
+
+struct arc_t {
+  int from{-1};
+  int to{-1};
+  int label{-1};
+  int col{-1};
+  double cost{0.0};
+};
+
+struct arc_flow_model_t {
```
nguidotti, line 48: Should the types here be templated, like the rest of the code?

cpp/src/mip_heuristics/structural/arc_flow.cu
```diff
@@ -0,0 +1,959 @@
...
+  {
+  }
+
+  std::vector<int64_t> col_entries;
+  std::vector<double> row_min_mag;
+  std::vector<double> row_max_mag;
+};
+
+template <typename i_t, typename f_t>
+struct host_problem_t {
```
nguidotti, line 111: I think `simplex::user_problem_t` has exactly the same variables. I would use that instead that creating another object for storing the problem.

cpp/src/mip_heuristics/structural/early_structural.cuh
```diff
@@ -0,0 +1,100 @@
...
+  std::unique_ptr<structural_heuristic_t<i_t, f_t>> active_;
+  std::atomic<bool> preemption_flag_{false};
+  bool task_launched_{false};
+};
+
+template <typename f_t>
+using structural_incumbent_callback_t =
+  std::function<void(const std::vector<f_t>& assignment, f_t objective)>;
+
+template <typename i_t, typename f_t>
```
nguidotti, line 76: What is the difference between "root structural" and "early structural"?

cpp/src/mip_heuristics/solve.cu
```diff
@@ -249,7 +251,13 @@ mip_solution_t<i_t, f_t> run_mip_solver(
     // optimization_problem_t). Its solver-space differs from both the first-pass FJ (original
     // problem) and B&B (post-trivial- presolve), so initial_upper_bound (user-space) is converted
     // via problem.get_solver_obj_from_user_obj.
+
```
nguidotti, line 254: I suggest grouping all the early heuristics into a single object with a simple interface (run, stop, etc.) instead of having everything inline in the `solve.cu`.

nguidotti: This can be done in a follow-up PR

nguidotti (commented):
Thanks for the hard work, Alice!

cpp/src/mip_heuristics/structural/arc_flow.cu
aliceb-nv: FMAs are relatively cheap to emulate in software for float128 and given that this isn't in the main solve loop, I thought it better to be safe than to emit wrong incumbents on numerically unscaled problems

cpp/src/mip_heuristics/structural/early_structural.cuh
aliceb-nv: Root runs at the root relaxation+cuts, early runs during the presolve passes. But yeah, I'll see if this can be unified into a single thing

cpp/src/mip_heuristics/structural/arc_flow.cu
aliceb-nv: user_problem_t is in CSC though, right? And in one-sided form
But I do agree duplication is not so great here. I will address this in my later CPUFJ PRs which faced the same sort of problem

cpp/src/mip_heuristics/structural/early_structural.cuh
aliceb-nv: Would be a bit tricky. I'll perform more simplification in later PRs

cpp/src/mip_heuristics/structural/arc_flow.cu
nguidotti: `user_problem_t` is in CSC, but it is in the range form

nguidotti: But you can get CSR quite easily (there are already routines for that)

aliceb-nv: I want to avoid an unnecessary transpose on instances where this structure is not matched (to keep latency to the first incumbent found low)

nguidotti (approved):
Thanks Alice!

ramakrishnap-nv (approved):

PR #1835, opened 2026-09-01
Tune Maven's own retry/concurrency to reduce Maven Central rate-limiting

ramakrishnap-nv:
## Summary

- `java-static-test` (from #1818) and, less often, `java-build` have hit `429 Too Many Requests` from Maven Central while resolving plugins like `maven-source-plugin` on a cold repository. `cuopt_mvn`'s retry loop (fixed in #1823 to actually run under `set -e`) retries the whole `mvn` invocation with backoff, but that's compensating for Maven's own resolver never being tuned for CI -- 4 attempts don't reliably outlast a sustained rate-limit window, and by the time the outer wrapper retries, the burst of parallel requests that likely triggered the 429 in the first place repeats.
- cuDF and cuVS already carry a fix for this exact problem in their own Java/Maven builds: a project-level `.mvn/maven.config`, auto-applied to every `mvn` invocation with no wrapper script needed, that caps concurrent downloads and adds a real backoff inside Maven's own transport-layer retry handler.

## Fix

Add `java/cuopt/.mvn/maven.config`, matching cuDF's (`java/.mvn/maven.config`) and cuVS's (`java/cuvs-java/.mvn/maven.config`) content exactly:

```
-e
-B
-Daether.connector.basic.downstreamThreads=1
-Daether.transport.http.retryHandler.count=5
-Daether.transport.http.retryHandler.interval=10000
-Dmaven.wagon.http.retryHandler.count=5
```

- `aether.connector.basic.downstreamThreads=1` caps Maven's own concurrent download threads, reducing the burst of parallel requests against Maven Central that likely triggers the rate-limiting in the first place.
- `aether.transport.http.retryHandler.interval=10000` adds a real 10s backoff inside Maven's own resolver, at the **transport** layer.

`cuopt_mvn`'s existing `-D` flags (`java/cuopt/scripts/maven.sh`) target the **connector**-layer retry handler (`aether.connector.http.retryHandler.*`), which recent Maven resolver versions may no longer consult now that retry logic lives at the transport layer -- this adds the layer that actually gets read. Verified via:

```
mvn org.apache.maven.plugins:maven-help-plugin:3.4.0:evaluate \
  -Dexpression=aether.transport.http.retryHandler.interval -q -DforceStdout
# -> 10000
```

`cuopt_mvn`'s outer shell-level retry loop is left in place as a second layer -- it's still useful for failures Maven's own retry can't cover (network drops mid-request, etc).

## Test plan

- [x] Confirmed the property resolves correctly via `maven-help-plugin:evaluate` (see above).
- [x] Ran the full `packaged-jar-tests` Maven Java suite locally with this config present; passes cleanly.
- [ ] `java-build` and `java-static-test` CI run clean without hitting Maven Central 429s.

Split out of #1818, where this was found while investigating an unrelated flaky Surefire crash -- this fix is independently useful for the existing `java-build` job today, not specific to that PR's self-contained classifier JAR work.


Iroy30 (approved):

java/cuopt/pom.xml
```diff
@@ -45,6 +45,63 @@ SPDX-License-Identifier: Apache-2.0
     <junit.jupiter.version>5.11.4</junit.jupiter.version>
   </properties>
 
+  <!--
+    Maven Central rate-limits NVIDIA-managed GitHub Actions runners: they egress through a small,
+    shared NAT'd IP range, so every RAPIDS repo's Java CI shares the same budget against Central
+    (see rapidsai/build-infra#370). Prefer the Google Cloud read-only mirror of Central, the same
+    one Apache ORC/Lucene/Spark use, and fall back to Central for anything not mirrored. Matches
+    the fix applied in rapidsai/kvikio#992.
+  -->
```
jameslamb, line 54:
```suggestion
  <!--
    Maven Central rate-limits NVIDIA-managed GitHub Actions runners: they egress through a small,
    shared NAT'd IP range, so every repo's Java CI shares the same budget against Central
    (see rapidsai/build-infra#370). Prefer the Google Cloud read-only mirror of Central, the same
    one Apache ORC/Lucene/Spark use, and fall back to Central for anything not mirrored. Matches
    the fix applied in rapidsai/kvikio#992.
  -->
```

If you want a comment documenting this that's fine, but it's not specific to "RAPIDS". This applies to every project using NVIDIA's hosted GitHub Actions runners.

jameslamb (approved):

PR #1837, opened 2026-09-01
Forward routing gRPC settings and map node name strings to ints

tmckayus:
This change fixes some gaps in the gRPC routing client:

1) string values for node name types in initial solutions were not turned into ints before sending across the wire
2) verbose_mode and error_logging settings flags were not being forwarded

### DIFF
--- a/python/cuopt/cuopt/grpc/client/grpc_client.pyx
+++ b/python/cuopt/cuopt/grpc/client/grpc_client.pyx
@@ -731,6 +731,52 @@ def _to_host(x):
     return np.asarray(x)
 
 
+_NODE_TYPE_NAMES = {
+    "Depot": 0,
+    "Pickup": 1,
+    "Delivery": 2,
+    "Break": 3,
+}
+
+
+def _is_node_type_name_array(values):
+    """True when values are node-type names rather than wire integers.
+
+    Integer and unsigned arrays (including uint8, the local DataModel storage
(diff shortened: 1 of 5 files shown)


python/cuopt/cuopt/grpc/client/grpc_client.pyx
```diff
@@ -965,6 +1013,42 @@ cdef _solution_to_py(cpu_routing_solution_t s):
     }
 
 
+# dump_best_results is intentionally not forwarded. The proto and C++ mapper
```
ramakrishnap-nv, line 1016: Lets create an issue in github and leave that here instead of adding comments, so we can follow-up later.

ramakrishnap-nv (approved):
Awesome work @tmckayus, have left a simple request, rest looks good

python/cuopt/cuopt/grpc/client/grpc_client.pyx
tmckayus: @ramakrishnap-nv I was trying to make developers aware that we've left out dump_best_results on purpose at this point, and it's not an oversight even though the proto technically supports it.  Should we leave the comment, what do you think?

ramakrishnap-nv: I was thinking if it is a follow-up task, may be it's better to also add an issue for this, and add it here. And we can also leave the comment for gist.

PR #1842, opened 2026-09-02
Fix O(n^2) loop in the flow cover cut generation

aliceb-nv:
Flow-cover generation rescanned every implied bound for each continuous-variable occurrence, causing effectively quadratic work on dense models. This change preprocesses bounds once per cut pass, caches shared a=0 candidates, groups direct candidates by controller, and safely rejects infeasible groups using monotonic alpha extrema.

Helps feasibilize `rd-rplusc` in 600s. We were previously stuck in the cut loop for the entire duration of the solve

## Checklist

- Testing
   - [ ] Added tests
   - [ ] NA
- Documentation
   - [ ] Added new documentation
   - [ ] NA


aliceb-nv:
@coderabbit review

cpp/src/cuts/cuts.cpp
```diff
@@ -1685,6 +1697,78 @@ flow_cover_generation_t<i_t, f_t>::flow_cover_generation_t(
...
+template <typename i_t, typename f_t>
+void flow_cover_generation_t<i_t, f_t>::preprocess_cut_pass(
+  const lp_problem_t<i_t, f_t>& lp,
+  const simplex_solver_settings_t<i_t, f_t>& settings,
+  const variable_bounds_t<i_t, f_t>& variable_bounds,
+  const std::vector<variable_type_t>& var_types,
+  const std::vector<f_t>& xstar)
+{
+  cut_pass_preprocessed = false;
+  cuopt_assert(var_types.size() >= (size_t)lp.num_cols, "");
```
rg20: It would be useful to have some message here

cpp/src/cuts/cuts.cpp
```diff
@@ -1685,6 +1697,78 @@ flow_cover_generation_t<i_t, f_t>::flow_cover_generation_t(
...
+    auto& preprocessed = use_upper_bound ? upper_implied_bounds : lower_implied_bounds;
+    cuopt_assert(offsets.size() >= (size_t)(lp.num_cols + 1), "");
+    cuopt_assert(variables.size() == weights.size(), "");
+    cuopt_assert(variables.size() == biases.size(), "");
+
+    preprocessed.bounds.resize(variables.size());
+    preprocessed.by_controller.clear();
+    preprocessed.by_controller.resize(lp.num_cols);
+    preprocessed.zero_candidate_cache.clear();
+    preprocessed.zero_candidate_cache.resize(lp.num_cols);
```
hlinsen: Ran some tests on large sparse models and allocation can become a bit expensive

Columns | Cold pass | Warm pass | RSS increase | Four-map storage
-- | -- | -- | -- | --
10K | 0.82 ms | 0.44 ms | noisy | 2.1 MiB
100K | 7.93 ms | 4.68 ms | noisy | 21.4 MiB
500K | 41.27 ms | 26.75 ms | 74 MiB | 106.8 MiB
1M | 83.53 ms | 64.65 ms | 204 MiB | 213.6 MiB
2M | 165.45 ms | 134.58 ms | 427 MiB | 427.2 MiB

I believe you can use a CSR like structure
```
struct group_t {
  i_t controller;
  i_t first_bound;
  i_t number_of_bounds;
  f_t minimum_alpha;
  f_t maximum_alpha;
};

std::vector<i_t> y_group_offsets;  // CSR-style offsets
std::vector<group_t> groups;       // only real (y, controller) groups
std::vector<i_t> bound_indices;    // contiguous bound IDs

std::vector<bound_state_t> state;  // active_bound/distance, refreshed per pass
```
The (y, controller) → bounds topology and alpha extrema are structural, so they could be built once per upper/lower side using a flat/CSR-style index, while only active_bound and distance are refreshed. The zero-candidate cache remains pass-dependent, but it could be sparse over only the queried (side, y, coefficient) keys instead of creating a map for every column.

hlinsen (approved):
thanks @aliceb-nv!
Made a control run that yields the same root gap closed ~8.50%. Performance and behavior is retained. Changes look good to me.
I added one comment.
I wouldn't block the PR on that it seems to affect only very large models but I left it here for referemce.

cpp/src/cuts/cuts.cpp
aliceb-nv: Addressed, thank you :)

nguidotti:
@aliceb-nv 
What is blocking for merging this PR?

aliceb-nv:
This is pending the merge of https://github.com/NVIDIA/cuopt/pull/1889. I will make this a P1

PR #1844, opened 2026-09-03
Avoid integer overflow with infinite time limit in CPU Feasibility Jump

vitor1001:
When in_time_limit is infinity, multiplying by 1000 and casting to an integer causes undefined behavior / integer overflow (trapping under UBSan). Guard against infinity and use std::chrono::milliseconds::max().

## Checklist

- Testing
   - [ ] Added tests
   - [ ] NA
- Documentation
   - [ ] Added new documentation
   - [ ] NA

### DIFF
--- a/cpp/src/mip_heuristics/feasibility_jump/fj_cpu.cu
+++ b/cpp/src/mip_heuristics/feasibility_jump/fj_cpu.cu
@@ -4,6 +4,8 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 /* clang-format on */
+ 
+#include <cuopt/error.hpp>
 
 #include <mip_heuristics/mip_constants.hpp>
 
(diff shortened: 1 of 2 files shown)


sylvesterkaczmarek (commented):
The new guard only special-cases positive infinity via `in_time_limit < +infinity`. `-infinity` still takes the `floor(... * 1000)` / integer-cast path and hits the same out-of-range conversion this PR is trying to remove, while NaN falls into the `milliseconds::max()` branch and is silently treated as unlimited. Could this use an explicit finiteness/non-finite policy (and validate NaN/negative infinity) with regressions for those inputs?

vitor1001:
Thanks for the quick review, did the suggested changes.

sylvesterkaczmarek:
Thanks for the quick fix. The NaN/negative-infinity policy I raised is addressed now with explicit validation and regression coverage. There is still the separate exact-boundary timing issue already flagged in review (`elapsed > time_limit` versus `>=`), so I would close that before treating the timing edge cases as fully done.

vitor1001:
> Thanks for the quick fix. The NaN/negative-infinity policy I raised is addressed now with explicit validation and regression coverage. There is still the separate exact-boundary timing issue already flagged in review (`elapsed > time_limit` versus `>=`), so I would close that before treating the timing edge cases as fully done.

Agreed, done!

sylvesterkaczmarek (commented):
Re-reviewed current `16aca33`. The non-finite policy is explicit now, oversized finite limits avoid the integer-conversion overflow, and the loop uses `elapsed >= time_limit`, which closes the exact-boundary/zero-limit case I had left open. The added corner-case coverage also pins NaN, negative infinity, positive infinity and oversized finite inputs. No remaining blocker from my earlier timing findings.

PR #1845, opened 2026-09-03
Guard against nullptr dereference when checking concurrent solver status

vitor1001:
If dual simplex or barrier did not set their solution pointers (e.g. on early termination, error, or limit), dereferencing sol_dual_simplex_ptr or sol_barrier_ptr directly causes a null pointer dereference.

## Checklist

- Testing
   - [ ] Added tests
   - [ ] NA
- Documentation
   - [ ] Added new documentation
   - [ ] NA

### DIFF
--- a/cpp/tests/linear_programming/pdlp_test.cu
+++ b/cpp/tests/linear_programming/pdlp_test.cu
@@ -56,6 +56,10 @@
 #include <utility>
 #include <vector>
 
+#ifdef _OPENMP
+#include <omp.h>
+#endif
+
 namespace cuopt::mathematical_optimization::test {
 
 constexpr double afiro_primal_objective = -464.0;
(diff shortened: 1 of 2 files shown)


sylvesterkaczmarek (commented):
`sol_dual_simplex_ptr` is still unconditionally dereferenced in the later `PDLP ConcurrentLimit` branch. If dual simplex produced no result, the new status guard maps it to `CONCURRENT_LIMIT`, but this branch then does `*sol_dual_simplex_ptr` anyway and can still crash. Please guard/fallback there too and add a null-result `ConcurrentLimit` regression.

vitor1001:
Great points, done!

sylvesterkaczmarek:
Thanks, appreciate the quick turnaround. The null ConcurrentLimit dereference I flagged is fixed, and the new regression exercises the no-dual-simplex path. I see the separate RAII test-cleanup comment as well; that's independent of my concern.

cpp/tests/linear_programming/pdlp_test.cu
```diff
@@ -56,6 +56,10 @@
 #include <utility>
 #include <vector>
 
+#ifdef _OPENMP
```
hlinsen: OpenMP is a hard requirement you can skip the macro check

cpp/tests/linear_programming/pdlp_test.cu
```diff
@@ -174,6 +178,59 @@ TEST(pdlp_class, concurrent_pdlp_exception_joins_worker_threads)
...
+
+  auto settings       = pdlp_solver_settings_t<int, double>{};
+  settings.method     = cuopt::mathematical_optimization::method_t::Concurrent;
+  settings.presolver  = cuopt::mathematical_optimization::presolver_t::None;
+  settings.inside_mip = true;
+
+  // inside_mip skips dual simplex. Setting threads to 1 ensures barrier is also disabled
+  // (< CUOPT_CONCURRENT_LP_BARRIER_REQUIRED_THREAD_COUNT), leaving both sol_dual_simplex_ptr
+  // and sol_barrier_ptr null.
+#ifdef _OPENMP
```
hlinsen: same here

cpp/tests/linear_programming/pdlp_test.cu
```diff
@@ -174,6 +178,59 @@ TEST(pdlp_class, concurrent_pdlp_exception_joins_worker_threads)
...
+  const int prev_threads = omp_get_max_threads();
+  omp_set_num_threads(1);
+#endif
```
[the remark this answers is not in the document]
hlinsen: Other exceptions will cause the test to fail and CI to fail.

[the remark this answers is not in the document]
hlinsen: @vitor1001 This is only an edge case, but we already have `cuopt::scope_guard` to make the restoration exception safe. Could you use it here to restore the OpenMP setting? Thanks for the PR!
```
#include <utilities/scope_guard.hpp>

const int prev_threads = omp_get_max_threads();
omp_set_num_threads(1);
const cuopt::scope_guard restore_threads{
  [prev_threads] { omp_set_num_threads(prev_threads); }};
```

cpp/tests/linear_programming/pdlp_test.cu
vitor1001: Agreed, done!

cpp/tests/linear_programming/pdlp_test.cu
vitor1001: Agreed, done!

cpp/tests/linear_programming/pdlp_test.cu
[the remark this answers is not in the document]
vitor1001: Good catch, done as suggested.

hlinsen (approved):
Thanks @vitor1001!

PR #1847, opened 2026-09-03
Provide std::mutex fallback for omp_mutex_t when OpenMP is disabled

vitor1001:
When compiling without OpenMP, provide standard library std::mutex fallback wrapper functions to allow compiling omp_mutex_t locks.

## Checklist

- Testing
   - [ ] Added tests
   - [ ] NA
- Documentation
   - [ ] Added new documentation
   - [ ] NA

### DIFF
--- a/cpp/src/utilities/omp_helpers.hpp
+++ b/cpp/src/utilities/omp_helpers.hpp
@@ -24,12 +24,46 @@ std::pair<i_t, i_t> calculate_index_range(i_t k, double total, double n)
 }  // namespace cuopt
 
 #ifdef _OPENMP
-
 #include <omp.h>
+#else
+#include <mutex>
+#endif
 #include <memory>
 #include <utility>
 
 namespace cuopt {
+
+#ifndef _OPENMP
+class omp_mutex_t {
+ public:
+  omp_mutex_t() : mutex(new std::mutex()) { }
+
+  omp_mutex_t(const omp_mutex_t&) = delete;
+
+  omp_mutex_t(omp_mutex_t&& other) { *this = std::move(other); }
+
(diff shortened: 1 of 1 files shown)


nguidotti:
OpenMP is required for running the solver.

vitor1001:
Hi @nguidotti  and thanks for the review!

We cannot use OpenMP on our internal monorepo at Google. cuOpt seem to build and run fine without OpenMP, as long as I apply this PR.

This file is the single place in the whole codebase that really assumes that the system supports OpenMP, so it's a relatively low-touch change. For me, personally, it would save me from re-applying this patch every time I sync to a new cuOpt release.

nguidotti:
Hello @vitor1001,

Can you share which solver are you using? There are a few parts of the library that requires OpenMP to work correctly (concurrent mode for LP, the entire MIP solver). Also, there is some operations that is not supported by `std::mutex` ilke `fetch_max/min` for floats/doubles.

nguidotti:
/ok to test [6fdfe4f](https://github.com/NVIDIA/cuopt/pull/1847/commits/6fdfe4ff4c2697c77528d66546342d00d427c24a)

vitor1001:
We don't use it in production, but I like cuOpt to be available for anyone that wants to experiment and/or benchmark with it. We had users that wanted to try both QP and MIP. MIP unit tests seem to pass without OpenMP.

nguidotti:
> We don't use it in production, but I like cuOpt to be available for anyone that wants to experiment and/or benchmark with it. We had users that wanted to try both QP and MIP. MIP unit tests seem to pass without OpenMP.

I am surprised that the MIP unit tests passes without OpenMP since we use it quite extensively within the solver for parallelism. Even if it works, the performance will be quite bad as it was not written for single-threaded execution

PR #1848, opened 2026-09-03
move server LPData transform into a standalone function

tmckayus:
This is one of a series of changes to refactor the cuopt http server so that a new proxy server can be added that shares the same data conversion code but delegates solves to the gRPC server. It will run as a sidecar.

Here, we are moving an LPData transform function out of a class into a standalone function. It converts lists to numpy arrays and "inf" and "ninf" to numpy infinity values, etc.

### DIFF
--- a/python/cuopt_server/cuopt_server/tests/test_lp_data_transformation.py
+++ b/python/cuopt_server/cuopt_server/tests/test_lp_data_transformation.py
@@ -1,0 +1,191 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+import copy
+
+import numpy as np
+
+from cuopt_server.utils.linear_programming.data_definition import LPData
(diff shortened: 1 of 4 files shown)


python/cuopt_server/cuopt_server/tests/test_lp_data_transformation.py
```diff
@@ -0,0 +1,191 @@
...
+
+import numpy as np
+
+from cuopt_server.utils.linear_programming.data_definition import LPData
+from cuopt_server.utils.linear_programming.data_transformation import (
+    transform_lp_data,
+)
+
+
+def _sample_lp_dict():
```
ramakrishnap-nv, line 14: Are we planning to delete this once we have merged the all the points and have end to end testing ?

ramakrishnap-nv (approved):
Minor comment, but rest looks good.

python/cuopt_server/cuopt_server/tests/test_lp_data_transformation.py
tmckayus: yes, we can prune the extra tests. This code is being used by the current server as well so in that sense there are end to end tests already.

PR #1855, opened 2026-09-04
Routing: Guard diagonal element lookup against negative offsets in top_k

vitor1001:
When row_id < col_offset, row_id - col_offset is negative. In C++, integer division with negative operands truncates toward zero (e.g. -1 / 16 == 0), causing thread 0 to mistakenly match threadIdx.x == 0 and index sort_cost[-1] or load_cost[-1].

Guard the diagonal replacement with `if (row_id >= col_offset)` so that negative offsets are not evaluated.

## Checklist

- Testing
   - [ ] Added tests
   - [ ] NA
- Documentation
   - [ ] Added new documentation
   - [ ] NA


akifcorduk (approved):
Thank you for the fix!

PR #1856, opened 2026-09-04
[DON'T MERGE] Bug fix: barrier determinism

yuwenchen95:
## Description

Fixes run-to-run nondeterminism in the barrier solver when `--cudss-deterministic 1` is set.

`get_spmv_alg` branches to `CUSPARSE_SPMV_CSR_ALG1` for cuSPARSE < 13.0, which is the path taken
on CUDA 12.9. The comment justified this with "ALG1 uses a deterministic row-split algorithm …
ALG1 is safe for reproducibility", but the cuSPARSE docs say the opposite:

| Algorithm | Documented behavior |
|---|---|
| `CUSPARSE_SPMV_CSR_ALG1` | "may produce slightly different results during different runs with the same input parameters" |
| `CUSPARSE_SPMV_CSR_ALG2` | "provides deterministic (bit-wise) results for each run" |

So on CUDA 12.x the solver was always using the non-deterministic algorithm. Measured on
cuSPARSE 12.5.10: ALG1 differed on 199/199 repeats (5 of 7990 entries, 1 ULP); ALG2 was
bit-identical over 200 repeats and across processes.

Select ALG2 when determinism is requested, and thread the flag through `cusparse_view_t` so both
`spmv()` and `transpose_spmv()` use it. The default path is unchanged.


chris-maes (approved):
LGTM. Thanks for the fix.

PR #1857, opened 2026-09-04
change Python grpc client so wait loop happens in Python

tmckayus:
If we use the C++ wait_result() method, the GIL is locked and no other threads run.  Run the wait loop in Python and call only the Cython status method to determine if the job has completed.

### DIFF
--- a/python/cuopt/cuopt/grpc/client/grpc_client.pyx
+++ b/python/cuopt/cuopt/grpc/client/grpc_client.pyx
@@ -93,6 +93,38 @@ class JobNotReadyError(GrpcError):
     pass
 
 
+# Matches the previous C++ wait() poll cadence. Keep this in Python
+# (time.sleep) so the GIL is released between short status RPCs.
+_WAIT_POLL_INTERVAL_S = 1.0
+
+
+def _wait_poll_loop(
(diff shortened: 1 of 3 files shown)


python/cuopt/cuopt/grpc/client/grpc_client.pyx
```diff
@@ -315,19 +352,25 @@ cdef class Client:
...
+        :meth:`start_log_stream` threads can therefore make progress during
+        the wait. Each :meth:`status` call is a unary RPC with a separate
+        60-second hang deadline; a hung poll can therefore outlast a short
+        ``timeout``.
         """
-        cdef int timeout_seconds = 0 if timeout is None else int(timeout)
-        cdef grpc_status_result_t wait_result = self._client.get().wait(
-            job_id.encode("utf-8"), timeout_seconds
+        timeout_seconds = 0 if timeout is None else int(timeout)
+        return _wait_poll_loop(
```
ramakrishnap-nv, line 371: Just had a question, we would still need wait from C++ side for C++ client, am I correct ?

ramakrishnap-nv (approved):
Just a question, rest looks good.

PR #1864, opened 2026-09-08
Improvement for qcqp canonicalization and symmetric check

yuwenchen95:
## Description

Quadratic-constraint COO canonicalization sized its scratch by the **problem dimension** rather than
by the constraint being processed: `marker` was `max_col - min_idx + 1`, so a ten-entry SOC
constraint in a 200k-variable model allocated and zeroed a 200k-element array. Separately,
`check_symmetric_offdiagonal_pairs` paid two comparison sorts whose comparators gather through the
COO arrays, which dominated on large dense blocks.

**Index compression.** Both entry points now reduce a constraint's indices to dense ranks in
`[0, buckets)` and order entries with two stable counting passes instead of comparison sorts. A cost
model picks between shifting by the minimum index and ranking via sort + unique. Canonicalization per quadratic constraint is Θ(min(nnz + range, nnz·log nnz)) in both time and memory, where range is the constraint's index span — so linear whenever indices are compact, and bounded by nnz log nnz otherwise. Previously it was Θ(nnz + dim), unbounded in the problem dimension.

**Scratch reuse.** A `coo_canonicalization_scratch_t` is declared once per loop and shared across
constraints, so models with many small quadratic constraints don't pay a fresh set of allocations
each time. Applied to every batch path: fast MPS parser, `canonicalize_quadratic_constraints`, MPS
writer, both gRPC decoders, and the reference parser's `flush_qcmatrix_block`.

**Dead pre-sort removed.** `finalize_qcmatrix_constraints` sorted each QCMATRIX block before calling
two functions that each derive their own ordering.

### Canonicalization workloads

Median of 9 samples, gcc 14.4 `-O3`. "before" is the algorithm on `main`.

| workload | entries | before | after | |
|---|---:|---:|---:|---|
| uniform sparse | 1,000,000 | 32.89 ms | **25.10 ms** | 1.31x |
| duplicate heavy | 1,000,000 | 18.37 ms | **16.29 ms** | 1.13x |
| few dense rows | 1,000,000 | 60.28 ms | **22.61 ms** | 2.67x |
| fully dense shuffled | 1,000,000 | 33.40 ms | **15.77 ms** | 2.12x |
| wide index range | 100,000 | 8.47 ms | 30.22 ms | **0.28x** |
| 1000x1000, 3 entries (x100k) | 3 | 1371 ns | **179 ns** | 7.66x |

The last row is per-matrix and is where scratch reuse shows: 344 ns with a fresh scratch per call,
179 ns with one shared. The ms rows are each a single large canonicalization, so reuse has nothing
to amortize there.

### Symmetric QCMATRIX blocks, end to end

Both functions, as the MPS path calls them.

| block | before | after | |
|---|---:|---:|---|
| dense 1000x1000 (nnz=1M) | 129.6 ms | **42.6 ms** | 3.04x |
| dense 2000x2000 (nnz=4M) | 623.9 ms | **196.0 ms** | 3.18x |
| dense 5000x5000 (nnz=25M) | 4670.7 ms | **1436.6 ms** | 3.25x |
| scattered, dim=200k, nnz=10 | 189.3 us | **0.93 us** | 204x |
| scattered, dim=200k, nnz=100 | 346.2 us | **7.50 us** | 46x |

Split for the 5000x5000 case: `check_symmetric_offdiagonal_pairs` 4127.5 -> 715.8 ms (5.8x),
`canonicalize_coo_matrix` 543.2 -> 720.8 ms (0.75x). Canonicalization is slower on a *squarish*
dense block, where two counting passes replace per-row sorts that stayed in cache; it wins where
rows are fewer and wider ("few dense rows" above).


cpp/src/io/mps_parser.cpp
```diff
@@ -156,6 +157,124 @@ void triples_to_csr_flat(const coo_entries_t<i_t, f_t>& entries,
   out_offsets = std::move(scratch.row_off);
 }
 
+// One stable counting pass of an LSD radix sort: reorders `src` by `key[entry]`, which must lie
+// in [0, buckets). O(nnz + buckets). Stability is load-bearing -- successive passes rely on it to
```
mlubin: Document what the function does, not what successive passes do and what's "load-bearing". The relationship between src and dst isn't explained.

cpp/src/io/mps_parser.cpp
```diff
@@ -156,6 +157,124 @@ void triples_to_csr_flat(const coo_entries_t<i_t, f_t>& entries,
...
+  if (v <= 1) { return 1; }
+  size_t bits = 0;
+  for (size_t x = v - 1; x != 0; x >>= 1) {
+    ++bits;
+  }
+  return bits;
+}
+
+// Reduce every entry to its canonical upper-triangle coordinates -- row = min(r,c), col = max(r,c)
+// -- as dense ranks in [0, buckets); returns the bucket count. Everything downstream is then sized
```
mlubin: I don't follow what the role of the buckets is. Is it the number of unique coordinates? If so, why not say so.

yuwenchen95: Rename `buckets` to `key_range`.

cpp/src/io/mps_parser.cpp
yuwenchen95: Rename `src`and `dst` to `order` and `scratch`. `scratch` is just the working space for `order`.

yuwenchen95 (commented):
Cleaned up comment and renamed some parameters.

cpp/src/grpc/codegen/generate_conversions.py
```diff
@@ -2209,6 +2209,12 @@ def _gen_proto_to_problem(registry, indent="  "):
         lines.append(f"{ind}if (pb_problem.{name}_size() > 0) {{")
         lines.append(f"{ind}  std::vector<{cpp_inner}> _entries;")
         lines.append(f"{ind}  _entries.reserve(pb_problem.{name}_size());")
+        if name == "quadratic_constraints":
```
rg20: @ramakrishnap-nv can you review this?

cpp/src/io/mps_data_model.cpp
```diff
@@ -464,8 +465,11 @@ template <typename i_t, typename f_t>
 void canonicalize_quadratic_constraints(
   std::vector<typename mps_data_model_t<i_t, f_t>::quadratic_constraint_t>& constraints)
 {
+  // One scratch for the whole batch, so a model with many small quadratic constraints does not pay
```
mlubin: Unnecessary comment

cpp/src/io/mps_parser.cpp
```diff
@@ -156,6 +157,134 @@ void triples_to_csr_flat(const coo_entries_t<i_t, f_t>& entries,
   out_offsets = std::move(scratch.row_off);
 }
 
+/**
+ * @brief One stable counting pass of an LSD radix sort: sorts the entry indices in `order` by
+ * `key[entry]`, which must lie in [0, key_range).
+ *
+ * @param scratch Ping-pong buffer, sized like `order`.
```
mlubin: What's a ping-pong buffer?

cpp/src/io/mps_parser.cpp
```diff
@@ -156,6 +157,134 @@ void triples_to_csr_flat(const coo_entries_t<i_t, f_t>& entries,
...
+  size_t bits = 0;
+  for (size_t x = v - 1; x != 0; x >>= 1) {
+    ++bits;
+  }
+  return bits;
+}
+
+/**
+ * @brief Maps each entry to its canonical upper-triangle coordinate -- row = min(r, c),
+ * col = max(r, c) -- with both endpoints replaced by an integer rank.
```
mlubin: what does endpoint mean here?

mlubin (approved):

cpp/src/io/mps_parser.cpp
yuwenchen95: It's just an idiom for double-buffering. Renamed it to `second buffer`.

cpp/src/io/mps_parser.cpp
yuwenchen95: They are `row_rank` and `col_rank`.

PR #1866, opened 2026-09-08
Clarify factorization issue causes residual error

yuwenchen95:
## Description

Hide residual info into debug mode, avoid show it in normal barrier run.


rg20 (approved):

chris-maes (changes requested):
let's handle this differently.

chris-maes (approved):
LGTM.

PR #1868, opened 2026-09-08
wire MIP initial solutions into gRPC

tmckayus:
MIP initial solutions were defined for gRPC but never wired into the data transfer. This change corrects that oversight.

### DIFF
--- a/cpp/include/cuopt/mathematical_optimization/cpu_optimization_problem.hpp
+++ b/cpp/include/cuopt/mathematical_optimization/cpu_optimization_problem.hpp
@@ -84,6 +84,8 @@ class cpu_optimization_problem_t : public optimization_problem_interface_t<i_t, 
   void set_problem_name(const std::string& problem_name) override;
   void set_variable_names(const std::vector<std::string>& variable_names) override;
   void set_row_names(const std::vector<std::string>& row_names) override;
+  void set_initial_primal_solution(const f_t* initial_primal_solution, i_t size);
+  void set_initial_dual_solution(const f_t* initial_dual_solution, i_t size);
 
   /**
    * @brief Transfer parsed MPS/QPS storage into this CPU problem without copying array/string data.
@@ -166,6 +168,8 @@ class cpu_optimization_problem_t : public optimization_problem_interface_t<i_t, 
   std::vector<f_t> get_constraint_upper_bounds_host() const override;
   std::vector<char> get_row_types_host() const override;
   std::vector<var_t> get_variable_types_host() const override;
+  std::vector<f_t> get_initial_primal_solution_host() const;
+  std::vector<f_t> get_initial_dual_solution_host() const;
 
   /**
    * @brief Write the optimization problem to an MPS file.
@@ -229,6 +233,8 @@ class cpu_optimization_problem_t : public optimization_problem_interface_t<i_t, 
   std::vector<f_t> constraint_upper_bounds_;
   std::vector<char> row_types_;
   std::vector<var_t> variable_types_;
+  std::vector<f_t> initial_primal_solution_;
+  std::vector<f_t> initial_dual_solution_;
 
   std::string objective_name_;
   std::string problem_name_;
--- a/cpp/include/cuopt/mathematical_optimization/cpu_optimization_problem.hpp
+++ b/cpp/include/cuopt/mathematical_optimization/cpu_optimization_problem.hpp
@@ -84,7 +84,25 @@ class cpu_optimization_problem_t : public optimization_problem_interface_t<i_t, 
   void set_problem_name(const std::string& problem_name) override;
   void set_variable_names(const std::vector<std::string>& variable_names) override;
   void set_row_names(const std::vector<std::string>& row_names) override;
+  /**
+   * @brief Copy an initial primal solution into host storage.
+   *
+   * Optional. Size and finiteness are not checked here; the solver validates
+   * those when the start is applied, matching local solve.
+   *
+   * @param[in] initial_primal_solution Host pointer; must be non-null when size != 0.
+   * @param[in] size Number of values to copy.
+   */
   void set_initial_primal_solution(const f_t* initial_primal_solution, i_t size);
+  /**
+   * @brief Copy an initial dual solution into host storage.
+   *
+   * Optional. Size and finiteness are not checked here; the solver validates
+   * those when the start is applied, matching local solve.
+   *
+   * @param[in] initial_dual_solution Host pointer; must be non-null when size != 0.
+   * @param[in] size Number of values to copy.
+   */
   void set_initial_dual_solution(const f_t* initial_dual_solution, i_t size);
 
   /**
@@ -168,7 +186,13 @@ class cpu_optimization_problem_t : public optimization_problem_interface_t<i_t, 
   std::vector<f_t> get_constraint_upper_bounds_host() const override;
   std::vector<char> get_row_types_host() const override;
   std::vector<var_t> get_variable_types_host() const override;
+  /**
+   * @brief Return a copy of the stored initial primal solution (empty if unset).
+   */
   std::vector<f_t> get_initial_primal_solution_host() const;
+  /**
+   * @brief Return a copy of the stored initial dual solution (empty if unset).
+   */
   std::vector<f_t> get_initial_dual_solution_host() const;
 
   /**
--- a/cpp/include/cuopt/mathematical_optimization/optimization_problem_utils.hpp
+++ b/cpp/include/cuopt/mathematical_optimization/optimization_problem_utils.hpp
@@ -39,6 +39,12 @@ inline constexpr char var_type_to_char(var_t variable_type)
   return 'C';
 }
 
+/**
+ * @brief Copy optional initial primal/dual arrays onto a CPU problem.
+ *
+ * No-op when both spans are empty, or when @p problem is not a
+ * cpu_optimization_problem_t (GPU problems do not store these host arrays).
+ */
 template <typename i_t, typename f_t>
 void copy_initial_solutions_to_cpu_problem(optimization_problem_interface_t<i_t, f_t>* problem,
                                            std::span<const f_t> primal,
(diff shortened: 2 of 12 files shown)


ramakrishnap-nv (approved):

python/cuopt/cuopt/tests/linear_programming/test_grpc_client.py
```diff
@@ -185,6 +185,47 @@ def test_mip_submit_and_result(self, grpc_server):
         assert solution.get_primal_objective() == pytest.approx(15.0, rel=1e-3)
         client.delete(job_id)
 
+    def test_mip_start_over_grpc(self, grpc_server):
```
ramakrishnap-nv, line 188: Neither of the two new tests can fail if the feature is reverted.

Here the MIP start is `x=5, y=5`, which is already the unique optimum: maximizing `x + 2y` under `x + y <= 10` and `x >= y` forces `y <= 5`, so the solver reaches `15.0` with no warm start at all. The assertion is the same `15.0` that `test_mip_submit_and_result` already asserts.

`test_lp_initial_solution_over_grpc` below has the same problem — an all-zero primal/dual start, asserting the same `0.36` as the existing demo-LP test.

I checked this concretely: revert the `skip_conversion` removal in `field_registry.yaml` (i.e. remove the feature) and both tests stay green. The C++ mapper round-trip tests do cover the wire format, so this is a gap in end-to-end coverage rather than no coverage — but something should assert the start actually reached the solver. Options: assert on the solver log / incumbent count, or pick a start that changes an observable (e.g. a feasible-but-suboptimal start on a problem where the log shows it as the first incumbent).

cpp/include/cuopt/mathematical_optimization/optimization_problem_utils.hpp
```diff
@@ -36,6 +39,61 @@ inline constexpr char var_type_to_char(var_t variable_type)
...
+ *
+ * No-op when both spans are empty, or when @p problem is not a
+ * cpu_optimization_problem_t (GPU problems do not store these host arrays).
+ */
+template <typename i_t, typename f_t>
+void copy_initial_solutions_to_cpu_problem(optimization_problem_interface_t<i_t, f_t>* problem,
+                                           std::span<const f_t> primal,
+                                           std::span<const f_t> dual)
+{
+  if (primal.empty() && dual.empty()) { return; }
```
ramakrishnap-nv, line 53: This early-out plus the per-array `if (!primal.empty())` / `if (!dual.empty())` guards below mean an empty array is a no-op rather than a reset, so this function can never *clear* a previously-set initial solution.

Two consequences, both latent today since every current caller passes a freshly constructed problem:

- Python's `_refresh_data_model_values` (`problem.py:1843`) calls `set_initial_primal_solution(np.array([]))` specifically to clear a prior warm start. If such a data model is ever populated onto a reused `cpu_optimization_problem_t`, the cleared start survives and gets silently re-sent.
- Asymmetric within a single call: non-empty primal + empty dual leaves a stale dual in place.

Worth either making empty mean "clear", or documenting that the function is populate-only and requires a fresh problem.

cpp/src/grpc/codegen/generated/generated_proto_to_problem.inc
```diff
@@ -74,6 +74,14 @@
     }
     cpu_problem.set_variable_types(variable_types.data(), static_cast<i_t>(variable_types.size()));
   }
+  {
+    std::vector<f_t> initial_primal_solution(pb_problem.initial_primal_solution().begin(), pb_problem.initial_primal_solution().end());
+    cpu_problem.set_initial_primal_solution(initial_primal_solution.data(), static_cast<i_t>(initial_primal_solution.size()));
```
ramakrishnap-nv, line 79: Nit / consistency: the unary path sets unconditionally, while the chunked generator emits an `if (!initial_primal_solution.empty())` guard for the same field (see `generated_chunked_arrays_to_problem.inc`).

No functional difference on a fresh problem — `cpu_optimization_problem_t::set_initial_primal_solution` explicitly guards its null check with `if (size != 0)`, so the empty case is handled — but it does reach `std::copy(p, p, dst)` with a `p` that may be null for an empty vector. Since these files are generated, the fix belongs in `generate_conversions.py`: emit the same emptiness guard on both paths.

ramakrishnap-nv (commented):
Review of the gRPC initial-solutions wiring. I regenerated the codegen from the modified `field_registry.yaml` and diffed against `cpp/src/grpc/codegen/generated/` — byte-identical, so the generated files are in sync. Wire plumbing is consistent across the unary, chunked, size-estimate and element-size paths, and the client path reaches the new copy.

Four comments below, none of them blocking. The main one is the test coverage: both new Python tests still pass if the feature is reverted.

One item that has no diff line to attach to — `cpp/include/cuopt/mathematical_optimization/solve_remote.hpp`: initial solutions are now serialized from `cpu_optimization_problem_t`, but `mip_solver_settings_t::initial_solutions` and `pdlp_solver_settings_t::initial_primal_solution_` have no `field_registry.yaml` entry. A C++ caller doing `settings.add_initial_solution(...); solve_mip_remote(problem, settings);` gets a warm start that works locally and is silently dropped remotely. Those members are `device_uvector`, so serializing them may well be out of scope for the CPU remote path — but a warning or rejection when they are set on a remote solve would beat a silent drop.

tmckayus:
@ramakrishnap-nv thanks, will follow up on feedback in additional PRs since none are blocking.

cpp/include/cuopt/mathematical_optimization/optimization_problem_utils.hpp
```diff
@@ -36,6 +39,61 @@ inline constexpr char var_type_to_char(var_t variable_type)
...
+  auto* cpu = dynamic_cast<cpu_optimization_problem_t<i_t, f_t>*>(problem);
```
rg20, line 54:
```suggestion
  auto* cpu_problem = dynamic_cast<cpu_optimization_problem_t<i_t, f_t>*>(problem);
```

cpp/include/cuopt/mathematical_optimization/cpu_optimization_problem.hpp
```diff
@@ -84,6 +84,26 @@ class cpu_optimization_problem_t : public optimization_problem_interface_t<i_t,
...
+  /**
+   * @brief Copy an initial primal solution into host storage.
+   *
+   * Optional. Size and finiteness are not checked here; the solver validates
+   * those when the start is applied, matching local solve.
+   *
+   * @param[in] initial_primal_solution Host pointer; must be non-null when size != 0.
+   * @param[in] size Number of values to copy.
+   */
+  void set_initial_primal_solution(const f_t* initial_primal_solution, i_t size);
```
mlubin, line 96: Why take raw pointers and sizes instead of spans? Same below.

cpp/include/cuopt/mathematical_optimization/optimization_problem_utils.hpp
tmckayus: changed the name to cpu_problem (includes additional following lines)

cpp/include/cuopt/mathematical_optimization/cpu_optimization_problem.hpp
tmckayus: changed to spans

python/cuopt/cuopt/tests/linear_programming/test_grpc_client.py
mlubin: Yeah it's generally hard to test if starting points are provided to the solver. For LP you could provide the optimal solution and test that the iteration count is ~zero. For MIP I think checking the log would be sufficient.

cpp/include/cuopt/mathematical_optimization/optimization_problem_utils.hpp
```diff
@@ -36,6 +39,59 @@ inline constexpr char var_type_to_char(var_t variable_type)
...
+ * Same contract as local solve: empty means unset; size/finiteness are left to
+ * add_initial_solution / problem_checking_t.
+ */
+template <typename i_t, typename f_t>
+void apply_initial_solutions_to_mip_settings(const cpu_optimization_problem_t<i_t, f_t>& problem,
+                                             mip_solver_settings_t<i_t, f_t>& settings)
+{
+  const auto primal = problem.get_initial_primal_solution_host();
+  if (!primal.empty()) {
+    settings.add_initial_solution(primal.data(), static_cast<i_t>(primal.size()));
```
rg20, line 72: May be out of scope, but we should move initial solutions to problem structure instead of doing this part of settings.

rg20 (approved):
Approving with minor comments.

cpp/include/cuopt/mathematical_optimization/optimization_problem_utils.hpp
tmckayus: we can add a follow up for this. As I understand it, this is a solver interface change.

python/cuopt/cuopt/tests/linear_programming/test_grpc_client.py
tmckayus: added a log line and test for MIP and an iteration count test for LP

python/cuopt/cuopt/tests/linear_programming/test_grpc_client.py
```diff
@@ -335,6 +340,67 @@ def test_mip_submit_and_result(self, grpc_server):
...
+        settings = SolverSettings()
+        settings.set_parameter(CUOPT_METHOD, SolverMethod.PDLP)
+        settings.set_parameter(CUOPT_PRESOLVE, 0)
+
+        client = Client("localhost", grpc_server)
+
+        def solve(problem):
+            job_id = client.submit(problem, settings)
+            try:
+                assert client.wait(job_id, timeout=120) == JobStatus.COMPLETED
```
mlubin: A toy LP shouldn't take 120 seconds, we should be able to use tighter time limits.

mlubin (approved):

python/cuopt/cuopt/tests/linear_programming/test_grpc_client.py
tmckayus: good catch, that was a cut and paste from client timeout tests. Set to 30s

PR #1871, opened 2026-09-09
Implement automatic switch from ADAT form to augmented form

rg20:
## Description

- When a QP barrier solve starts on the ADAT (normal-equations) system, monitor the GMRES-IR residual each iteration. If the relative residual (||ADAT·dy − h|| / max(1, ||h||)) exceeds 1e-2, switch the remainder of the solve to the augmented KKT formulation, which factors A directly and never forms A D⁻¹ Aᵀ.
- A static pre-iteration check cannot catch this: D is uniform at iteration 0 and only becomes ill-conditioned as variables approach bounds. The switch rebuilds the augmented CSR, runs a fresh cuDSS symbolic analysis, and tears down the ADAT factorization first so peak GPU memory does not spike.
- This fallback is QP-only. LPs stay on ADAT; flipping them to augmented was a net regression. SOCPs and non-diagonal-Q problems that already start augmented are unchanged.

## Benchmarks
`twod` and `QPLIB_9008` models are now solved to suboptimality.  Previously these were stopper because of numerical errors

## Checklist

- Testing
   - [ ] Added tests
   - [ ] NA
- Documentation
   - [ ] Added new documentation
   - [x] NA

### DIFF
--- a/cpp/src/barrier/barrier.cu
+++ b/cpp/src/barrier/barrier.cu
@@ -3261,16 +3261,18 @@ i_t barrier_solver_t<i_t, f_t>::gpu_compute_search_direction(iteration_data_t<i_
           settings.log.printf("||ADAT*dy - h|| %e after IR\n", adat_solve_err);
         }
 
-        // Reactive fallback: even GMRES-IR on top of the direct Cholesky solve cannot
-        // fix a Newton direction once the normal-equations system itself has become
-        // this ill-conditioned. Rather than let the barrier loop keep grinding on
-        // ADAT solves that are only getting worse (see the diagonal-ratio growth this
-        // is a proxy for), give up on ADAT for the rest of this solve and switch to
-        // the augmented KKT formulation, which factors A directly and never forms
-        // A*D^{-1}*A^T. This one call's (dx, dy) is still whatever the degraded ADAT
-        // solve produced -- switching only takes effect starting the next call -- but
-        // that is a single degraded step against 900s of a solve that would otherwise
-        // never recover.
+        // Reactive fallback (QP only): even GMRES-IR on top of the direct Cholesky
+        // solve cannot fix a Newton direction once the normal-equations system itself
+        // has become this ill-conditioned. Rather than let the barrier loop keep
+        // grinding on ADAT solves that are only getting worse (see the diagonal-ratio
+        // growth this is a proxy for), give up on ADAT for the rest of this solve and
+        // switch to the augmented KKT formulation, which factors A directly and never
+        // forms A*D^{-1}*A^T. LPs stay on ADAT: the ill-conditioning that motivates
+        // this switch is the huge dynamic range that Q induces in D, and flipping LPs
+        // to augmented has been a net regression. This one call's (dx, dy) is still
+        // whatever the degraded ADAT solve produced -- switching only takes effect
+        // starting the next call -- but that is a single degraded step against 900s of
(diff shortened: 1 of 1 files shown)


cpp/src/barrier/barrier.cu
```diff
@@ -3197,6 +3260,40 @@ i_t barrier_solver_t<i_t, f_t>::gpu_compute_search_direction(iteration_data_t<i_
...
+        // starting the next call -- but that is a single degraded step against 900s of
+        // a solve that would otherwise never recover.
+        //
+        // Relative to ||h||, not absolute: some problems (e.g. QPLIB_9002) start from a
+        // badly-scaled initial point with ||h|| ~1e18 that self-corrects within the first
+        // couple of iterations on plain ADAT. An absolute threshold misfires there --
+        // adat_solve_err ~3000 against ||h|| ~1e18 is a relative residual of ~1e-15,
+        // excellent, not a sign of real trouble -- triggering a needless (and, for that
+        // problem, harmful) switch before the normal self-correction gets a chance.
+        const bool has_Q             = data.Q.x.size() > 0;
```
yuwenchen95, line 3283: Can we also do this fallback for LP?

cpp/src/barrier/barrier.cu
```diff
@@ -3197,6 +3260,40 @@ i_t barrier_solver_t<i_t, f_t>::gpu_compute_search_direction(iteration_data_t<i_
         if (adat_solve_err > 1e-1) {
           settings.log.printf("||ADAT*dy - h|| %e after IR\n", adat_solve_err);
         }
+
+        // Reactive fallback (QP only): even GMRES-IR on top of the direct Cholesky
```
yuwenchen95, line 3264: Comment should be cleaned up. Only need to mention the switch as the backup plan.

cpp/src/barrier/barrier.cu
```diff
@@ -1083,6 +1083,69 @@ class iteration_data_t {
     }
   }
 
+  // One-time, one-directional fallback from the ADAT (normal-equations) formulation to
```
yuwenchen95, line 1086: Comment needs to be cleaned up.

cpp/src/barrier/barrier.cu
```diff
@@ -1083,6 +1083,69 @@ class iteration_data_t {
...
+    // Free the ADAT factorization's cuDSS handle/workspace before allocating the
+    // augmented one, so the two never coexist and inflate peak GPU memory.
+    chol.reset();
+    chol =
+      std::make_unique<sparse_cholesky_cudss_t<i_t, f_t>>(handle_ptr, settings_, augmented_size);
+    chol->set_positive_definite(false);
+
+    // Build the augmented CSR structure and run symbolic analysis against the fresh
+    // chol -- exactly the construction-time sequence for a solve that starts augmented.
+    form_augmented(true);
```
yuwenchen95, line 1143: Simply setting `true` here will restart the augmented path from the beginning rather than the current iterate. If we want the augmented path as the backup, we should restarting barrier from current iterate.

cpp/src/barrier/barrier.cu
rg20: I need to experiment more for LP. It only triggers for one problem and it has a regression. That is also very non-deterministic.

rg20:
Closing this now as #1885 addresses the underlying issue.

Issue #1768, opened 2026-08-22
[BUG] Solve() aborts the process (OMP kmp_alloc assertion) when each solve runs on a freshly created thread


mal84emma:
> [!NOTE]
> The investigation and this report were prepared by an AI agent (Claude) on behalf of the reporter.

**Describe the bug**

When each `Solve()` call is made from a **newly created Python thread**, cuOpt aborts the process after a small number of sequential solves:

```
OMP: Error #13: Assertion failure at kmp_alloc.cpp(2725).
```

(exit 134), or a bare SIGSEGV with no message at all (exit 139). There is no Python exception and no traceback — the interpreter dies. Running the identical solves on the main thread, or on a single long-lived worker thread, is completely clean.

This looks like per-thread state in the bundled LLVM OpenMP runtime (`libomp-8fe85495.so`, a `NEEDED` of `libcuopt.so`, resolving to `libcuopt_cu12.libs/libomp-8fe85495.so`) that is not released when the thread that entered the solver exits, so each new solver-entering thread leaks a thread record until the runtime's allocator invariant breaks.

**Both MILP and LP are affected.** MILP merely reaches the threshold fastest — consistently on the 2nd solve here, versus the 5th–13th for LP. The failing signature is identical in both cases.

This is a hard, uncatchable process abort, and it is triggered by an entirely ordinary pattern: thread-per-request services, and worker-thread wrappers used to keep the calling thread interruptible.

**Steps/Code to reproduce bug**

Self-contained; only `cuopt` and `numpy` are needed. A small knapsack (20 variables, 3 `<=` rows) that is trivially feasible (`x = 0`) and bounded, solved N times sequentially, with a `--mode` switch controlling which thread each solve runs on.

```python
#!/usr/bin/env python3
"""Minimal reproduction: cuOpt Solve() aborts the process when each solve runs
on a freshly created Python thread."""

import argparse
import queue
import sys
import threading

import numpy as np

from cuopt.linear_programming import DataModel, Solve, SolverSettings
from cuopt.linear_programming.solver.solver_parameters import CUOPT_TIME_LIMIT


def build_model(integer=True, n=20, seed=0):
    """Small, definitely feasible and bounded knapsack.

    maximize sum_i v_i x_i  s.t.  sum_i w_ij x_i <= cap_j  (j = 0..2),
    0 <= x_i <= 1, x_i integer (or continuous with --lp).
    """
    rng = np.random.default_rng(seed)
    m = 3
[...]

PR #1875, opened 2026-09-09
[TEST] Use system GNU libgomp instead of bundled LLVM libomp for libcuopt wheel

ramakrishnap-nv:
## Summary

Experiment for #1219 (dual-OpenMP-runtime SIGSEGV; also relevant to #1768). The LLVM libomp switch in #1429 was a workaround for an outdated system GCC/libgomp on the Rocky Linux wheel build image. Since the build already requires a C++20-capable GCC (newer toolset via SCL), that constraint may no longer hold — worth testing directly rather than assuming.

This removes the `OpenMP_gomp_LIBRARY` override and the LLVM toolset install in `ci/build_wheel_libcuopt.sh`, letting libcuopt link the toolset's own GNU libgomp, and excludes `libgomp.so.*` from the auditwheel bundle (same treatment as OpenSSL) so it resolves from the host at runtime instead of being vendored/hash-renamed. This matches what cuDSS's `libcudss_mtlayer_gomp.so.0` already loads, eliminating the dual-runtime conflict.

Draft PR to run wheel-build/wheel-test CI and check for regressions (build success, fast MPS parser correctness/perf, sequential-solve stability) before deciding whether to land this for real.

## Test plan
- [ ] `wheel-build-libcuopt` / `wheel-build-cuopt` succeed
- [ ] `wheel-tests-cuopt` passes, including fast MPS parser tests
- [ ] Manually check the built wheel no longer bundles a hash-renamed libgomp, and only one `libgomp.so.1` is loaded at runtime alongside cuDSS

### DIFF
--- a/ci/build_wheel_libcuopt.sh
+++ b/ci/build_wheel_libcuopt.sh
@@ -28,13 +28,29 @@ fi
 # Install Protobuf + gRPC (protoc + grpc_cpp_plugin)
 bash ci/utils/install_protobuf_grpc.sh
 
-# EXPERIMENT: use the toolset's GNU libgomp instead of bundling LLVM libomp, to test whether
-# the constraint that motivated the LLVM switch (an outdated system GCC/libgomp on the Rocky
-# Linux wheel build image) still holds now that the build already requires a GCC new enough
-# for C++20. Matching the system libgomp also matches what cuDSS's threading layer
-# (libcudss_mtlayer_gomp.so.0) loads at runtime, avoiding a dual-OpenMP-runtime conflict.
-# See https://github.com/NVIDIA/cuopt/issues/1219
-export SKBUILD_CMAKE_ARGS=""
+# EXPERIMENT (#1219): use GNU libgomp instead of bundling LLVM libomp, to test whether the
+# constraint that motivated the LLVM switch in #1429 still holds. It doesn't fully: GNU
+# libgomp has implemented the OpenMP 5.0 detached-task runtime (omp_fulfill_event) that the
+# fast MPS parser needs since GCC 9, but the *default* '-lgomp' resolution on the Rocky Linux
+# 8 wheel build image picks up the base OS's pre-5.0 libgomp rather than the SCL gcc-toolset
+# compiler's own newer one. Resolve a libgomp.so that actually exports the symbol we need,
+# the same way the (now-removed) LLVM libomp override resolved a concrete versioned ELF file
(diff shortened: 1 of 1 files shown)


ramakrishnap-nv:
## Findings — closing this experiment

Confirmed by inspecting the actual CI container (`rapidsai/ci-wheel:26.12-cuda12.9.2-rockylinux8-py3.14`) directly, rather than iterating further on CI:

- The active `g++` on this image is **system GCC 8.5.0**, not gcc-toolset-14 (CMake only uses gcc-toolset-14 because `CC`/`CXX` are set explicitly elsewhere in the build).
- gcc-toolset-14's own `libgomp.so` is **not a real library** — it's a linker script: `INPUT ( /usr/lib64/libgomp.so.1 )`, pointing straight back to the system one.
- The system `libgomp.so.1` is owned by RPM `libgomp-8.5.0-28.el8_10`, documented as **"GCC OpenMP v4.5 shared support library"** — predates OpenMP 5.0 / `omp_fulfill_event` (needed by the fast MPS parser's detached-task pipeline in `mps_section_scanner.cpp`).
- `dnf provides */libgomp.so.1` on this image only resolves to 8.5.0 variants. There is no newer GNU libgomp available from the image's configured repos, and gcc-toolset-14 doesn't carry its own private copy (toolsets only override the compiler/stdlib, not libgomp).

So Alice's original read (#1429 discussion) was right, and still holds today: the Rocky Linux 8 wheel build image's libgomp is genuinely outdated, and gcc-toolset-14 doesn't fix that. Sourcing a newer libgomp from elsewhere (e.g. conda-forge, confirmed locally to have the symbol) would mean either bundling it into the wheel — which recreates the original dual-runtime conflict with cuDSS's dlopen'd system libgomp, just GNU-vs-GNU instead of GNU-vs-LLVM — or requiring it as a host-provided runtime dependency, raising the minimum libgomp version wheel users need (same class of requirement as the existing libssl.so.3 one). Given the build image has no conda/mamba available, going down that path is more build-system surface area than this issue warrants.

**Closing this PR.** Next step: keep LLVM libomp for the wheel, and address the dual-runtime crash directly — `dlopen(libcudss_mtlayer_gomp.so.0, RTLD_NOW | RTLD_NODELETE)` early in cuOpt so cuDSS's own later `dlclose()` during teardown can never actually unmap it, removing the race that causes #1219's SIGSEGV. Will open a new PR for that.

PR #1880, opened 2026-09-10
[CPUFJ PR 1] Improve time-to-first-incumbent, unify solution callback publication, scaffolding changes

aliceb-nv:
This PR introduces changes to improve the latency to the first incumbent, along with a unified solution publication object to ensure user callbacks receive a new incumbent as soon as it is found.

Also included are scaffolding changes to prepare for the coming CPUFJ primal integral improvements:

Highway (SIMD library) integration + THIRD-PARTY-LICENCE changes
run_mip.cpp harness changes to log incumbents as received by the user callback
Common util functions (e.g. numerically compensated dot product)
This PR will be the first of a series of 4-5 PRs destined to merge the primal integral improvements into main.

Benchmark results are neutral. This is scaffolding that will matter for the coming PRs

## Checklist

- Testing
   - [ ] Added tests
   - [ ] NA
- Documentation
   - [ ] Added new documentation
   - [ ] NA


benchmarks/linear_programming/cuopt/run_mip.cpp
```diff
@@ -151,8 +415,12 @@ int run_single_file(std::string file_path,
                     double work_limit,
                     bool deterministic)
 {
-  const raft::handle_t handle_{};
+  (void)cudaFree(0);
```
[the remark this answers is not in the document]
aliceb-nv: This is a benchmark harness intended for developer use only. If CUDA was broken on the system, it will be apparent in a myriad of ways.

benchmarks/linear_programming/cuopt/run_mip.cpp
```diff
@@ -206,42 +490,74 @@ int run_single_file(std::string file_path,
...
-  settings.tolerances.absolute_tolerance = 1e-6;
-  settings.presolver                     = cuopt::mathematical_optimization::presolver_t::Default;
-  settings.reliability_branching         = reliability_branching;
-  settings.clique_cuts                   = -1;
-  settings.seed                          = 42;
   cuopt::mathematical_optimization::benchmark_info_t benchmark_info;
-  settings.benchmark_info_ptr = &benchmark_info;
-  auto start_run_solver       = std::chrono::high_resolution_clock::now();
-  auto solution = cuopt::mathematical_optimization::solve_mip(&handle_, mps_data_model, settings);
+  if constexpr (!use_c_api) { settings.benchmark_info_ptr = &benchmark_info; }
```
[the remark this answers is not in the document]
aliceb-nv: Intended. Again, this is a developer-facing-only benchmark utility harness.

benchmarks/linear_programming/cuopt/run_mip.cpp
```diff
@@ -206,42 +490,74 @@ int run_single_file(std::string file_path,
...
+    start_run_solver = std::chrono::high_resolution_clock::now();
+    incumbent_tracker =
+      std::make_unique<incumbent_tracker_t>(start_run_solver, mps_data_model.get_n_variables());
+    cuOptSetMIPGetSolutionCallback(
+      c_api.settings, c_api_incumbent_callback, incumbent_tracker.get());
+    cuOptSolve(c_api.problem, c_api.settings, &c_api.solution);
+    cuOptGetTerminationStatus(c_api.solution, &solution.termination_status);
+    cuOptGetObjectiveValue(c_api.solution, &solution.objective_value);
+    cuOptGetSolutionBound(c_api.solution, &solution.solution_bound);
+    cuOptGetMIPGap(c_api.solution, &solution.mip_gap);
```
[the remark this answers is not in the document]
aliceb-nv: Again, this is a developer-facing-only benchmark utility harness. I would much prefer if you focused on the actual meat of the PR instead of wasting your time and tokens on a developer utility. I have already made this clear before in the previous review.

cpp/CMakeLists.txt
```diff
@@ -304,6 +305,25 @@ set(BUILD_SHARED_LIBS OFF)
 FetchContent_MakeAvailable(pslp)
 set(BUILD_SHARED_LIBS ${BUILD_SHARED_LIBS_SAVED})
 
+FetchContent_Declare(
+        highway
+        GIT_REPOSITORY "https://github.com/google/highway.git"
+        GIT_TAG "1.4.0"
```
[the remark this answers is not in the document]
aliceb-nv: We already have plenty of precedents of using GIT_TAG in the repo. This is clearer. This concern is minor.

cpp/src/pdlp/optimization_problem.cu
```diff
@@ -285,14 +288,39 @@ void optimization_problem_t<i_t, f_t>::set_variable_types(const var_t* variable_
...
+                                  variable_types_.end(),
+                                  [] __host__ __device__(var_t val) {
+                                    return val == var_t::INTEGER || val == var_t::SEMI_CONTINUOUS;
+                                  });
+    has_semi_continuous_variables = thrust::count_if(handle_ptr_->get_thrust_policy(),
+                                                     variable_types_.begin(),
+                                                     variable_types_.end(),
+                                                     [] __host__ __device__(var_t val) {
+                                                       return val == var_t::SEMI_CONTINUOUS;
+                                                     }) > 0;
```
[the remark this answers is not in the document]
aliceb-nv: Can you name any scenario in production where a handle_ptr could reach this function as nullptr? I believe that if handle_ptr is nullptr, we have much bigger problems anyways. Might as well fail loudly with a segfault that can then be gdb-attached rather than hide it with a cuopt_expects or a silent fail or whatnot.

benchmarks/linear_programming/cuopt/row_audit.hpp
```diff
@@ -0,0 +1,143 @@
+/* clang-format off */
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+/* clang-format on */
+
+#pragma once
```
akifcorduk, line 8: There are some functions in https://github.com/NVIDIA/cuopt/blob/main/benchmarks/linear_programming/cuopt/initial_problem_check.hpp 
Maybe unify them?

benchmarks/linear_programming/cuopt/run_mip.cpp
```diff
@@ -86,6 +92,181 @@ void write_to_output_file(const std::string& out_dir,
...
+  double solution_bound{};
+  double mip_gap{};
+  cuopt_int_t termination_status{};
+
+  double get_objective_value() const { return objective_value; }
+  double get_solution_bound() const { return solution_bound; }
+  double get_mip_gap() const { return mip_gap; }
+};
+
+static void configure_c_api_settings(
```
akifcorduk, line 125: Why do we need C API here?

benchmarks/linear_programming/cuopt/run_mip.cpp
```diff
@@ -86,6 +92,181 @@ void write_to_output_file(const std::string& out_dir,
...
+    c_settings, CUOPT_MIP_RELIABILITY_BRANCHING, settings.reliability_branching);
+  cuOptSetIntegerParameter(c_settings, CUOPT_MIP_CLIQUE_CUTS, settings.clique_cuts);
+  cuOptSetIntegerParameter(c_settings, CUOPT_RANDOM_SEED, settings.seed);
+  cuOptSetParameter(
+    c_settings, CUOPT_MIP_HEURISTICS_ONLY, settings.heuristics_only ? "true" : "false");
+  cuOptSetParameter(c_settings, CUOPT_LOG_TO_CONSOLE, settings.log_to_console ? "true" : "false");
+  cuOptSetParameter(c_settings, CUOPT_LOG_FILE, settings.log_file.c_str());
+}
+
+static bool verify_solution(
```
akifcorduk, line 150: We could put this to another file I think. We shouldn't bloat the benchmarking file.

benchmarks/linear_programming/cuopt/run_mip.cpp
akifcorduk, line 418: Do we really need that? If we need that, it means something has gone wrong on the previous problem on the same process right?

benchmarks/linear_programming/cuopt/run_mip.cpp
```diff
@@ -288,17 +596,27 @@ int run_single_file(std::string file_path,
   std::stringstream ss;
   int decimal_places = 2;
```
akifcorduk, line 597: Let's also change this decimal place to 5, so that scripts that infer metrics from logs are correct.

cpp/include/cuopt/mathematical_optimization/cpu_optimization_problem.hpp
```diff
@@ -124,6 +124,12 @@ class cpu_optimization_problem_t : public optimization_problem_interface_t<i_t,
   std::string get_objective_name() const override;
   std::string get_problem_name() const override;
   problem_category_t get_problem_category() const override;
+  /**
+   * @brief Whether any variable type is SEMI_CONTINUOUS.
+   *
+   * Cached in set_variable_types() / adopt_from_mps_data_model().
+   */
```
akifcorduk, line 131: Other functions don't have comments, let's be consistent.

cpp/include/cuopt/mathematical_optimization/optimization_problem.hpp
```diff
@@ -284,6 +284,12 @@ class optimization_problem_t : public optimization_problem_interface_t<i_t, f_t>
   std::string get_objective_name() const override;
   std::string get_problem_name() const override;
   problem_category_t get_problem_category() const override;
+  /**
+   * @brief Whether any variable type is SEMI_CONTINUOUS.
+   *
+   * Cached in set_variable_types(); used to skip SC reformulation host probes.
+   */
+  bool has_semi_continuous_variables() const noexcept;
```
akifcorduk, line 292: Same.

cpp/src/mip_heuristics/local_search/local_search.cu
```diff
@@ -78,6 +78,7 @@ void local_search_t<i_t, f_t>::start_cpufj_scratch_threads(population_t<i_t, f_t
     cpu_fj->improvement_callback =
       [this, &population, problem_ptr = context.problem_ptr](
         f_t obj, const std::vector<f_t>& h_vec, double /*work_units*/) {
+        context.solution_publication.publish_if_better(problem_ptr, h_vec, obj);
```
akifcorduk, line 81: Shouldn't these calls be part of `add_solution` or `add_external` solution ?

cpp/src/mip_heuristics/solution_publication.cuh
```diff
@@ -0,0 +1,173 @@
...
+ private:
+  void audit_integrality(problem_t<i_t, f_t>* problem_ptr, const std::vector<f_t>& assignment)
+  {
+    // The B&B thread may never have selected a device of its own.
+    RAFT_CUDA_TRY(cudaSetDevice(device_id_));
+    const auto integer_variables =
+      cuopt::host_copy(problem_ptr->integer_indices, handle_->get_stream());
+    for (i_t variable : integer_variables) {
+      const f_t value = assignment[variable];
+      if (value == std::round(value)) continue;
```
akifcorduk, line 120: I think we should check integrality tolerance.

cpp/src/mip_heuristics/solution_publication.cuh
```diff
@@ -0,0 +1,173 @@
...
+  }
+
+ private:
+  void audit_integrality(problem_t<i_t, f_t>* problem_ptr, const std::vector<f_t>& assignment)
+  {
+    // The B&B thread may never have selected a device of its own.
+    RAFT_CUDA_TRY(cudaSetDevice(device_id_));
+    const auto integer_variables =
+      cuopt::host_copy(problem_ptr->integer_indices, handle_->get_stream());
+    for (i_t variable : integer_variables) {
```
akifcorduk, line 118: Just a nit: it is a common function that might be repeated in many places. We can reuse maybe?

cpp/src/mip_heuristics/utils.hpp
```diff
@@ -0,0 +1,46 @@
...
+#else
+#define CUOPT_MIP_HOST_DEVICE inline
+#endif
+
+namespace cuopt::mathematical_optimization::mip {
+
+// Ogita-Rump-Oishi Dot2. TwoProduct recovers the rounding of each coefficient-value product, which
+// a compensated summation over already-multiplied terms cannot see.
+// https://epubs.siam.org/doi/abs/10.1137/030601818
+template <typename UIt, typename VIt>
```
akifcorduk, line 24: I don't think we need a new file for this. There are various util places. Example:
https://github.com/NVIDIA/cuopt/blob/main/cpp/src/mip_heuristics/utils.cuh

cpp/src/pdlp/optimization_problem.cu
```diff
@@ -54,6 +54,8 @@
 
 namespace cuopt::mathematical_optimization {
 
+constexpr size_t host_variable_type_summary_limit = 50'000;
```
akifcorduk, line 57: Shouldn't this be configurable via a setting?

cpp/src/pdlp/optimization_problem.cu
```diff
@@ -285,14 +288,39 @@ void optimization_problem_t<i_t, f_t>::set_variable_types(const var_t* variable_
...
+                                  });
+    has_semi_continuous_variables = thrust::count_if(handle_ptr_->get_thrust_policy(),
+                                                     variable_types_.begin(),
+                                                     variable_types_.end(),
+                                                     [] __host__ __device__(var_t val) {
+                                                       return val == var_t::SEMI_CONTINUOUS;
+                                                     }) > 0;
+  }
+  has_semi_continuous_variables_ = has_semi_continuous_variables;
   if (n_discrete == size) {
```
akifcorduk, line 324: Now we are counting semi-continuous as discrete, does this if still hold? We don't have a differenciation between MIP and IP, but still I am not sure if we can categorize it as IP.

akifcorduk (approved):
Thanks Alice! Approving with some comments. I have thumbs uped one AI comment.

benchmarks/linear_programming/cuopt/run_mip.cpp
aliceb-nv: I added this to get a solve path that is a close as possible to what GAMS is using on the benchmarks, in case some bugs or behavior changes are specific to the API path

benchmarks/linear_programming/cuopt/run_mip.cpp
aliceb-nv: I'm just using that as an idiom to get CUDA to initialize :) cudaFree(0) is a no-op, but it gets the driver to get all set up so that we don't include CUDA initialization in any timings

cpp/src/mip_heuristics/utils.hpp
aliceb-nv: I did that because it will get used in host-only files in later PRs, utils.cuh brings in a lot of CUDA context that will cause GCC to fail the build

cpp/src/pdlp/optimization_problem.cu
aliceb-nv: I suppose so, but this is more of a minor tuning that's very unlikely to affect user runs in any perceivable way. I'd be more afraid of users being misled into thinking this is major

cpp/src/pdlp/optimization_problem.cu
aliceb-nv: I think main treats an IP with semi-continuous as IP still 😅 I don't think it makes sense, but I did not dare change this as I don't know how this could affect the gRPC paths and such.

benchmarks/linear_programming/cuopt/initial_problem_check.hpp
```diff
@@ -5,6 +5,263 @@
...
+    const _Float128 term = (_Float128)values[entry] * (_Float128)solution[indices[entry]];
+    activity += term;
+    if (term > 0) positive += term;
+  }
+  return {activity, positive};
+}
+
+inline double scaled_tolerance(double absolute_tolerance, double first, double second)
+{
+  return absolute_tolerance * std::max({1.0, std::fabs(first), std::fabs(second)});
```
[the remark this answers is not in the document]
aliceb-nv: This is on purpose, disregard

benchmarks/linear_programming/cuopt/initial_problem_check.hpp
```diff
@@ -5,6 +5,263 @@
...
+                                  << " lower_tolerance=" << lower_bound_tolerance
+                                  << " upper_tolerance=" << upper_bound_tolerance << "\n";
+      return false;
+    }
+    if (variable_types[variable] == 'I' && value != std::round(value)) {
+      std::osyncstream(std::cerr) << std::setprecision(17) << "Incumbent " << incumbent
+                                  << " variable " << variable
+                                  << " is not exactly integral: value=" << value
+                                  << " residual=" << std::abs(value - std::round(value)) << "\n";
+      return false;
```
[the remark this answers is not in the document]
aliceb-nv: This is on purpose (stricter checks), disregard

benchmarks/linear_programming/cuopt/run_mip.cpp
nguidotti: When using the cuOpt, I would expect the all API has similar behaviour. Maybe we should address the divergence

nguidotti (approved):
Thanks Alice! It looks good to me. I have a few questions:
- Can you explain how we are handling the solution callbacks? 
- For Highway, you are doing some form of dynamic dispatch, right? How this interaction with the compile flags (like `-mavx2`)?

aliceb-nv:
Hey Nicolas,

Solution callbacks haven't changed much. The only difference now is that instead of waiting for the solution to enter the GPU heuristic population and the various processing steps, they are reported via this publication mechanism at the source. That lets us report solutions faster, since sometimes seconds pass before the GPU heuristics get a chance to process the solution queue

Highway lets you build specialized functions that run on any ISA of your choosing regardless of compiler flags, it does this automatically via macro and GCC extensions magic :) At program initialization it looks at your supported ISAs and fixes the function pointers to point to the most optimal version for your current CPU.

For the C API, this is just a defensive approach. I want to make sure that an API divergence bug, if any, cannot slip past my eyes

ramakrishnap-nv (approved):

PR #1885, opened 2026-09-10
Presolve: remove empty columns with a diagonal quadratic term

rg20:
## Description

Empty-column presolve previously skipped any variable that appeared in `Q`. That is necessary for off-diagonal coupling, but a variable with only a diagonal term `q_jj` is an unconstrained 1-D QP and can be fixed.
For an empty `A` column, fix `x_j` by minimizing `c_j x_j + (1/2) q_jj x_j^2` over `[l_j, u_j]`:
- `q_jj > 0`: clip the vertex `-c_j / q_jj` to the bounds
- `q_jj < 0`: take the better finite bound (leave the column if a bound is infinite)
- `q_jj = 0`: existing linear rule (`l_j` if `c_j >= 0`, `u_j` if `c_j <= 0`)
The 1-D objective is folded into `obj_constant`. `Q` is compacted by dropping the removed rows and remapping column indices. Linear empty columns are unchanged.

This is enough to solve QPLIB instances `twod` and `QPLIB_9008` to optimality.

## Checklist

- Testing
   - [ ] Added tests
   - [ ] NA
- Documentation
   - [ ] Added new documentation
   - [x] NA

### DIFF
--- a/cpp/src/dual_simplex/presolve.cpp
+++ b/cpp/src/dual_simplex/presolve.cpp
@@ -26,6 +26,75 @@ static i_t linear_variable_count(const lp_problem_t<i_t, f_t>& problem)
   return problem.second_order_cone_dims.empty() ? problem.num_cols : problem.cone_var_start;
 }
 
+template <typename f_t>
+static f_t quadratic_1d_obj(f_t x, f_t c, f_t q)
+{
+  return c * x + 0.5 * q * x * x;
+}
+
+// Minimizer of c*x + (1/2) q x^2 over [lower, upper]. False if unbounded.
+template <typename f_t>
+static bool unconstrained_1d_qp_minimizer(f_t c, f_t q, f_t lower, f_t upper, f_t& x)
+{
+  if (q > 0) {
+    x = std::min(upper, std::max(lower, -c / q));
+    return std::isfinite(x);
+  }
+  if (q < 0) {
+    if (lower <= -inf || upper >= inf) { return false; }
+    x = (quadratic_1d_obj(lower, c, q) <= quadratic_1d_obj(upper, c, q)) ? lower : upper;
+    return true;
+  }
+  if (c >= 0 && lower > -inf) {
+    x = lower;
+    return true;
+  }
+  if (c <= 0 && upper < inf) {
+    x = upper;
+    return true;
+  }
+  return false;
+}
+
+template <typename i_t, typename f_t>
+static void collect_diagonal_quadratic(const csr_matrix_t<i_t, f_t>& Q,
+                                       std::vector<f_t>& q_diag,
+                                       std::vector<bool>& q_coupled)
+{
+  const i_t n = static_cast<i_t>(q_diag.size());
+  for (i_t i = 0; i < Q.m; ++i) {
+    for (i_t p = Q.row_start[i]; p < Q.row_start[i + 1]; ++p) {
+      const i_t col = Q.j[p];
+      if (i == col) {
+        q_diag[i] += Q.x[p];
+      } else {
+        q_coupled[i] = true;
+        if (col >= 0 && col < n) { q_coupled[col] = true; }
+      }
+    }
+  }
+}
+
+// Drop marked variables from square Q (rows and matching column indices).
+template <typename i_t, typename f_t>
+static void remove_variables_from_Q(csr_matrix_t<i_t, f_t>& Q,
+                                    std::vector<i_t>& col_marker,
+                                    const std::vector<i_t>& col_old_to_new,
+                                    i_t new_cols)
+{
+  csr_matrix_t<i_t, f_t> Qout(0, 0, 0);
+  Q.remove_rows(col_marker, Qout);
+  const i_t nnz = Qout.row_start[Qout.m];
+  for (i_t p = 0; p < nnz; ++p) {
+    const i_t new_col = col_old_to_new[Qout.j[p]];
+    assert(new_col != -1);
+    Qout.j[p] = new_col;
+  }
+  Qout.n      = new_cols;
+  Qout.nz_max = nnz;
+  Q           = std::move(Qout);
+}
+
 template <typename i_t, typename f_t>
 i_t remove_empty_cols(lp_problem_t<i_t, f_t>& problem,
                       i_t& num_empty_cols,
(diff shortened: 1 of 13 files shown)


cpp/src/dual_simplex/presolve.cpp
```diff
@@ -87,14 +139,13 @@ i_t remove_empty_cols(lp_problem_t<i_t, f_t>& problem,
   presolve_info.remaining_variables.reserve(new_cols);
 
   problem.A.remove_columns(col_marker);
-  // Clean up objective, lower, upper, and col_names
```
chris-maes, line 90: Leave comment?

cpp/src/dual_simplex/presolve.cpp
```diff
@@ -26,51 +26,103 @@ static i_t linear_variable_count(const lp_problem_t<i_t, f_t>& problem)
...
-        presolve_info.removed_values.push_back(problem.lower[j]);
-        problem.obj_constant += problem.objective[j] * problem.lower[j];
-        remove_var = true;
-      } else if (problem.objective[j] <= 0 && problem.upper[j] < inf && !non_removable) {
-        presolve_info.removed_values.push_back(problem.upper[j]);
-        problem.obj_constant += problem.objective[j] * problem.upper[j];
+    if (j < linear_cols && problem.A.col_length(j) == 0 && !q_coupled[j]) {
+      f_t x_fix;
+      if (unconstrained_1d_qp_minimizer(
+            problem.objective[j], q_diag[j], problem.lower[j], problem.upper[j], x_fix)) {
```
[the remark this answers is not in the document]
chris-maes: @rg20, I think coderabbit is right.  We do need to satisfy

```
A^T y + z = c + Q*x

on the original problem.

If x_j is the variable we are fixing then A*e_j = 0, so e_j^T A^T = 0 so e_j^T (A^T y + z = c + Q*x) gives 

z_j = c_j + e_j^T * Q * x,

if Q is diagonal than e_j^T Q = Q(j, :) = Q(j, j)*e_j^T so we have

z_j = c_j + Q(j, j) *e_j^T x = c_j + q_diag[j] * x_fix

```

chris-maes (approved):
I think you need to fix the reduced cost issue pointed out by coderabbit, but otherwise it looks good to me.

cpp/src/dual_simplex/presolve.cpp
```diff
@@ -26,59 +26,114 @@ static i_t linear_variable_count(const lp_problem_t<i_t, f_t>& problem)
...
+
+// Minimizer of c*x + (1/2) q x^2 over [lower, upper]. False if unbounded.
+template <typename f_t>
+static bool unconstrained_1d_qp_minimizer(f_t c, f_t q, f_t lower, f_t upper, f_t& x)
+{
+  if (q > 0) {
+    x = std::min(upper, std::max(lower, -c / q));
+    return std::isfinite(x);
+  }
+  if (q < 0) {
```
yuwenchen95, line 43: The internal problem has already been turned into a minimization problem. Since we only solve a convex QP by now, we don't need to branch on `q<0`.

cpp/src/dual_simplex/presolve.cpp
```diff
@@ -26,59 +26,114 @@ static i_t linear_variable_count(const lp_problem_t<i_t, f_t>& problem)
...
-    }
-  }
+  const i_t linear_cols = linear_variable_count(problem);
+  std::vector<f_t> q_diag(problem.num_cols, 0.0);
+  std::vector<bool> q_coupled(problem.num_cols, false);
+  if (problem.Q.n > 0) { collect_diagonal_quadratic(problem.Q, q_diag, q_coupled); }
 
   std::vector<i_t> col_marker(problem.num_cols);
   i_t new_cols = 0;
   for (i_t j = 0; j < problem.num_cols; ++j) {
```
yuwenchen95, line 118: Shall we loop only over `linear_cols` instead of `problem.num_cols`? We have already setup SOCP here and don't presolve over cone variables.

cpp/src/dual_simplex/presolve.cpp
```diff
@@ -26,59 +26,114 @@ static i_t linear_variable_count(const lp_problem_t<i_t, f_t>& problem)
...
-      bool non_removable = has_quadratic_term[j];
-      if (problem.objective[j] >= 0 && problem.lower[j] > -inf && !non_removable) {
-        presolve_info.removed_values.push_back(problem.lower[j]);
-        problem.obj_constant += problem.objective[j] * problem.lower[j];
-        remove_var = true;
-      } else if (problem.objective[j] <= 0 && problem.upper[j] < inf && !non_removable) {
-        presolve_info.removed_values.push_back(problem.upper[j]);
-        problem.obj_constant += problem.objective[j] * problem.upper[j];
+    f_t removed_z   = problem.objective[j];
+    if (j < linear_cols && problem.A.col_length(j) == 0 && !q_coupled[j]) {
```
yuwenchen95, line 121: `j < linear_cols` can be removed once we loop over only `linear_cols`.

cpp/src/dual_simplex/presolve.cpp
```diff
@@ -26,59 +26,114 @@ static i_t linear_variable_count(const lp_problem_t<i_t, f_t>& problem)
...
+      f_t x_fix;
+      if (unconstrained_1d_qp_minimizer(
```
yuwenchen95, line 123: If `unconstrained_1d_qp_minimizer` returns false, isn't it a sign that the problem is unbounded below and we can claim dual infeasibility?

yuwenchen95 (changes requested):
Left some comment regarding algorithmic efficiency and code simplification. Possible dual infeasibility can also be derived if `unconstrained_1d_qp_minimizer` returns false.

cpp/src/dual_simplex/presolve.cpp
rg20: I think it is still okay to keep this for the sake of completeness.

yuwenchen95: Maybe we can add an assert `q >=0` instead of branching? It would also be clear to track the boundary of code safety.

yuwenchen95 (approved):

PR #1886, opened 2026-09-10
add cancel and status for grpc routing client

tmckayus:
Add status and cancel APIs to the gRPC routing client. This is for general use, but also to support the http proxy server.


ramakrishnap-nv (approved):
Think code rabbit changes would be the ones needs to be addressed, rest seems to be good shape.

PR #1889, opened 2026-09-11
Primal and dual simplex improvements

chris-maes:
This PR include many improvements to primal and dual simplex. It also makes primal simplex a first class method for solving LPs.

It results in a **7.26% improvement in solve time** and a **8.02% improvement in MIP gap** (geomean over 240 MIPLIB benchmark problems with 300 second time limit on GH200). **5 more problems solve to optimality** within the time limit, and the number of problems with feasible solutions remains the same.

On the 240 MIPLIB LP relaxations this PR is **1.44X faster** (raw geomean) or 1.28X faster (geomean shifted by 1s). When compared with HiGHS the 1s shifted geomean is **0.99X** (i.e. the solvers are equivalent).

On NETLIB primal simplex is about 2.3X slower than dual simplex. So while we enable primal simplex to be used to solve LPs, we don't recommend it. 

We now give a brief description of the changes:

Dual Simplex:
* We now always use primal simplex if we are not able to remove a perturbation (before we just ignored the perturbation)
* This allows us to apply stronger initial perturbation; in particular we perturb more when the initial point is highly dual degenerate
* However, too much perturbation at the end of the solve can force us to go into the much slower primal simplex
* So we try to remove perturbation throughout the solve when variables leave the basis
* We also try to flip bounded variables that are dual infeasible after we remove the perturbation and continue with dual simplex, rather than switching to primal
* When a problem has many bounded variables that are dual degenerate, phase I doesn't tell us what bound these variables should take. We now try three different initial points and choose the one with the smallest number of primal infeasibilities (we also look at the magnitude of the primal infeasibility)
* The bound flipping ratio test was also completely rewritten: 
     - It now decides on which bounds to flip (before this was done by a separate routine). 
     - We also fixed a bug where we double counting the slope contribution from the first selected variable
     - We also removed the O(n*log(n)) algorithm and replaced it with a coarse filter and then group variables into    buckets with the same Harris ratio. The new algorithm should be approximately O(n), but may be O(n^2) in the worse case.
    - We now select the variable in the final Harris bucket (i.e. the longest step-length) that maximizes the pivot value, while ensuring we are within 1/10th of the largest pivot value over all breakpoints. 
 * When solving an LP alone (outside of MIP) we apply row-scaling
* Improved work estimates in right-looking LU and in many other places. Work estimates should now be within 5% of the time percentage.

Primal Simplex:
* The previous primal simplex algorithm was disabled in the code. We made many changes to enable it as a standalone solver and as a clean up routine for dual simplex
* We fixed cycling issues and incorrect leaving variable status
* We limited steps by the entering variables bound
* Fixed issues around the Phase-I / Phase-II transtion
* Add work estimates throughout primal simplex
* Added a two pass Harris ratio test
* Added Devex pricing 

Crossover
* We added primal simplex as a clean up when the crossover basis is primal feasible but dual infeasible
* Fixed a tolerance mismatch that was causing crossover to declare numerical error more than it should

Acknowledgement:
* The following improvements were proposed by the [Hiverge](https://www.hiverge.ai/) AI discovery engine (cc: [@kerry-hiverge](https://github.com/kerry-hiverge) ).
   - Prefer larger pivots within Harris-ratio buckets  
   - Row equilibration for imbalanced LPs
   - Stronger perturbations for highly dual-degenerate bases


cpp/include/cuopt/mathematical_optimization/constants.h
```diff
@@ -211,7 +214,8 @@
 #define CUOPT_METHOD_PDLP         1
 #define CUOPT_METHOD_DUAL_SIMPLEX 2
 #define CUOPT_METHOD_BARRIER      3
-#define CUOPT_METHOD_UNSET        4
+#define CUOPT_METHOD_PRIMAL       4
+#define CUOPT_METHOD_UNSET        5
```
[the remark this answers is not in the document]
chris-maes: I think it's fine to change constants from version to version. We don't say that you won't have to recompile with a new version.

cpp/src/dual_simplex/crossover.cpp
```diff
@@ -1423,8 +1424,11 @@ crossover_status_t crossover(const lp_problem_t<i_t, f_t>& lp,
   } else if (dual_feasible && !primal_feasible) {
     i_t dual_iter = 0;
     std::vector<f_t> edge_norms;
-    dual_status_t status =
-      dual_phase2(2, 0, start_time, lp, settings, vstatus, solution, dual_iter, edge_norms);
+    f_t work_estimate                                 = 0.0;
+    simplex_solver_settings_t<i_t, f_t> dual_settings = settings;
+    dual_settings.iteration_limit                     = std::numeric_limits<i_t>::max();
```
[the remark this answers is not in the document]
chris-maes: It's intentional iteration limits apply to the main solve, not crossover.

cpp/src/dual_simplex/phase2.cpp
```diff
@@ -2969,6 +3461,63 @@ dual_status_t dual_phase2_with_advanced_basis(i_t phase,
...
+                                                                            false);
+          if (primal_status == primal_status_t::OPTIMAL) {
+            settings.log.printf("Primal cleanup successful. Iterations %d\n", iter - dual_iter);
+            objective = lp.objective;
+          } else {
+            settings.log.printf("Primal cleanup failed.\n");
+            const f_t dual_infeas = phase2::dual_infeasibility(
+              lp, settings, vstatus, sol.z, settings.tight_tol, settings.dual_tol);
+            if (dual_infeas > 10.0 * settings.dual_tol) { return dual_status_t::NUMERICAL; }
+          }
```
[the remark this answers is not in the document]
chris-maes: I don't think we could have reached that point in the code if we were not primal feasible.

cpp/src/dual_simplex/scaling.cpp
```diff
@@ -253,6 +253,39 @@ i_t scaling(const lp_problem_t<i_t, f_t>& unscaled,
...
   }
 
+  // MIP performs integer-aware row scaling before presolve, while QP and SOCP
+  // use the Ruiz path above. Apply this simpler equilibration only to LPs.
+  const bool use_lp_row_scaling =
+    !settings.inside_mip && unscaled.second_order_cone_dims.empty() && unscaled.Q.n == 0;
+  if (use_lp_row_scaling) {
+    csr_matrix_t<i_t, f_t> Arow(0, 0, 0);
+    scaled.A.to_compressed_row(Arow);
+    std::vector<f_t> row_norm(m, 1.0);
```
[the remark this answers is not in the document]
chris-maes: I think these are 1.0 to avoid dividing by zero.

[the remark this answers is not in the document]
chris-maes: I think these are 1.0 to avoid dividing by zero.

cpp/src/dual_simplex/right_looking_lu.cpp
```diff
@@ -224,7 +225,7 @@ class trailing_matrix_t {
             if (markowitz <= markowitz_lower_bound) { break; }
           }
         }
-        work_estimate_ += 3 * (c_end - c_start);
+        work_estimate_ += 3 * (p - c_start);
```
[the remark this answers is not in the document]
chris-maes: work estimates are meant to be rough estimates. How much could this undercount?

cpp/src/dual_simplex/solve.cpp
```diff
@@ -61,6 +61,53 @@ void write_matlab(const std::string& filename, const simplex::lp_problem_t<i_t,
...
+  switch (status) {
+    case primal_status_t::OPTIMAL: return lp_status_t::OPTIMAL;
+    case primal_status_t::PRIMAL_UNBOUNDED: return lp_status_t::UNBOUNDED;
+    case primal_status_t::PRIMAL_INFEASIBLE: return lp_status_t::INFEASIBLE;
+    case primal_status_t::TIME_LIMIT: return lp_status_t::TIME_LIMIT;
+    case primal_status_t::ITERATION_LIMIT: return lp_status_t::ITERATION_LIMIT;
+    case primal_status_t::CONCURRENT_LIMIT: return lp_status_t::CONCURRENT_LIMIT;
+    case primal_status_t::NUMERICAL:
+    case primal_status_t::NOT_LOADED:
+    default: return lp_status_t::NUMERICAL_ISSUES;
```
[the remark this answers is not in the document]
chris-maes: WORK_LIMIT is not an external status code. You can only set work limits from within the solver.

cpp/src/dual_simplex/scaling.cpp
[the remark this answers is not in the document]
chris-maes: I'm not sure you want to scale by a number less than 1.0.

cpp/src/dual_simplex/phase2.cpp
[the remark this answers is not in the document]
chris-maes: Ok. I addressed this.

cpp/src/dual_simplex/primal.cpp
```diff
@@ -403,150 +876,601 @@ primal_status_t primal_phase2(i_t phase,
...
+            primal_inf);
+          settings.log.printf(" Iter     Objective           Num Inf.  Sum Inf.       Time\n");
+          switched_phase = true;
+          continue;
+        }
+        if (num_dual_inf > 0) {
+          // The refreshed reduced costs contain a candidate visible at the active
+          // pricing tolerance.
+          continue;
+        }
```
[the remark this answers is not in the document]
chris-maes: I'm hesitant to make such a large change as you suggest as it would invalidate my benchmark results. I have added limit checks at the top of the loop. I believe this would prevent the solver from hanging.

cpp/src/dual_simplex/solve.cpp
```diff
@@ -291,12 +352,19 @@ lp_status_t solve_linear_program_with_advanced_basis(
...
     }
     constexpr bool primal_cleanup = false;
     if (status == dual_status_t::OPTIMAL && primal_cleanup) {
+      settings.log.printf("Running primal cleanup\n");
       primal_phase2(2, start_time, lp, settings, vstatus, solution, iter);
+      // TODO: We need to update ft if the basis changed
+    }
+    if (settings.inside_mip == 1 && settings.concurrent_halt != nullptr) {
+      settings.log.debug("Setting concurrent halt to 1 inside_mip\n");
+      *settings.concurrent_halt = 1;
```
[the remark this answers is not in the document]
chris-maes: A concurrent solve should not be affected by work limit. If a user sets an iteration limit, it seems appropriate to stop as well.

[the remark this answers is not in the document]
chris-maes: My point is WORK_LIMIT does not happen in the context of a concurrent solve.

mlubin:
Haven't completed a review, but here are some initial agent findings that seem legitimate:

```
 1. [P1] Initial objective perturbation can cause an incorrect cutoff and potentially incorrect MIP pruning.
     cpp/src/dual_simplex/phase2.cpp:3147, cpp/src/dual_simplex/phase2.cpp:4232

     The PR activates initial cost perturbation, but cutoff testing still uses the original-cost objective at the current iterate. That value is not necessarily a valid lower bound for the original LP while simplex is solving the perturbed problem.

     Reproduced with:

     minimize 99.001*x + 99*y
     x + y = 1
     0 <= x,y <= 2
     cutoff = 99.0005

     The solver returns CUTOFF at objective 99.001, although the original optimum is 99, below the cutoff. Cutoff and bound reporting need an original-problem-valid
     bound.

  2. [P2] Primal simplex compares raw work counts against scaled work units.
     cpp/src/dual_simplex/primal.cpp:966, cpp/src/dual_simplex/primal.cpp:975

     The new checks compare the raw operation estimate directly with settings.work_limit; elsewhere, work units are raw estimates divided by 1e8. Primal cleanup also
     receives the dual solver’s raw accumulator.

     A reproduction with work limit 100 stopped at approximately 1.324e-5 actual work units, despite otherwise solving quickly. Finite work limits can therefore
     terminate solves prematurely.

  3. [P2] The new primal method cannot be serialized through gRPC.
     cpp/src/grpc/codegen/generated/generated_enum_converters_settings.inc:29

     The PR adds method_t::Primal = 4, but the gRPC registry/schema/converters still support only the previous methods. Serializing primal settings throws Unknown
     method_t: 4. The protocol mapping needs updating.
```

chris-maes:
@mlubin Thanks. The cutoff issue has been around for a while and is known.  I will work on a fix for that. We can discuss whether it should go into this PR. 

I'm not sure I understand the complaint about primal simplex and raw vs scaled work limits. A subsequent PR uses primal simplex work limits and work correctly. I can look into it. But more details would be helpful here. 

Changes for gRPC are beyond the scope of this PR. Maybe @tmckayus can help with updating gRPC after this PR lands.

cpp/src/dual_simplex/basis_updates.cpp
```diff
@@ -2009,6 +2009,34 @@ i_t basis_update_mpf_t<i_t, f_t>::u_solve(sparse_vector_t<i_t, f_t>& rhs) const
...
+}
+
+// Sparse-in/sparse-out overload of u_multiply. Same semantics as the dense version.
+template <typename i_t, typename f_t>
+void basis_update_mpf_t<i_t, f_t>::u_multiply(const sparse_vector_t<i_t, f_t>& x,
+                                              sparse_vector_t<i_t, f_t>& y) const
+{
+  const i_t m = L0_.m;
+  // Scatter x into a dense workspace, compute U0 * x, gather back to sparse.
+  std::vector<f_t> x_dense;
```
nguidotti: Is converting to dense, doing the matvec and then converting back to dense faster than working directly with the sparse vector?

nguidotti (commented):
Congrats for the performance improvements! I suggest adding some tests to exercise the primal simplex and the perturbation work

cpp/src/dual_simplex/scaling.cpp
```diff
@@ -253,6 +253,39 @@ i_t scaling(const lp_problem_t<i_t, f_t>& unscaled,
     return 0;
   }
 
+  // MIP performs integer-aware row scaling before presolve, while QP and SOCP
+  // use the Ruiz path above. Apply this simpler equilibration only to LPs.
+  const bool use_lp_row_scaling =
+    !settings.inside_mip && unscaled.second_order_cone_dims.empty() && unscaled.Q.n == 0;
+  if (use_lp_row_scaling) {
+    csr_matrix_t<i_t, f_t> Arow(0, 0, 0);
```
mlubin: Consider looping through the CSC copy of the matrix to accumulate the row norms instead of making a new copy CSR of the matrix.

cpp/src/dual_simplex/primal.cpp
```diff
@@ -352,47 +808,64 @@ primal_status_t primal_phase2(i_t phase,
...
+  assert(lp.rhs.size() == m);
 
-  set_primal_variables_on_bounds(lp, settings, z, vstatus, x);
+  std::vector<f_t>& x = sol.x;
+  std::vector<f_t>& y = sol.y;
+  std::vector<f_t>& z = sol.z;
 
-  const f_t init_dual_inf = dual_infeasibility(lp, vstatus, z);
-  settings.log.printf("Initial dual infeasibility %e\n", init_dual_inf);
+  std::vector<f_t> incoming_x                     = x;
```
mlubin: (Pre-existing) This vector appears to be unused.

cpp/src/dual_simplex/primal.cpp
```diff
@@ -352,47 +808,64 @@ primal_status_t primal_phase2(i_t phase,
...
+  std::vector<variable_status_t> incoming_vstatus = vstatus;
```
mlubin: (Pre-existing) This vector appears to be unused.

cpp/src/dual_simplex/primal.cpp
```diff
@@ -352,47 +808,64 @@ primal_status_t primal_phase2(i_t phase,
...
+  work_estimate += 2.0 * n;
+  settings.log.printf("Primal Simplex\n");
+  settings.log.printf("Pricing: %s\n", settings.primal_pricing == 1 ? "Devex" : "Dantzig");
+  // Nonbasics must be on their bounds before forming B x_B = b - A_N x_N.
+  // Setting them after the solve leaves ||A*x - b|| large whenever x_N != 0.
+  set_primal_variables_on_bounds(lp, settings, vstatus, x, work_estimate);
 
   std::vector<f_t> rhs = lp.rhs;
```
mlubin: This block seems to be a duplication of `compute_basic_primal_variables`.

cpp/src/dual_simplex/phase2.cpp
```diff
@@ -2324,61 +2459,300 @@ f_t amount_of_perturbation(const lp_problem_t<i_t, f_t>& lp, const std::vector<f
   return perturbation;
 }
 
+// Attempt to remove perturbation at optimality of the perturbed problem.
+// Returns:
+//   0 (OPTIMAL)        - perturbation fully removed, solution is optimal for original problem
```
mlubin, line 2464: Consider using an enum instead 0,1,2.

cpp/src/dual_simplex/phase2.cpp
```diff
@@ -2324,61 +2459,300 @@ f_t amount_of_perturbation(const lp_problem_t<i_t, f_t>& lp, const std::vector<f
...
                         const lp_problem_t<i_t, f_t>& lp,
                         const simplex_solver_settings_t<i_t, f_t>& settings,
                         basis_update_mpf_t<i_t, f_t>& ft,
                         const std::vector<f_t>& objective,
-                        const std::vector<i_t>& basic_list,
-                        const std::vector<i_t>& nonbasic_list,
-                        const std::vector<variable_status_t>& vstatus,
+                        // Primal cleanup below pivots, so the basis, the statuses
+                        // and the iteration count are updated in place.
+                        std::vector<i_t>& basic_list,
```
mlubin: ft, basic_list, nonbasic_list, max_val, and y aren't used in this function.

cpp/src/dual_simplex/phase2.cpp
```diff
@@ -2324,61 +2459,300 @@ f_t amount_of_perturbation(const lp_problem_t<i_t, f_t>& lp, const std::vector<f
...
 void prepare_optimality(i_t info,
                         f_t orig_primal_infeas,
                         const lp_problem_t<i_t, f_t>& lp,
                         const simplex_solver_settings_t<i_t, f_t>& settings,
                         basis_update_mpf_t<i_t, f_t>& ft,
                         const std::vector<f_t>& objective,
-                        const std::vector<i_t>& basic_list,
-                        const std::vector<i_t>& nonbasic_list,
-                        const std::vector<variable_status_t>& vstatus,
+                        // Primal cleanup below pivots, so the basis, the statuses
```
mlubin: Is this comment accurate? is there pivoting below?

cpp/src/dual_simplex/phase2.cpp
```diff
@@ -2324,61 +2459,300 @@ f_t amount_of_perturbation(const lp_problem_t<i_t, f_t>& lp, const std::vector<f
...
+                        // and the iteration count are updated in place.
+                        std::vector<i_t>& basic_list,
+                        std::vector<i_t>& nonbasic_list,
+                        std::vector<variable_status_t>& vstatus,
```
mlubin: vstatus can be const&

cpp/src/dual_simplex/phase2.cpp
```diff
@@ -2324,61 +2459,300 @@ f_t amount_of_perturbation(const lp_problem_t<i_t, f_t>& lp, const std::vector<f
...
                         int phase,
                         f_t start_time,
                         f_t max_val,
-                        i_t iter,
+                        f_t& work_estimate,
```
mlubin: work_estimate and iter aren't modified, can be taken by value

mlubin:
> I'm not sure I understand the complaint about primal simplex and raw vs scaled work limits. A subsequent PR uses primal simplex work limits and work correctly. I can look into it. But more details would be helpful here.

It looks like the primal and dual simplex codes treat the work limits with different scaling. Dual divides by 1e8 and primal doesn't.
```
  // Dual: report converted work, then check the context total.
  context->record_work_sync_on_horizon(
      (phase2_work_estimate - last_work_reported) / 1e8);
  last_work_reported = phase2_work_estimate;

  if (context->global_work_units_elapsed >= settings.work_limit)
      return dual_status_t::WORK_LIMIT;
```
```
  // Primal: compare the unconverted operation count directly.
  if (work_estimate > settings.work_limit)
      return primal_status_t::WORK_LIMIT;
```

cpp/src/dual_simplex/phase2.cpp
```diff
@@ -3122,13 +3683,17 @@ dual_status_t dual_phase2_with_advanced_basis(i_t phase,
...
         phase2_work_estimate += bfrt.work_estimate();
         if (entering_index == RATIO_TEST_NUMERICAL_ISSUES) {
           settings.log.printf("Numerical issues encountered in ratio test.\n");
           return dual_status_t::NUMERICAL;
         }
-        timers.bfrt_time += timers.stop_timer();
+        timers.bfrt_time += timers.stop_timer(phase2_work_estimate + ft.work_estimate());
+        // BFRT diagnostics
+        timers.bfrt_calls++;
+        if (step_length == 0.0) { timers.bfrt_zero_steps++; }
```
mlubin: Agent reports that `step_length` may be uninitialized at this point. "step_length is uninitialized when passed into BFRT. If no breakpoints exist, BFRT returns without assigning it"

cpp/src/dual_simplex/primal.cpp
```diff
@@ -403,150 +876,608 @@ primal_status_t primal_phase2(i_t phase,
...
+          // After compute_delta_z and update_z, delta_z[j] still holds the raw
+          // pivot row entries (update_z multiplies by dual_step_length into z, not delta_z)
+          for (i_t k = 0; k < n - m; ++k) {
+            const i_t j         = nonbasic_list[k];
+            const f_t a_j       = delta_z[j];
+            const f_t candidate = (a_j * a_j / pivot_sq) * w_enter;
+            if (candidate > devex_weight[j]) { devex_weight[j] = candidate; }
+          }
+          // Weight for leaving variable (now nonbasic)
+          devex_weight[leaving_index] = std::max(1.0 / pivot_sq, f_t(1e-4));
```
mlubin, line 1315: Astra is flagging an inconsistency in the devex weights but I don't fully follow. Copying here in case it's a real issue:
```
 The leaving-variable weight update drops the pre-pivot entering weight. For pivot (\alpha), the new leaving edge is the old entering edge divided by (-\alpha). Its squared reference norm therefore scales by (1/\alpha^2). The code uses w_enter when updating
  other weights, but assigns the leaving weight just 1/pivot_sq.

  A two-pivot example:

  minimize -2a - b
  a - 2b + s0 = 1
  a -  b + s1 = 2
  a, b, s0, s1 >= 0; b <= 2

  Start with basis (s0, s1), solution (a,b,s0,s1)=(0,0,1,2), reference set {a,b}, and unit weights.

  1. a enters, s0 leaves, pivot 1. The solution becomes (1,0,0,1), and the code updates w_b to 4.
  2. b enters, s1 leaves, pivot 1. The entering edge is (2,1,0,-1); the new leaving edge is its negative. Both have squared reference norm 2²+1²=5. Propagating the stored approximation would give w_s1=4, but the code assigns 1.

  Resetting devex_weight[b]=1 afterward does not resolve this: b is now basic, so that resets an unused pricing slot, not the reference framework. The leaving weight must use the old entering edge’s weight, or its recomputed reference norm, consistently with
  the other updates.
```

cpp/src/pdlp/solve.cu
```diff
@@ -602,9 +602,12 @@ std::tuple<simplex::lp_solution_t<i_t, f_t>, simplex::lp_status_t, f_t, f_t, f_t
   f_t norm_rhs            = vector_norm2<i_t, f_t>(user_problem.rhs);
 
   simplex::simplex_solver_settings_t<i_t, f_t> dual_simplex_settings;
-  dual_simplex_settings.time_limit      = settings.time_limit;
-  dual_simplex_settings.iteration_limit = settings.iteration_limit;
-  dual_simplex_settings.concurrent_halt = settings.concurrent_halt;
+  dual_simplex_settings.time_limit           = settings.time_limit;
+  dual_simplex_settings.iteration_limit      = settings.iteration_limit;
+  dual_simplex_settings.concurrent_halt      = settings.concurrent_halt;
+  dual_simplex_settings.initial_perturbation = settings.initial_perturbation;
```
mlubin, line 644: These new options aren't piped through to PDLP and barrier, which impacts crossover if used.

mlubin (commented):
Agreed with @nguidotti, since we're exposing the primal algorithm it should have a similar level of coverage in unit tests as dual simplex.

cpp/src/dual_simplex/basis_updates.cpp
chris-maes: Thanks for catching this. This was a stub left in. I made it work with a sparse vector. It leads to a small performance improvement on a large problem.

chris-maes:
> It looks like the primal and dual simplex codes treat the work limits with different scaling. Dual divides by 1e8 and primal doesn't.
> 
> ```
>   // Dual: report converted work, then check the context total.
>   context->record_work_sync_on_horizon(
>       (phase2_work_estimate - last_work_reported) / 1e8);
>   last_work_reported = phase2_work_estimate;
> 
>   if (context->global_work_units_elapsed >= settings.work_limit)
>       return dual_status_t::WORK_LIMIT;
> ```
> 
> ```
>   // Primal: compare the unconverted operation count directly.
>   if (work_estimate > settings.work_limit)
>       return primal_status_t::WORK_LIMIT;
> ```

Ah. This is just different use of the same work estimates. We use work estimates in two different ways: 

1) Primal and Dual Simplex now support WORK_LIMIT which allows internal algorithms to set a limit on the work estimate.
2) Dual simplex passes its work estimate (divided by 1e8) into record_work_sync_on_horizon for use with deterministic MIP. 

I can talk with @aliceb-nv about implementing 2 inside Primal Simplex as well. But this only affects the deterministic MIP path. Primal clean ups inside the branch and bound nodes should be extremely rare (since we don't apply much perturbation).  Maybe I can address this in a follow up PR.

cpp/src/dual_simplex/primal.cpp
chris-maes: Thanks. Fixed.

cpp/src/dual_simplex/primal.cpp
chris-maes: Thanks. Fixed.

cpp/src/dual_simplex/phase2.cpp
chris-maes: Thanks. Fixed.

cpp/src/dual_simplex/phase2.cpp
chris-maes: No. This is left over from a refactor. Fixed.

cpp/src/dual_simplex/phase2.cpp
chris-maes: Thanks. Fixed.

cpp/src/dual_simplex/scaling.cpp
chris-maes: Thanks. Fixed.

cpp/src/dual_simplex/primal.cpp
chris-maes: Ah yes. Good catch. Fixed. Thanks.

cpp/src/dual_simplex/phase2.cpp
chris-maes: Thanks. Guarded the step_length check. Fixed.

cpp/src/pdlp/solve.cu
chris-maes: Ok. Fixed. Thanks.

mlubin:
> Primal and Dual Simplex now support WORK_LIMIT which allows internal algorithms to set a limit on the work estimate.

Is there an inconsistency in the units used for settings.work_limit? It looks like dual checks the equivalent of `work_estimate / 1e8 > settings.work_limit` while primal checks `work_estimate > settings.work_limit`. Is that intended?

chris-maes:
> > Primal and Dual Simplex now support WORK_LIMIT which allows internal algorithms to set a limit on the work estimate.
> 
> Is there an inconsistency in the units used for settings.work_limit? It looks like dual checks the equivalent of `work_estimate / 1e8 > settings.work_limit` while primal checks `work_estimate > settings.work_limit`. Is that intended?

Everything is as intended. The deterministic path divides the work_estimate by 1e8 for syncing threads (@aliceb-nv can explain why, I think it's just a way to have the units correspond to roughly 1 second of work). The work limit I added is based on the actual work estimates. This is because you often want to base your limit on how much work a known operation did (e.g. the work done in the root LP solve). They are just two different ways of using the work estimates.

mlubin (approved):

ramakrishnap-nv (approved):
May be a follow-up is needed to add constants to docs, but rest looks good.

nguidotti (approved):

PR #1892, opened 2026-09-11
add basic http proxy server that forwards to grpc

tmckayus:
This initial version of the server is LP/MIP only. VRP and additional features will be added in a series of changes.

The server was tested locally by taking the current LP/MIP tests from cuopt_sh CLI tests from CI and running them against the proxy server, plus additional tests to cover the entire API surface of the proxy server. 67 tests in all.  Those tests are not include here -- in a future PR, once the proxy server is fully implemented, we'll switch the CLI tests in CI to run against the proxy server / grpc server combination instead of the legacy http server.

### DIFF
--- a/python/cuopt_server/cuopt_server/cuopt_proxy.py
+++ b/python/cuopt_server/cuopt_server/cuopt_proxy.py
@@ -1,0 +1,182 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""gRPC-backed HTTP proxy for legacy LP/MILP clients (Series C, C10).
+
+This process speaks the self-hosted HTTP API and forwards solves to
+``cuopt_grpc_server``. It does not run a local job queue or worker pool.
+"""
+
+import argparse
+import logging
+import os
+import sys
+
+
+import cuopt_server.utils.settings as settings
+from cuopt_server._version import __version__
+from cuopt_server.utils.logutil import message_init
+
+log_fmt = "%(asctime)s.%(msecs)03d %(levelname)s %(message)s"
+date_fmt = "%Y-%m-%d %H:%M:%S"
+
+
+def parse_args(argv=None):
+    ip = os.environ.get("CUOPT_SERVER_IP", "0.0.0.0")
+    port = os.environ.get("CUOPT_SERVER_PORT", 8000)
+    grpc_host = os.environ.get("CUOPT_GRPC_HOST", "127.0.0.1")
+    grpc_port = os.environ.get("CUOPT_GRPC_PORT", 5001)
+    log_level = os.environ.get("CUOPT_SERVER_LOG_LEVEL", "info")
+    log_file = os.environ.get("CUOPT_SERVER_LOG_FILE", "")
+    datadir = os.environ.get("CUOPT_DATA_DIR", "")
+    resultdir = os.environ.get("CUOPT_RESULT_DIR", "")
+    maxresult = os.environ.get("CUOPT_MAX_RESULT", 250)
+    resultmode = os.environ.get("CUOPT_RESULT_MODE", "644")
+    ssl_certfile = os.environ.get("CUOPT_SSL_CERTFILE", "")
+    ssl_keyfile = os.environ.get("CUOPT_SSL_KEYFILE", "")
+
+    levels = {
+        "critical": logging.CRITICAL,
+        "error": logging.ERROR,
+        "warning": logging.WARNING,
+        "info": logging.INFO,
+        "debug": logging.DEBUG,
+    }
+
+    parser = argparse.ArgumentParser(
+        prog="cuopt_proxy",
+        description=(
+            "HTTP proxy for legacy cuOpt LP/MILP clients. "
+            "Forwards jobs to cuopt_grpc_server."
+        ),
+    )
+    parser.add_argument(
+        "-i",
+        "--ip",
+        type=str,
+        help="Bind address (CUOPT_SERVER_IP)",
+        default=ip,
+    )
+    parser.add_argument(
+        "-p",
+        "--port",
+        type=int,
+        help="HTTP listen port (CUOPT_SERVER_PORT)",
+        default=int(port),
+    )
+    parser.add_argument(
+        "--grpc-host",
+        type=str,
+        help="cuopt_grpc_server host (CUOPT_GRPC_HOST)",
+        default=grpc_host,
+    )
+    parser.add_argument(
+        "--grpc-port",
+        type=int,
+        help="cuopt_grpc_server port (CUOPT_GRPC_PORT)",
+        default=int(grpc_port),
+    )
+    parser.add_argument(
+        "-l",
+        "--log-level",
+        type=str,
+        choices=list(levels.keys()),
+        help="Log level (CUOPT_SERVER_LOG_LEVEL)",
+        default=log_level,
+    )
+    parser.add_argument(
+        "-f",
+        "--log-file",
+        type=str,
+        help="Log filename (CUOPT_SERVER_LOG_FILE)",
+        default=log_file,
+    )
+    parser.add_argument(
+        "-d",
+        "--data-dir",
+        type=str,
+        help="CUOPT_DATA_DIR for CUOPT-DATA-FILE uploads",
+        default=datadir,
+    )
+    parser.add_argument(
+        "-r",
+        "--result-dir",
+        type=str,
+        help="CUOPT_RESULT_DIR for large result files",
+        default=resultdir,
+    )
+    parser.add_argument(
+        "-m",
+        "--max-result",
+        type=int,
+        help="Result size threshold in KB (CUOPT_MAX_RESULT)",
+        default=int(maxresult),
+    )
+    parser.add_argument(
+        "-mo",
+        "--mode",
+        type=str,
+        help="Octal mode for result files (CUOPT_RESULT_MODE)",
+        default=resultmode,
+    )
+    parser.add_argument(
+        "--ssl-certfile",
+        type=str,
+        help="SSL certificate file (CUOPT_SSL_CERTFILE)",
+        default=ssl_certfile,
+    )
+    parser.add_argument(
+        "--ssl-keyfile",
+        type=str,
+        help="SSL key file (CUOPT_SSL_KEYFILE)",
+        default=ssl_keyfile,
+    )
+    args = parser.parse_args(argv)
+    args._log_level_int = levels[args.log_level]
+    return args
+
+
+def _configure_logging(args):
+    message_init()
+    handlers = []
+    if args.log_file:
+        handlers.append(logging.FileHandler(args.log_file))
+    else:
+        handlers.append(logging.StreamHandler(sys.stdout))
+    logging.basicConfig(
+        level=args._log_level_int,
+        format=log_fmt,
+        datefmt=date_fmt,
+        handlers=handlers,
+        force=True,
+    )
+
+
+def main(argv=None):
+    args = parse_args(argv)
+    _configure_logging(args)
+    logging.info(f"cuOpt HTTP proxy {__version__}")
+
+    if args.data_dir:
+        settings.set_data_dir(args.data_dir)
+    settings.set_result_dir(args.result_dir, args.max_result, args.mode)
+
+    from cuopt.grpc.linear_programming import Client
+
+    from cuopt_server.proxy_webserver import run_server, set_grpc_client
+
+    logging.info(
+        f"Connecting to cuopt_grpc_server at {args.grpc_host}:{args.grpc_port}"
+    )
+    set_grpc_client(Client(args.grpc_host, args.grpc_port))
+    run_server(
+        args.ip,
+        args.port,
+        args.log_level,
+        args.ssl_certfile,
+        args.ssl_keyfile,
+    )
+
+
+if __name__ == "__main__":
+    main()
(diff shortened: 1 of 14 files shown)


python/cuopt_server/cuopt_server/proxy_webserver.py
```diff
@@ -0,0 +1,752 @@
...
+        _require_uuid(id)
+        if meta is not None and meta.get("validation_only"):
+            return encode(meta["validation_result"], accept, job_result=True)
+        client = get_grpc_client()
+        status = client.status(id)
+        if _is_status(status, "NOT_FOUND"):
+            raise HTTPException(status_code=404, detail=f"id {id} not found")
+        if _is_status(status, "QUEUED", "PROCESSING"):
+            return encode({"reqId": id}, accept)
+        sol = client.result(
```
ramakrishnap-nv, line 581: `client.result()` raises `GrpcError` for FAILED/CANCELLED jobs (per its docstring), but only NOT_FOUND/QUEUED/PROCESSING are special-cased above. A client polling a failed/cancelled solve will get an opaque 500 instead of a structured error response.

python/cuopt_server/cuopt_server/proxy_webserver.py
```diff
@@ -0,0 +1,752 @@
...
+    accept: str = Header(default="application/json"),
+):
+    try:
+        accept = _resolve_accept(accept)
+        _require_uuid(id)
+        # gRPC deletes logs with the job. HTTP DELETE log is a no-op 200.
+        return Response(status_code=200)
+    except HTTPException as e:
+        return encode(http_exception_handler(e), accept)
+    except Exception as e:
```
ramakrishnap-nv, line 411: This handler returns `exception_handler(e)` directly, skipping the `encode(...)` wrapper every other endpoint uses. On error, msgpack/zlib clients silently get plain JSON instead of their negotiated content type.

python/cuopt_server/cuopt_server/utils/http_codec.py
```diff
@@ -113,6 +113,23 @@ def deserialize(ctype, buf):
...
+
+    Callers allocate ``buf`` from ``Content-Length`` (a ``bytearray`` or a
+    shared-memory view) so Starlette does not assemble a second copy via
+    ``request.body()``. Pydantic validation happens later, after
+    :func:`deserialize`.
+    """
+    pos = 0
+    try:
+        async for chunk in request.stream():
+            buf[pos : pos + len(chunk)] = chunk
```
ramakrishnap-nv, line 133: `buf` is sized from the client-supplied Content-Length, but slice-assignment silently grows the bytearray if the actual stream is larger — so Content-Length isn't actually enforced as a bound, contrary to the docstring above.

python/cuopt_server/cuopt_server/proxy_webserver.py
```diff
@@ -0,0 +1,752 @@
...
+                    "validation_result": envelope,
+                },
+            )
+            return encode({"reqId": job_id}, accept)
+
+        client = get_grpc_client()
+        job_id = client.submit(
+            data_model,
+            solver_settings,
+            enable_incumbents=bool(incumbent_solutions),
```
ramakrishnap-nv: Passing an explicit `enable_incumbents=bool(...)` bypasses the underlying `mip and settings.get_mip_callbacks()` gate in `grpc_client.pyx`. An LP-only job with `incumbent_solutions=true` will now enable incumbent tracking on the gRPC worker even though it's MIP-only functionality.

python/cuopt_server/cuopt_server/proxy_webserver.py
```diff
@@ -0,0 +1,752 @@
...
+        meta = _get_job(id)
+        if meta is not None and meta.get("validation_only"):
+            return encode(
+                [{"solution": [], "cost": None, "bound": None}], accept
+            )
+        client = get_grpc_client()
+        status = client.status(id)
+        if _is_status(status, "NOT_FOUND"):
+            raise HTTPException(status_code=404, detail=f"id {id} not found")
+        from_index = 0 if meta is None else meta.get("incumbent_next_index", 0)
```
ramakrishnap-nv: Read (here) → gRPC fetch → write-back (`_update_job` below) isn't atomic across the two separate `_jobs_lock` acquisitions. Two concurrent polls for the same job can read the same `from_index` and both deliver duplicate incumbents to their callers.

python/cuopt_server/cuopt_server/proxy_webserver.py
```diff
@@ -0,0 +1,752 @@
...
+        if file_path:
+            data = load_optimization_file(file_path, warnings)
+        else:
+            now = time.time()
+            buf = bytearray(sz)
+            await get_data(buf, request)
+            logging.debug(f"time to receive data {time.time() - now}")
+            data = deserialize(ctype, buf)
+
+        lp_data, data_model, solver_settings = _prepare_lp(data, warnings)
```
ramakrishnap-nv: This CPU-bound LP validation/conversion runs synchronously inline in an `async def` handler with no executor/thread offload, unlike the legacy server which does this work in a separate worker process. Under load this blocks the single event loop, serializing health checks/status polls behind conversion work.

python/cuopt_server/cuopt_server/proxy_webserver.py
```diff
@@ -0,0 +1,752 @@
...
+app = FastAPI(
+    title="NVIDIA cuOpt HTTP proxy",
+    version=__version__,
+    docs_url="/cuopt/docs",
+    redoc_url="/cuopt/redoc",
+    openapi_url="/cuopt/openapi.json",
+)
+
+_grpc_client = None
+_jobs = {}
```
ramakrishnap-nv, line 90: `_jobs` has no TTL/eviction — leaks unboundedly for clients that never call DELETE. It's also process-local, so `--workers N>1` or a restart drops metadata (variable names, Accept format, incumbent cursor) for in-flight jobs on other workers.

python/cuopt_server/cuopt_server/utils/client_version.py
```diff
@@ -0,0 +1,32 @@
...
+
+import logging
+import os
+
+from cuopt_server._version import __version__
+
+
+def check_client_version(client_vers):
+    logging.debug(f"client_vers is {client_vers} in check")
+    if os.environ.get("CUOPT_CHECK_CLIENT", True) in ["True", True]:
```
ramakrishnap-nv: Only the literal string `"True"` is recognized — `"true"`, `"1"`, `"yes"` etc. silently disable the version check. Pre-existing behavior, but this module is now shared by both the legacy server and the new proxy, so the blast radius doubles.

python/cuopt_server/cuopt_server/proxy_webserver.py
```diff
@@ -0,0 +1,752 @@
...
+        )
+        return encode({"reqId": job_id}, accept)
+
+    except (RequestValidationError, ValidationError) as e:
+        return encode(validation_exception_handler(e), accept)
+
+    except HTTPException as e:
+        return encode(http_exception_handler(e), accept)
+
+    except Exception as e:
```
ramakrishnap-nv, line 799: This try/except boilerplate (HTTPException → encode, Exception → encode) is duplicated verbatim across ~10 endpoints instead of going through the already-registered `@app.exception_handler(Exception)`. That drift is exactly what let the `deletesolverlogs` handler above skip the `encode()` wrapper — worth factoring into a shared decorator.

python/cuopt_server/cuopt_server/utils/http_codec.py
```diff
@@ -113,6 +113,23 @@ def deserialize(ctype, buf):
...
+            pos = pos + len(chunk)
+    except Exception:
+        logging.debug("exception in get_data", exc_info=True)
```
ramakrishnap-nv: This was a `print` before and is now `logging.debug` — truncated/disconnected uploads become invisible at the default `info` log level, surfacing only as a confusing downstream 422 with no trace of the real cause.

python/cuopt_server/cuopt_server/cuopt_proxy.py
```diff
@@ -0,0 +1,182 @@
...
+
+import cuopt_server.utils.settings as settings
+from cuopt_server._version import __version__
+from cuopt_server.utils.logutil import message_init
+
+log_fmt = "%(asctime)s.%(msecs)03d %(levelname)s %(message)s"
+date_fmt = "%Y-%m-%d %H:%M:%S"
+
+
+def parse_args(argv=None):
```
ramakrishnap-nv: This duplicates `cuopt_service.py`'s CLI flags, env var names/defaults, and log-level mapping almost verbatim. Worth sharing a common arg-parsing helper so the two entry points don't silently drift (e.g. if `CUOPT_MAX_RESULT`'s default changes).

ramakrishnap-nv (changes requested):
Automated review pass — flagging a few correctness/robustness issues and some maintenance risk from duplicated boilerplate. Full write-up (including missing test cases) available on request; these are the inline, code-anchored points.

python/cuopt_server/cuopt_server/proxy_webserver.py
tmckayus: fixed.

python/cuopt_server/cuopt_server/proxy_webserver.py
tmckayus: fixed.

python/cuopt_server/cuopt_server/utils/http_codec.py
tmckayus: fixed.

python/cuopt_server/cuopt_server/proxy_webserver.py
tmckayus: fixed.

python/cuopt_server/cuopt_server/proxy_webserver.py
tmckayus: fixed.

python/cuopt_server/cuopt_server/proxy_webserver.py
tmckayus: Fixed. Moved deserialization, conversion, and submit rpc into a thread with asyncio.to_thread

python/cuopt_server/cuopt_server/proxy_webserver.py
tmckayus: this is no different from the legacy server, same behavior. We also don't anticipate multiple uvicorn workers in the proxy server, it is just a thin sidecar that forwards information to gRPC.

python/cuopt_server/cuopt_server/utils/client_version.py
tmckayus: added "true" and True to match legacy server.

python/cuopt_server/cuopt_server/proxy_webserver.py
tmckayus: deferred refactor

python/cuopt_server/cuopt_server/utils/http_codec.py
tmckayus: fixed, changed to log at warn level

python/cuopt_server/cuopt_server/cuopt_proxy.py
tmckayus: deferred cleanup. The legacy server is meant to disappear anyway, so a refactor is not out of the question but ultimately the problem is self-correcting. And we expect the flags to be stable.

python/cuopt_server/cuopt_server/proxy_webserver.py
```diff
@@ -0,0 +1,834 @@
...
+            ctype,
+            buf,
+            file_path,
+            warnings,
+            validation_only,
+            incumbent_solutions,
+            solver_logs,
+            accept,
+            result_file,
+        )
```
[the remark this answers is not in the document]
tmckayus: deferred for a follow up

python/cuopt_server/cuopt_server/proxy_webserver.py
```diff
@@ -0,0 +1,752 @@
...
+                status_code=422, detail="Data length is zero and reqId not set"
+            )
+
+        if ctype not in [mime_json, mime_msgpack, mime_zlib, mime_pickle]:
+            raise HTTPException(
+                status_code=415,
+                detail=f"Unsupported Content-Type value {ctype}, "
+                f"supported values are "
+                f"{[mime_json, mime_msgpack, mime_zlib, mime_pickle]}",
+            )
```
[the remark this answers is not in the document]
tmckayus: skipped, same situation as the legacy server. SafeUnpickler is the guard against bad loads.

python/cuopt_server/cuopt_server/cuopt_proxy.py
```diff
@@ -0,0 +1,182 @@
...
+    settings.set_result_dir(args.result_dir, args.max_result, args.mode)
+
+    from cuopt.grpc.linear_programming import Client
+
+    from cuopt_server.proxy_webserver import run_server, set_grpc_client
+
+    logging.info(
+        f"Connecting to cuopt_grpc_server at {args.grpc_host}:{args.grpc_port}"
+    )
+    set_grpc_client(Client(args.grpc_host, args.grpc_port))
```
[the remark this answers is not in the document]
tmckayus: Supported deployment is proxy and cuopt_grpc_server in the same container/pod (loopback). We do not treat a non-localhost hostname as “must TLS”; that breaks in-cluster DNS. Operators who split them across a network can already set CUOPT_TLS_*. We will not enable gRPC TLS by default for the sidecar hop; we will if we ever document exposing the gRPC port.

ramakrishnap-nv (approved):

PR #1897, opened 2026-09-14
Don't abort the process when cudssSetThreadingLayer fails

### LINKED ISSUE #1896: Barrier/cuDSS: a failed cudssSetThreadingLayer aborts the whole process instead of degrading gracefully
## Summary

`sparse_cholesky_cudss_t`'s constructor calls `cudssSetThreadingLayer()` via `CUDSS_CALL_AND_CHECK_EXIT`, which `exit()`s the whole process on any failure (`cpp/src/barrier/sparse_cholesky.cuh`). Since this is a native abort, it bypasses any surrounding try/except in embedding applications (Python, server, etc.) entirely — a missing/misconfigured threading library takes the whole host process down instead of just cuDSS running single-threaded.

## Real-world report

A downstream application hit this in production on a DGX Spark GB10 host: a background GPU warmup call aborted the whole app container with
```
FAILED: CUDSS call ended unsuccessfully with status = 3, details: "cudssSetThreadingLayer"
```
right after startup completed — independently reproduced on a second DGX Spark host. The app installs `cuopt-cu12` via pip into its own custom base image (not NVIDIA's official runtime image), so this is a third distinct manifestation of the libgomp/cuDSS threading-layer fragility already tracked in #1219 (teardown race, fixed by #1883) and #1670/#1888 (NVIDIA's own Docker image missing `libgomp1`, fixed by #1888) — neither of those fixes this failure mode, since here the *first* call to `cudssSetThreadingLayer` fails outright (not a teardown race), on an image we don't control.

Per NVIDIA's cuDSS team (comment from `ES` on the report): status 3 means the threading library path was invalid or not found, and cuDSS logs a specific diagnostic (`[CUDSS][Error][cudssSetThreadingLayer] Failed to open library from <path>`) that should say why — that line isn't visible in the report we have, which may just be a log-capture gap (native stderr not captured by the app's logging setup) rather than truly absent.

Reproduced locally: forcing `CUDSS_THREADING_LIB` to an invalid path reproduces the identical error message and exit code (254) on current `main`.

## Proposed fix

1. Don't `exit()` the process when `cudssSetThreadingLayer` fails. Fall back to single-threaded cuDSS and log a clear, actionable warning (including the library path/env var that was tried) instead — a missing library should mean a slower solve, not an aborted job.
2. Consider surfacing cuDSS's own native log output more visibly (or capturing/forwarding it) so a failure like this is diagnosable without needing native log access on the host.

## References
- #1219 (dual-runtime teardown race, fixed by #1883)
- #1670 / #1888 (NVIDIA runtime image missing `libgomp1`)

ramakrishnap-nv:
Closes #1896. A missing/unresolvable cuDSS threading library currently `exit()`s the whole process, bypassing try/except in embedding apps. Falls back to single-threaded cuDSS with a warning instead. Verified against the real `26.8.0-cu12` image: reproduces the reported crash, and this fix turns it into a completed solve (exit 254 → 0). Added a regression test.


ramakrishnap-nv:
Fixed the stale-fixture test gap (removed the SetUp fixture before invoking the CLI, added --solution-file explicitly since the CLI only writes output when that's passed) and trimmed comments. Verified locally: all 8 CLI_TEST cases pass.

cpp/src/barrier/sparse_cholesky.cuh
```diff
@@ -259,9 +259,19 @@ class sparse_cholesky_cudss_t : public sparse_cholesky_base_t<i_t, f_t> {
     }
 
     if (cudss_mt_lib_file != nullptr) {
-      settings.log.printf("cuDSS Threading layer       : %s\n", cudss_mt_lib_file);
-      CUDSS_CALL_AND_CHECK_EXIT(
-        cudssSetThreadingLayer(handle, cudss_mt_lib_file), status, "cudssSetThreadingLayer");
+      // A missing/misconfigured threading library shouldn't take the whole process down (#1896).
```
chris-maes: No need to add a comment about what shouldn't happen in the code.

chris-maes (approved):
LGTM. Thanks for the fix.

ramakrishnap-nv:
Dropped the comment per Chris's feedback — the code and the runtime log message already make the fallback behavior clear.

cpp/src/barrier/sparse_cholesky.cuh
ramakrishnap-nv: Dropped it — the code and the runtime log message already make the fallback behavior clear.

PR #1898, opened 2026-09-14
simplify test for incumbent solution arrival during wait

tmckayus:
A test was added to ensure that incumbent processing threads in Python could receive solutions during a Client.wait() gRPC call. Previously, a C++ level wait was used with locked the GIL and prevented the thread from running.

The original test was too complex and fragile -- the test should only ever assert that at least 1 incumbent solution is received while wait() is pending. This change simplifies the test.


chris-maes (approved):
LGTM.

ramakrishnap-nv (approved):

Issue #1, opened 2025-04-10
[FEA] Add a C API for cuOpt LP and MIP Solver


afender:
Most optimization solvers in the ecosystem are written in C, and the others provide C interfaces for interoperability and integration into systems written in C/C++, Rust, Julia, and other languages.

PR #1905, opened 2026-09-15
Give cuDSS a threading layer built against cuOpt's own libgomp (wheels)

ramakrishnap-nv:
Unifies cuOpt's OpenMP runtime with cuDSS's (#1219), confirmed as the right approach by NVIDIA's cuDSS team.

#1875 showed Rocky 8's own GNU libgomp is too old (no OpenMP 5.0 detached-task support), so wheel builds now resolve and fetch the newest libgomp from conda-forge satisfying an explicit `>=9` floor (verified against conda-forge's own sha256, not the download itself), and compile/link against it, replacing LLVM libomp. Wheel builds also compile our own cuDSS threading layer (`libcudss_mtlayer_cuopt.so`) against that same libgomp, installed alongside `libcuopt.so` with `$ORIGIN` RPATH, replacing cuDSS's prebuilt `libcudss_mtlayer_gomp.so.0`. cuDSS and cuOpt now genuinely share one libgomp instance instead of two separate copies of the same flavor, verified via `/proc/<pid>/maps` during a live solve.

Conda is untouched (its libgomp is already shared naturally). Source builds using the conda dev environment also already get a matching libgomp. If threading setup fails for any reason, #1897's graceful fallback (single-threaded + log) still applies.

Also adds an explicit `libgomp >=9` version floor to `dependencies.yaml` / `conda/recipes/libcuopt/recipe.yaml`, and documents the `libgomp.so.1` / `CUDSS_THREADING_LIB` requirements in system-requirements.rst, faq.rst, and convex-settings.rst.


ramakrishnap-nv:
Found the CI failure cause: -DOpenMP_gomp_LIBRARY only points the compiler/linker at the fetched modern libgomp; auditwheel repair does its own separate dependency resolution over the system search path and, without LD_LIBRARY_PATH set, silently vendored Rocky 8's old system libgomp instead (confirmed: the vendored copy was missing OMP_5.0.1). Fixed by exporting LD_LIBRARY_PATH before the repair step.

Also surfaced a second, separate wrinkle: one test failure resolved against libraft's own separately-published wheel's bundled (old) libgomp, not ours. Re-testing to see if that's still an issue after this fix before digging further.

cpp/src/barrier/sparse_cholesky.cuh
```diff
@@ -21,8 +21,38 @@
...
 
+#include <dlfcn.h>
+#include <mutex>
+#include <string>
+
 namespace cuopt::mathematical_optimization::barrier {
 
+namespace detail {
+
+// Preload cuOpt's bundled modern libgomp under its literal soname before cuDSS's threading
```
mlubin: This seems fishy. What if you're running cuopt in a python process that called cuDSS for some other reason before loading cuOpt?

ci/utils/install_modern_libgomp.sh
```diff
@@ -0,0 +1,75 @@
...
+        ;;
+    *)
+        echo "Unsupported architecture for modern libgomp fetch: $(arch)" >&2
+        exit 1
+        ;;
+esac
+
+# Pinned for build reproducibility -- bump deliberately, not by tracking "latest".
+version="16.2.0"
+pkg="libgomp-${version}-${build}.conda"
```
mlubin: How long are outdated conda artifacts retained for?

ramakrishnap-nv:
Two replies:

**libgomp preload**: the RTLD_NOLOAD guard means if libgomp.so.1 is already loaded when our code runs (e.g. cuDSS was used for something else first) — that existing instance is honored, we skip our own preload entirely, no double-load, no override. Only when nothing has loaded it yet do we set ours as the process's one instance.

**conda-forge retention**: since we fetch-and-bundle at build time (not at install time), conda-forge's retention policy only matters for future CI rebuilds against this exact pin, not for anyone using an already-published wheel. If the pinned build ever ages out, the fix is just bumping the version/build/sha256 in the script — a known, low-cost maintenance item, not a runtime risk.

ramakrishnap-nv:
Added sha256 verification (from conda-forge's own repodata.json) before extraction — fixed.

Declining the "throw on preload failure" suggestion on sparse_cholesky.cuh though: the dlopen there is expected to fail harmlessly on conda/local-dev builds (no co-located libgomp.so.1 exists — CUOPT_BUNDLED_LIBGOMP is wheel-only), so throwing would break every non-wheel build. It's intentionally best-effort; cudssSetThreadingLayer's own failure path already degrades gracefully (#1897).

mlubin:
> that existing instance is honored, we skip our own preload entirely, no double-load, no override

Sure, but it sounds like we won't fix the cause of the crashes in that case, right?

Have we explored the fixes suggested in #1219?

> Unify OpenMP runtimes: Ensure cuDSS threading layer uses the same libgomp as the main cuOpt library (not the system one). This can be done by setting LD_LIBRARY_PATH or rebuilding cuDSS to bundle its own libgomp.

> Avoid dlclose in cudssDestroy: If cuDSS unloads its threading layer during destruction, this can corrupt shared OpenMP state. The threading layer should remain loaded for the lifetime of the process.

cpp/src/barrier/sparse_cholesky.cuh
ramakrishnap-nv: Fair pushback. To be precise about what's actually covered:

- #1219's suggested fix #2 ("avoid dlclose in cudssDestroy, threading layer should remain loaded for the process lifetime") is exactly what #1883's RTLD_NODELETE pin does — and that pins the mtlayer *plugin* itself, independent of which libgomp backs it. So the specific reproduced crash (dlclose-during-active-threads SIGSEGV) is fixed unconditionally, regardless of load order.
- #1219's suggested fix #1 ("unify OpenMP runtimes, same libgomp as main cuOpt library") is what this PR does — but only when our own code is first to touch libgomp.so.1. If something else in the process loaded a different one first, we back off rather than force a second instance, so unification doesn't happen for that run.

So you're right that the "two independent OpenMP thread pools" architectural risk isn't fully eliminated in that edge case — just the specific crash mechanism we've actually reproduced. Making the preload unconditional would close that gap but reintroduce the exact double-load risk your first comment flagged. No free lunch here; open to how you'd want to resolve that tension.

ci/utils/install_modern_libgomp.sh
ramakrishnap-nv: Since we fetch-and-bundle at build time (not install time), conda-forge's retention only matters for future CI rebuilds against this exact pin, not for anyone using an already-published wheel. If the pinned build ever ages out, the fix is bumping version/build/sha256 in the script — low-cost maintenance, not a runtime risk.

cpp/src/barrier/sparse_cholesky.cuh
ramakrishnap-nv: Else we can leave it open and don't need to bundle, in that case if anything is set or available it will be used, else it will run single thread with warning there is no multi thread library.

ci/utils/install_modern_libgomp.sh
mlubin: Introducing yet another reason why the CI can break without warning isn't low-cost. We should be proactive and know how long these builds will be available for.

ramakrishnap-nv: Fair — checked rather than assumed. A libgomp build from January 2020 (linux-64, 7.3.0-h24d8f2e_3.tar.bz2, still listed via the anaconda.org files API) returns HTTP 200 and downloads fine right now — 6+ years, no purge. conda-forge/anaconda.org doesn't appear to garbage-collect old builds. Given that, I'd keep the bundling approach rather than drop it over this specific risk.

cpp/src/barrier/sparse_cholesky.cuh
ramakrishnap-nv: Dropped the preload entirely — pushed. cuDSS now just resolves its threading library normally (no forced shared instance), and #1897 already covers the case where it can't find one gracefully. This PR is back to scoped to just compiling libcuopt.so against a modern, sha256-verified libgomp instead of LLVM libomp. Also documented libgomp1/CUDSS_THREADING_LIB as a runtime requirement in system-requirements.rst/faq.rst/convex-settings.rst.

docs/cuopt/source/faq.rst
```diff
@@ -102,6 +102,18 @@ General FAQ
 
    After installing, re-run ``pip install`` (or restart the Python process) and the import should succeed.
 
+.. dropdown:: Why does my MIP/LP solve run single-threaded, or log a "could not load cuDSS threading layer" warning?
+
+   cuDSS's threading layer (used by the barrier method) loads ``libgomp.so.1`` from the host at runtime; the cuOpt wheel does not bundle it. If the host doesn't provide it, cuDSS falls back to single-threaded execution and logs a warning instead of failing the solve.
+
+   Install your distribution's GNU OpenMP runtime package, typically ``libgomp1`` (Debian/Ubuntu) or ``libgomp`` (RHEL/Rocky/Fedora), then re-run the solve.
+
+   If ``libgomp.so.1`` is installed in a non-standard location, or you want cuDSS to use a specific threading-layer library, set the ``CUDSS_THREADING_LIB`` environment variable to its absolute path before running:
```
nguidotti: I think we should specify that cuOpt and cuDSS threading layer should match. Otherwise, we will run at the same all over again

ramakrishnap-nv: Good catch — added a warning that whatever CUDSS_THREADING_LIB points at must be built against the same OpenMP runtime cuOpt uses, or it reintroduces #1219. Also worth knowing: we're now prototyping the actual fix for the default case too (a custom cuDSS threading layer built against the same libgomp cuOpt links against, replacing cuDSS's prebuilt one for wheel builds) — verified locally via /proc/maps that it results in a genuinely single shared libgomp instance.

ramakrishnap-nv:
@coderabbitai review

ci/utils/install_modern_libgomp.sh
ramakrishnap-nv: conda-forge doesn't delete packages from Anaconda.org -- broken ones get relabeled `broken`, not removed, so the artifact itself isn't the risk. Fair point on the general failure surface though: latest commit resolves the version dynamically from conda-forge's API (always takes the newest build satisfying the `dependencies.yaml` floor, so we're never depending on one specific historical pin). If that lookup itself fails, the build fails loudly and can be re-run, same as any other transient CI failure -- kept it to one code path rather than adding a fallback pin as well.

docs/cuopt/source/faq.rst
```diff
@@ -102,6 +102,18 @@ General FAQ
 
    After installing, re-run ``pip install`` (or restart the Python process) and the import should succeed.
 
+.. dropdown:: Why does my MIP/LP solve run single-threaded, or log a "could not load cuDSS threading layer" warning?
+
+   cuDSS's threading layer (used by the barrier method) loads ``libgomp.so.1`` from the host at runtime; the cuOpt wheel does not bundle it. If the host doesn't provide it, cuDSS falls back to single-threaded execution and logs a warning instead of failing the solve.
```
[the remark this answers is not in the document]
ramakrishnap-nv: Updated -- distinguished pip (bundled, no host requirement) from conda/source (still needs host libgomp) instead of a blanket statement either way.

docs/cuopt/source/faq.rst
ramakrishnap-nv: Moved the 'must match' requirement to the top of the dropdown instead of only in the trailing warning box, and now also note it applies to CUDSS_THREADING_LIB overrides specifically since the wheel's own bundled libgomp isn't necessarily what a custom override links against.

docs/cuopt/source/convex-settings.rst
```diff
@@ -217,6 +217,8 @@ Barrier Solver Settings
 
 The following settings control the behavior of the barrier (interior-point) method:
 
+.. note:: The barrier method uses cuDSS for sparse Cholesky factorization, which requires the host to provide ``libgomp.so.1`` (GNU OpenMP runtime) for multi-threaded execution. If missing, the solve falls back to single-threaded cuDSS instead of failing. ``CUDSS_THREADING_LIB`` can point cuDSS at a specific threading-layer library, but it must be built against the same OpenMP runtime cuOpt itself uses, or the two can conflict. See :doc:`system-requirements` and :doc:`faq`.
```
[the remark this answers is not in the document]
ramakrishnap-nv: Confirmed by reading auditwheel's wheel_abi.py: it scans every ELF file in the wheel (not just ones reachable from the main extension's NEEDED graph), so libcudss_mtlayer_cuopt.so's own libgomp.so.1 reference gets discovered and relinked to the same vendored copy as libcuopt.so. Updated faq/system-requirements/convex-settings to say pip installs need no host libgomp, conda/source still do.

cpp/src/barrier/sparse_cholesky.cuh
```diff
@@ -21,8 +21,38 @@
...
+// layer loads, so cuDSS resolves the same instance instead of a missing/outdated host one
+// (#1219). Skips if any libgomp.so.1 is already loaded, to avoid mapping a second instance.
+inline void preload_bundled_libgomp()
+{
+  static std::once_flag once;
+  std::call_once(once, []() {
+    if (dlopen("libgomp.so.1", RTLD_NOW | RTLD_NOLOAD) != nullptr) { return; }
```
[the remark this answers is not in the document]
ramakrishnap-nv: The preload approach this was on has been fully reverted -- sparse_cholesky.cuh no longer has any dlopen/preload code. Replaced by the custom-threading-layer design (each side just uses the libgomp it's actually linked against, nothing dlopen'd or preloaded).

cpp/src/barrier/sparse_cholesky.cuh
ramakrishnap-nv: Good catch, and moot now -- the preload was reverted entirely. Current design doesn't preload anything into an existing process; cuOpt and the custom cuDSS threading layer just happen to link the same libgomp at build time, so there's no dependency on load order or who touched cuDSS first.

ci/utils/install_modern_libgomp.sh
ramakrishnap-nv: This actually happened -- api.anaconda.org failed TLS certificate verification in the Rocky 8 wheel-build image (conda.anaconda.org, which we already use for the download, works fine there). Switched the lookup to query conda-forge's repodata on that same already-working host instead of a second domain. Also fixed a real bug this surfaced: a failed lookup was silently continuing with empty variables instead of stopping the script.

ci/utils/install_modern_libgomp.sh
```diff
@@ -0,0 +1,107 @@
...
+min_major="$(grep -oP '(?<=- libgomp >=)[0-9]+' "${deps_yaml}" | head -1)"
+if [[ -z "${min_major}" ]]; then
+    echo "Could not find a 'libgomp >=N' floor in ${deps_yaml}" >&2
+    exit 1
+fi
+
+python3 -m pip install --quiet zstandard
+
+workdir="$(mktemp -d)"
+trap 'rm -rf "${workdir}"' EXIT
```
gforsyth:
```suggestion
```

This script is already re-implementing conda fetches, we can just let the tmpdir die on its own

ci/utils/install_modern_libgomp.sh
```diff
@@ -0,0 +1,107 @@
...
+    (fn, v) for fn, v in data["packages.conda"].items()
+    if v["name"] == "libgomp" and int(v["version"].split(".")[0]) >= min_major
+]
+if not candidates:
+    sys.exit(f"No libgomp >={min_major} build found in {repodata_path}")
+
+fn, v = max(candidates, key=lambda c: (tuple(map(int, c[1]["version"].split("."))), c[1]["timestamp"]))
+print(fn, v["sha256"])
+PYEOF
+)"; then
```
gforsyth: LLMs love these inlined code blocks.  I do not.  This should be split out into a separate script.

ci/utils/install_modern_libgomp.sh
```diff
@@ -0,0 +1,107 @@
...
+# The floor lives in dependencies.yaml (also used for conda/recipes/libcuopt/recipe.yaml), read
+# from there so it can't drift from the version resolved below.
+deps_yaml="$(dirname "${BASH_SOURCE[0]}")/../../dependencies.yaml"
+min_major="$(grep -oP '(?<=- libgomp >=)[0-9]+' "${deps_yaml}" | head -1)"
+if [[ -z "${min_major}" ]]; then
+    echo "Could not find a 'libgomp >=N' floor in ${deps_yaml}" >&2
+    exit 1
+fi
+
+python3 -m pip install --quiet zstandard
```
gforsyth:
```suggestion
python -m pip install --quiet zstandard
```

nit, but it's 2026 and we don't need to specify `python3` anymore

ci/utils/install_modern_libgomp.sh
```diff
@@ -0,0 +1,107 @@
...
+import zipfile
+
+import zstandard
+
+pkg_path, out_dir = sys.argv[1], sys.argv[2]
+zf = zipfile.ZipFile(pkg_path)
+pkg_name = next(n for n in zf.namelist() if n.startswith("pkg-"))
+raw = zf.read(pkg_name)
+tar_bytes = zstandard.ZstdDecompressor().decompress(raw, max_output_size=200 * 1024 * 1024)
+tarfile.open(fileobj=io.BytesIO(tar_bytes)).extractall(out_dir)
```
gforsyth: This should be part of the same external python file, not inlined

ci/utils/install_modern_libgomp.sh
```diff
@@ -0,0 +1,107 @@
...
+esac
+
+# The floor lives in dependencies.yaml (also used for conda/recipes/libcuopt/recipe.yaml), read
+# from there so it can't drift from the version resolved below.
+deps_yaml="$(dirname "${BASH_SOURCE[0]}")/../../dependencies.yaml"
+min_major="$(grep -oP '(?<=- libgomp >=)[0-9]+' "${deps_yaml}" | head -1)"
+if [[ -z "${min_major}" ]]; then
+    echo "Could not find a 'libgomp >=N' floor in ${deps_yaml}" >&2
+    exit 1
+fi
```
gforsyth: I'd like to see this rewritten in Python -- I _really_ don't like us re-implementing conda fetching for this purpose, but I can't think of a workaround that doesn't involve installing conda.

That said, this is not logic to write in Bash.

docs/cuopt/source/convex-settings.rst
```diff
@@ -217,6 +217,8 @@ Barrier Solver Settings
 
 The following settings control the behavior of the barrier (interior-point) method:
 
+.. note:: The barrier method uses cuDSS for sparse Cholesky factorization, whose threading layer must be built against the same GNU OpenMP runtime as cuOpt itself. pip installs bundle both, so this is handled automatically; conda and source builds need the host to provide ``libgomp.so.1``. If missing, or if a mismatched ``CUDSS_THREADING_LIB`` override is set, the solve falls back to single-threaded cuDSS instead of failing. See :doc:`system-requirements` and :doc:`faq`.
```
gforsyth: You've added `libgomp` to the `libcuopt` recipe's runtime dependencies so a `conda` build doesn't need the host to provide the shared object

docs/cuopt/source/faq.rst
```diff
@@ -102,6 +102,22 @@ General FAQ
 
    After installing, re-run ``pip install`` (or restart the Python process) and the import should succeed.
 
+.. dropdown:: Why does my MIP/LP solve run single-threaded, or log a "could not load cuDSS threading layer" warning?
+
+   cuDSS's threading layer (used by the barrier method) must be built against the *same* GNU OpenMP runtime as cuOpt itself. If it isn't, cuDSS falls back to single-threaded execution and logs a warning instead of failing the solve.
+
+   **pip installs**: the wheel bundles its own GNU OpenMP runtime and a cuDSS threading layer built against it, so this shouldn't happen out of the box -- no host ``libgomp`` is required.
+
+   **conda and source builds**: cuDSS's threading layer loads ``libgomp.so.1`` from the host at runtime. If the host doesn't provide it, install your distribution's GNU OpenMP runtime package, typically ``libgomp1`` (Debian/Ubuntu) or ``libgomp`` (RHEL/Rocky/Fedora), then re-run the solve.
```
gforsyth: same comment re: conda builds

docs/cuopt/source/system-requirements.rst
```diff
@@ -34,6 +34,11 @@ Dependencies are installed automatically when using the pip and Conda installati
       - On RHEL/Rocky/Alma 8, install the ``openssl3`` package from EPEL (see :doc:`faq`)
       - On Ubuntu 20.04, install OpenSSL 3 from a PPA/backport or use the cuOpt container
 
+   * GNU OpenMP runtime (for multi-threaded cuDSS/barrier solves):
+      - pip installs bundle their own; no host ``libgomp`` needed
+      - conda and source builds need ``libgomp.so.1`` on the host -- install via ``libgomp1`` (Debian/Ubuntu) or ``libgomp`` (RHEL/Rocky/Fedora)
```
gforsyth: same comment re: conda buidls

gforsyth (changes requested):
I was (am) concerned about `glibc` mismatches since you are grabbing `libgomp.so` from conda-forge which has a more modern glibc toolchain as compared to Rocky 8.

At least as of today, and `conda-forge/libgomp/15.3.0/linux-64/libgomp-15.3.0-he0feb66_5.conda`, it looks like `libgomp.so.1.0.0` version requirements are met by glibc 2.28 (which Rocky 8 ships):

```sh
$ objdump -T libgomp.so.1.0.0 | grep -o 'GLIBC_[0-9.]*' | sort -Vu
GLIBC_2.2.5
GLIBC_2.3
GLIBC_2.3.4
GLIBC_2.6
GLIBC_2.14
GLIBC_2.17
```

ci/utils/install_modern_libgomp.sh
ramakrishnap-nv: Rewrote it as ci/utils/install_modern_libgomp.py, invoked with 'python' from build_wheel_libcuopt.sh. Same repodata-based resolution, no conda CLI needed, but no more inline heredocs.

ci/utils/install_modern_libgomp.sh
ramakrishnap-nv: Moved to install_modern_libgomp.py.

ci/utils/install_modern_libgomp.sh
ramakrishnap-nv: Same file now, not inlined.

ci/utils/install_modern_libgomp.sh
ramakrishnap-nv: Done -- build_wheel_libcuopt.sh now calls 'python ci/utils/install_modern_libgomp.py', matching the rest of ci/utils/*.py.

ci/utils/install_modern_libgomp.sh
ramakrishnap-nv: Dropped the explicit cleanup in the Python rewrite -- agreed, nothing to do here.

docs/cuopt/source/convex-settings.rst
ramakrishnap-nv: Right, fixed -- conda pulls libgomp in automatically as a package dependency; only actual source builds need it provided manually. Updated the wording here and in faq/system-requirements.

docs/cuopt/source/faq.rst
ramakrishnap-nv: Fixed -- split into pip / conda (automatic) / source build (manual) instead of lumping conda in with source builds.

docs/cuopt/source/system-requirements.rst
ramakrishnap-nv: Fixed, same correction.

gforsyth (approved):

gforsyth (approved):

cpp/CMakeLists.txt
```diff
@@ -1025,6 +1025,29 @@ if (TARGET KaMinPar)
         add_dependencies(cuopt_objs KaMinPar)
 endif ()
 
+# CUOPT_BUILD_CUSTOM_CUDSS_MTLAYER: build our own cuDSS threading-layer plugin against the same
```
nguidotti: NITPICK: Let us shorten the comment to 1 or 2 lines. This is too verbose.

nguidotti (approved):

cpp/CMakeLists.txt
ramakrishnap-nv: Trimmed to 3 lines.

Iroy30 (approved):
All looks good, might need a little clean up on docs and comments.

docs/cuopt/source/faq.rst
```diff
@@ -102,6 +102,24 @@ General FAQ
...
 
+.. dropdown:: Why does my MIP/LP solve run single-threaded, or log a "could not load cuDSS threading layer" warning?
+
+   cuDSS's threading layer (used by the barrier method) must be built against the *same* GNU OpenMP runtime as cuOpt itself. If it isn't, cuDSS falls back to single-threaded execution and logs a warning instead of failing the solve.
+
+   **pip installs**: the wheel bundles its own GNU OpenMP runtime and a cuDSS threading layer built against it, so this shouldn't happen out of the box -- no host ``libgomp`` is required.
+
+   **conda installs**: ``libgomp`` is a runtime dependency of the ``libcuopt`` package, so ``conda install`` pulls it in automatically -- no manual step needed.
+
+   **source builds**: cuDSS's threading layer loads ``libgomp.so.1`` from the host at runtime. If the host doesn't provide it, install your distribution's GNU OpenMP runtime package, typically ``libgomp1`` (Debian/Ubuntu) or ``libgomp`` (RHEL/Rocky/Fedora), then re-run the solve.
```
hlinsen: For the source build, OpenMP::OpenMP_CXX links cuOpt against conda’s libgomp.so.1. Which is also reused by cuDSS. If a user builds and runs with cuopt_dev env activated they shouldn't need the system install.

docs/cuopt/source/system-requirements.rst
```diff
@@ -34,6 +34,11 @@ Dependencies are installed automatically when using the pip and Conda installati
       - On RHEL/Rocky/Alma 8, install the ``openssl3`` package from EPEL (see :doc:`faq`)
       - On Ubuntu 20.04, install OpenSSL 3 from a PPA/backport or use the cuOpt container
 
+   * GNU OpenMP runtime (for multi-threaded cuDSS/barrier solves):
+      - pip installs bundle their own; conda installs pull in ``libgomp`` automatically as a package dependency -- neither needs a manual step
+      - source builds need ``libgomp.so.1`` on the host -- install via ``libgomp1`` (Debian/Ubuntu) or ``libgomp`` (RHEL/Rocky/Fedora)
```
hlinsen: same comment as above, source build is picking up the right libgomp instance from conda.

docs/cuopt/source/faq.rst
ramakrishnap-nv: Good catch, fixed. Source builds using the conda dev environment already get a matching libgomp via OpenMP::OpenMP_CXX, same as cuDSS's own conda-provided one. Narrowed the guidance to only apply to a build done entirely outside that conda environment.

docs/cuopt/source/system-requirements.rst
ramakrishnap-nv: Fixed, same correction.

hlinsen (approved):

PR #1910, opened 2026-09-15
add a proxy server CPU path for routing conversion to DataModel

tmckayus:
The proxy server does not solve locally, and so does not need DataModel built with GPU data types. To prevent any instability, keep the algorithm exactly the same but use numpy/pandas types instead of cudf types to convert routing JSON to DataModel in the proxy server. Move a copy of the original conversion with GPU types to deprecated/ subdirectory for the legacy server.


python/cuopt_server/cuopt_server/utils/routing/host_optimization_data_model.py
```diff
@@ -0,0 +1,368 @@
...
+from cuopt_server.utils.routing.validation_solver_config import (
+    validate_solver_config,
+)
+from cuopt_server.utils.routing.validation_task_data import validate_task_data
+
+
+class HostOptimizationDataModel(OptimizationDataModel):
+    """OptimizationDataModel whose request data uses host containers."""
+
+    def set_cost_matrix(self, cost_matrix):
```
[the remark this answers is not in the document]
tmckayus: skipping, no different than the legacy server.

python/cuopt_server/cuopt_server/utils/routing/host_optimization_data_model.py
```diff
@@ -0,0 +1,368 @@
...
+            cost_matrix,
+            is_travel_time=False,
+            updating=False,
+            comparison_matrix=None,
+        )
+        if is_valid[0]:
+            self.is_route_detail_set = True
+            self.cost_matrix = {}
+            for v_type, matrix in cost_matrix.items():
+                np_cost_matrix = np.array(matrix, dtype=np.float32)
```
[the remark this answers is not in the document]
tmckayus: skipping, no different than the legacy server

python/cuopt_server/cuopt_server/utils/deprecated/routing/conversion.py
```diff
@@ -0,0 +1,431 @@
+# SPDX-FileCopyrightText: Copyright (c) 2022-2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
ramakrishnap-nv:
```suggestion
# SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```

python/cuopt_server/cuopt_server/utils/routing/conversion.py
```diff
@@ -174,25 +177,17 @@ def populate_optimization_data(
 
 
 def create_data_model(
-    optimization_data: OptimizationDataModel,
+    optimization_data: HostOptimizationDataModel,
     cost_matrix: Optional[dict] = None,
     travel_time_matrix: Optional[dict] = None,
 ):
     warnings = []
```
ramakrishnap-nv, line 184: The RMM pool-allocator sanity assertion from the old `create_data_model` was dropped just above here, but this still calls `routing.DataModel(...)`, which allocates GPU memory. Under the default (non-pool) allocator this now fails silently instead of fast — worth keeping the check.

python/cuopt_server/cuopt_server/utils/deprecated/routing/solver.py
```diff
@@ -13,8 +13,8 @@
     OutOfMemoryError,
 )
 
+from cuopt_server.utils.deprecated.routing.conversion import create_data_model
 from cuopt_server.utils.routing.conversion import (  # noqa: F401
```
ramakrishnap-nv, line 17: This now imports `create_solver`/`prep_optimization_data`/`warn_on_objectives` from the live (host/pandas) `routing/conversion.py`, while `create_data_model` comes from the deprecated cudf module. A future change to the live functions for pandas semantics could silently break this GPU path — nothing marks or tests the cross-module dependency.

python/cuopt_server/cuopt_server/utils/routing/host_optimization_data_model.py
```diff
@@ -0,0 +1,399 @@
...
+serialization. Waypoint graph preparation may still produce device matrices.
+"""
+
+import numpy as np
+import pandas as pd
+
+from cuopt_server.utils.routing.optimization_data_model import (
+    OptimizationDataModel,
+    get_none_for_empty_list,
+    get_objectives_as_lists,
```
ramakrishnap-nv, line 17: This subclasses `OptimizationDataModel`, which imports `cudf`, so the "CPU-only" path still unconditionally imports cudf on load. Probably fine given cudf is already a mandatory dependency — just confirming that's intentional.

ramakrishnap-nv (commented):
A few things worth a look before merge — flagging the two I'd prioritize (dropped RMM check, cross-module import drift) plus some lower-priority notes. One more finding (latent AttributeError in `optimization_data_model.py`'s inherited getters using `.to_arrow()`, unreachable today but real once a host model hits that code path) is in an unchanged file so it's noted here rather than inline.

python/cuopt_server/cuopt_server/utils/routing/host_optimization_data_model.py
```diff
@@ -0,0 +1,399 @@
+# SPDX-FileCopyrightText: Copyright (c) 2022-2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
ramakrishnap-nv, line 1: This file and `deprecated/routing/conversion.py` are largely line-for-line duplicates of their GPU counterparts (pandas/numpy vs cudf). Docstring notes this is meant to be temporary, but until then a validation fix in one copy not mirrored to the other will silently diverge proxy vs. legacy request handling.

python/cuopt_server/cuopt_server/utils/routing/host_optimization_data_model.py
```diff
@@ -0,0 +1,399 @@
...
+
+class HostOptimizationDataModel(OptimizationDataModel):
+    """OptimizationDataModel whose request data uses host containers.
+
+    Incremental ``update_*`` methods that wrap cudf on the parent are
+    unimplemented here. When the GPU model is deprecated, this class can
+    absorb ``OptimizationDataModel`` rather than keep a parallel type.
+    """
+
+    def update_cost_matrix(self, *args, **kwargs):
```
ramakrishnap-nv, line 39: All `update_*` methods raise `NotImplementedError`. Fine for now, but a landmine if incremental-update requests are ever routed to the proxy.

python/cuopt_server/cuopt_server/utils/routing/host_optimization_data_model.py
```diff
@@ -0,0 +1,399 @@
...
+            cost_matrix,
+            is_travel_time=False,
+            updating=False,
+            comparison_matrix=None,
+        )
+        if is_valid[0]:
+            self.is_route_detail_set = True
+            self.cost_matrix = {}
+            for v_type, matrix in cost_matrix.items():
+                np_cost_matrix = np.array(matrix, dtype=np.float32)
```
ramakrishnap-nv, line 76: Builds an intermediate `np.array` before wrapping in `pd.DataFrame`, unlike other `set_*` methods here that pass the matrix directly with `dtype=`. Minor, but an avoidable extra copy on large matrices.

python/cuopt_server/cuopt_server/utils/routing/conversion.py
tmckayus: No, because the DataModel defers construction of GPU structures until a local solve, and the proxy server will never do a local solve. This is fine.

python/cuopt_server/cuopt_server/utils/deprecated/routing/solver.py
tmckayus: This is intentional. create_solver only handles scalar settings; prep_optimization_data is container-compatible; warn_on_objectives is a no-op. After the legacy server is deprecated for a release cycle, it will be removed anyway. Not worth freezing copies of additional routines.

python/cuopt_server/cuopt_server/utils/routing/conversion.py
tmckayus: Additionally, a follow-on PR to be submitted as soon as #1910 is merged will create an RMM pool when the request has a waypoint graph, not on dense matrices.  For this PR waypoint conversion will happen in the proxy server on the default allocator (but that will be short-lived)

python/cuopt_server/cuopt_server/utils/routing/host_optimization_data_model.py
tmckayus: This is known and intentional, the HostOptimizationDataModel and the OptimizationDataModel will be merged and will be CPU-only when the legacy server is removed after deprecation.

python/cuopt_server/cuopt_server/utils/routing/host_optimization_data_model.py
tmckayus: Intentional. Chosen for stability in the proxy server which is for backward compat. The JSON contract is frozen, and we didn't want to introduce any unintentional jitter from modified conversion code. As noted above, after removal of the legacy server there will be only the CPU path, no duplication.

python/cuopt_server/cuopt_server/utils/routing/host_optimization_data_model.py
tmckayus: They never will be. API is static. There will not be any updates, only solves from a single submit.

python/cuopt_server/cuopt_server/utils/routing/host_optimization_data_model.py
tmckayus: Matches the legacy conversion sequence. Choosing stability over optimization.

ramakrishnap-nv (approved):

PR #1911, opened 2026-09-15
[BUG] Correctly stop sub-MIP when main solve finishes

nguidotti:
Since a recursive sub-MIP runs inline with the calling thread, it can ignore the stop signal received from the main solve if it is stuck on solving a degenerate node in the deep level of the recursion. This fixes by storing all active solvers and calling `node_concurrent_halt = true` in all of them when the main solve finishes.

## Checklist

- Testing
   - [ ] Added tests
   - [ ] NA
- Documentation
   - [ ] Added new documentation
   - [ ] NA


chris-maes (approved):
LGTM. Thanks for the fix. Let's merge it.

PR #1913, opened 2026-09-16
Gpu computation from ruiz scaling to kkt build

yuwenchen95:
## Description
Moves the barrier solver's Ruiz equilibration to the GPU and removes redundant copies and
conversions of `A` and `Q` between scaling and KKT assembly.

**GPU Ruiz equilibration.** Adds `scaling_ruiz_gpu` (`cpp/src/dual_simplex/scaling_gpu.cu`), a
device port of `scaling()`'s Ruiz branch, used for the barrier's SOCP/QP path. It mirrors the
CPU implementation step for step: the loop contains only `abs`/`max`/`sqrt`/`*`//`, no
summations, so results are bit-identical to the CPU path. Selected by
`gpu_ruiz_nnz_threshold` (default 500000 — see note below). The CPU `scaling()` was also
changed to compute row inf-norms by scatter-max directly off CSC, dropping two host CSR builds
of `A`.

**Fewer host↔device round trips.** `A` and `A^T` are now uploaded and converted once per solve
rather than rebuilt per consumer:
- device CSC→CSR replaced (merge sort → scatter + segmented sort), reused for a new on-device
  transpose, so the augmented path no longer builds and uploads a host `A^T`;
- `cusparse_view_` builds its CSR on device and borrows `device_A_csc_`/`device_AT_csc_` instead
  of owning a second CSR + CSC-transpose; the Q view likewise borrows `device_Q_csc_` (valid
  because `optimization_problem_t` stores only the symmetrized `H = Q + Q^T`, so CSC(Q) is CSR(Q));
- on the SOCP path the scaled `A`/`Q` stay resident: `scaling_ruiz_gpu` hands them to the barrier
  via `lp_problem_t::device_A`/`device_Q` instead of downloading and immediately re-uploading;
- the coefficient-range reduction used for logging runs on device instead of scanning the
  just-downloaded host array.

**Device memory on the ADAT path.** When there are no dense columns `AD == A`, so `device_A` and
`d_original_A_values` are no longer allocated — the SpGEMM's left operand points at
`device_AT_csc_` (already CSR(A)) and `form_adat` restores unscaled values from
`device_A_csc_.x`. `device_AD` is seeded device-to-device, dropping the host `AD = lp.A` copy.
This takes `A`'s device footprint on that path from roughly 60·nnz to 40·nnz, which matters when
the barrier runs concurrently with PDLP on large LPs. `initialize_cusparse_data` now takes A's
CSR pieces rather than a matrix so either source can be passed; `multiply_kernels`' `A`/`DAT`
parameters were dead (it uses only the descriptors) and are removed.

**Test results:** 14% acceleration on SOCP benchmark, no regression on QP benchmarks.


cpp/src/dual_simplex/presolve.hpp
```diff
@@ -69,6 +76,14 @@ struct lp_problem_t {
   i_t cone_var_start{0};
   std::vector<i_t> second_order_cone_dims;
 
+  // Set by scaling_ruiz_gpu when it leaves the scaled A on device; `A` then keeps its sparsity
+  // pattern but has an empty `x`, and the barrier takes this over instead of uploading `A`.
+  std::shared_ptr<barrier::device_csc_matrix_t<i_t, f_t>> device_A;
```
chris-maes: lp_problem_t is a CPU-only data structure. I don't think we should have device_A and device_Q be part of this data structure. Can these be constructed and stored elsewhere?

cpp/src/dual_simplex/scaling.hpp
```diff
@@ -23,6 +23,14 @@ i_t scaling(const lp_problem_t<i_t, f_t>& unscaled,
             std::vector<f_t>& column_scaling,
             std::vector<f_t>& row_scaling);
 
+// GPU-based Ruiz scaling.
+template <typename i_t, typename f_t>
```
chris-maes: I think this belongs in a scaling_gpu.cuh file. Not in scaling.hpp.

cpp/src/barrier/scaling_gpu.cu
```diff
@@ -0,0 +1,454 @@
+/* clang-format off */
```
chris-maes, line 1: I think this file should move inside the barrier directory. The dual_simplex directory doesn't contain any .cu files. It is CPU only.

cpp/src/dual_simplex/solve.cpp
```diff
@@ -480,7 +480,15 @@ lp_status_t solve_linear_program_with_barrier(
                                     presolved_lp.A.col_start[presolved_lp.num_cols]);
   std::vector<f_t> column_scales;
   std::vector<f_t> row_scales;
-  scaling(presolved_lp, barrier_settings, barrier_lp, column_scales, row_scales);
+  const bool is_ruiz_candidate =
+    !presolved_lp.second_order_cone_dims.empty() || presolved_lp.Q.n > 0;
+  const i_t presolved_nnz = presolved_lp.A.col_start[presolved_lp.num_cols] +
+                            (presolved_lp.Q.n > 0 ? presolved_lp.Q.row_start[presolved_lp.Q.m] : 0);
+  if (is_ruiz_candidate && presolved_nnz >= barrier_settings.gpu_ruiz_nnz_threshold) {
+    scaling_ruiz_gpu(presolved_lp, barrier_settings, barrier_lp, column_scales, row_scales);
```
chris-maes: I think this should return device_A and device_Q

cpp/src/dual_simplex/solve.cpp
chris-maes: You can pass in device_A and device_Q here.

yuwenchen95: Moved device_A and device_Q out.

cpp/src/barrier/barrier.cu
```diff
@@ -445,11 +445,44 @@ class barrier_reduce_helper_t {
 template <typename i_t, typename f_t>
 class iteration_data_t {
  public:
+  /** The device Q to factorize: adopted from the scaling when already there, uploaded otherwise. */
+  static device_csc_matrix_t<i_t, f_t> make_device_Q(
+    std::shared_ptr<device_csc_matrix_t<i_t, f_t>> scaled_device_Q,
```
chris-maes: Can we avoid using a `shared_ptr` here?

cpp/src/barrier/barrier.cu
```diff
@@ -445,11 +445,44 @@ class barrier_reduce_helper_t {
...
+    return empty;
+  }
+
   iteration_data_t(const lp_problem_t<i_t, f_t>& lp,
                    i_t num_upper_bounds,
                    const std::vector<i_t>& direct_free_variables,
                    const csc_matrix_t<i_t, f_t>& Qin,
-                   const simplex_solver_settings_t<i_t, f_t>& settings)
+                   const simplex_solver_settings_t<i_t, f_t>& settings,
+                   std::shared_ptr<device_csc_matrix_t<i_t, f_t>> scaled_device_A,
```
chris-maes: Do these need to be shared_ptrs? Can they be unique_ptrs?

cpp/src/barrier/barrier.cu
```diff
@@ -2431,10 +2491,18 @@ void cholesky_debug_check(const iteration_data_t<i_t, f_t>& data,
...
 template <typename i_t, typename f_t>
-barrier_solver_t<i_t, f_t>::barrier_solver_t(const lp_problem_t<i_t, f_t>& lp,
-                                             const simplex::presolve_info_t<i_t, f_t>& presolve,
-                                             const simplex_solver_settings_t<i_t, f_t>& settings)
-  : lp(lp), settings(settings), presolve_info(presolve), stream_view_(lp.handle_ptr->get_stream())
+barrier_solver_t<i_t, f_t>::barrier_solver_t(
+  const lp_problem_t<i_t, f_t>& lp,
+  const simplex::presolve_info_t<i_t, f_t>& presolve,
+  const simplex_solver_settings_t<i_t, f_t>& settings,
+  std::shared_ptr<device_csc_matrix_t<i_t, f_t>> device_A,
```
chris-maes: Can these be unique_ptrs?

cpp/src/barrier/barrier.hpp
```diff
@@ -117,6 +127,9 @@ class barrier_solver_t {
   const simplex::simplex_solver_settings_t<i_t, f_t>& settings;
   const simplex::presolve_info_t<i_t, f_t>& presolve_info;
   cuda::stream_ref stream_view_;
+  // Handed over to iteration_data_t by solve(), which empties them.
+  std::shared_ptr<device_csc_matrix_t<i_t, f_t>> device_A_;
```
chris-maes: unique_ptrs

cpp/src/barrier/scaling_gpu.cu
```diff
@@ -0,0 +1,465 @@
...
+  thrust::for_each(rmm::exec_policy(stream),
+                   thrust::make_counting_iterator(i_t(0)),
+                   thrust::make_counting_iterator(A.nz_max),
+                   [x = A.x.data(), row = A.i.data(), rmax = row_norm.data()] __device__(i_t p) {
+                     const f_t a = raft::abs(x[p]);
+                     // Read first: most nonzeros lose to the running max, so the CAS loop
+                     // inside myAtomicMax is usually skipped entirely. Collisions are rare
+                     // anyway -- a CSC column cannot repeat a row, so consecutive threads
+                     // within a column target distinct accumulators.
+                     if (a > rmax[row[p]]) { raft::myAtomicMax(rmax + row[p], a); }
```
chris-maes: I think the argument to myAtomicMax would be clearer if it was

myAtomicMax(&rmax[row[p]], a)

cpp/src/barrier/scaling_gpu.cuh
```diff
@@ -0,0 +1,39 @@
...
+// `scaled.A` then keeps its sparsity pattern but has an empty `x` and this is the only copy of the
+// values); `device_Q` receives the scaled Q. Both are null when there is nothing to hand over, and
+// the barrier uploads from `scaled` instead.
+template <typename i_t, typename f_t>
+i_t scaling_ruiz_gpu(const lp_problem_t<i_t, f_t>& unscaled,
+                     const simplex_solver_settings_t<i_t, f_t>& settings,
+                     lp_problem_t<i_t, f_t>& scaled,
+                     std::vector<f_t>& column_scaling,
+                     std::vector<f_t>& row_scaling,
+                     std::shared_ptr<barrier::device_csc_matrix_t<i_t, f_t>>& device_A,
```
chris-maes: Do these need to be shared_ptrs. Can these be unique_ptrs instead?

cpp/src/dual_simplex/solve.cpp
```diff
@@ -480,7 +481,19 @@ lp_status_t solve_linear_program_with_barrier(
                                     presolved_lp.A.col_start[presolved_lp.num_cols]);
   std::vector<f_t> column_scales;
   std::vector<f_t> row_scales;
-  scaling(presolved_lp, barrier_settings, barrier_lp, column_scales, row_scales);
+  // Scaled A/Q that the GPU scaling may leave on device for the barrier to adopt; null otherwise.
+  std::shared_ptr<barrier::device_csc_matrix_t<i_t, f_t>> device_A;
```
chris-maes: unique_ptrs?

chris-maes (changes requested):
it would be good to avoid the use of shared_ptrs. These are basically globals. Please see if you can convert the code to using unique_ptrs instead.

yuwenchen95:
> it would be good to avoid the use of shared_ptrs. These are basically globals. Please see if you can convert the code to using unique_ptrs instead.

Replaced `shared_ptr` with `unique_ptr`, wrapped in a `scaled_device_matrices_t` struct
(`barrier.hpp`) so the deletion is compiled in `barrier.cu` where the matrix type is complete.

cpp/src/utilities/reduce_ops.cuh
```diff
@@ -0,0 +1,36 @@
...
+  }
+};
+
+template <typename item_t>
+struct min_op_t {
+  __host__ __device__ item_t operator()(const item_t& lhs, const item_t& rhs) const
+  {
+    return lhs < rhs ? lhs : rhs;
+  }
+};
```
hlinsen, line 34: You can directly use cuda::maximum<item_t>{} https://nvidia.github.io/cccl/unstable/libcudacxx/extended_api/functional/maximum_minimum.html#floating-point-behavior

cpp/src/barrier/scaling_gpu.cu
```diff
@@ -0,0 +1,573 @@
...
+                      d_col_scale.data() + n,
+                      c.data(),
+                      d_col_scale.data(),
+                      cuda::std::multiplies<f_t>{});
+    thrust::for_each(
+      rmm::exec_policy(stream),
+      thrust::make_counting_iterator(i_t(0)),
+      thrust::make_counting_iterator(n),
+      [lower = d_lower.data(), upper = d_upper.data(), c = c.data()] __device__(i_t j) {
+        if (lower[j] > f_t(-1e20)) lower[j] /= c[j];
```
hlinsen, line 491: Is it fine to hardcode?

cpp/src/barrier/scaling_gpu.cu
```diff
@@ -0,0 +1,573 @@
...
+        thrust::make_counting_iterator(dQ.m),
+        [x = dQ.x.data(), rs = dQ.row_start.data(), col = dQ.j.data(), c = c.data()] __device__(
+          i_t i) {
+          for (i_t p = rs[i]; p < rs[i + 1]; ++p) {
+            x[p] *= c[i] * c[col[p]];
+          }
+        });
+    }
+
+    if (max_deviation < 0.1) break;
```
hlinsen, line 510: Can we move the hard coded value to a constant

cpp/src/barrier/scaling_gpu.cu
```diff
@@ -0,0 +1,573 @@
...
+                       OffsetEndIt end_offsets,
+                       i_t num_segments,
+                       f_t* out,
+                       rmm::device_buffer& temp_storage,
+                       rmm::cuda_stream_view stream)
+{
+  if (num_segments == 0) return;
+  auto abs_it  = thrust::make_transform_iterator(values, cuopt::abs_value_transform_t<f_t>{});
+  size_t bytes = 0;
+  cub::DeviceSegmentedReduce::Reduce(nullptr,
```
hlinsen, line 122: CCCL 3.4+ has a single-call overload that removes the sizing call and explicit scratch buffer:
```
RAFT_CUDA_TRY(cub::DeviceSegmentedReduce::Reduce(
  abs_it, out, num_segments, begin_offsets, end_offsets,
  cuopt::max_op_t<f_t>{}, f_t(0), stream.get()));
```

cpp/src/barrier/scaling_gpu.cu
```diff
@@ -0,0 +1,573 @@
...
+  std::vector<f_t> col_scale_host(n, 1.0);
+  rmm::device_uvector<f_t> d_rhs       = cuopt::device_copy(scaled.rhs, stream);
+  rmm::device_uvector<f_t> d_objective = cuopt::device_copy(scaled.objective, stream);
+  rmm::device_uvector<f_t> d_lower     = cuopt::device_copy(scaled.lower, stream);
+  rmm::device_uvector<f_t> d_upper     = cuopt::device_copy(scaled.upper, stream);
+  rmm::device_uvector<f_t> d_row_scale = cuopt::device_copy(row_scaling, stream);
+  rmm::device_uvector<f_t> d_col_scale = cuopt::device_copy(col_scale_host, stream);
+
+  // Per-nonzero column ids for A, built once (sparsity pattern is fixed across iterations).
+  dA.form_col_index(stream);  // dA.col_index[p] = column of nonzero p
```
hlinsen, line 238: Agent flagged a potential bug, could you check

> Fix column IDs after an empty CSC column. GPU scaling newly calls form_col_index(). For col_start = [0, 2, 2, 4], that builder marks the last two entries as column 1, although they belong to column 2. Scaling then applies the wrong column factor. Mark each nonempty column start with its column ID and use an inclusive maximum scan; add a CPU/GPU parity case with an interior empty column.
> 
> Example — empty column 1
> 
> col_start = [0, 2, 2, 4]
> 
> column 0: entries 0, 1
> column 1: empty
> column 2: entries 2, 3
> 
> entry:          0  1  2  3
> actual column:  0  0  2  2
> code's label:   0  0  1  1   ✗
> 
> A.x:                 [10, 20, 30, 40]
> scales c:             [ 1,  2,  3]
> correct scaled A.x:  [10, 20, 90, 120]
> code's scaled A.x:   [10, 20, 60,  80]

PR #1914, opened 2026-09-16
Speedup barrier second-order cone conversion in general path

yuwenchen95:
## Description
Two changes for QCQP -> SOCP conversion:
1) Reuse general-path scratch buffers: The LDLT path allocated two buffers sized by the whole model's variable count on every quadratic constraint, even though each constraint only uses a small slice. Hoisting them above the loop and reuse them across constraints .

2) Build the SOC Hessian directly from COO: The Hessian was built via a dense n_local×n_local accumulator scanned three times, which degrades to O(q_nnz²) since Q is already duplicate-free after canonicalization. Feeding the canonicalized COO straight to coo_to_csc removes the dense accumulator and its redundant scans entirely.

These changes cut hub1's time-to-barrier-start from 14.32s to 0.66s. No regression on SOCP and 16% acceleration on QCP benchmarks.


chris-maes (approved):
LGTM. Congrats on the speed up

PR #1917, opened 2026-09-16
proxy server add VRP request and solution over gRPC

tmckayus:
This change adds VRP support to the proxy server (gRPC server already has VRP support)


python/cuopt_server/cuopt_server/utils/routing/conversion.py
```diff
@@ -524,3 +525,124 @@ def prep_optimization_data(optimization_data):
...
+
+
+def _node_type_name(value):
+    try:
+        return _NODE_TYPE_NAMES[int(value)]
+    except (ValueError, TypeError, KeyError):
+        return str(value)
+
+
+def solution_to_legacy_http(
```
ramakrishnap-nv: Think we don't need to mention legacy since that would our standard and not planning to changee

ramakrishnap-nv (approved):
Rest looks good, but I would suggest not to use legacy since we will not be replacing it. May be we can just keep it normal.

python/cuopt_server/cuopt_server/utils/routing/conversion.py
tmckayus: great point! will fix

tmckayus: done

PR #1922, opened 2026-09-16
Handle PSLP false-infeasible on badly-scaled LPs

rg20:
## Description

PSLP presolve can flag a problem infeasible and return a zero-sized reduced problem, but this call is unreliable on badly-scaled problems (coefficient range spanning >= 1e9). In that regime, don't trust the result: log a warning and continue solving the original, unreduced problem instead of terminating the solve on PSLP's say-so. Well-scaled problems keep the previous behaviour of trusting PSLP directly.

Add pslp_postsolve_skip_ so undo_pslp() safely no-ops in this fallback case, since pslp_presolver_ holds state from the aborted run rather than a real reduction.

Add PSLP presolve unit tests covering the well-scaled (trusted) and badly-scaled (untrusted, falls back, end-to-end solve still detects infeasibility) cases.

## Checklist

- Testing
   - [x] Added tests
   - [ ] NA
- Documentation
   - [ ] Added new documentation
   - [x] NA


Kh4ster (approved):
That safe guard makes sense, all good!

PR #1923, opened 2026-09-17
presolve: refresh the cached linear column count after removing empty…

yuwenchen95:
## Description

`presolve()` caches `linear_variable_count()` in `linear_cols`, but `remove_empty_cols()`
then shrinks `num_cols` and replaces `objective`/`lower`/`upper` with shorter vectors. The
cache is never refreshed, so the later free-variable loops read past the end and push
out-of-range column indices into `direct_free_variables`. The barrier indexes
`num_cols`-sized arrays with them, corrupting the heap — surfacing as
`malloc(): corrupted top size` or `cudaErrorLaunchFailure`. `PRIMAL1`, `PRIMAL2` and `PRIMAL4` abort computation.

Fix: recompute `linear_cols` after the removal. All three problems can be solved again.


cpp/src/dual_simplex/presolve.cpp
```diff
@@ -1389,6 +1388,9 @@ i_t presolve(const lp_problem_t<i_t, f_t>& original,
   if (num_empty_cols > 0) {
     settings.log.printf("Presolve attempt to remove %d empty cols\n", num_empty_cols);
     remove_empty_cols(problem, num_empty_cols, presolve_info);
+    // remove_empty_cols drops columns and shrinks objective/lower/upper, need to recompute
+    // linear_cols
+    linear_cols = linear_variable_count(problem);
```
rg20: This should be done by `remove_empty_cols`

yuwenchen95: Good point.

rg20 (approved):

PR #1928, opened 2026-09-17
build(conda): split libcuopt into per-component packages

ramakrishnap-nv:
Splits the single `libcuopt` conda output into one package per component, using the install components added in #1926.

| Package | Contents | Beyond rmm / rapids-logger |
|---|---|---|
| `libcuopt-client` | `libcuopt_client.so` + shared headers | grpc, protobuf, abseil — no CUDA math libs |
| `libcuopt-mathopt` | `libcuopt_mathopt.so`, `cuopt_cli` | cublas, cusparse, cudss, nccl, tbb |
| `libcuopt-routing` | `libcuopt_routing.so` | cublas |
| `cuopt-grpc-server` | `cuopt_grpc_server` | the three libraries |
| `libcuopt` | `libcuopt.so` ld script | metapackage pinning the three |

A routing-only install no longer pulls cusparse, cudss or nccl. Run dependencies were split by measuring `DT_NEEDED` per library rather than by dividing the existing list.

`conda install libcuopt` still gets everything, so the `cuopt` recipe that depends on it is unaffected.

Wheels are not included here — see the comment below.

Draft: stacked on #1926.


dependencies.yaml
```diff
@@ -417,6 +441,78 @@ dependencies:
...
+              cuda: "12.*"
+              cuda_suffixed: "true"
+            packages:
+              - libcuopt-client-cu12==26.10.*,>=0.0.0a0
+          - matrix:
+              cuda: "13.*"
+              cuda_suffixed: "true"
+            packages:
+              - libcuopt-client-cu13==26.10.*,>=0.0.0a0
+          - {matrix: null, packages: [*libcuopt_client_unsuffixed]}
```
KyleFromNVIDIA: This will fail the check that I'm introducing in https://github.com/NVIDIA/cuopt/pull/1915. Please wait until that's merged. Same for the others below.

KyleFromNVIDIA (changes requested):

dependencies.yaml
KyleFromNVIDIA: Hmmm. Looks like `verify-dependencies` doesn't check the suffixing of cuopt, because it's not in https://github.com/rapidsai/rapids-metadata.

ramakrishnap-nv: #1915 merged earlier today, so I have rebased onto main and run the new hook against these changes:

```
verify-dependencies......................................................Passed
```

Full `pre-commit run --all-files` is clean too. Matches your follow-up — the suffixed `libcuopt-*` names are not in rapids-metadata, so the hook has nothing to check them against. Happy to add them there if you would rather these be verified rather than merely not rejected.

conda/recipes/libcuopt/recipe.yaml
```diff
@@ -127,37 +125,90 @@ outputs:
...
+        - tbb-devel
+      run:
+        - ${{ pin_compatible("cuda-version", upper_bound="x", lower_bound="x") }}
+        - ${{ pin_subpackage("libcuopt-client", exact=True) }}
+        - libboost-devel
+        - librmm =${{ minor_version }}
+        - cuda-nvrtc
+        - libcudss >=0.7,<0.8
+        - libgomp >=9
+        - nccl >=2.19
```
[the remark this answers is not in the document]
ramakrishnap-nv: The premise is right — `libcuopt_mathopt.so` does carry `libcusparse.so.12` in `DT_NEEDED` — but this is pre-existing rather than something the split introduced.

The monolithic `libcuopt` output did not list `libcusparse` in `run` either, and it explicitly carried `libcusparse` under `ignore_run_exports`, so the `libcusparse-dev` host dependency never became a run dependency. I preserved both verbatim when splitting the output, so the resolved run set for mathopt is identical to what `libcuopt` shipped before.

Leaving it as is here on purpose: changing which CUDA runtime packages are declared is a dependency-semantics change, and folding that into a packaging split would make both harder to review and to revert. The `ignore_run_exports` entry looks deliberate rather than accidental, so I would rather not reverse it without understanding why it was added.

Worth its own issue — I will open one referencing this thread unless someone knows the history offhand.

dependencies.yaml
```diff
@@ -435,6 +459,78 @@ dependencies:
...
```
KyleFromNVIDIA:
```suggestion
          - matrix:
              cuda_suffixed: "false"
            packages:
              - *libcuopt_client_unsuffixed
```

dependencies.yaml
```diff
@@ -435,6 +459,78 @@ dependencies:
...
+              cuda: "12.*"
+              cuda_suffixed: "true"
+            packages:
+              - libcuopt-mathopt-cu12==26.10.*,>=0.0.0a0
+          - matrix:
+              cuda: "13.*"
+              cuda_suffixed: "true"
+            packages:
+              - libcuopt-mathopt-cu13==26.10.*,>=0.0.0a0
+          - {matrix: null, packages: [*libcuopt_mathopt_unsuffixed]}
```
KyleFromNVIDIA:
```suggestion
          - matrix:
              cuda_suffixed: "false"
            packages:
              - *libcuopt_mathopt_unsuffixed
```

dependencies.yaml
```diff
@@ -435,6 +459,78 @@ dependencies:
...
+              cuda: "12.*"
+              cuda_suffixed: "true"
+            packages:
+              - libcuopt-routing-cu12==26.10.*,>=0.0.0a0
+          - matrix:
+              cuda: "13.*"
+              cuda_suffixed: "true"
+            packages:
+              - libcuopt-routing-cu13==26.10.*,>=0.0.0a0
+          - {matrix: null, packages: [*libcuopt_routing_unsuffixed]}
```
KyleFromNVIDIA:
```suggestion
          - matrix:
              cuda_suffixed: "false"
            packages:
              - *libcuopt_routing_unsuffixed
```

KyleFromNVIDIA (changes requested):
Please make `cuda_suffixed` explicit even though the hook doesn't currently require it.

dependencies.yaml
ramakrishnap-nv: The three unsuffixed entries now match on `cuda_suffixed: "false"` as suggested, applied upstream in cefc51db1. Generator runs clean with no drift and `verify-dependencies` passes.

dependencies.yaml
ramakrishnap-nv: Applied — expanded all three  shorthands to the explicit  form.

dependencies.yaml
ramakrishnap-nv: Applied.

dependencies.yaml
ramakrishnap-nv: Applied.

dependencies.yaml
ramakrishnap-nv: Applied — expanded all three `{matrix: null, ...}` shorthands to the explicit `cuda_suffixed: "false"` form to match the existing style in the file.

conda/recipes/libcuopt/recipe.yaml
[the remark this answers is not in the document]
ramakrishnap-nv: Updating — the CI RTLD_NOW load test just confirmed the finding. The previous response was wrong to defer: while the `ignore_run_exports` suppression correctly avoids the CUDA-minor pin from `libcusparse-dev`'s run export, it also removed `libcusparse` from the resolved test environment entirely, causing `libcuopt_mathopt.so` to fail to load (`libcusparse.so.12: cannot open shared object file`). Fixed in e4166af3: added `libcusparse` explicitly to `run` (unversioned, same approach as `libcublas`), while keeping it in `ignore_run_exports` so the cache's pinned run export is still suppressed.

KyleFromNVIDIA (approved):

tmckayus (approved):

PR #1932, opened 2026-09-17
Add NNZ cutoff for concurrent Barrier

hlinsen:


cpp/src/pdlp/pdlp_constants.hpp
```diff
@@ -38,6 +40,18 @@ inline constexpr bool deterministic_batch_pdlp = true;
 
 inline constexpr bool enable_batch_resizing = true;
 
+inline constexpr std::size_t concurrent_barrier_nnz_cutoff = 50'000'000;
```
mlubin: Consider adding a TODO to delete this cutoff once the specific issues in cuDSS are resolved.

chris-maes: I think you need `CONCURRENT_BARRIER_NNZ_CUTOFF` to be a parameter that users can change. In case someone wants to enable barrier.

cpp/include/cuopt/mathematical_optimization/mip/solver_settings.hpp
```diff
@@ -147,6 +147,7 @@ class mip_solver_settings_t {
   i_t strong_branching_simplex_iteration_limit = -1;
   i_t num_gpus                                 = 1;
   method_t method{method_t::Concurrent};
+  i_t concurrent_barrier_nnz_cutoff{50'000'000};
```
yuwenchen95: Just out of curiosity, how does this value determined? Is it possible to check available memory dynamically based on the memory use of PDLP?

cpp/src/pdlp/solve.cu
```diff
@@ -1620,10 +1626,21 @@ optimization_problem_solution_t<i_t, f_t> run_concurrent(
   // library init is now recovered by manual_cuda_graph_t::run, so the previous main-thread
   // preflight (eager handle construction + cuDSS warmup) is no longer needed.
   std::unique_ptr<raft::handle_t> barrier_handle_ptr;
-  if (!enable_barrier) {
-    CUOPT_LOG_DEBUG("MIP: skipping concurrent barrier, %d threads available < %d required.",
+  if (skip_cpu_solvers) {
+    CUOPT_LOG_CONDITIONAL_INFO(
+      !settings.inside_mip,
+      "Skipping concurrent Barrier and dual simplex: reduced problem has %zu nonzeros (cutoff: "
+      "%d).",
```
chris-maes: Maybe this should be: 
Skipping concurrent barrier and dual simplex: reduced problem has XXX nonzeros (CONCURRENT_NNZ_CUTOFF: YYY)

So we point users to the parameter name that needs to change.

chris-maes (approved):
Minor change in comment displayed to user. Otherwise LGTM. Thank you

PR #1936, opened 2026-09-18
Barrier presolve remove fixed variables

yuwenchen95:
## Description
Variables with equal lower and upper bounds contribute a constant to the objective and a constant to each row activity, so they can be folded into obj_constant and the RHS and dropped before the barrier forms its KKT system.

remove_fixed_variables already existed in [presolve.cpp](vscode-webview://06vj0jbc9r8l8dkbr8fqud2co2jp64rpjrltbbetfj6gkj2qeatu/cpp/src/dual_simplex/presolve.cpp#L245) but was never called from anywhere — it was dead code, and it did not record the bookkeeping postsolve needs. This PR rewrites it to be postsolve-correct and wires it into presolve() behind a new setting.

On db-joint-soerensen this removes 167,184 of 2,477,529 columns and 2.5% of the nonzeros.


mlubin:
Why do this for barrier only?

yuwenchen95:
> Why do this for barrier only?

It's a byproduct of another pending PR to improve robustness of barrier on hard SOCP problems. I thought MIP's presolve is handled to PSLP, please correct me if I'm incorrect.

PR #1938, opened 2026-09-18
test(conda): guard the per-component package boundaries

ramakrishnap-nv:
Stacked on #1928.

#1928 already loads each component library with `RTLD_NOW` in an environment holding only that package and its declared dependencies. Four gaps remain:

| Output | Gap | Added |
|---|---|---|
| `libcuopt-mathopt` | `cuopt_cli` only checked for existence | run `cuopt_cli --help` |
| `libcuopt-routing` | "no longer pulls cusparse, cudss or nccl" is unasserted | assert absent from the solved env |
| `libcuopt-client` | "no CUDA math libs" is unasserted | assert cublas/cusparse/cudss/nccl absent |
| `libcuopt` | `libcuopt.so` is an ld script, so `CDLL` cannot cover it | parse `INPUT()`, load each library |
| `libcuopt` | `cuopt-config.cmake` unguarded | add it and `cuopt-targets.cmake` to `package_contents` |
| `libcuopt-tests` | no `tests:` block | `package_contents` for three test binaries |

The load tests pass whether or not the heavy CUDA libraries are present, so they cannot catch the split regressing. The new assertions run against the *solved environment*, so a transitive `run_export` is caught too. `rattler-build` gives each `tests:` entry its own environment, so this needs no extra CI job.

### Verification

Recipe renders all six outputs under `rattler-build` 0.57.2 (CI's version). The ld-script parser and `conda-meta` lookup were run against an installed tree: the parser returns the three components and all three `CDLL`-load; the lookup detects `libcusparse`, `libcudss`, `nccl`, `libcublas` and `librmm` when present, so the assertions fail rather than pass vacuously. `cuopt_cli --help` exits 0 with no GPU.

Not verified: the tests have not run in the real per-output environments, which needs a full conda C++ build.

## Checklist

- Testing
- Documentation
   - [x] NA


ramakrishnap-nv:
@coderabbitai review

ramakrishnap-nv:
Closing — the maintenance cost outweighs the value in this shape.

Keeping the context here for whoever picks this up.

**What prompted it.** The component packages in #1928 as actually built pull the libraries the split is meant to remove. Solving a routing-only environment from the CI artifact gives `libcublas`, `libcudss`, `libcusparse` and `nccl` — 834 MB. The client-only environment pulls the same four.

**Root cause.** Every output inherits `run_exports` from the shared top-level `cache:` host environment, not just from its own `host:`. The per-output `ignore_run_exports` lists only suppress the output's own. Verified with a minimal recipe under rattler-build 0.57.2: an output declaring *no requirements at all* still inherits the cache's exports.

That is also why CI went red, though only by luck — one of the ~18 leaked exports was `cuda-version`, pinned to the build's CUDA minor, which broke the test-environment solve. The `libcusparse` / `libcudss` / `nccl` leak on its own resolves fine and would have shipped silently.

**Why not this PR.** The routing and client checks were denylists of forbidden package names. They fail correctly today, but they decay silently: a heavy dependency added later that isn't on the list is simply not covered, and nothing ever goes red to say so.

**Better fix.** `ignore_run_exports` on the `cache:` block. Verified that it suppresses only the cache's exports — an output declaring a package in its own `host:` still gets the correct ABI-tight pin. That closes the invisible channel, so a component can only gain a dependency via a line someone wrote in the recipe, which review catches. Weaker than an assertion, but no list to maintain.

Worth revisiting are the two items here with nothing to maintain: `cuopt_cli --help` (the executable has its own `DT_NEEDED` closure that loading the `.so` does not exercise) and `cuopt-config.cmake` / `cuopt-targets.cmake` in `package_contents` (5112f48 exists because those files were dropped once already).

PR #1941, opened 2026-09-18
Add update apis

Iroy30:
## Checklist

- Testing
   - [x] Added tests
   - [ ] NA
- Documentation
   - [ ] Added new documentation
   - [x] NA


cpp/src/barrier/barrier_transform.hpp
```diff
@@ -43,6 +45,12 @@ struct barrier_transform_t {
   std::vector<double> row_scales;
   // Barrier linear objective minus crush(user c) from the first solve (Q*ell shift, etc.).
   std::vector<double> linear_obj_shift;
+  // Barrier RHS minus crush(user b) from the first solve (fixed/lower-bound shifts).
+  std::vector<double> rhs_shift;
+  // False when range rows or folding put the user RHS somewhere other than barrier_lp->rhs.
+  bool rhs_update_supported{false};
+  // Absolute primal tolerance of the first solve, used to test rows presolve dropped as empty.
+  double primal_tol{1e-6};
```
yuwenchen95, line 53: We don't need to save this primal tolerance and it should be decoupled from existing`settings.primal_tol` since they are for different purpose. It is only used in empty-row check, we can localize the precision there with a much tighter value.

cpp/src/barrier/barrier_cache.cu
```diff
@@ -36,6 +59,8 @@ struct barrier_cache_t::impl {
   std::unique_ptr<barrier_transform_t> transform;
   barrier_iteration_data_ptr iteration_data;
   bool c_dirty{false};
```
yuwenchen95, line 61: Shall we better rename them as `linear_objective_dirty` and `rhs_dirty`?

PR #1944, opened 2026-09-18
Enable compilation errors on unused-parameter and unused-variable

mlubin:


mlubin:
/ok to test [47c3ca5](https://github.com/NVIDIA/cuopt/pull/1944/commits/47c3ca5a4d80aa5148ceae8bd845603c7280e52d)

ramakrishnap-nv:
@coderabbitai review

ramakrishnap-nv (approved):

mlubin:
/ok to test [47004d6](https://github.com/NVIDIA/cuopt/pull/1944/commits/47004d63f9e621cf2177cfb78626fc38a462bbe0)

cpp/tests/CMakeLists.txt
```diff
@@ -95,8 +95,19 @@ function(ConfigureTest CMAKE_TEST_NAME)
         "${CUOPT_TEST_DIR}/../src"
         "${CUOPT_TEST_DIR}/../src/io"
         "${CUOPT_TEST_DIR}"
-        "${papilo_SOURCE_DIR}/src"
-        "${papilo_BINARY_DIR}"
+    )
+    if(CMAKE_CXX_COMPILER_ID STREQUAL "Clang")
```
bdice, line 99: Can you explain why the behavior differs for clang? Maybe leave a comment with that.

There are a couple related comments in `cpp/CMakeLists.txt` but one was deleted. Maybe restore that?

bdice (approved):

aliceb-nv:
That's a lot of code churn, I see many locations where [[maybe_unused]] is applied to RAII objects which (to my knowledge) is unecessary. Shouldn't we consider dummy implementations for the CUOPT_LOG_* macros and cuopt_assert calls?

cpp/tests/CMakeLists.txt
mlubin: The historical issue is "Adding Papilo as a system include messes up clang's include resolution if papilo is already installed as a conda package" but the current workaround isn't correct. Papilo needs to be treated as a system library otherwise clang will fail to compile because of the new checks. Likely needs a separate PR to change the clang behavior.

mlubin:
> I see many locations where [[maybe_unused]] is applied to RAII objects which (to my knowledge) is unecessary.

Sorry that was AI slop. Fixed.

> Shouldn't we consider dummy implementations for the CUOPT_LOG_* macros and cuopt_assert calls?

Yeah I'll look into that.

chris-maes (changes requested):
Let's hold merging until after the release and after discussion with the team.

